PR #16, opened 2014-12-30
Follow changes in std / rustc

tomprogrammer:
This PR contains the changes to std::num currently.
> I expect to add the changes of the second pass stabilization of std::string very soon.

String stabilization commit added.


erickt:
thanks again!

PR #20, opened 2015-02-18
Make the JSON parser accept trailing whitespace

inrustwetrust:
Vim (and, I suspect, other editors as well) automatically adds a newline at the end of text files. This caused the JSON parser to fail due to trailing characters.


erickt:
Wonderful! Thanks!

Issue #9, opened 2014-09-08
Implement support for default fields


kvark:
Needed for https://github.com/kvark/scene-rs/tree/master/src/gltf

hugoduncan:
Another approach might be to only allow `Option` type fields to be missing, and default them to `None`.

erickt:
@kvark: ack! I just realized I haven't come up with a way to handle missing fields in serde2. I need to investigate how to get that to work :)

erickt:
We now do have the ability for deserializers like json to represent missing fields as a null or `None` type. We can  do the same thing for deserializees if they want to call `Default` by adding an annotation in the same vein as `#[serde(alias=...)]`.

erickt:
This has actually been done, but I'm leaving it open until we add a test case for it.

PR #21, opened 2015-02-20
Add serde default attribute for fields

hugoduncan:
When annotated with #[serde(default)], a field gets a default value using
std:default::Default, and will not error when decoding if the field is missing
from the input.

Addresses #9

### DIFF
--- a/serde2/serde2_macros/src/lib.rs
+++ b/serde2/serde2_macros/src/lib.rs
@@ -6,6 +6,7 @@ extern crate rustc;
 use syntax::ast::{
     Ident,
     MetaItem,
+    MetaItem_,
     Item,
     Expr,
     MutMutable,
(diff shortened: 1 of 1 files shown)


erickt:
This is awesome! Unfortunately there is a conflict. Could you rebase?

hugoduncan:
Rebased.  The only change in the conflicting regions was for the type of the `push` argument of `expand_derive_deserialize` and `expand_derive_serialize`, which seems to have changed.

I'm using rustc 81bce5290 2015-02-16.

serde2/serde2_macros/src/lib.rs
```diff
@@ -842,18 +850,50 @@ fn declare_map_field_deserializer(
...
+
+fn default_value(field: &ast::StructField) -> bool {
+    field.node.attrs.iter()
+        .any(|sa|
+             match &sa.node.value.node {
+                 &ast::MetaItem_::MetaList(ref n, ref vals) => {
+                     if n == &"serde" {
+                         vals.iter()
+                             .map(|mi|
+                                  match &mi.node {
```
tomprogrammer: These `match`es could be refactored into `if let`, that would remove 2 levels of indentation.

Issue #13, opened 2014-12-11
per field attributes


seanmonstar:
Brought up by WalrusPony in IRC. Fields in JSON may be conflict with keywords in Rust. One solution might be to append an attribute on a field, specifying the name that should be used in deriving.

```rust
#[deriving(Serializable, Deserializable)]
struct Foo {
    bar: u32,
    #[serde(name="mod")]
    mod_: String,
}
```

hugoduncan:
This could also be useful if the json contains camel cased keys, and you would like to use snake case field names.

erickt:
Done! Thanks @hugoduncan!

PR #22, opened 2015-02-20
Add aliases on struct fields

hugoduncan:
Allow specification of an alias on a struct field with
the #[serde(alias="SomeAlias"] attribute.

Addresses #13

Depends on https://github.com/erickt/rust-serde/pull/21


erickt:
Awesome! Could you rebase?

hugoduncan:
Rebased, and squashed the `if let` refactoring.

Issue #23, opened 2015-02-23
Serde2 fails to compile on latest nightly


hugoduncan:
Serde2 fails to compile with `rustc (2b01a37ec 2015-02-21)`.

Some of the errors are coming from templated structs with no data members.  eg.  `pub struct PrimitiveVisitor<T>;` is giving "error: parameter `T` is never used".  

I have been unable to come up with a way of imlpementing the functionality required of these structs, and was wondering if anyone had found a workaround?

hugoduncan:
Looks like adding  a `PhantomData<T>` data member, and implementing `Default` gets this to work, but it isn't very elegant.

hugoduncan:
Oh, I didn't see the pull request.  Not having to implement `Default` makes this look a little better.

PR #24, opened 2015-02-24
Update serde2 to compile on nightly 2015-02-21

tomprogrammer:
The biggest change in this patchset is the intoduction of `PhantomData`s. I think I've used them correctly, but please review that code first, before some nasty variance errors occur with this implementation.


tomprogrammer:
Fixes #23

PR #37, opened 2015-03-16
MapVisitor::end and SeqVisitor::end were mostly never called

oli-obk:


oli-obk:
once aster is up to date on crates.io this will compile

erickt:
Thanks! I forgot to push up the aster and quasi crates. Just did. I originally intended that it was optional to call end, but you're right it's probably for the best to do this :)

oli-obk:
well it still is optional to _implement_ end. you could just add a function that only contains `Ok(())` and then use `next` instead of `peek`

PR #54, opened 2015-04-11
Remove `impl Into<Vec<u8>> for ByteBuf` since it causes a compile error

apoelstra:
As @BurntSushi observes, there is an `impl<T, U> Into<U> for T where U: From<T>`
in libcore. We have `From<Vec<u8>>` for `ByteBuf` since we have implemented
`From<T> for ByteBuf where T: Into<Vec<u8>>`, so this is redundant anyway.


erickt:
Thanks!

PR #58, opened 2015-04-16
separate out the line/column counting from character iteration

oli-obk:
I thought that most deserializers could make use of this

this is technically a breaking change, since it also changes the way line/col positions are reported.

1. before the `\n` was character at col pos 1, but the first line's first char was the first char of the document.
2. on eof the col-counter is not increased anymore

both can be observed by the changes in the test suite

### DIFF
--- a/src/iterator.rs
+++ b/src/iterator.rs
@@ -1,0 +1,46 @@
+use std::iter::Peekable;
+use std::io;
+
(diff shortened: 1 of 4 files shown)


erickt:
That's nice! Travis is unhappy with it though.

oli-obk:
haha forgot to add the new file

oli-obk:
travis is happy now

erickt:
Thanks!

Issue #61, opened 2015-04-21
Multiple external field names


hugoduncan:
I'm using serde to interact with Amazon's EC2 API.  The API uses url-encoded parameters on input and returns XML.  I am using serde for both serialisation and deserialisation.  However, the field names expected by EC2 on input are different from the field names returned in the XML.

I can see several ways of solving this:
  a) Extend serde to allow renaming fileds differently on serialisation and on deserialisation, possibly via `rename_ser` and `rename_de` annotations.
  b) Extend serde to allow `Serialiser` specific field encodings.
  c) Create parallel data structures for input and output data.

Option c) obviously does not require any change in serde, but I'm reluctant to go this route, as it involves significant extra code (there are approx 650 structs used in the EC2 api), particularly in order to hide the duplication in the exposed api, and compile times are already awful (>5mins).

Option a) seems relatively simple to implement, but adds some complexity to serde.  Cognitive complexity can be limited by keeping the `rename` annotation for the case where the field names are identical for both serialisation and deserialisation.

Option b) seems the most flexible, but would probably add more complexity to serde, espescially for implementors of `Serializer`, and would likely incur extra runtime overhead.

Would either of a) or b) be considered for serde?

erickt:
Of course EC2 would have inconsistent names. Why make things simple for us?

So (a) looks good in general, though it's a bit strange that we can't roundtrip the format using this scheme. Do you think if we had specialization it could work for this? 

And yuck! 5 minutes? That could be a useful stress test to help optimize the type checker. Do you have it online somewhere?

hugoduncan:
I pushed the [rust AWS lib](https://bitbucket.org/hugoduncan/rust-awaze/src/a8ee6bedad3635cac38f4997e490d232cad0f019/ec2/?at=master).  Note this is still very much a work in progress.  It depends on #62 and #39.

erickt:
@hugoduncan: #69 landed, so is this safe to close?

hugoduncan:
I think so.

PR #62, opened 2015-04-21
Add serialize, deserialize specific rename

hugoduncan:
Adds the rename_ser and rename_de field attributes to specify serialisation
and deserialisation specific renames.

Addresses #61

### DIFF
--- a/tests/test_annotations.rs
+++ b/tests/test_annotations.rs
@@ -20,6 +20,13 @@ struct Rename {
     a2: i32,
 }
 
+#[derive(Debug, PartialEq, Serialize, Deserialize)]
+struct DirectionRename {
+    a1: i32,
+    #[serde(rename_serialize="a3", rename_deserialize="a4")]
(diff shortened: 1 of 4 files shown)


tests/test_annotations.rs
```diff
@@ -20,6 +20,13 @@ struct Rename {
     a2: i32,
 }
 
+#[derive(Debug, PartialEq, Serialize, Deserialize)]
+struct DirectionRename {
+    a1: i32,
+    #[serde(rename_ser="a3", rename_de="a4")]
```
erickt: Overall this looks good, but I think this attribute should probably be `rename_serialize` and `rename_deserialize`. That'll be more familiar with folks that aren't used to the short names (which should probably also be renamed).

hugoduncan:
Attribute names changed, and commits squashed.

erickt:
Thanks!

PR #64, opened 2015-04-23
missing field errors displayed original field name instead of renamed

### LINKED ISSUE #63: MapVisitor::missing_field gets the original field name, not the renamed name
Deserializing the following struct calls `missing_field` with `field=="myval"` instead of `field=="blubber"`

```rust
struct Test {
    #[serde(rename="blubber")]
    myval: String,
}
```

oli-obk:
closes #63


oli-obk:
travis-failure is due to rustc-serialize failing

erickt:
Thanks!

PR #67, opened 2015-04-26
Allow the pretty printer character to be changed

erickt:
This unfortunately loses the simd-ish whitespace printer, but since pretty printing shouldn't be on a hot path, this shouldn't really matter.

Partially addresses #65.

@Byron: how's this look?

### DIFF
--- a/src/json/ser.rs
+++ b/src/json/ser.rs
@@ -24,7 +24,7 @@ impl<W> Serializer<W>
     }
 }
 
-impl<W> Serializer<W, PrettyFormatter<'static>>
+impl<'a, W> Serializer<W, PrettyFormatter<'a>>
     where W: io::Write,
 {
     /// Creates a new JSON pretty print serializer.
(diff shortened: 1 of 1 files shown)


Byron:
I think this will work just fine, and leave room for additional options in future.

The way I see it, I would get pretty-printing indentation options with something like this ...
```Rust
Serializer::new_with_formatter(writer, PrettyFormatter::with_indent(b"\t"));
```
... which at some point might offer more options like ...
```Rust
Serializer::new_with_formatter(writer, PrettyFormatter::with_indent(b"\t").skip_null(true));
```

Byron:
By the way: It seems odd that I have to call `Serializer::new_with_formatter()` - I would expect either `Serializer::with_formatter()` or `PrettyFormatter::new_with_indent()`. As far as I can tell, it's idiomatic to  use the `with_something` form over `new_with_something`.

erickt:
That's a good point. Let me tweak the method.

erickt:
@Byron: How about now?

Byron:
Besides what is mentioned in the line comments, it looks good to me :).

PR #69, opened 2015-04-28
Adds serializer format specific field names

hugoduncan:
Allows different field names to be used for different external formats.

Field names are specified using the `rename` field attribute, e.g:

    #[serde(rename(xml= "a4", json="a5"))]

The implementation adds a `fmt` method to the `Serializer` and
`Deserializer` traits, allowing them to name the format that
produce/consume.

If no format specific renaming is used, the code produced should be the
same as before this change.

The code does not allow for a global rename and a format specific rename
at the same time, so either you use the global rename, or you have to
explicitly add a rename for each format used.

This still needs to be completed to correctly report missing field names
as per 1748831.

Reverts #62

Addresses #61


hugoduncan:
@erickt Is it possible that this will be accepted in some form? If so I'll finish it up.

src/de/mod.rs
```diff
@@ -113,6 +113,10 @@ pub trait Deserializer {
     {
         self.visit(visitor)
     }
+
+    fn fmt() -> &'static str {
```
erickt: Could this be renamed to `format()`?

src/json/value.rs
```diff
@@ -571,6 +571,11 @@ impl ser::Serializer for Serializer {
 
         Ok(())
     }
+
+    #[inline]
+    fn fmt() -> &'static str {
+        "value"
```
erickt: Shouldn't this be `"json"`?

erickt:
Why did you revert #62?

serde_macros/src/ser.rs
```diff
@@ -517,12 +517,29 @@ fn serialize_struct_visitor<I>(
...
-    let key_exprs = struct_field_strs(cx, builder, struct_def, Direction::Serialize);
+    let key_exprs = struct_field_strs(cx, builder, struct_def);
 
-    let arms: Vec<ast::Arm> = key_exprs.iter()
+    let arms: Vec<ast::Arm> = key_exprs.into_iter()
         .zip(value_exprs)
         .enumerate()
-        .map(|(i, (key_expr, value_expr))| {
+        .map(|(i, (field, value_expr))| {
+            let key_expr = match field {
```
erickt: Could this logic be moved into `field.rs`?

serde_macros/src/field.rs
```diff
@@ -5,19 +7,17 @@ use syntax::ptr::P;
 
 use aster;
 
-pub enum Direction {
-    Serialize,
-    Deserialize,
+pub enum Rename<'a> {
+    None,
+    Global(&'a ast::Lit),
+    // Format(HashMap<InternedString, &'a ast::Lit>)
```
erickt: Could you remove this commented line?

serde_macros/src/de.rs
```diff
@@ -639,11 +712,16 @@ fn deserialize_map(
     let extract_values: Vec<P<ast::Stmt>> = field_names.iter()
         .zip(struct_def.fields.iter())
         .map(|(field_name, field)| {
-            let rename = field::field_rename(field, &field::Direction::Deserialize);
+            let rename = field::field_rename(builder, field);
             let name_str = match (rename, field.node.kind) {
```
erickt: Could you also move this into `field.rs`?

serde_macros/src/de.rs
```diff
@@ -548,14 +551,80 @@ fn deserialize_field_visitor(
         )
         .build();
 
+
```
erickt: Nit: Could you remove this extra whitespace?

serde_macros/src/field.rs
```diff
@@ -30,47 +30,99 @@ pub fn field_rename<'a>(
...
                 vals.iter().fold(None, |v, mi| {
-                    if let ast::MetaNameValue(ref n, ref lit) = mi.node {
-                        if n == &"rename" || n == &dir_attr {
-                            Some(lit)
-                        } else {
-                            v
-                        }
-                    } else {
-                        v
+                    match mi.node {
```
erickt: Could we factor this out too? Maybe into an `attr.rs` file? These syntax extension files are already so messy :)

serde_macros/src/field.rs
```diff
@@ -30,47 +30,99 @@ pub fn field_rename<'a>(
...
+                        ast::MetaList(ref n, ref items) => {
+                            if n == &"rename" {
+                                let mut m = HashMap::new();
+                                m.extend(
+                                    items.iter()
+                                        .filter_map(
+                                            |item|
+                                            match item.node {
+                                                ast::MetaNameValue(ref n, ref lit) =>
+                                                    Some((// (n.to_owned(), lit)
```
erickt: Remove the commented out code.

hugoduncan:
I implemented this as #62 ended up being insufficient, in particular due to the lack of round-tripping when it was used (and I wanted to use json with rust field names, as well as XML and urlencoded query arguments).

I don't have a use case for different field names on serialisation and deserialisation in the same format.

Supporting both this and #62 would complicate the code.

Having said that, I am willing to put #62 back into this code if you think it should stay.

erickt:
Overall I like what you're doing here. The one other general thought I have is how we should grow this attribute grammar if we get some other format-specific attributes. For example, say we wanted to annotate if a field should be an xml attribute or an element. Would it be better then if we supported:

```rust
struct Person {
     #[serde(xml(attribute, rename="Gender"), json(rename="Gender"))]
     gender: String,

     #[serde(xml(element, rename="Name"), json(rename="Gender"))]
     name String,
}
```

To produce:

```xml
<person Gender="female">
    <Name>Emily</Name>
</person>
```

or:

```json
{
  "Gender": "female",
  "Name": "Emily"
}
```

?

erickt:
@hugoduncan: I wasn't crazy about the lack of round-trip-ability of #62, so if you don't need it, then I'm fine pulling it out until someone else asks for it :)

hugoduncan:
I think another question is whether format independent attributes are required.

```rust
struct Person {
     #[serde(rename="Gender", formats(xml(attribute, rename="XmlGender"))]
     gender: String,
}
```

hugoduncan:
@erickt I have refactored things to address your comments. 

The attribute grammer is the remaining issue.

serde_macros/Cargo.toml
```diff
@@ -12,5 +12,11 @@ plugin = true
 
 [dependencies]
 aster = "*"
-quasi = "*"
-quasi_macros = "*"
+
+[dependencies.quasi]
```
erickt: I don't think this is necessary anymore.

serde_macros/src/de.rs
```diff
@@ -154,14 +158,25 @@ fn deserialize_item_struct(
...
     trait_generics: &ast::Generics,
-) -> (P<ast::Item>, P<ast::Ty>, P<ast::Expr>) {
-    if trait_generics.ty_params.is_empty() {
+    forward_ty_params: Vec<ast::TyParam>,
+    forward_tys: Vec<P<ast::Ty>>
+) -> (P<ast::Item>, P<ast::Ty>, P<ast::Expr>, ast::Generics) {
+    if trait_generics.ty_params.is_empty() && forward_tys.is_empty() {
         (
-            builder.item().tuple_struct("__Visitor").build(),
+           builder.item().tuple_struct("__Visitor").build(),
```
erickt: Nit: minor whitespace change here.

serde_macros/src/de.rs
```diff
@@ -565,26 +670,30 @@ fn deserialize_field_visitor(
...
                     }
 
-                    deserializer.visit(__FieldVisitor)
+                    deserializer.visit(
+                        __FieldVisitor::<D>{ phantom: PhantomData })
                 }
             }
         ).unwrap(),
-    ]
+        ]
```
erickt: Nit: whitespace

serde_macros/src/lib.rs
```diff
@@ -10,9 +10,10 @@ use syntax::ext::base::Decorator;
 use syntax::parse::token;
 use rustc::plugin::Registry;
 
-mod ser;
-mod de;
+mod attr;
```
erickt: Nit: can you remove the newline below this line?

src/de/impls.rs
```diff
@@ -82,6 +93,10 @@ macro_rules! impl_deserialize_num_method {
     }
 }
 
+pub struct NumericVisitor<T> {
```
erickt: How does this differ from the `PrimitiveVisitor`?

src/de/mod.rs
```diff
@@ -113,12 +113,16 @@ pub trait Deserializer {
     {
         self.visit(visitor)
     }
+
+    fn format() -> &'static str {
```
erickt: Can you add a docstring here?

src/iter.rs
```diff
@@ -1,7 +1,8 @@
 use std::io;
+use std::iter::Peekable;
 
 pub struct LineColIterator<Iter: Iterator<Item=io::Result<u8>>> {
-    iter: Iter,
+    iter: Peekable<Iter>,
```
erickt: Rather than making this always Peekable, could you add an impl for it to be peekable if the underlying iterator is peekable?

hugoduncan:
Oops, looks like I pushed a couple of commits to this branch in error (the XML related changes). I've removed these.

serde_macros/src/lib.rs
hugoduncan: I'm not sure which newline you mean here.

hugoduncan:
As a general question (for quasia and aster too), should this be targeting 1.0, or 1.1-nightly?

erickt:
@hugoduncan: looks like the tests failed with an odd error:

```
examples/json.rs:11:28: 11:39 error: unexpected token: `,`
examples/json.rs:11 #[derive(Debug, Serialize, Deserialize)]
```

Any idea what could be causing that?

On quasi and aster, they need to be nightly, but do try to stay close to 1.0 otherwise. I'm close to getting syntex working, which would let us use serde_macros on stable :)

hugoduncan:
@erickt I'm still trying to pin it down (and am having trouble building a simple repro case), but quasi seems to have started disliking `,`s between (and after the last) match arms, and also in one case rejecting a `,` between two expressions used as arguments in a function call expression.

hugoduncan:
Fixed compilation issues, and rebased.  Looks like travis timed out the compilation.

Issue #68, opened 2015-04-26
Move serde into it's own organization


erickt:
@Byron / @oli-obk / @hugoduncan / @tomprogrammer / @kvark: FYI, if you haven't noticed, serde now has a new github organization. I want to be a little more formal about reviewing my patches with the rest of the serde community.

Byron:
A great move ! This will definitely give it (even) more velocity !
Thank you for taking care !

frewsxcv:
This can be closed, right?

erickt:
Yep!

oli-obk:
do you want serde (de-)serializer-repos in that orga or do you prefer the devs to put them in their own repos?

erickt:
@oli-obk: I'd be happy to have them in serde.rs. Want one for xml? What would you like it called?

@Byron: want one for yaml?

oli-obk:
Just `xml`? Since it'll already be inside the serde-rs orga it doesn't need any annotations like `rust` or `serde`.

Byron:
@erickt Sorry for the late reply. It seems there is a repo for yaml already. Right now I don't know how you see the separation to be done,but I guess I can just look at the json implementation once it is there. In the meanwhile, I will keep working in my fork and communicate through the PR, until there is something usable.

oli-obk:
@Byron: the xml-deserialization is already working as its own repo. Maybe that'll help set it up for yaml.

Byron:
@oli-obk Thank you, I will do that ! I'd still have a lot of questions about how collaboration is supposed to work. Maybe a guide exists somewhere and I didn't find it. For reasons of lazyness, I'd just push to `serde-rs/yaml` directly, without setting up a PR to allow code-reviews. The latter, however, would be good to have, and I would love to get other eyes on the code at least once certain milestones are met. While writing this, I changed my mind and now believe a PR based workflow would be the way to go ... so let's do it that way :).

oli-obk:
technically we should set up homu to get the `r+` abilities

oli-obk:
i am still pushing directly, as it's so early in development, but if you think reviews are better, I can start doing PRs

Byron:
I am not sure who will have time for reviews anyway, however, I just want to get used to that workflow, and if in doubt, will just merge my own PRs.

oli-obk:
we can do mutual reviewing. Not sure how I can set github up to get notifications about PRs though

frewsxcv:
[...]

PR #39, opened 2015-03-19
Xml

oli-obk:
would you be so kind as to review the `serde_macros` and `Deserializer` changes?

current state:
- [x] xml to struct deserialization
- [x] deserialize bool, int, string from `<anytagname>value</anythingelse>`
- [x] deserialize sequences (tuple, array, vector) as struct member
- [x] deserialize escaped chars (`&abcd;`)
- [x] deserialize CDATA
- [ ] deserialize enumerations
- [ ] deserialize arrays of enumerations
- [ ] deserialize errors instead of assertions
- [ ] more deserialize tests
- [x] parse to dom tree
- [ ] remove debug print! messages
- [ ] struct to xml serialization

anti-features (just ignoring stuff I don't like):
- [x] ignore namespaces
- [x] skip xml comments
- [x] skip xml version tag
- [x] ignoring xml-attributes

things that would be nice to have but might not be possible
- [ ] xsd verification
- [ ] sequences of sequences (how would these even look like in xml?)
- [ ] attributes to collapse xml elements that only contain a single type of element.

PR #70, opened 2015-04-28
changes needed for xml parsing

oli-obk:
I'm starting to work on a separate crate at https://github.com/oli-obk/serde-xml

cc #39 
cc #68

### DIFF
--- a/benches/bench_log.rs
+++ b/benches/bench_log.rs
@@ -15,6 +15,7 @@ use serde::de::{self, Deserialize, Deserializer};
 use serde::json::ser::escape_str;
 use serde::json;
 use serde::ser::{self, Serialize, Serializer};
+use std::str::FromStr;
 
 use rustc_serialize::Encodable;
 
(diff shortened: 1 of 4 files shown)


oli-obk:
rebased over 206e19edb40a8e3df29fc8d3862d92b02176f3dc

erickt:
It's a shame that `NumericVisitor` and `PrimitiveVisitor` are nearly identical. I suppose we should just drop `PrimitiveVisitor` since most things will need to be parsable from a string.

On making `LineColIterator` always peekable, could you instead make it that it's only peekable if it's wrapping a peekable iterator?

Other than that looks good. Sorry it took me so long to get to this!

oli-obk:
done, but this makes using the `get_mut` and `get_ref` functions quite dangerous when used with a peekable `Iterator`

oli-obk:
@erickt I remember why I added NumericVisitor instead of modifying PrimitiveVisitor: enums!

```
benches/bench_log.rs:85:15: 85:56 error: the trait `core::str::FromStr` is not implemented for the type `HttpProtocol` [E0277]
benches/bench_log.rs:85         state.visit(de::impls::PrimitiveVisitor::new())
```

erickt:
Thanks and sorry for the long delay!

PR #72, opened 2015-05-08
Improved support for byte strings

borman:
Currently working on a serialization library for a format where all strings are byte strings. Thus, I find byte string support requiring some attention.

The proposed changes are:

1. Allow to disassemble `Bytes`/`ByteBuf` into underlying representation
2. Add a nice-looking `Debug` implementation for `Bytes`/`ByteBuf`
3. Add methods to construct strings from byte strings via str::from_utf8
4. Add default `visit_byte_buf` implementation as a fallback to `visit_bytes`
5. Implement `ValueDeserializer` for `Bytes`/`ByteBuf`
6. Recognize struct field names as byte strings.

### DIFF
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -1,6 +1,9 @@
 //! Helper module to enable serializing bytes more efficiently
 
 use std::ops;
+use std::fmt;
+use std::ascii;
+use std::char;
 
 use ser;
 use de;
@@ -8,8 +11,15 @@ use de;
 ///////////////////////////////////////////////////////////////////////////////
 
 /// `Bytes` wraps a `&[u8]` in order to serialize into a byte array.
+#[derive(Clone, Copy, Eq, Hash, PartialEq, PartialOrd, Ord)]
 pub struct Bytes<'a> {
     bytes: &'a [u8],
+}
+
+impl<'a> fmt::Debug for Bytes<'a> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "b\"{}\"", escape_bytestring(self.bytes))
+    }
 }
 
 impl<'a> From<&'a [u8]> for Bytes<'a> {
@@ -25,6 +35,12 @@ impl<'a> From<&'a Vec<u8>> for Bytes<'a> {
         Bytes {
             bytes: &bytes,
         }
+    }
+}
+
+impl<'a> Into<&'a [u8]> for Bytes<'a> {
+    fn into(self) -> &'a [u8] {
+        self.bytes
     }
 }
 
@@ -46,7 +62,7 @@ impl<'a> ser::Serialize for Bytes<'a> {
 ///////////////////////////////////////////////////////////////////////////////
 
 /// `ByteBuf` wraps a `Vec<u8>` in order to hook into serialize and from deserialize a byte array.
-#[derive(Clone, Debug, Eq, Hash, PartialEq, PartialOrd, Ord)]
+#[derive(Clone, Eq, Hash, PartialEq, PartialOrd, Ord)]
 pub struct ByteBuf {
     bytes: Vec<u8>,
 }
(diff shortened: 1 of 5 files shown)


src/bytes.rs
```diff
@@ -63,6 +79,16 @@ impl ByteBuf {
             bytes: Vec::with_capacity(cap)
         }
     }
+
+    pub fn as_vec(self) -> Vec<u8> {
+        self.bytes
+    }
+}
```
oli-obk: why not also use `Into` here?

src/bytes.rs
```diff
@@ -172,3 +198,15 @@ impl de::Deserialize for ByteBuf {
...
     }
 }
+
+///////////////////////////////////////////////////////////////////////////////
+
+fn escape_bytestring(bytes: &[u8]) -> String {
+    let mut result = String::new();
+    for &b in bytes {
+        for esc in ascii::escape_default(b) {
+            result.push(char::from_u32(esc as u32).unwrap());
```
oli-obk: isn't `as char` fine here, since `esc` can never be larger than `127`?

src/bytes.rs
borman: Would conflict with `From<T>` at line 94:
```rust
impl Into<Vec<u8>> for ByteBuf {...}
impl<T> From<T> for ByteBuf where T: Into<Vec<u8>> {...}
```

```
src/bytes.rs:100:1: 106:2 error: conflicting implementations for trait `core::convert::From` [E0119]
src/bytes.rs:100 impl<T> From<T> for ByteBuf where T: Into<Vec<u8>> {
src/bytes.rs:101     fn from(bytes: T) -> Self {
src/bytes.rs:102         ByteBuf {
src/bytes.rs:103             bytes: bytes.into(),
src/bytes.rs:104         }
src/bytes.rs:105     }
                 ...
src/bytes.rs:100:1: 106:2 note: conflicting implementation in crate `core`
src/bytes.rs:100 impl<T> From<T> for ByteBuf where T: Into<Vec<u8>> {
src/bytes.rs:101     fn from(bytes: T) -> Self {
src/bytes.rs:102         ByteBuf {
src/bytes.rs:103             bytes: bytes.into(),
src/bytes.rs:104         }
src/bytes.rs:105     }
                 ...
```

borman:
Updated; will squash commits after review.

src/bytes.rs
borman: Fixed.

src/bytes.rs
borman: I would suggest removing 
```rust
impl<T> From<T> for ByteBuf where T: Into<Vec<u8>>
```
in favor of simpler
```rust
impl From<Vec<u8>> for ByteBuf
```
Though that would be a breaking change.

oli-obk: does removing the `From` impl and adding the `Into` impl work?
Then the `From` impl is generated in the standard library if I read this correctly.

borman: > Then the From impl is generated in the standard library if I read this correctly.
Yes, the offending implementation is 
```rust
impl<T> From<T> for T
```
http://doc.rust-lang.org/src/core/convert.rs.html#152

oli-obk: nope it's that one: http://doc.rust-lang.org/src/core/convert.rs.html#144 ;)

oli-obk: what I mean is, you don't need to implement `Into`, it's generated, since `From` exists.

borman: Well, the message actually mentions `From` implementations, not `Into`.

Anyway, committed this change.

borman: It would be generated if there was `impl From<ByteBuf> for Vec<u8>`, not `impl From<Vec<u8>> for ByteBuf`

oli-obk: oh right... I got very confused with all those `From` and `Into`... sorry about that

erickt:
Thanks! I'm prepping for a release, so the WIPs are fine for me right now :)

PR #74, opened 2015-05-16
Add serialization/deserialization for fixed size arrays

tomprogrammer:
Unfortunately this PR increases build times for the `serde` crate about 70%.

The implementation is similar the the tuple visitors and should be self-explanatory.


erickt:
Yikes about those compile times though :(

PR #76, opened 2015-05-18
implement lookup method for json::Value

blaenk:
The toml crate has a very cool [lookup](http://alexcrichton.com/toml-rs/toml/enum.Value.html#method.lookup) method on its `toml::Value` enum which accepts a string and yields the value at that path, if any exists. In my opinion it's much more ergonomic than using `find_path` that we inherited from rustc-serialize, compare `lookup("user.name.last")` with `find_path(&["user", "name", "last"])`.

This implementation simply splits the input on `'.'` and then feeds the result into `find_path`.

Hopefully this addition isn't too controversial. Curious to hear what you think.

### DIFF
--- a/src/json/value.rs
+++ b/src/json/value.rs
@@ -46,6 +46,22 @@ impl Value {
         Some(target)
     }
 
+    /// Looks up a value by path.
+    ///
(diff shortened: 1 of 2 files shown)


erickt:
Thanks!

PR #80, opened 2015-05-20
Fix #77, integer overflow when parsing JSON scientific notation number.

### LINKED ISSUE #77: Integer overflow when parsing JSON scientific notation number
This is a different issue from #75

### Code

```rust
#![feature(plugin)]
#![plugin(afl_coverage_plugin)]

extern crate afl_coverage;

extern crate serde;

use std::io::{self, Read, Cursor};

use serde::json::{self, Value};


fn main() {
    let mut input = String::new();
    let result = io::stdin().read_to_string(&mut input);
    if result.is_ok() {
        if let Ok(j) = json::from_str::<json::Value>(&input) {
            let _ = json::to_string(&j);
        }
    }
}
```

### Input

```
[7E-7777777777]
```

### Crash

```
root@vultr:~/afl-staging-area2# cargo run < outputs/crashes/id\:000008*
     Running `target/debug/afl-staging-area2`
thread '<main>' panicked at 'arithmetic operation overflowed', /root/serde/serde/src/json/de.rs:270
An unknown error occurred

```

This bug was found using https://github.com/kmcallister/afl.rs 👍

ProtectedMode:
I think this should fix it, but make sure to review it. :smiley:

### DIFF
--- a/serde/src/json/de.rs
+++ b/serde/src/json/de.rs
@@ -249,7 +249,7 @@ impl<Iter> Deserializer<Iter>
     fn parse_exponent(&mut self, mut res: f64) -> Result<f64, Error> {
         try!(self.bump());
 
-        let mut exp = 0;
+        let mut exp: u64 = 0;
         let mut neg_exp = false;
 
         if self.ch_is(b'+') {
@@ -267,8 +267,16 @@ impl<Iter> Deserializer<Iter>
         while !self.eof() {
             match self.ch_or_null() {
                 c @ b'0' ... b'9' => {
-                    exp *= 10;
-                    exp += (c as i32) - (b'0' as i32);
+                    macro_rules! try_or_invalid {
+                        ($e: expr) => {
+                            match $e {
+                                Some(v) => v,
+                                None => { return Err(self.error(ErrorCode::InvalidNumber)); }
+                            }
+                        }
+                    }
+                    exp = try_or_invalid!(exp.checked_mul(10));
+                    exp = try_or_invalid!(exp.checked_add((c as u64) - (b'0' as u64)));
 
                     try!(self.bump());
                 }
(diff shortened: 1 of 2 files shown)


serde/src/json/de.rs
```diff
@@ -267,8 +267,16 @@ impl<Iter> Deserializer<Iter>
         while !self.eof() {
             match self.ch_or_null() {
                 c @ b'0' ... b'9' => {
-                    exp *= 10;
-                    exp += (c as i32) - (b'0' as i32);
+                    macro_rules! try_or_invalid {
```
erickt: Could you pull this macro out to the top level since I also use it for `parse_decimal`?

erickt:
Thanks! Could you also add a test for this?

tomprogrammer:
Could you expand your patch to include Lines https://github.com/ProtectedMode/serde/blob/master/serde/src/json/de.rs#L238-239. I guess they can panic on invalid input also.

ProtectedMode:
@erickt, I'm unsure about how the test_de works. Does it convert the given Rust value to JSON, and then match it with the expected token output? If so, how do you let it output a scientific notation for a given number?

@tomprogrammer, since `dec` and `res` there are floating-point numbers, they can't overflow as far as I am aware.

tomprogrammer:
@ProtectedMode You are right, there can't be a panic in these lines.

erickt:
@ProtectedMode: Check out https://github.com/serde-rs/serde/blob/master/serde_tests/tests/test_json.rs#L705, which was the test I added for parsing decimal overflow. You probably could just add

```rust
("1e777777777777777777777777777", Error::SyntaxError(ErrorCode::InvalidNumber, 1, 22)),
```

ProtectedMode:
@erickt, it seems I was confused and missed that file, now it makes sense.

I added your test, I will squash the commits if you want.

frewsxcv:
:clap:

PR #84, opened 2015-06-01
Fix for building with nightly compiler.

Detegr:
Apparently the syntax extension API has changed a little and the build fails when compiled with nightly compiler. This PR changes the type of `annotatable` from `Annotatable` to `&Annotatable` for `expand_derive_deserialize` and `expand_derive_serialize`, making the compilation to succeed again.

### DIFF
--- a/serde_codegen/src/de.rs
+++ b/serde_codegen/src/de.rs
@@ -24,11 +24,11 @@ pub fn expand_derive_deserialize(
     cx: &mut ExtCtxt,
     span: Span,
     meta_item: &MetaItem,
-    annotatable: Annotatable,
+    annotatable: &Annotatable,
     push: &mut FnMut(Annotatable)
 ) {
     let item = match annotatable {
-        Annotatable::Item(item) => item,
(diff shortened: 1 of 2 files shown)


Detegr:
Forgot to mention, the compiler I used was `rustc 1.2.0-nightly (b77d60adb 2015-05-29) (built 2015-05-30)`.

Detegr:
It looks like this change was reverted as it compiles correctly without these changes with `rustc 1.2.0-nightly (613e57b44 2015-06-01) (built 2015-06-02)`.

Detegr:
This goes beyond my Rust build systems knowledge: it seems that without this PR, serde_codegen compiles with `cargo build`, and with this PR it does not. However, when I add serde and serde_macros as dependencies for a project, this PR is required to compile them correctly.
As said, I'm not familiar enough with cargo to figure out what is happening, so I'll leave this PR closed.

erickt:
@Detegr: This should be fixed now :)

PR #92, opened 2015-06-20
Add impls for Cow

SkylerLipthay:
I found (de)serializing `Cow` to be useful after working with rustc-serialize and found myself missing the same impls in Serde.


erickt:
Thanks!

PR #102, opened 2015-07-10
Implement serialization for `Result`.

pcwalton:
Servo wants this.


pcwalton:
@erickt Do you have time to review this soon? Two Servo IPC PRs are now blocked on this.

erickt:
I suspect we can simplify but it still looks good enough right now.

PR #105, opened 2015-07-13
Make BoolVisitor public so it is documented as being in existence

jnicholls:
This is consistent with the other visitors.  Even though they technically do not need to be public, it's handy to have it public for 1) documentation and 2) custom deserializer implementations that can pick and choose existing visitors a la carte.


erickt:
Thanks!

PR #106, opened 2015-07-14
Implement serialization for `NonZero` values in nightly.

pcwalton:
This adds a new Cargo feature so that you can opt into this feature on nightly if you need it while maintaining compatibility with stable Rust.

### DIFF
--- a/serde/src/de/impls.rs
+++ b/serde/src/de/impls.rs
@@ -9,6 +9,11 @@ use std::sync::Arc;
 
 use num::FromPrimitive;
 
+#[cfg(feature = "nightly")]
+use core::nonzero::{NonZero, Zeroable};
+#[cfg(feature = "nightly")]
+use std::num::Zero;
+
 use de::{
     Deserialize,
     Deserializer,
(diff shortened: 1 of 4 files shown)


serde/src/ser/impls.rs
```diff
@@ -545,3 +548,10 @@ impl Serialize for path::PathBuf {
         self.to_str().unwrap().serialize(serializer)
     }
 }
+
+impl<T> Serialize for NonZero<T> where T: Serialize + Zeroable {
```
erickt: Doesn't this impl also need a `#cfg(nightly)]`?

serde/src/de/impls.rs
```diff
@@ -865,3 +870,18 @@ impl<'a, T: ?Sized> Deserialize for Cow<'a, T> where T: ToOwned, T::Owned: Deser
         Ok(Cow::Owned(val))
     }
 }
+
+///////////////////////////////////////////////////////////////////////////////
+
+impl<T> Deserialize for NonZero<T> where T: Deserialize + PartialEq + Zeroable + Zero {
```
erickt: Doesn't this impl also need a `#cfg(nightly)]`?

PR #109, opened 2015-07-17
Make enum variants generic

dswd:
I am working on Msgpack implementation for serde and for this, I want to be able to map variants to numbers rather than strings.

### DIFF
--- a/serde/src/json/ser.rs
+++ b/serde/src/json/ser.rs
@@ -3,6 +3,7 @@ use std::num::FpCategory;
 use std::string::FromUtf8Error;
 
 use ser;
+use ser::Serialize;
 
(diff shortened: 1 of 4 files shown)


dswd:
This seems to break the json part, I am trying to fix it...

dswd:
Ok this became a much bigger change than I intended. Lots of the JSON implementation relies on enum variants being encoded as strings.
I hope that my changes do not break anything. Especially https://github.com/serde-rs/serde/pull/109/files#diff-d10a4862a2935aa58736fc9c43538657L216 which I an very unsure about.

dswd:
I think I am stuck. The enum string encoding is rooted to deep in the code generator for me to change.

PR #111, opened 2015-07-17
tuple enums with single element should not be a json-array

### LINKED ISSUE #104: Enum-Tuple-Variant with single element should not be a sequence
```rust
#[Deserialize]
enum A {
    B(i32),
    C(Vec<i32>),
}
```

The json for this requires additional square brackets for another array layer.

```json
{"C":[[5, 6, 7]]}

or

{"B":[42]}
```

Unit variants are already handles specially. I propose to also handle single element variants specially, so the following would be legal:

```json
{"C":[5, 6, 7]}

or

{"B":42}
```

Anyone that wishes to have the previous behavior, could use an explicit tuple.

```rust
#[Deserialize]
enum A {
    B((i32,)),
    C((Vec<i32>,)),
}
```

Alternatively deserialization might accept both, but that might be more of a mess than it's worth

oli-obk:
closes #104

### DIFF
--- a/serde/src/de/mod.rs
+++ b/serde/src/de/mod.rs
@@ -571,6 +571,11 @@ pub trait VariantVisitor {
         Err(Error::syntax_error())
     }
 
+    /// `visit_simple` is called when deserializing a variant with a single value.
+    fn visit_simple<T: Deserialize>(&mut self) -> Result<T, Self::Error> {
+        Err(Error::syntax_error())
+    }
+
     /// `visit_seq` is called when deserializing a tuple-like variant.
     fn visit_seq<V>(&mut self, _visitor: V) -> Result<V::Value, Self::Error>
         where V: Visitor
@@ -599,6 +604,10 @@ impl<'a, T> VariantVisitor for &'a mut T where T: VariantVisitor {
 
     fn visit_unit(&mut self) -> Result<(), T::Error> {
         (**self).visit_unit()
+    }
+
+    fn visit_simple<D: Deserialize>(&mut self) -> Result<D, T::Error> {
+        (**self).visit_simple()
     }
 
     fn visit_seq<V>(&mut self, visitor: V) -> Result<V::Value, T::Error>
--- a/serde/src/ser/mod.rs
+++ b/serde/src/ser/mod.rs
@@ -125,6 +125,14 @@ pub trait Serializer {
         self.visit_unit()
     }
 
+    #[inline]
+    fn visit_enum_simple<T>(&mut self,
+                            _name: &str,
+                            _variant: &str,
+                            _value: T,
+                            ) -> Result<(), Self::Error>
+        where T: Serialize;
+
     fn visit_none(&mut self) -> Result<(), Self::Error>;
 
     fn visit_some<V>(&mut self, value: V) -> Result<(), Self::Error>
--- a/serde_codegen/src/de.rs
+++ b/serde_codegen/src/de.rs
@@ -555,6 +555,12 @@ fn deserialize_variant(
                 Ok($type_ident::$variant_ident)
             })
         }
+        ast::TupleVariantKind(ref args) if args.len() == 1 => {
+            quote_expr!(cx, {
+                let val = try!(visitor.visit_simple());
+                Ok($type_ident::$variant_ident(val))
+            })
+        }
         ast::TupleVariantKind(ref args) => {
             deserialize_tuple_variant(
                 cx,
--- a/serde_codegen/src/ser.rs
+++ b/serde_codegen/src/ser.rs
@@ -296,7 +296,25 @@ fn serialize_variant(
                     )
                 }
             )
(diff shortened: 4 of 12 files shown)


serde/src/json/de.rs
```diff
@@ -632,28 +632,28 @@ impl<Iter> de::VariantVisitor for Deserializer<Iter>
...
+    }
 
+    fn visit_simple<T: de::Deserialize>(&mut self) -> Result<T, Error> {
         de::Deserialize::deserialize(self)
     }
 
     fn visit_seq<V>(&mut self, visitor: V) -> Result<V::Value, Error>
         where V: de::Visitor,
     {
-        try!(self.parse_object_colon());
```
oli-obk: sorry about the noise

serde/src/de/mod.rs
```diff
@@ -568,6 +568,11 @@ pub trait VariantVisitor {
         Err(Error::syntax_error())
     }
 
+    /// `visit_simple` is called when deserializing a variant with a single value.
+    fn visit_simple<T: Deserialize>(&mut self) -> Result<T, Self::Error> {
+        Err(Error::syntax_error())
```
oli-obk: instead of erroring this could try to do a `visit_seq` by default to get back the original behavior for downstream `Deserializer` implementors

serde/src/ser/mod.rs
```diff
@@ -124,6 +124,14 @@ pub trait Serializer {
         self.visit_unit()
     }
 
+    #[inline]
+    fn visit_enum_simple<T>(&mut self,
+                            _name: &str,
+                            _variant: &str,
+                            _value: T,
+                            ) -> Result<(), Self::Error>
+        where T: Serialize;
```
oli-obk: same here, instead of requiring this to be implemented, it could call visit_seq on some locally generated sequence visitor with a single element

erickt:
fyi, I haven't had a chance to really dig into this because of #114. There are a few things we should consider before landing this:

* In #114 I have a series of fallbacks where by default enum tuple variants call `visit_tuple_struct`, which calls `visit_tuple`, which finally calls `visit_seq`. So,
* We should expose the same functionality for tuple structs. This has the nice benefit of allowing newtypes to get rid of their wrapper type. However,
* Does this mean single element tuples should also get rid of their tuple-ness?
* Should we decay single element sequences into a single value? That might be going a step too far.
* Does this need to be expressed in serde itself, or is there a way that the {,de}serializer can implement this functionality without needing to count the values? If we trust the tuple sequence length, we could check if we're serializing just one value and do something special for it.
* I'm not crazy about the `simple` in `visit_enum_simple`. Maybe `visit_tuple_single`, `visit_tuple_struct_single`, `visit_enum_seq_single`?

oli-obk:
> Does this mean single element tuples should also get rid of their tuple-ness?

no, otherwise you need a `Vec` just to cause a json-array wrapper

> Should we decay single element sequences into a single value? That might be going a step too far.

I agree, a sequence is a sequence, this would only end up being very suprising for users

> Does this need to be expressed in serde itself, or is there a way that the {,de}serializer can implement this functionality without needing to count the values? If we trust the tuple sequence length, we could check if we're serializing just one value and do something special for it.

As i wrote above, that would eliminate the ability to actually enforce serialization of tuples when we want them. I believe a single element tuple to be different from a single element tuple struct or a single element enum tuple variant. A single element tuple is something you only create for the purpose of having a tuple around it. The others have mostly use cases where you don't want the tuple-ness.


> I'm not crazy about the simple in visit_enum_simple. Maybe visit_tuple_single, visit_tuple_struct_single, visit_enum_seq_single?

I just threw something at it so i had something to compile. And the alternatives are so long :( During deserialization it could be `VariantVisitor::visit_single_value`. but for serialization that would be `visit_enum_single_value`?

a single element tuple struct might be better represented by `visit_newtype`.

tomprogrammer:
> Does this mean single element tuples should also get rid of their tuple-ness?

I agree with oli-obk, we shouldn't special case tuples, only single element tuple structs and simle element enum tuple variants.

I use tuples to represent short heterogeneous sequences which should be wrapped in an array. The following macro code works for all lengths of the sequence:
```rust
                    ( $id, ( $(&self.$Field),+, ) ).serialize(serializer)
```

erickt:
@oli-obk: could you rebase this on master?

oli-obk:
I was waiting for 0.5.0

oli-obk:
rebased, but I can't even compile serde_json...

erickt:
@oli-obk yikes, what's the error you're getting?

oli-obk:
ah don't worry, I just was in a hurry and forgot sth in the rebase.

oli-obk:
fixed

oli-obk:
travis likes me again

erickt:
woo! there are a few more things I want to do, but this is a great foundation.

PR #114, opened 2015-07-22
Major refactor: Rename methods, variant index, struct fields, deserializing structs from seq, pulling out json, and cleanup

### LINKED ISSUE #110: Make enum variants generic
The current implementation forces serializers to represent enum variants as strings. However the deserializer API already uses generic types for enum variants.
I am working on a msgpack backend for serde which should represent enum variants as numbers rather than strings.
I tried to do the changes on my own and create a pull request but I got stuck as the assumption that enum variants are strings is all over the code, especially the json part and the codegen part.
You can see my approach at https://github.com/dswd/serde/tree/patch-2

erickt:
This is a series of breaking changes in order to better support formats like bincode:

* Bump version number to 0.5.0
* Renames methods like `visit_named_{map,seq}` to `visit_{struct,tuple,tuple_struct}`. This better reflects how their used.
* Passes the struct fields names to the deserializer.
* Allow structs to be deserialized from sequences.
* Added constructor to tuple deserializer visitor
* Moved json into it's own crate
* Other minor cleanup

Closes #110

cc @pcwalton

### DIFF
--- a/serde_json/src/lib.rs
+++ b/serde_json/src/lib.rs
@@ -92,6 +92,9 @@
 //! }
 //! ```
 
+extern crate num;
+extern crate serde;
+
 pub use self::de::{Deserializer, from_str};
 pub use self::error::{Error, ErrorCode};
 pub use self::ser::{
--- a/serde_json/src/lib.rs
+++ b/serde_json/src/lib.rs
@@ -6,7 +6,7 @@
 //! encode structured data in a text format that can be easily read by humans.  Its simple syntax
 //! and native compatibility with JavaScript have made it a widely used format.
 //!
-//! Data types that can be encoded are JavaScript types (see the `serde::json:Value` enum for more
+//! Data types that can be encoded are JavaScript types (see the `serde_json:Value` enum for more
 //! details):
 //!
 //! * `Boolean`: equivalent to rust's `bool`
@@ -16,7 +16,7 @@
 //! * `String`: equivalent to rust's `String`
 //! * `Array`: equivalent to rust's `Vec<T>`, but also allowing objects of different types in the
 //!    same array
-//! * `Object`: equivalent to rust's `BTreeMap<String, serde::json::Value>`
+//! * `Object`: equivalent to rust's `BTreeMap<String, serde_json::Value>`
 //! * `Null`
 //!
 //! An object is a series of string keys mapping to values, in `"key": value` format.  Arrays are
@@ -48,8 +48,8 @@
 //! `serde::Deserialize` trait.  Serde provides provides an annotation to automatically generate
 //! the code for these traits: `#[derive(Serialize, Deserialize)]`.
 //!
-//! The JSON API also provides an enum `serde::json::Value` and a method `to_value` to serialize
-//! objects.  A `serde::json::Value` value can be serialized as a string or buffer using the
+//! The JSON API also provides an enum `serde_json::Value` and a method `to_value` to serialize
+//! objects.  A `serde_json::Value` value can be serialized as a string or buffer using the
 //! functions described above.  You can also use the `json::Serializer` object, which implements the
 //! `Serializer` trait.
 //!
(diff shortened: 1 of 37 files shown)


erickt:
Some open questions:

* Since I'm queueing some some breaking changes, what other breaking changes are there that should be done in 0.5.0?
* Argument lists are starting to get a bit noisy. Visiting enums now ends up passing in 4 arguments: the variant index, the enum name, the variant name, then the next visitor. Deserializing structs now passes in the struct field arguments. Is it worth while trying to either capture these arguments in some context struct, or do some fancy trait trickery to get these arguments exposed? I actually implemented passing this info in a traits, but I ended up having to generate 5 impls, which seemed excessive.
* Is it worth passing in the field index to the serializer? It can already track the index on it's side, but it could be slightly more consistent with enum variants passing in the index.

pcwalton:
I don't have much of an opinion on any of those questions, but this PR seems great!

serde_json/src/lib.rs
```diff
@@ -92,6 +92,9 @@
 //! }
```
tomprogrammer: The doctest above refers to serde::json instead of serde_json.

erickt: Fixed, thanks!

tomprogrammer:
Great PR, finally serde is a generic serialization framework with json split out in a own repo.

oli-obk:
you got #115 mangled up in this PR

hugoduncan:
> other breaking changes are there that should be done in 0.5.0?

I can't think of any.  Converting nested enum/structs to a flat list of "dotted" keys currently requires some work e.g [for AWS requests](https://bitbucket.org/hugoduncan/rust-awaze/src/a8ee6bedad3635cac38f4997e490d232cad0f019/awaze-core/src/awsencoded.rs?at=master), but I'm not sure if that points to a need for some sort of cumulative "context" to be passed to each visitor.

> Argument lists are starting to get a bit noisy

The argument lists don't seem too bad to me, and I don't think they warrant the added complexity of trait trickery or defining a composite type to pass.

> Is it worth passing in the field index to the serializer? It can already track the index on it's side, but it  could be slightly more consistent with enum variants passing in the index.

What is the use case for passing the enum variant index, and does that use case exist for the field index?  or for the sequence element index?

erickt:
@hugoduncan: The enum variant index is necessary for a format like bincode, that uses it to more efficiently encode which variant a value is. For structs, it actually encodes them as a tuple, so the field position is implied in the tuple ordering. The only scenario I could think of where you'd want to encode the field index is if you're deserializing an out of order `Map<u32, T>` value, and you want to map it to a struct. This PR adds supports for that. I can't really think of another scenario, so I think we're probably good right now.

erickt:
@hugoduncan: On passing a context object around, I've thought about that and did end up experimenting with it back in the day. I ended up deciding that it was sufficient to store the state on the `Serializer` object. It felt like a cleaner api than having many formats having to thread a `()` object through the process. I'm always open to re-evaluating of course.

And that aws format looks to be quite screwy :) Amazon, json is a hierarchical file format, why not take advantage of it!

oli-obk:
> other breaking changes are there that should be done in 0.5.0?

how about passing length to the deserialization visit function for fixed size sequences like tuples? (and fixed size arrays?)

erickt:
@oli-obk: as in this method? https://github.com/serde-rs/serde/blob/master/serde/src/de/mod.rs#L223. How do you think it'd be used? Hm. Maybe for formats that prefix a sequence with a length could exit early if they don't match up with the destination type.

oli-obk:
in xml it's necessary to be able to deserialize sequences of sequences, as a sequence is just loads of elements with the same tag. So a sequence of n tuples of m arity is n*m elements with the same tag. no way to deserialize without knowing the arity. of course Vec<Vec<T>> still doesn't work

PR #115, opened 2015-07-22
Fix serializing maps/sequences with no size hint

### LINKED ISSUE #101: Incorrect JSON is generated when serializing an empty map with no up-front size provided
It'll generate things like `{"foo":{}"bar":{}}` (note no comma between `{}` and `"bar"`), which causes errors in deserialization. Supplying up-front sizes to all maps in the serialization code works around the problem.

erickt:
Closes #101


oli-obk:
lgtm

I had this bug for maps when I implemented a pandoc-ast parser just a few days ago

PR #116, opened 2015-07-22
Use new travis containers

erickt:


oli-obk:
I thought that was the default... did they change it?

erickt:
@oli-obk: I saw the serde jobs were getting a legacy alert that directed me to this [page](http://docs.travis-ci.com/user/migrating-from-legacy/?utm_source=legacy-notice&utm_medium=banner&utm_campaign=legacy-upgrade). In short, they said that putting `sudo: false` in the `travis.yml` lets us play on much faster containers.

PR #118, opened 2015-07-23
Add `#[serde(skip_serializing)] to skip serializing some fields

### LINKED ISSUE #99: Add annotation that never serializes a particular value
This is particularly useful when a structure contains a non-serializable cached value.

erickt:
Closes #99 

cc @kostya: Is this what you're looking for?

### DIFF
--- a/serde_codegen/src/attr.rs
+++ b/serde_codegen/src/attr.rs
@@ -16,14 +16,20 @@ pub enum FieldNames {
 
 /// Represents field attribute information
 pub struct FieldAttrs {
+    skip_serializing_field: bool,
     names: FieldNames,
     use_default: bool,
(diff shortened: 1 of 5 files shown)


kostya:
thanks, how to use it?
```
json.rs/src/main.rs:22:3: 22:38 error: The attribute `serde` is currently unknown to the compiler and may have meaning added to it in the future
json.rs/src/main.rs:22   #[serde(skip_serializing, default)]
```

erickt:
@kostya: If you're using `serde_macros`, try adding `#[feature(custom_attribute)]`. If you're using `syntex` and `serde_codegen` you shouldn't need to do anything, it'll add that for you.

PR #120, opened 2015-07-29
Add enum fields and tuple length to deserialization visitor methods, renamed some more methods

erickt:
Some more bits inspired by #114.

This adds the expected length of tuples, tuple structs, and tuple variants to the visitors as well as adds the variant names to the `visit_enum` method. Also renames `VariantVisitor::visit_{map,seq}` to `visit_{struct,tuple}` as well as renaming `Serializer::visit_enum_{unit,map,seq}` to `visit_{unit,struct,tuple}_variant` to be consistent with the other methods.

cc @oli-obk, @tomprogrammer, @hugoduncan

### DIFF
--- a/serde/src/de/impls.rs
+++ b/serde/src/de/impls.rs
@@ -565,7 +565,7 @@ array_impls! {
 
 macro_rules! tuple_impls {
     () => {};
-    ($($visitor:ident => ($($name:ident),+),)+) => {
+    ($($len:expr => $visitor:ident => ($($name:ident),+),)+) => {
         $(
             pub struct $visitor<$($name,)+> {
                 marker: PhantomData<($($name,)+)>,
@@ -610,7 +610,7 @@ macro_rules! tuple_impls {
                 fn deserialize<D>(deserializer: &mut D) -> Result<($($name,)+), D::Error>
                     where D: Deserializer,
                 {
-                    deserializer.visit_tuple($visitor::new())
+                    deserializer.visit_tuple($len, $visitor::new())
                 }
             }
         )+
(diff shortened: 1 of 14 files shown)


serde/src/de/impls.rs
```diff
@@ -916,17 +916,19 @@ impl<T, E> Deserialize for Result<T, E> where T: Deserialize, E: Deserialize {
             {
                 match try!(visitor.visit_variant()) {
                     Field::Ok => {
-                        let (value,) = try!(visitor.visit_seq(TupleVisitor1::new()));
+                        let (value,) = try!(visitor.visit_tuple(1, TupleVisitor1::new()));
```
oli-obk: shouldn't this now directly deserialize the value instead of a single element tuple?

serde/src/ser/impls.rs
```diff
@@ -631,7 +631,7 @@ impl<T, E> Serialize for Result<T, E> where T: Serialize, E: Serialize {
                     }
                 }
 
-                serializer.visit_enum_seq("Result", 0, "Ok", Visitor(Some(value)))
+                serializer.visit_tuple_variant("Result", 0, "Ok", Visitor(Some(value)))
```
oli-obk: same here, directly serialize the value

serde/src/de/impls.rs
erickt: Yeah, I got that in the next patch :)

erickt:
@oli-obk: I updated the PR to include also changing `visit_enum_simple` to pass in the variant index as well as switched to using it for the `Result<T, E>` serializer.

serde/src/de/impls.rs
oli-obk: :+1:

oli-obk: uhm... I see the serialization change, but the deserialization here is still the same.

oli-obk:
So, a hypothetical format could serialize the variant id only, deserialize it, and use the variant names array to find the name, and then pass that name as the variant name. LGTM

erickt:
@oli-obk: Updated the pr to also include having `Result<T, E>` deserialize from a simple enum, and added a few more tests.

erickt:
@oli-obk: totally agree on the documentation. serde is a bit of a beast :/

PR #121, opened 2015-07-30
Rename *enum_simple to enum_newtype, add hooks for serializing newtype structs

erickt:
This enables formats like JSON to serialize newtype wrapper structs without the normal object wrapper. Consider types like:

```rust
struct Point {
    x: u32,
    y: u32,
}

struct MyPoint(Point);
```

Before this patch, `MyPoint(1,2)` would get serialized as `[{"x":1,"y":2}]`, but now it gets serialized as `{"x":1,"y"2}`.

### DIFF
--- a/serde/src/de/mod.rs
+++ b/serde/src/de/mod.rs
@@ -208,8 +208,8 @@ pub trait Deserializer {
         self.visit(visitor)
     }
 
-    /// This method hints that the `Deserialize` type is expecting a named unit. This allows
-    /// deserializers to a named unit that aren't tagged as a named unit.
+    /// This method hints that the `Deserialize` type is expecting a unit struct. This allows
+    /// deserializers to a unit struct that aren't tagged as a unit struct.
     #[inline]
     fn visit_unit_struct<V>(&mut self,
                             _name: &'static str,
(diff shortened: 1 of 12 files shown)


erickt:
cc @oli-obk, @tomprogrammer, @hugoduncan

serde/src/de/mod.rs
```diff
@@ -219,6 +219,17 @@ pub trait Deserializer {
         self.visit_unit(visitor)
     }
 
+    /// This method hints that the `Deserialize` type is expecting a . This allows
```
oli-obk: this description needs to be updated

serde/src/de/mod.rs
```diff
@@ -579,8 +596,8 @@ pub trait VariantVisitor {
         Err(Error::syntax_error())
     }
 
-    /// `visit_simple` is called when deserializing a variant with a single value.
-    fn visit_simple<T>(&mut self) -> Result<T, Self::Error>
+    /// `visit_newtype` is called when deserializing a variant with a single value.
+    fn visit_newtype<T>(&mut self) -> Result<T, Self::Error>
```
oli-obk: :+1:

oli-obk:
was this closed by accident?

tomprogrammer:
The commit history contains the commits listed here, so erickt just rebased and closed the PR.

oli-obk:
oh... :)

PR #126, opened 2015-08-05
Update json number parsing to switch to floats if the number gets too big

erickt:
This patch switches json to parse a large number as a `f64` if it grows bigger than one that can fit into `i64`.

Along the way it also adds a `json::Result` helper type as well.

### DIFF
--- a/serde_json/src/de.rs
+++ b/serde_json/src/de.rs
@@ -108,7 +108,13 @@ impl<Iter> Deserializer<Iter>
                 try!(self.parse_ident(b"alse"));
                 visitor.visit_bool(false)
             }
-            b'0' ... b'9' | b'-' => self.parse_number(visitor),
+            b'-' => {
+                try!(self.bump());
+                self.parse_number(false, visitor)
+            }
+            b'0' ... b'9' => {
+                self.parse_number(true, visitor)
+            }
             b'"' => {
                 try!(self.parse_string());
                 let s = str::from_utf8(&self.str_buf).unwrap();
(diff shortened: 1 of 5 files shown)


serde_json/src/de.rs
```diff
@@ -145,95 +151,157 @@ impl<Iter> Deserializer<Iter>
...
-        }
-
-        let res = try!(self.parse_integer());
+        match self.ch_or_null() {
+            b'0' => {
+                try!(self.bump());
 
-        if self.ch_is(b'.') || self.ch_is(b'e') || self.ch_is(b'E') {
-            let mut res = res as f64;
+                // There can be only one leading '0'.
```
oli-obk: is this additional complexity really necessary? Does it matter for deserialization if there are multiple leading zeros?

serde_json/src/de.rs
```diff
@@ -145,95 +151,157 @@ impl<Iter> Deserializer<Iter>
...
 
-                // Make sure we didn't underflow.
-                if res > 0 {
-                    Err(self.error(ErrorCode::InvalidNumber))
-                } else {
-                    visitor.visit_i64(res)
+                            // We need to be careful with overflow. If we can, try to keep the
+                            // number as a `u64` until we grow too large. At that point, switch to
+                            // parsing the value as a `f64`.
+                            match res.checked_mul(10) {
```
oli-obk: res.checked_mul(10).and_then(|val| val.checked_add(digit)) instead of nesting?

serde_json/src/de.rs
erickt: Yeah, this is just from the [json](http://json.org/) grammar. serde_json strictly implements the spec, so it has to reject numbers like `00`.

oli-obk: I'm totally with you during serialization. but if the deserialization accepts more input and the code is simpler, does it really matter if it's not perfectly by the spec?

PR #129, opened 2015-08-07
Remove "_error" from de::Error methods, add string argument to Error::syntax()

### LINKED ISSUE #125: Extend `de::Error` to signal more cases of errors
Maybe we could strip off the `_error` suffix, it's already clear due to the name of the trait. I usually call these method using `de::Error::syntax_error()`.

I often parse some data coming in via `visit_str` and other visitors into specialized types. There are additional error cases when a constraint is not met, for example parsing a string as UUID. How to handle such errors?

One could argue that these errors are not related to serialization/deserialization, but validation. That means that validation errors should be embedded (like it's done in `io::Error`?).

Otherwise we could add some more methods to `de::Error`. I'd like to change `fn syntax_error() -> Self` to `fn syntax(msg: &str) -> Self` because I (ab)use it to signal validation errors and have no way to transport which constraint was violated. Which error conditions you like to see included?

Current trait definition of `de::Error`:
```rust
pub trait Error {
    fn syntax_error() -> Self;

    fn end_of_stream_error() -> Self;

    fn unknown_field_error(field: &str) -> Self;

    fn missing_field_error(field: &'static str) -> Self;
}
```

erickt:
Closes #125

### DIFF
--- a/serde/src/de/impls.rs
+++ b/serde/src/de/impls.rs
@@ -85,7 +85,7 @@ impl Visitor for BoolVisitor {
         match s.trim() {
             "true" => Ok(true),
             "false" => Ok(false),
-            _ => Err(Error::syntax_error()),
+            _ => Err(Error::syntax()),
         }
     }
 }
@@ -108,7 +108,7 @@ macro_rules! impl_deserialize_num_method {
         {
             match FromPrimitive::$from_method(v) {
                 Some(v) => Ok(v),
-                None => Err(Error::syntax_error()),
+                None => Err(Error::syntax()),
             }
         }
     }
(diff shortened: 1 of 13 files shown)


serde/src/de/impls.rs
```diff
@@ -200,12 +200,12 @@ impl Visitor for CharVisitor {
         let mut iter = v.chars();
         if let Some(v) = iter.next() {
             if iter.next().is_some() {
-                Err(Error::syntax_error())
+                Err(Error::syntax("sepected a character"))
```
tomprogrammer: typo: expected

erickt: fixed!

PR #130, opened 2015-08-07
json deser from Value adjusted to be equal to deser from strings

oli-obk:
the default impl from the trait deserializes from a single element tuple. This caused a confusing bug in my code where deserialization failed when deserializing from `Value`, but worked if i turned the `Value` into a string in-between.

### DIFF
--- a/serde_json/src/value.rs
+++ b/serde_json/src/value.rs
@@ -723,6 +723,15 @@ impl de::Deserializer for Deserializer {
     }
 
     #[inline]
+    fn visit_newtype_struct<V>(&mut self,
+                               _name: &'static str,
+                               mut visitor: V) -> Result<V::Value, Self::Error>
+        where V: de::Visitor,
+    {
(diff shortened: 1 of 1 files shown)


oli-obk:
Should I adjust the unit tests to run every test normally AND through an intermediate `Value`?

erickt:
Nice catch! I'm going to land this now because I want to release 0.5.0 any minute now, but could you add a test in a separate PR?

oli-obk:
that was the intent. yay for 0.5!

oli-obk:
do you mean a test for this specific case? or adjusting all tests to also test for failures due to intermediate `Value`s?

erickt:
Do you think this fix caught all these issues? If so, just a test for this specific case should probably be good.

oli-obk:
i have no idea. The idea behind doing it for all tests is to prevent any regressions. But it'll blow up the test-runtime a lot probably

PR #139, opened 2015-08-14
Reject serializing maps to JSON with non-string keys

### LINKED ISSUE #122: Serializing a map with non-string keys will silently generate invalid JSON
For example, a `HashMap<u32, u32>` will produce invalid JSON when serialized.

At a minimum, the json serializer should error rather than produce invalid JSON.

Ideally however, it would fall back to serializing the map as a list of pairs:
```
[[0, 1], [2, 0], ...]
```

erickt:
Closes #122. This is a breaking change since it modifies the return type of serde_json::to_vec{,pretty}, so it'll require a major version bump of serde_json.

This also black boxes the json log benchmarks, which unfortunately cuts `bench_log::bench_deserializer` from about 120MB/s to 85MB/s :/

### DIFF
--- a/serde_json/src/error.rs
+++ b/serde_json/src/error.rs
@@ -2,6 +2,7 @@ use std::error;
 use std::fmt;
 use std::io;
 use std::result;
+use std::string::FromUtf8Error;
 
 use serde::de;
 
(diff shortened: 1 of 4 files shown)


oli-obk:
lgtm

PR #147, opened 2015-08-28
readme: fix example code so it builds

jmesmon:
Might be worthwhile to look at trying to automate testing the code in
the README (perhaps by generating the README from some special tests).


erickt:
Thanks! I've been meaning to look into that actually...

PR #165, opened 2015-10-05
Improve error forwarding and `ValueDeserializer` usability

tomprogrammer:
There are two changes in this PR, I'm citing the commit messages here:

> When using a `ValueDeserializer` errors produced with the methods
> `type_mismatch`, `length_mismatch` and `syntax` were not forwarded
> correctly. These methods now store all information in the enum
> `de::value::Error`.

and

> Motivation for this change is to enable easy forwarding of visitor calls
> to other types implementing `Deserialize`:
>
> ```rust
> fn visit_u8<E>(&mut self, value: u8) -> Result<Self::Value, E>
>     where E: de::Error
> {
>     try!(Deserialize::deserialize(&mut value.into_deserializer())
>          .map_err(From::from))
> }
> ```
>
> Because `E` is generic one can't implement `From<de::value::Error>` for
> the non-local type `E`.

I don't know if adding a trait constraint is the right thing to do, but `de::value::Error` is a important part of the API of serde which may justify the additional constraint. The Json library already implemented it.

Issue #146, opened 2015-08-28
[Question] Delegate deserialization to a generic `Deserialize`able in a visitor method.


tomprogrammer:
I haven't figured out how I can achieve the following using `serde`, therefore I appreciate your help.

# Introduction
The data structure I want to deserialize appears in two forms. It contains either a single string or an array of objects.
```json
["This string maps to the error variant"]
[{keys: values}, {keys: values}, …]
```

The `enum VariantData<T>` should represent this data structure where `T` is a generic `Deserialize`able encapsulating the knowlegde how to deserialize itself.
```rust
enum VariantData<T> {
    Content(Vec<T>),
    Error(String),
}
```

In `VecVisitor<T>` calling `SeqVisitor::visit()` is enough, because every element is of type `T` and the method delegates deserialization to `T`.

I can use the same approach for `VariantData<T>` deserializing a single element to a helper `enum FirstValue<T>`. I guess an alternative would be to use a `EnumVisitor` directly but I'm not yet familiar enough with it.
```rust
enum FirstValue<T> {
    Content(T),
    Error(String),
}
```

# The problem
In both cases I need to distinguish the type I want deserialize to based on the kind of structure in the raw data. If a map follows the target type is `T` (or `Vec<T>` if directly serializing the array instead of looping over it), if a plain string follows the target type is `String`.

My first idea was: (The code is incomplete since its enough to show my problem.)
```rust
impl<T> Deserialize for VariantContent<T> {
    fn deserialize<D: Deserializer>(deserializer: &mut D) -> Result<Variant<T>, D::Error> {
        struct VariantContentVisitor<T> {
            data: PhantomData<T>,
        }

        impl<T> de::Visitor for VariantContentVisitor<T> where T: Deserialize {
            type Value = VariantContent<T>;

            fn visit_seq<V>(&mut self, mut visitor: V) -> Result<Self, V::Error>
                where V: de::SeqVisitor
            {
                let first: FirstValue<T> = try!(visitor.visit());
                match first {
                    FirstValue::Content(x) => {
                        // Loop over the remaining Ts and add them to `Vec<T>`
                        let elem: T = try!(visitor.visit());
                        unimplemented!()
                    },
                    FirstValue::Error(x) => VariantContent::Error(x),
                }
            }
        }

        deserializer.visit(VariantContentVisitor { data: PhantomData })
    }
}

// …
impl<T> de::Visitor for FirstValueVisitor<T> where T: Deserialize {
[...]

PR #166, opened 2015-10-05
Improve error forwarding and `ValueDeserializer` usability – Take 2

tomprogrammer:
There are two changes in this PR:

When using a ValueDeserializer errors produced with the methods
type_mismatch, length_mismatch and syntax were not forwarded
correctly. These methods now store all information in the enum
de::value::Error.

In constrast to https://github.com/serde-rs/serde/pull/165 this PR makes it obsolete to add the `From<de::value::Error>` trait bound to the `de::Error` trait. Even the conversion of `de::value::Error` to the deserializer specific error is not needed with this PR, the ValueDeserializer is monomorphized to use the deserializer's error type.

An example for using a `ValueDeserializer` to forward deserializing to
another type's visitor:

```rust
fn visit_u8<E>(&mut self, value: u8) -> Result<Self::Value, E>
    where E: de::Error
{
    try!(Deserialize::deserialize(&mut value.into_deserializer()))
}
```

Please note that there is an regression in serde-rs/json which needs to
be fixed.

### DIFF
--- a/serde/src/de/value.rs
+++ b/serde/src/de/value.rs
@@ -22,7 +22,13 @@ use bytes;
 #[derive(Clone, Debug, PartialEq)]
 pub enum Error {
     /// The value had some syntatic error.
-    SyntaxError,
+    SyntaxError(String),
+
+    /// The value had an incorrect type.
+    TypeError(de::Type),
+
+    /// The value had an invalid length.
+    LengthError(usize),
 
     /// EOF while deserializing a value.
     EndOfStreamError,
@@ -35,7 +41,9 @@ pub enum Error {
 }
 
 impl de::Error for Error {
-    fn syntax(_: &str) -> Self { Error::SyntaxError }
+    fn syntax(msg: &str) -> Self { Error::SyntaxError(String::from(msg)) }
+    fn type_mismatch(type_: de::Type) -> Self { Error::TypeError(type_) }
+    fn length_mismatch(len: usize) -> Self { Error::LengthError(len) }
     fn end_of_stream() -> Self { Error::EndOfStreamError }
     fn unknown_field(field: &str) -> Self { Error::UnknownFieldError(String::from(field)) }
     fn missing_field(field: &'static str) -> Self { Error::MissingFieldError(field) }
--- a/serde/src/de/value.rs
+++ b/serde/src/de/value.rs
@@ -362,6 +362,38 @@ impl<T, E> ValueDeserializer<E> for HashSet<T>
 
 ///////////////////////////////////////////////////////////////////////////////
 
+/// A helper deserializer that deserializes a sequence using a `SeqVisitor`.
+pub struct SeqVisitorDeserializer<V_, E> {
+    visitor: V_,
+    marker: PhantomData<E>,
+}
+
+impl<V_, E> SeqVisitorDeserializer<V_, E>
+    where V_: de::SeqVisitor<Error = E>,
+          E: de::Error,
+{
+    /// Construct a new `SeqVisitorDeserializer<V_, E>`.
+    pub fn new(visitor: V_) -> Self {
+        SeqVisitorDeserializer{
+            visitor: visitor,
+            marker: PhantomData
+        }
+    }
+}
+
+impl<V_, E> de::Deserializer for SeqVisitorDeserializer<V_, E>
+    where V_: de::SeqVisitor<Error = E>,
+          E: de::Error,
+{
+    type Error = E;
+
+    fn visit<V: de::Visitor>(&mut self, mut visitor: V) -> Result<V::Value, Self::Error> {
+        visitor.visit_seq(&mut self.visitor)
+    }
+}
+
+///////////////////////////////////////////////////////////////////////////////
+
 /// A helper deserializer that deserializes a map.
 pub struct MapDeserializer<I, K, V, E>
     where I: Iterator<Item=(K, V)>,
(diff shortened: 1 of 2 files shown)


GitCop:
There were the following issues with your Pull Request

 * Commit: 0a651536349480a47ea80cb137ec450d0b708b6e
  * Commits must be in the following format: **%{type}(%{scope}): %{description}**

 * Commit: c1104700dedd1cd4bb46cf247eed86d6bcc695c3
  * Commits must be in the following format: **%{type}(%{scope}): %{description}**

 * Commit: 2d609bb8f114d9aff0141526982777598385bb00
  * Commits must be in the following format: **%{type}(%{scope}): %{description}**

 * Commit: abea54b8482c70b14d4e9e98a5ead0883f22de36
  * Commits must be in the following format: **%{type}(%{scope}): %{description}**

Guidelines are available at https://github.com/serde-rs/serde

---
This message was auto-generated by https://gitcop.com

tomprogrammer:
This change may cause problems due to insufficient inference capabilities of the compiler. In https://github.com/serde-rs/json/pull/11 you can see an example of an additional type annotation which is required to avoid compile errors.

GitCop:
There were the following issues with your Pull Request

 * Commit: 0a651536349480a47ea80cb137ec450d0b708b6e
  * Commits must be in the following format: **%{type}(%{scope}): %{description}**

 * Commit: e0429cdd96afa5022b8258f86212e2b6cfcf5543
  * Commits must be in the following format: **%{type}(%{scope}): %{description}**

 * Commit: 7413bbb7bf2b92d7f92fba2fe1dde97f1933155c
  * Commits must be in the following format: **%{type}(%{scope}): %{description}**

Guidelines are available at https://github.com/serde-rs/serde

---
This message was auto-generated by https://gitcop.com

GitCop:
There were the following issues with your Pull Request

 * Commit: 0a651536349480a47ea80cb137ec450d0b708b6e
  * Commits must be in the following format: **%{type}(%{scope}): %{description}**

 * Commit: e0429cdd96afa5022b8258f86212e2b6cfcf5543
  * Commits must be in the following format: **%{type}(%{scope}): %{description}**

 * Commit: 7413bbb7bf2b92d7f92fba2fe1dde97f1933155c
  * Commits must be in the following format: **%{type}(%{scope}): %{description}**

 * Commit: b7d93b6e110215617420f5778dfe7e0fe6a5d7f1
  * Your commit message body contains a line that is longer than 72 characters
  * Commits must be in the following format: **%{type}(%{scope}): %{description}**

Guidelines are available at https://github.com/serde-rs/serde

---
This message was auto-generated by https://gitcop.com

tomprogrammer:
I added two deserializers that can now be integrated in the `ValueDeserializer` infrastructure. I've used them in https://github.com/serde-rs/serde/issues/146 . This PR makes it possible to upstream these.

GitCop:
There were the following issues with your Pull Request

 * Commit: 0a651536349480a47ea80cb137ec450d0b708b6e
  * Commits must be in the following format: **%{type}(%{scope}): %{description}**

 * Commit: e0429cdd96afa5022b8258f86212e2b6cfcf5543
  * Commits must be in the following format: **%{type}(%{scope}): %{description}**

 * Commit: 7413bbb7bf2b92d7f92fba2fe1dde97f1933155c
  * Commits must be in the following format: **%{type}(%{scope}): %{description}**

 * Commit: 0b5e91d4300ec08625f242a3e6e4c2692ba4b272
  * Your commit message body contains a line that is longer than 72 characters
  * Commits must be in the following format: **%{type}(%{scope}): %{description}**

Guidelines are available at https://github.com/serde-rs/serde

---
This message was auto-generated by https://gitcop.com

serde/src/de/value.rs
```diff
@@ -22,31 +23,39 @@ use bytes;
...
 pub enum Error {
     /// The value had some syntatic error.
-    SyntaxError,
+    Syntax(String),
+
+    /// The value had an incorrect type.
+    Type(de::Type),
+
+    /// The value had an invalid length.
+    Length(usize),
```
oli-obk: where would this be raised? When a 3 element tuple is expected but a 4 element sequence is given?

tomprogrammer: The `ValueDeserializer`s are generic in respect to the error now, so the `Length` variant is indeed not used by the existing implementations. This is a remnant of take one to avoid unifying `Length` into `Syntax`, because all nested deserializers are required to go through `value::Error` there which is not the case in take two. In conclusion this variant isn't required by the existing implementations. If we remove it and someone calls `de::Error::length_mismatch` the error would be unified into the `Syntax` variant though, but I guess this is intentional because it's explicitly stated in the `de::Error` trait.

tomprogrammer: There are three possible solutions that I see here (the last option being my preferred):
- Remove `value::Error` because `ValueDeserializer`s are assumed to be used in a nested manner.
- Remove just the `Length` variant because existing implementations don't use it.
- Keep the status quo and support all possibilities, for example other implementations which may use `de::Error::length_mismatch`.

What do you think is sensible?

erickt:
@tomprogrammer: Merged it in after fixing a test failure. Overall I like the direction this is going, but I do wish the ergonomics was a little easier. I'm starting to toy around with using a [traverse](https://github.com/reem/rust-traverse) style approach for {,de}serializing compound structures, which might help make cases like this a little cleaner to use.

Issue #60, opened 2015-04-17
allow deserialization to work with missing attributes


nicolai86:
When parsing JSON I'd like to only parse parts of the payload, e.g. 

given JSON like `{ 'x': 32, 'y': 12 }` I'd expect serde to deserialize it using the following struct:

``` rust
#[derive(Serialize, Deserialize)]
#[allow(non_snake_case)]
struct Point {
    y: f32,
}
```

However, this fails with an `UnknownField("x")` error.

This is important if you are only interested in parts of the payload, or if working with dynamic payload.

jgallagher:
I think this is a dupe of #44, although that issue mentions adding an annotation to allow skipping unknown fields. I think this issue's request should be the default; that's how most JSON parsers in other languages (and `rustc-serialize`) work.

softprops:
+1

doomsplayer:
+1

erickt:
This was implemented in #204, and will be a part of serde 0.7.0 when it comes out, hopefully within the week!

PR #201, opened 2016-01-06
Allow unknown fields (with flag to reinstate errors)

### LINKED ISSUE #44: Allow struct deserialization to ignore unknown fields
Right now `#[derive(Deserialize)]` on structs and struct variants raises an error if it's passed an unknown field. We should add an annotation that allows users to just skip these fields.

JohnHeitmann:
Closes #44 and #60. Silently allow unknown fields for formats where this is possible. Adds a new off-by-default ``#[serde(disallow_unknown)]`` attribute that will reinstate old behavior.

Allowing unknown by default is both more convenient during development and more safe in production. Service owners expect to be able to add new response fields without a version change and without breaking clients.

Test updates: https://github.com/JohnHeitmann/json/commit/7d6946772eb8f63daeb038a91c4a50319566caf1

### DIFF
--- a/README.md
+++ b/README.md
@@ -585,6 +585,8 @@ Annotations
 `serde_codegen` and `serde_macros` support annotations that help to customize
 how types are serialized. Here are the supported annotations:
 
+Field Annotations:
+
 | Annotation                                   | Function                                                       |
 | ----------                                   | --------                                                       |
 | `#[serde(rename(json="name1", xml="name2"))` | Serialize this field with the given name for the given formats |
@@ -593,6 +595,13 @@ how types are serialized. Here are the supported annotations:
 | `#[serde(skip_serializing)`                  | Do not serialize this value                                    |
 | `#[serde(skip_serializing_if_empty)`         | Do not serialize this value if `$value.is_empty()` is `true`   |
 | `#[serde(skip_serializing_if_none)`          | Do not serialize this value if `$value.is_none()` is `true`    |
+
+Structure Annotations:
+
+| Annotation                  | Function                                                                                                                                           |
+| ----------                  | --------                                                                                                                                           |
+| `#[serde(disallow_unknown)` | Always error during serialization when encountering unknown fields. When absent, unknown fields are ignored for self-describing formats like JSON. |
+
 
 Serialization Formats Using Serde
 =================================
--- a/serde/src/de/impls.rs
+++ b/serde/src/de/impls.rs
@@ -898,6 +898,110 @@ impl<T, E> Deserialize for Result<T, E> where T: Deserialize, E: Deserialize {
 
 ///////////////////////////////////////////////////////////////////////////////
 
+/// A target for deserializers that want to ignore data. Implements
+/// Deserialize and silently eats data given to it.
+pub struct IgnoredAny;
+
+impl Deserialize for IgnoredAny {
+    #[inline]
+    fn deserialize<D>(deserializer: &mut D) -> Result<IgnoredAny, D::Error>
+        where D: Deserializer,
+    {
+        struct IgnoredAnyVisitor;
+
+        impl Visitor for IgnoredAnyVisitor {
+            type Value = IgnoredAny;
+
+            #[inline]
+            fn visit_bool<E>(&mut self, _: bool) -> Result<IgnoredAny, E> {
+                Ok(IgnoredAny)
+            }
+
+            #[inline]
+            fn visit_i64<E>(&mut self, _: i64) -> Result<IgnoredAny, E> {
+                Ok(IgnoredAny)
+            }
+
+            #[inline]
+            fn visit_u64<E>(&mut self, _: u64) -> Result<IgnoredAny, E> {
+                Ok(IgnoredAny)
+            }
+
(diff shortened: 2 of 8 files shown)


JohnHeitmann:
CI failure looks unrelated to my change. rand-0.3.12 doesn't compile wth rust nightly.

erickt:
Thanks @JohnHeitmann! Yeah, those test failures are from nightly breaking the plugins. I'll look it over in a few minutes.

README.md
```diff
@@ -593,6 +593,7 @@ how types are serialized. Here are the supported annotations:
 | `#[serde(skip_serializing)`                  | Do not serialize this value                                    |
 | `#[serde(skip_serializing_if_empty)`         | Do not serialize this value if `$value.is_empty()` is `true`   |
 | `#[serde(skip_serializing_if_none)`          | Do not serialize this value if `$value.is_none()` is `true`    |
+| `#[serde(disallow_unknown)`                  | Always error during serialization when encountering unknown fields. When absent, unknown fields are ignored for self-describing formats like JSON.  |
```
erickt: Note to self: we'll have to bump the major version for this change.

serde/src/de/impls.rs
```diff
@@ -898,6 +898,109 @@ impl<T, E> Deserialize for Result<T, E> where T: Deserialize, E: Deserialize {
...
+            }
+
+            #[inline]
+            fn visit_map<V>(&mut self, mut visitor: V) -> Result<IgnoredAny, V::Error>
+                where V: MapVisitor,
+            {
+                while let Some((_, _)) = try!(visitor.visit::<String, IgnoredAny>()) {
+                    // Gobble
+                }
+                try!(visitor.end());
```
erickt: minor nit: could you put a newline here line you did on 974?

serde_codegen/src/field.rs
```diff
@@ -15,3 +15,11 @@ pub fn struct_field_attrs(
...
         .collect()
 }
+
+pub fn container_attrs(
+    _cx: &ExtCtxt,
+    container: &ast::Item,
+) -> ContainerAttrs {
+    ContainerAttrsBuilder::new().attrs(container.attrs()).build()
+}
+
```
erickt: minor nit: can you remove the trailing newline here?

serde_codegen/src/de.rs
```diff
@@ -82,6 +82,9 @@ fn deserialize_body(
     impl_generics: &ast::Generics,
     ty: P<ast::Ty>,
 ) -> P<ast::Expr> {
+
```
erickt: minor nit: remove this newline.

README.md
erickt: Now that we have container annotations, we should tweak these docs. Can you refactor this section to something like:

```
Structure Annotations:

| Annotation                  | Function                                                                                                                                           |
| ----------                  | --------                                                                                                                                           |
| `#[serde(disallow_unknown)` | Always error during serialization when encountering unknown fields. When absent, unknown fields are ignored for self-describing formats like JSON. |

Field Annotations:

| Annotation                                   | Function                                                       |
| ----------                                   | --------                                                       |
| `#[serde(rename(json="name1", xml="name2"))` | Serialize this field with the given name for the given formats |
| `#[serde(default)`                           | If the value is not specified, use the `Default::default()`    |
| `#[serde(rename="name")`                     | Serialize this field with the given name                       |
| `#[serde(skip_serializing)`                  | Do not serialize this value                                    |
| `#[serde(skip_serializing_if_empty)`         | Do not serialize this value if `$value.is_empty()` is `true`   |
| `#[serde(skip_serializing_if_none)`          | Do not serialize this value if `$value.is_none()` is `true`    |
```

erickt:
Overall this looks great, but we need some tests here (instead of in the json repo). Check out [tests/test_annotations.rs](https://github.com/serde-rs/serde/blob/master/serde_tests/tests/test_annotations.rs) if you need some references.

Thanks!

JohnHeitmann:
Yeah, drat, I added those tests and found a bug where I assume map keys are always strings in IgnoredAny::visit_map(). Trying to work that one out...

JohnHeitmann:
Ok, this is 99% ready to go. Much better test coverage from serde_tests now. I _think_ that integer underflow guard I added to test/tokens is papering over a test harness bug, not an implementation bug. I'll take a second look at it tomorrow.

JohnHeitmann:
Length hinting in the token deserializer feeds off of the rust type, not the token stream. Since the rust type has fewer fields than streams with unknown tokens, it would underflow under the old logic. I believe the underflow hack I added is fine because length hinting is for externally defined formats like bincode  and those formats can't support unknown anyway. I'd appreciate it if you would take a look.

erickt:
@JohnHeitmann: Overall this looks great. The only thing left is you should apply these patches to the `master` branch, not `0.6.x`, since this will be a breaking change. I'll then cut 0.7.0.

JohnHeitmann:
Moved to a new PR against master: https://github.com/serde-rs/serde/pull/204

(Googling didn't turn up a way to update this in place)

PR #204, opened 2016-01-11
Implemented disallow_unknown

JohnHeitmann:
* Added codegen for disallow_unknown
* ... with new default to ignore unknown values during deserialization
* Added ContainerAttrs

Original pull request against v0.6.x: https://github.com/serde-rs/serde/pull/201

### DIFF
--- a/README.md
+++ b/README.md
@@ -585,6 +585,8 @@ Annotations
 `serde_codegen` and `serde_macros` support annotations that help to customize
 how types are serialized. Here are the supported annotations:
 
+Field Annotations:
+
 | Annotation                                   | Function                                                       |
 | ----------                                   | --------                                                       |
 | `#[serde(rename(json="name1", xml="name2"))` | Serialize this field with the given name for the given formats |
@@ -593,6 +595,13 @@ how types are serialized. Here are the supported annotations:
 | `#[serde(skip_serializing)`                  | Do not serialize this value                                    |
 | `#[serde(skip_serializing_if_empty)`         | Do not serialize this value if `$value.is_empty()` is `true`   |
 | `#[serde(skip_serializing_if_none)`          | Do not serialize this value if `$value.is_none()` is `true`    |
+
+Structure Annotations:
+
+| Annotation                  | Function                                                                                                                                           |
+| ----------                  | --------                                                                                                                                           |
+| `#[serde(disallow_unknown)` | Always error during serialization when encountering unknown fields. When absent, unknown fields are ignored for self-describing formats like JSON. |
+
 
 Serialization Formats Using Serde
 =================================
--- a/README.md
+++ b/README.md
@@ -600,7 +600,7 @@ Structure Annotations:
 
 | Annotation                  | Function                                                                                                                                           |
 | ----------                  | --------                                                                                                                                           |
-| `#[serde(disallow_unknown)` | Always error during serialization when encountering unknown fields. When absent, unknown fields are ignored for self-describing formats like JSON. |
+| `#[serde(deny_unknown_fields)` | Always error during serialization when encountering unknown fields. When absent, unknown fields are ignored for self-describing formats like JSON. |
 
 
 Serialization Formats Using Serde
--- a/serde/src/de/impls.rs
+++ b/serde/src/de/impls.rs
@@ -898,6 +898,110 @@ impl<T, E> Deserialize for Result<T, E> where T: Deserialize, E: Deserialize {
 
 ///////////////////////////////////////////////////////////////////////////////
 
+/// A target for deserializers that want to ignore data. Implements
+/// Deserialize and silently eats data given to it.
+pub struct IgnoredAny;
+
+impl Deserialize for IgnoredAny {
+    #[inline]
+    fn deserialize<D>(deserializer: &mut D) -> Result<IgnoredAny, D::Error>
+        where D: Deserializer,
+    {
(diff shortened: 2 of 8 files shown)


hauleth:
:+1: for that. I am parsing GH API output and including hell lot of unneeded fields to satisfy serde would be terrible.

oli-obk:
What's the reason for making the default ignore unknown fields? If you implement it the other way around, then this doesn't even need a major version bump.

hauleth:
I think that the main reason behind that is that this behaviour is consistent with any other serialization library that I am aware of.

oli-obk:
I don't like hiding something like that. Even if it's the standard everywhere else, it's a guarantee for bugs to allow by default. The correct way to go forward would be to analyze serde-use-cases and see which default is less often used. (like `mut` vs `const` in variable mutability). Side-note: "allow by default" is more like "mut by default", but that's more of an opinion ;)

hauleth:
I think that in 90% of use cases of deserialization author of the library will not be in control of source of data. That is why allowing to drop unused fields by default will be reasonable. 

I think it should allow extra fields but without such radiacal change. There should be additional `#[allow_unused]` with deprecation warning on fields without any flag. 

But still we should allow unused fields with 0.7.

--
Łukasz Jan Niemier

Dnia 11 sty 2016 o godz. 15:49 Oliver Schneider <notifications@github.com> napisał(a):

erickt:
@JohnHeitmann: Another tweak. How about `#[serde(deny_unknown_fields)]`? That'd blend a bit better with the `#[deny(...)]` lint attributes?

erickt:
@oli-obk: A couple things. First, we really do need a major bump to rename `visit` to `serialize`. I've been sitting on that branch for far too long. The advantage with making this switch is that I've seen a number of APIs, ([Spark's REST api](https://spark.apache.org/docs/latest/monitoring.html) off the top of my head) that consider adding fields to requests/responses to be a non-breaking change. For our users, this means they'd have to jump through a lot of hoops to interact with multiple services that may or may not be returning everything. So I think it'd be more common for our users to want us to be liberal on what we parse.

oli-obk:
I'm fully aware of this. But I still consider this anecdotal evidence. I only encounter situations where it should be a breaking change. Savegames, Rust's metadata, Configuration files (the reason I wrote serde-xml). I'm not sure how to collect a sound dataset for making a decision, since I'm biased myself. If we'd use an `allow_unknown_fields`, we could at least start collecting this data by counting serde users using the attribute and serde users not using it. By allowing by default, we cannot collect this data, since it will work for people that should be using the `deny_unknown_fields` attribute.

JohnHeitmann:
I come at this from a web services background. The status quo in the web services world is for service providers to assume that they can add new fields in responses without breaking clients (and without versioning).

Disallowing unknown by default in that context is dangerously brittle. With that default in every situation where the producer and consumer are different processes (not just client/server situations, but even the same process across restarts) you would have to version any change that added a field or break consumers. Versioning is hard, a huge pain for both client and server, and is easy to forget to do entirely or get wrong. Maintaining multiple versions over time is costly for everyone involved, and itself carries risks because of the increased surface area of things to get wrong.

>  I only encounter situations where it should be a breaking change. Savegames, Rust's metadata, Configuration files

I'm not convinced all of those should be breaking by default. Maybe something non-critical like save games. But consider a servo config file: you release servo version 9 that has a new config field. Users upgrade. Some users for whatever reason need to downgrade back to version 8 (or maybe they copy their config to a different machine). Should servo 8 refuse to start because of the extra field?

> Another tweak. How about #[serde(deny_unknown_fields)]? That'd blend a bit better with the #[deny(...)] lint attributes?

Good idea.

JohnHeitmann:
One more thing on save games even though it's way out of my domain: they're probably going to be a non-self-describing format, in which case this setting doesn't apply. An extra field in a packed binary save file is in practice still not allowed.

oli-obk:
a third alternative would be to require an annotation, but that might be overly annoying for such a minor issue. 

Let's go with your way.

erickt:
Thanks @JohnHeitmann! I'll cut a release soon :)

softprops:
> What's the reason for making the default ignore unknown fields?

Wanted to chime in. In many cases you are writing clients for servers you don't control and in many of those cases you may only care for a subset of the the total fields available.

doomsplayer:
:+1:

jwilm:
Going to need this for redis-serde. Thanks for implementing it! Looking forward to that 0.7 release, now.

softprops:
Is there an eta on this release?

softprops:
Bump

jwilm:
@softprops do you have an urgent need? I've got a .7 compat branch of redis-json if that's helpful to you.

softprops:
Eager is a better word :) I've got handful of rustc serialize based projects that Ive been holding off on porting to to serde because have been blocked by this issue. Serde will clean up so much custom serializer code I've had to write but I've been blocked so far. I'm so excited that it finally landed! The waiting is killing me!

mitsuhiko:
Definitely needs a release with this, but I noticed that currently the json-serde library is very much broken for 0.7. This should probably be fixed before serde 0.7 actually lands.

softprops:
:/

softprops:
My use of serde is exclusively json-centric

ernestrc:
+1

softprops:
How can I help to make that release happen?

PR #205, opened 2016-01-12
Remove extraneous flag

JohnHeitmann:
Getting this out of the way in front of a bigger patch it would have gotten lost in. This change should do nothing. I think the intention here was to say, "Yes, take the default format if needed in renames", but this default is about data defaulting (and rename defaults are always taken anyway if needed).


erickt:
Yep, I think you're right. I checked, and this didn't seem to do anything.

Issue #195, opened 2015-12-19
Equivalent of the "rename" annotation for enums?


jimmycuadra:
I have an enum that I'd like to serialize with a different format for the variants than it appears in the Rust source code. For example, a value of this enum:

``` rust
enum Example {
  FirstExample,
  SecondExample,
}
```

would be serialized as the string "first_example" or "second_example". For structs, there is the annotation `#[serde(rename="first_example")]` that can be used for a field's name, but I'm wondering if there is an equivalent for enum variants to avoid having to implement serialization and deserialization manually. I think this would only make sense for enum variants without data.

erickt:
Unfortunately there isn't support for that yet, but it's a reasonable feature request.

jimmycuadra:
If I were to take a shot at implementing it, would you want to reuse the "rename" annotation or would you prefer something else?

erickt:
Dup of #182. Thanks though!

PR #207, opened 2016-01-15
Support rename for enum variants.

### LINKED ISSUE #182: Enum variants can't be renamed
I would expect this to work, but it doesn't appear to:

```rust
#[derive(Deserialize)]
enum Foo {
    #[serde(rename = "BAR")]
    Bar,
}
```

JohnHeitmann:
Closes #182, #195
Obviates PR https://github.com/serde-rs/serde/pull/205

Unfortunately there are some minor snags/quibbles:

The old behavior was to silently ignore the attribute on variants, so this could theoretically be a breaking change for someone who tried it out, but it didn't work, but they left it in production code. I don't have a strong opinion about what to do here. This change is the easiest route: keep the name. If you want to do something different (pick a different name; or release validation to forbid the old behavior for a few months and hold this in reserve) I'd be happy to bolt that on. I'm working on attr validation anyway, which is why the attr.rs change is anemic here.

I ran into a glitch with format names. Deserializer format name is determined very locally, which meant any Deserializer that uses a ValueDeserializer as a helper wouldn't get format-specific renames applied as intended. My first inclination was to pass the parent Deserializer type down from the top level of the codenned nest all the way into leaf nestings, so that the root Deserializer always governed the format name. My rust-fu wasn't up to the task. Instead, this lets Deserializer authors control nested format names. More flexible, but more easy to forget. I can take another crack at my original plan if that's what you'd prefer.

### DIFF
--- a/serde/src/de/impls.rs
+++ b/serde/src/de/impls.rs
@@ -1001,6 +1001,245 @@ impl Deserialize for IgnoredAny {
     }
 }
 
+///////////////////////////////////////////////////////////////////////////////
+
+/// See: RenamedDeserializer::rename_deserializer
+pub struct RenamedDeserializer<MAIN, NAME> {
+    main: MAIN,
+    new_name: PhantomData<NAME>,
+}
+
+impl <MAIN, NAME> RenamedDeserializer<MAIN, NAME> {
+    /// Returns a new Deserializer wrapping main_deserializer, which forwards all methods
+    /// to main_deserializer, except for format(), which it forwards to the NAME type.
+    ///
+    /// This is useful for Deserializer authors who use helper deserializers, but want
+    /// those helper serializers to share the same name as the parent deserializers
+    /// for the purposes of, for example, field renaming.
+    pub fn rename_deserializer(main_deserializer: MAIN) -> RenamedDeserializer<MAIN, NAME> {
+        RenamedDeserializer{ main: main_deserializer, new_name: PhantomData }
+    }
+}
+
+#[allow(missing_docs)]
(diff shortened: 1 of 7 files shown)


JohnHeitmann:
Is there a way to see line-by-line coverage coloring in Travis?

erickt:
This would go in a major version bump, so I'm not too worried about breaking people's code if they forgot about an attribute. Reading through the patch now...

erickt:
I've actually been considering getting rid of `#[serde(rename(format1=..., format2=...))]`. I'm not quite sure if it's pulling it's weight. Would that remove the need for `RenamedDeserializer`?

JohnHeitmann:
If you removed Deserializer::format entirely would it remove the need for RenamedDeserializer.

If you just removed the format-specific rename, then folks could/might still match on format() in their Deserialize types to handle different serialized shapes per format.

Whether it carries its weight.... dunno. My current vague take on it is that it may be more powerful to let people control serde declaratively external to the type itself. That would let them control serde on complex types out of their control, and would let them change shapes per format, or per anything really.

erickt:
@JohnHeitmann: To a certain extent, if end users want more control, they could manually implement `Serialize`, and have nearly as much control as they want. The problem is though is that it's a pain in the butt to write the `Deserialize` implementations. We really ought to come up with some mechanisms that can make that easier.

erickt:
I just pushed up #233, which removes the problematic `Serializer::format`, and adds support for variant renaming that was heavily inspired by you. I'm going to close this, and favor the other patch because it avoids the work you had to get it to work here.

PR #208, opened 2016-01-15
Allow to use 'skip_serialization_if_empty' attribute for Option's

target-san:
This simple trick allows to make one of serialization skipping attributes deprecated.
The only problem I've encountered is that I was unable to suppress import warning. Adding attribute text to function, to import, and constructing it as syntax tree proved useless.

### DIFF
--- a/serde_codegen/src/de.rs
+++ b/serde_codegen/src/de.rs
@@ -559,7 +559,6 @@ fn deserialize_item_enum(
                 let expr = builder.expr().str(variant.node.name);
                  attr::FieldAttrsBuilder::new(builder)
                     .name(expr)
-                    .default()
                     .build()
             })
             .collect(),


erickt:
I initially was leaning towards the approach I mentioned in #213, but I'm starting to come around to this approach. I like that we could have one obvious trait `serde_codegen` could generate code for, rather than just assuming that the methods `is_empty()` or `is_none()` exist. Could you update this patch to implement `Empty` for the other containers, and update `serde_codegen` to use this trait instead of assuming those methods exist?

serde_codegen/src/ser.rs
```diff
@@ -57,12 +57,43 @@ pub fn expand_derive_serialize(
     );
 
     let where_clause = &impl_generics.where_clause;
+    /*
```
erickt: You should remove this commented out code.

erickt:
The relevant location to change is [here](https://github.com/serde-rs/serde/blob/master/serde_codegen/src/ser.rs#L600-L606). I bet you could just replace it with something like:

```rust
            let stmt = if field.skip_serializing_field_if_empty() {
                quote_stmt!(cx, if ::serde::ser::empty::Empty::is_empty(&$value_expr) { continue; })
            } else if field.skip_serializing_field_if_none() {
                quote_stmt!(cx, if ($value_expr).is_none() { continue; })
            } else {
                quote_stmt!(cx, {})
            };
```

We'll have to figure out how to deprecate `#[serde(skip_serializing_field_if_none)]`, although 0.7.0 is already on it's way to have a number of breaking changes.

serde_codegen/src/ser.rs
target-san: It wasn't intended for immediate merge. I left those two commented blocks as examples of how I attempted to add warning suppression attributes, in hopes someone would give good advice how to do this properly.

erickt: :) [this](https://github.com/serde-rs/serde/pull/208#issuecomment-172601055) should help with that.

target-san:
Using single explicit trait would require everyone to explicitly implement it. Wouldn't that be a problem?

target-san:
I accidentally closed this one by moving commits to separate branch. Continued in #214

PR #214, opened 2016-01-18
Unify 'skip_serialization_if_empty' and 'skip_serialization_if_none' via trait

target-san:
It seems I accidentally closed #208 by moving everything to separate branch


target-san:
We cannot cover standard collections with Empty trait. HashMap cannot be extended with custom traits on stable, because its third generic arg is unstable both in its bound and default.

oli-obk:
You don't need to specify the third arg, you can just leave it at its default:

```rust
use std::collections::HashMap;
use std::hash::Hash;

trait Dummy {
    fn is_cake(&self) -> bool;
}

impl<K: Eq + Hash, V> Dummy for HashMap<K, V> {
    fn is_cake(&self) -> bool { self.is_empty() }
}

fn main() {
}
```

target-san:
Thanks, didn't know this

target-san:
I've stumbled upon pretty stupid problem when implementing explicit trait Empty.
The problem is, `::serde::ser::empty::Empty::is_empty($value_type)` can support only either reference field or value field. But not both. Because `::serde::ser::empty::Empty::is_empty(&$value_type)` tries to pass &&-refernce when supplied &-reference. Using ` $value_type as Empty` also doesn't do any good, for the same reason. I guess our only option is to use duck-typing.

erickt:
This was alternatively implemented in #238. Thanks though!

Issue #57, opened 2015-04-15
`Path` serialization can panic


tbu-:
The `Serialize` implementation of `Path` incorrectly assumes that all paths can be represented as Unicode strings.

erickt:
Good point! Should be simple enough to have it return a syntax error if it gets a non-unicode string.

erickt:
err, rather tries to serialize to a non-unicode string. So it could either just be serialized as a byte string, or we first try to encode it as a string, then fail back to a byte string if it's non-unicode.

tbu-:
What should happen if you serialize the path `\xff` on Linux and try to deserialize that on Windows? Similarly, what happens if you serialize a lone surrogate on Windows and try to deserialize that on Linux?

Actually, what about path delimiters? :( Should serialization/deserialization even work across systems? What happens for `usize`-incompatiblities?

erickt:
@tbu-: ah the madness of cross platform things. This is one area where I'm looking forward to #198, which eases overriding which serializer gets used for a field. That'd allow someone with more specific knowledge than serde about how something would get used.

For a default, I'd just rather have serde be able to serialize and deserialize the value and get the same results. It's more up to the serializers/deserializers/users to decide if they need to conform the serialized format to some well defined schema.

erickt:
This was implemented in #224 for version 0.7.

PR #215, opened 2016-01-19
Use deserializer.deserialize_string for PathBuf

TyOverby:
It looks like PathBuf always serializes as a string, so why not skip the indirection?

The motivation for this is [this bug](https://github.com/TyOverby/bincode/issues/53) in bincode.


erickt:
You know, we probably should instead be serializing paths as a `&[u8]` instead of as a `String`. Or at least falling back to bytes if the `PathBuf` includes non-utf-8 characters.

erickt:
Previous comment really wants #57. In the meantime this patch makes sense with the current implementation.

PR #221, opened 2016-01-23
Bump clippy dependency to compile on 1.8.

SimonPersson:
The build fails on rustc 1.8 as clippy 0.0.36 no longer compiles. This bumps up the version of clippy to make it compile on nightly again.


thibault-ml:
Thanks for that, that's just what I needed :-)

erickt:
I pulled in this patch by way of the v0.6.x branch, and pushed up v0.6.11. Thanks!

PR #222, opened 2016-01-27
Fixes for README

jwilm:
Two issues have been fixed in README.md.

- Examples with `serializer` and `deserializer` were incorrectly calling `visit_*` methods
- Rustdoc must have moved at some point because all the documentation links were broken. These were updated.

The former issue was super confusing when trying to figure out how the internals work.


jwilm:
I'm actually quite confused now having read through parts of JSON serde. The `Serializer` and `Deserializer` traits specify a number of `serialize_*` and `deserialize_*` methods. Implementations for `serde::Serializer` and `serde::Deserializer` in JSON include a bunch of `visit_*` methods. These are not actually part of the trait, yet [there they are](https://github.com/serde-rs/json/blob/master/json/src/de.rs#L524-L534). What am I missing here? I assume at this point that my first commit made the documentation incorrect instead of fixing it, but I have no idea why.

jwilm:
It looks like the traits have been changed from having `visit_` methods to `de/serialize_` methods since the documentation was updated or since serde-json or serde-rmp have been updated. I resubmit that my first commit is actually a fix :).

erickt:
Hey @jwilm, sorry about all that. The `master` branch has some features on it that haven't been released yet. All of the serde ecosystem is still tracking the `0.6.x` branch. That branch still has those `visit_` method. I just haven't had time to push this `0.7.x` over the last few hurdles. Thanks for fixing these links!

PR #228, opened 2016-02-06
fix(clippy): make clippy optional

compressed:
I noticed when building `serde_macros` that it was pulling in `clippy` as required. It seems like this should be an optional dependency here too, like in the other serde crates.

Also, I noticed setting the version such as `clippy = "^0.0.37"` will require updates as clippy/nightly break. If desired, I can submit another PR to change the clippy dependency to `clippy = "^0.0"` to allow all future `0.0.*` clippy versions (I can do this across all the serde crates). This way you won't have to change the `Cargo.toml` when another clippy version is released.

Let me know. Thanks!

### DIFF
--- a/serde_macros/Cargo.toml
+++ b/serde_macros/Cargo.toml
@@ -13,7 +13,7 @@ name = "serde_macros"
 plugin = true
 
 [dependencies]
-clippy = "^0.0.37"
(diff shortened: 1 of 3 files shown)


erickt:
Thanks! For now I'd rather track clippy by hand, just to be safe.

PR #233, opened 2016-02-08
Remove support for format-specific renames, replace with ser or de specific renames

### LINKED ISSUE #211: Consider removing format specific renames
cc @hugoduncan.

I did a quick search through github, and I couldn't find anyone using `#[serde(rename(json=..., xml=...))]` and etc. I'm not sure if it's pulling it's weight and I'm considering removing it in the upcoming 0.7.0 release. This is also adding some complication in #207. Does anyone have any arguments towards keeping it?

erickt:
This effectively reverts #69, and restores the functionality added in #62. This feature that allows format-specific renames has never been used, and it's complicating the implementation of #207.

According to #61, if one uses serde to serialize requests that pass url-encoded parameters to a server, it might get responses back with a different capitalization scheme. This patch restores the behavior implemented in #62.

This closes #211.

### DIFF
--- a/serde/src/de/mod.rs
+++ b/serde/src/de/mod.rs
@@ -424,15 +424,6 @@ pub trait Deserializer {
     {
         self.deserialize(visitor)
     }
-
-    /// Specify a format string for the deserializer.
-    ///
-    /// The deserializer format is used to determine which format
-    /// specific field attributes should be used with the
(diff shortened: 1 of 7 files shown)


erickt:
This also adds support for renaming containers and variants.

Closes #207, #182, #195

Issue #216, opened 2016-01-19
Consider adding support for #[serde(default = $expr)]


erickt:
There is a bit of a dance that happens when deserializing a struct, and one of the fields is missing. `serde_codegen` effectively does:

* Was the field attributed with `#[serde(default)]`? If so, set the value `Default::default()`, assuming that the field type actually implements `Default`.
* Otherwise, call the `Deserializer`'s `missing_field()` method. If the format has some built-in support for a value for a missing field, such as JSON considering a missing value as a `null` value, try to deserialize from that value, or return an error.

This unfortunately means that as a consumer, you may have to newtype yourself an implementation of `Default`. Instead, it might be better if we optionally allowed for `#[serde(default = $expr)]`, where if the value is not present, we evaluate some expression. This would cut down on some of the boilerplate necessary for using `serde_codegen`.

Open question: If we do implement this, is this expression being evaluated in the same scope as `Deserialize::deserialize`? On one level it'd be potentially interesting that it'd be able to access one of the other values to implement itself, but I consider the `serde_codegen` generated bodies to not be a stable interface. So I'm leaning towards us putting the `$expr` into a sub-function to keep it from accessing the internal deserialization state.

cc #198, which has a similar idea for specifying alternative serializers/deserializers.

erickt:
This was implemented in #238!

Issue #90, opened 2015-06-15
deserialize: flexible defaults for missing struct fields


anguslees:
Currently I have to define a newtype (eg: "BoolDefaultTrue") every time I want an optional struct field to have a non-Default::default() default, which is clumsy.  Ideally I could just say something like:

```rust
struct Foo {
    #[serde(default=true)]
    enabled: bool,
}
```

... and this would fill in the value `true` if it was missing from the serialised string.  Super-awesome would be if serde also generated a `Default` trait implementation for `Foo` that respected the same values, but I could understand that being considered out of scope.

oli-obk:
If you create a manual `Default` impl for `Foo`, serde still uses the `Default` impl of the fields` type.

Example: this impl doesn't cause the `#[serde(default)]` to set the field to `true`.

```rust
impl Default for Foo {
    fn default() -> Foo {
        Foo {
            enabled: true,
        }
    }
}
```

@erickt: is using the surrounding type's `Default` impl (if present) too much magic? Alternatively requiring that the type impls `Default` if any field is annotated would be a breaking change and maybe not what one expects when annotating a field.

erickt:
This was implemented in #238!

dtolnay:
Yep, this is now doable with the `#[serde(default="...")]` attribute. I think this issue can be closed.

PR #237, opened 2016-02-16
Add codegen expr magic for default, skip serializing, and alternative serializers

erickt:
This is an alternative implementation of #214, #198, #208 and implements #90 and #216. This allows one to write:

```rust
struct Struct {
    #[serde(default="123")]
    a: i32,
    #[serde(skip_serializing_if="self.b == 123")]
    b: i32,
    #[serde(serialize_with="(self.b == 123).serialize(serializer)"]
    c: i32,
    #[serde(deserialize_with="Ok(if try!(bool::deserialize(deserializer) { 123) } else { 0 })"]
    d: i32,
}
```

cc @oli-obk, @target-san, @arcnmx


erickt:
arg, wrong branch.

erickt:
Pushed up on the right branch to https://github.com/serde-rs/serde/pull/238.

PR #240, opened 2016-02-21
Fix cargo `nightly` feature

kevinmehall:
517c2f79b70 renamed the `nightly` feature to `nightly-testing` to
reflect that the `clippy` dependency is only required when testing.
However, the code also uses `#[cfg(feature = "nightly")]` to enable
trait impls for feature-gated types. This commit restores that
functionality and fixes a few `cfg_attr`s that refer to clippy lints.


erickt:
Thanks!

PR #243, opened 2016-02-22
feat(de): Remove dependency on the `num` crate.

erickt:
Serde really should only be depending on the standard library. Other implementations should be done in upstream crates.


serde/src/de/from_primitive.rs
```diff
@@ -0,0 +1,406 @@
...
+use std::{usize, u8, u16, u32, u64};
+use std::{isize, i8, i16, i32, i64};
+use std::{f32, f64};
+use std::mem::size_of;
+
+/// Numbers which have upper and lower bounds
+pub trait Bounded {
+    // FIXME (#5527): These should be associated constants
+    /// returns the smallest finite number this type can represent
+    fn min_value() -> Self;
```
oli-obk: the standard library implements these https://doc.rust-lang.org/std/primitive.u16.html#method.min_value

oli-obk:
While I find this to be a good change, I'm skeptical about the `Deserialize` impls we have right now. If a `Deserializer` says "here have a u8", should the `Deserialize` impl really extend it to any other numeric type that can take the actual runtime-value? Out of scope for this PR, but just a thought too keep in mind.

serde/src/de/from_primitive.rs
erickt: Yeah, this was more of a short term hack to divorce us from `num` that'll need some cleaning up in the future.

erickt:
@oli-obk: The main usecase is to support formats like json that don't express all of the different integer types. It serializes `u8`s into `u64`s, so we need some mechanism to reverse the process. We could use the type hints, which might help us cut down on some casts, but that's more of an implementation detail.

PR #245, opened 2016-02-23
feat(serde): Add hooks for fixed-sized arrays

### LINKED ISSUE #244: Figure out if there are sensible hooks for serializing fixed length arrays
Fixed length arrays don't yet have a hook to hint that they're fixed length, so they're serialized like a generic sequence. It might be nice to add one for backends that might want to do something special with them.

One potential challenge here is how do we trust that the `Serialize` is actually serializing the right number of values, but we have the same challenge for tuples.

erickt:
Closes #244


oli-obk:
lgtm, I'll implement it for xml once this is merged

PR #248, opened 2016-02-23
Resolves #183 - Serialize and Deserialize for PhantomData

### LINKED ISSUE #183: Deserialize` is not implemented for the type `core::marker::PhantomData<_>`
This makes it difficult to deserialize structs with phantom types. Serialization also has the same problem but it seems to be avoidable by #[serde(skip_serializing)], so maybe it could be fixed by adding a #[serde(skip_deserializing)], or making skip_serializing also skip deserializing?

Yoric:
This should implement Serialize and Deserialize for PhantomData. I have not been able to find unit tests, so I can't be 100% sure, though.


Yoric:
Cc @erickt

PR #250, opened 2016-02-24
YAML serde based on yaml-rust

dtolnay:
Just opening this PR for discussion.

I noticed that [serde-rs/yaml](https://github.com/serde-rs/yaml/) seems pretty dead. The missing pieces are a YAML emitter (partially implemented) and parser (not started).

I found a reasonably complete YAML emitter and parser at [yaml-rust](https://crates.io/crates/yaml-rust) so I threw together [a serde that uses it](https://github.com/dtolnay/serde-yaml).

Is this something that would be valuable to contribute to serde-rs so that YAML is supported despite being theoretically slower than possible? Later we can swap in our own emitter and parser as the pieces materialize.

### DIFF
--- a/README.md
+++ b/README.md
@@ -704,10 +704,10 @@ Field Annotations:
 Serialization Formats Using Serde
 =================================
 
-| Format      | Name                                               |
-| ------      | ----                                               |
-| Bincode     | [bincode](https://crates.io/crates/bincode)        |
-| JSON        | [serde\_json](https://crates.io/crates/serde_json) |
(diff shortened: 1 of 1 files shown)


erickt:
Sure I'd be happy to link to your project! Have you also considered trying to merge your serializer into [yaml-rust](https://chyh1990.github.io/yaml-rust/)? That's what I've done in a few other places, like [toml](https://github.com/alexcrichton/toml-rs).

cc @Byron

PR #255, opened 2016-02-26
tests(readme): Use skeptic to test the readme

erickt:
This can't actually be merged in until serde_json 0.7 can be released.


dtolnay:
Why is this test added in serde_macros instead of serde_tests?

PR #258, opened 2016-02-27
Automatically publish docs from master branch

dtolnay:
For whatever reason it seems the doc-upload step has not been running recently - the last commit on gh-pages is from 6 months ago.

This is the solution I have been using for [serde-yaml](https://github.com/dtolnay/serde-yaml).

Here is what it looks like on my fork: https://github.com/dtolnay/serde/commits/gh-pages

Note that the encrypted ssh key in this PR is one that works for my fork. Before merging, one of the serde-rs owners will need to:

1. Generate an ssh key with no password:

        ssh-keygen -t rsa -b 4096 -C "docs@serde-rs.github.io"

2. Add the public key (id_rsa.pub) as a [deploy key for this repo](https://github.com/serde-rs/serde/settings/keys).

3. Encrypt the private key:

        travis encrypt-file /path/to/id_rsa

4. Update my PR to replace upload_key.enc with the one for this repo.

5. Update my PR to replace SSH_KEY_TRAVIS_ID with the one given by `travis encrypt-file`.


erickt:
I think I might have fixed it. I regenerated the GH_TOKEN according to http://www.hoverbear.org/2015/03/07/rust-travis-github-pages/#givingtravispermissions, so hopefully it'll get it updated. Thanks!

Issue #259, opened 2016-02-28
#[serde(deserialize_with=...)] should not require a field to implement Deserialize


erickt:
Unfortunately there's a gap in the current implementation of `#[serde(deserialize_with="...")]`. First, it's not getting used in the `Visitor::visit_seq` method, and two, it's interacting poorly with `serde::de::MapVisitor::missing_field`, which is called when a value is missing. This is necessary for formats like json, which typically interpret missing fields as a `null` value.

dtolnay:
> First, it's not getting used in the `Visitor::visit_seq` method

I fixed this in https://github.com/serde-rs/serde/pull/285/commits/886670134a5382bc181cd44b93592712e7c01076#diff-34bac2079ee371621845dc9670e2e20fR521.

> two, it's interacting poorly with `serde::de::MapVisitor::missing_field`

@erickt if you have any thoughts of how to resolve this, I have some time to implement it.

dtolnay:
I have one possible solution to this in https://github.com/dtolnay/serde/pull/1 (it conflicts with #308 so I am holding off on making a PR here).

The solution is, if a field is missing:

1. If there is a default expression, use the default expression (this includes `skip_deserializing`).
2. If the field has `deserialize_with`, fail with `Visitor::Error::missing_field(name)`.
3. Otherwise call `visitor.missing_field(name)` to possibly handle the field.

This means `deserialize_with` no longer goes through the visitor's `missing_field`, instead failing directly unless there is a default. Thus fields with `deserialize_with` do not require `Deserialize` as a bound. The missing field behavior can be controlled by setting a default for the field.

dtolnay:
Fixed in #311, can be closed.

Issue #197, opened 2015-12-31
skip_serializing does not inhibit generic bounds


arcnmx:
The following will fail because the `derive` will automatically add a `T: Serialize/Deserialize` bound despite the attributes. It would be nice if the codegen were smart enough to figure out the bound isn't necessary.

Also note that `default` is not quite correct in this test case, as it doesn't imply that the type should never be deserialized. A `skip_deserializing` attribute should also exist (see also #183).

Also related: the `: Default` bound should not be necessary, and automatically added to the derived `impl Deserialize` in the same manner.

```rust
#![feature(plugin, custom_derive, custom_attribute)]
#![plugin(serde_macros)]

extern crate serde;

#[derive(Serialize, Deserialize)]
struct A<T: Default> {
	#[serde(skip_serializing, default)]
	t: T,
}

#[derive(Default)]
struct B;

fn main() {
	fn assert<T: serde::Serialize + serde::Deserialize>(_: T) { }

	assert(A { t: B })
}
```

erickt:
@arcnmx: `#[serde(skip_deserializing)]` doesn't make sense since it'd be impossible to create the structure. It would be neat if we could skip adding bounds if we know need them.

arcnmx:
@erickt the difference between `default` and `skip_deserializing` is that `default` still implies that type will be deserialized if possible. `skip_deserializing` would mean to *never even try*, just unconditionally initialize it from a given or default value and don't look for the key at all, etc.

arcnmx:
(and would therefore mean a `T: Deserialize` bound is unneeded and wrong etc.)

erickt:
Sure, that makes sense. Bumping this off the 0.7.0 release since it could be implemented backwards compatibly.

dtolnay:
Using the changes in #285, the code above now compiles if you replace `#[serde(skip_serializing, default)]` with `#[serde(skip_serializing, skip_deserializing)]` (which is I think how @arcnmx intended for it to work).

arcnmx:
@dtolnay awesome work, looks like that will indeed kill this issue off for good! I'd add that perhaps it's worth discussing whether `#[serde(skip_deserializing, default)]` should be required for consistency(ish) reasons, but I don't have a personal opinion on that anyway.

dtolnay:
Thanks for bringing that up. My opinion is no - `skip_deserializing` should imply `default` unless you specify a different default with `default=...`.

dtolnay:
[...]

PR #260, opened 2016-02-29
feat(codegen): Inhibit generic bounds if skip_serializing

dtolnay:
Addresses part of #197.

The generated code for a struct like:

    struct Test<A, B, C> {
        a: X<A>,
        #[serde(skip_serializing)]
        b: B,
        #[serde(serialize_with="...")]
        c: C,
    }

Used to be:

    impl<A, B, C> Serialize for Test<A, B, C>
        where A: Serialize,
              B: Serialize,
              C: Serialize,
    { ... }

Now it is:

    impl<A, B, C> Serialize for Test<A, B, C>
        where X<A>: Serialize,
    { ... }

Both `skip_serializing` and `serialize_with` mean the type does not need to implement `Serialize`.

The remaining work for #197 is to add a `skip_deserializing` attribute that does the equivalent for `Deserialize` (also filed as #259), and to infer `T: Default` appropriately.

### DIFF
--- a/serde_codegen/src/attr.rs
+++ b/serde_codegen/src/attr.rs
@@ -382,7 +382,7 @@ fn get_renames(cx: &ExtCtxt,
     Ok((ser_name, de_name))
 }
 
-fn get_serde_meta_items(attr: &ast::Attribute) -> Option<&[P<ast::MetaItem>]> {
+pub fn get_serde_meta_items(attr: &ast::Attribute) -> Option<&[P<ast::MetaItem>]> {
     match attr.node.value.node {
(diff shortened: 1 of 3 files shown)


dtolnay:
It looks like this works on beta and nightly but not stable because it depends on https://github.com/rust-lang/rust/pull/30389. I will need to either omit bounds for types that do not involve type parameters, or revert to the old behavior on <1.7.0.

dtolnay:
I have added additional filtering to make this work with <1.7.0.

erickt:
@dtolnay: This was awesome, thanks!

PR #266, opened 2016-03-09
add clippy badge

oli-obk:
~~Not very useful as just a badge, but a great way to find odd things in our code.~~

We already run against clippy in travis... Just a badge then... Who doesn't like badges?


erickt:
Badges for everyone! Thanks!

PR #273, opened 2016-03-30
Silence unused_variables warning with empty enums

nox:
Deriving Serialize or Deserialize on Void-like enums triggered an unused_variables warning.

### DIFF
--- a/serde_codegen/src/ser.rs
+++ b/serde_codegen/src/ser.rs
@@ -80,7 +80,7 @@ fn serialize_item(
 
     Ok(quote_item!(cx,
         impl $impl_generics ::serde::ser::Serialize for $ty $where_clause {
-            fn serialize<__S>(&self, serializer: &mut __S) -> ::std::result::Result<(), __S::Error>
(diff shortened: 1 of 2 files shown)


erickt:
Thanks! Looks good. Can you add a test for this?

nox:
Where? And do we even have compile tests?

oli-obk:
You can add a normal test and add `#![deny(unused_variables)]`

nox:
Oh right, good call.

nox:
@oli-obk Done.

oli-obk:
I'm not sure if that test does what it's supposed to (you can test if it errors before your changes, then it's fine). I suggest to add the `deny` with a `!` to the entire module to make sure nothing in the tests ever produces unused variables.

nox:
A  `deny` on the whole module without any specific type to test this can easily not test what really needs to be tested. And I would still need to allow `dead_code` because that empty enum will still not be used.

As for your first remark, that's exactly what triggered the warning in some code I wrote.

oli-obk:
> As for your first remark, that's exactly what triggered the warning in some code I wrote.

ah, wonderful

PR #274, opened 2016-03-30
pass the renamed `deserialize` field name to missing_field

oli-obk:
I don't think we should ever access the rust-name directly

needed to port serde-xml to 0.7

### DIFF
--- a/serde_codegen/src/attr.rs
+++ b/serde_codegen/src/attr.rs
@@ -30,11 +30,6 @@ impl Name {
         }
     }
 
-    /// Return the string expression of the field ident.
-    pub fn ident_expr(&self) -> P<ast::Expr> {
-        AstBuilder::new().expr().str(self.ident)
-    }
-
     /// Return the container name for the container when serializing.
     pub fn serialize_name(&self) -> InternedString {
         match self.serialize_name {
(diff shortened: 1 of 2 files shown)


oli-obk:
ping @erickt

erickt:
Heh, I *literally* just pulled this PR to check it out on my train into work :)

oli-obk:
did you get a chance to look at it?

franklinhu:
Is it possible to add a unit test to make sure this doesn't happen in the future? Thanks!

oli-obk:
> Is it possible to add a unit test to make sure this doesn't happen in the future?

Good idea. Regression test has been added.

erickt:
@oli-obk: Overall this looks good. Unfortunately this is a breaking change, so we're going to have to spin this out in v0.8 in case anyone was relying on the old behavior :(

oli-obk:
I'd say it's a bugfix. I don't see how you can do anything useful with the original field name. It's not like anyone but deserializer-implementors has access to `missing_field`, and I highly doubt anyone is relying on error messages.

erickt:
@oli-obk: fair enough. I just used github's code search and couldn't find any users of it :)

PR #280, opened 2016-04-07
Fix documentation link

bombless:


dtolnay:
I fixed both in #264.

bombless:
Thanks @dtolnay , do you know why these docs are missing? https://serde-rs.github.io/serde/serde_codegen/serde_codegen/index.html

dtolnay:
`serde_codegen` is the implementation of `#[derive(Serialize, Deserialize)]`. There used to be rustdoc for it, but the truth is the actual structs and functions that are behind the implementation are not interesting to document. The public API is really just `#[derive(Serialize, Deserialize)]` + the [annotations](https://github.com/serde-rs/serde#annotations).

Issue #287, opened 2016-04-14
Cannot deserialize into Box<[T]>


alexcrichton:
I think maybe there's a missing impl for this?

e.g. https://github.com/rust-lang-nursery/rustc-serialize/blob/master/src/serialize.rs#L422-L427

erickt:
Landed it!

PR #290, opened 2016-04-18
fix(impls): Deserialize impl for Box<[T]>

dtolnay:
+tests for serializing Box<[T]>, which already worked. Addresses #287.


alexbool:
Could you please make a release on crates.io with this feature?

dtolnay:
@erickt or @pcwalton will need to do that. They are the owners on crates.io.

Issue #59, opened 2015-04-17
visiting a struct field/map key twice silently overwrites the previous field value


oli-obk:
```json
{
    "inner": [ {
        "a": null,
        "b": 2, 
        "c": ["abc", "xyz"]
    } ],
    "inner": [ {
        "a": null,
        "b": 42,
        "c": ["cake", "lie"]
    } ]
}
```

parses to

```rust
Outer {
    inner: vec![
        Inner { a: (), b: 42, c: vec!["cake".to_string(), "lie".to_string()] }
    ]
}
```

without any troubles. For multimaps that is alright, as it would get appended, but normal maps and structs should error out.

dtolnay:
Fixed in #293, can be closed.

PR #285, opened 2016-04-13
feat(codegen): Infer Default and Deserialize bounds correctly

dtolnay:
This PR addresses the remaining part of #197 (following up on my previous PRs #260 and #265) by inferring a `T: Default` bound for fields that (1) have `#[serde(default)]`, or (2) have `#[serde(skip_deserializing)]` and not `#[serde(default=...)]`. I factored out the generics-building code from #260 to share it between ser and de.

This PR also makes progress toward #259, although I have not attempted to resolve the `MapVisitor::missing_field` problem mentioned in that issue.

I would like to add more tests before merging, but the implementation is ready for review.

PR #293, opened 2016-04-19
feat(codegen): Detect repeated struct field when deserializing

dtolnay:
Addresses #59. Let me know whether you think we need an escape hatch to opt out of this check.

### DIFF
--- a/serde_tests/tests/token.rs
+++ b/serde_tests/tests/token.rs
@@ -410,6 +410,7 @@ pub enum Error {
     UnknownFieldError(String),
     UnknownVariantError(String),
     MissingFieldError(&'static str),
+    DuplicateFieldError(&'static str),
     InvalidName(&'static str),
     InvalidValue(String),
     UnexpectedToken(Token<'static>),
@@ -429,6 +430,10 @@ impl de::Error for Error {
 
     fn end_of_stream() -> Error { Error::EndOfStreamError }
 
+    fn invalid_value(msg: &str) -> Error {
+        Error::InvalidValue(msg.to_owned())
+    }
+
     fn unknown_field(field: &str) -> Error {
         Error::UnknownFieldError(field.to_owned())
     }
@@ -439,6 +444,10 @@ impl de::Error for Error {
 
     fn missing_field(field: &'static str) -> Error {
         Error::MissingFieldError(field)
+    }
+
+    fn duplicate_field(field: &'static str) -> Error {
+        Error::DuplicateFieldError(field)
     }
 }
 
--- a/serde/src/de/mod.rs
+++ b/serde/src/de/mod.rs
@@ -46,6 +46,12 @@ pub trait Error: Sized + error::Error {
     fn missing_field(field: &'static str) -> Self {
         Error::custom(format!("Missing field `{}`", field))
     }
+
+    /// Raised when a `Deserialize` struct type received more than one of the
+    /// same struct field.
+    fn duplicate_field(field: &'static str) -> Self {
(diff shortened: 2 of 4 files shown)


oli-obk:
> Let me know whether you think we need an escape hatch to opt out of this check.

I don't think we need one, but technically this is a breaking change. I wonder if anyone relies on it. Maybe toml?

dtolnay:
@alexcrichton any idea whether the TOML serde relies on the behavior that the last value for a particular struct field overrides any previous values for the same struct field? Using JSON syntax, this PR proposes that deserializing `{"a": 1, "a": 2}` should be an error rather than `MyStruct { a: 2 }`.

erickt:
cc @alexcrichton on the toml question. Would this cause you problems?

alexcrichton:
Ah yeah I don't think that this is a problem for TOML, the format already disallows duplicate keys in a table, so it's actually impossible to have a TOML document with two key/value pairs that have the same key.

So sounds good to me!

dtolnay:
I have resolved the merge conflict with #285.

dtolnay:
Because of the `use ::serde::de::Error` there is a semantic merge conflict with #298 so one of these PRs will break the other. It looks like @erickt has reviewed #298 already so let's merge that first and then @erickt or I can rebase this one.

dtolnay:
I am debugging why this does not compile after the merge commit.

dtolnay:
I narrowed this down to the following difference. This compiles:

```rust
extern crate serde;
use serde::de::Error;
fn main() {}
```

This does not compile:

```rust
fn main() {
    extern crate serde;
    use serde::de::Error;
}
```

The compiler diagnostic says "Did you mean \`self::serde::de\`?"

If I change it to `self::serde::de`, it still fails to compile and the diagnostic says "Did you mean \`self::self::serde::de\`?"

If I change it to `self::self::serde::de` (which is nonsensical), the diagnostic says "Maybe a missing \`extern crate self\`?"

Has anyone seen this before or know a workaround?

dtolnay:
Resolved with the help of [this thread](https://users.rust-lang.org/t/use-trait-from-crate-in-the-same-lexical-scope/5671) on the users forum. The remaining failure is related to [quasi](https://travis-ci.org/serde-rs/serde/jobs/127422523) not this PR. Ready for another review @oli-obk or @erickt.

serde_tests/tests/token.rs
```diff
@@ -429,6 +429,10 @@ impl de::Error for Error {
 
     fn end_of_stream() -> Error { Error::EndOfStreamError }
 
+    fn invalid_value(msg: &str) -> Error {
+        Error::InvalidValue(msg.to_owned())
```
oli-obk: I think `de::Error` deserves a new method `duplicate_field`. We already have `unknown_field`. The default impl could forward to `invalid_value` for now, and we can add an issue for 0.8 to remove that default impl.

dtolnay: Sounds reasonable. I pushed the change. Why do you say we need to get rid of the default impl in 0.8?

oli-obk: Because I was confused on how Error is implemented ;)

dtolnay:
The Travis failure is due to clippy.

dtolnay:
I pushed a merge commit to fix the conflict with #308.

dtolnay:

dtolnay:
Rebased, per https://github.com/serde-rs/serde/pull/315#issuecomment-218154729.

oli-obk:
@homu r+

Issue #159, opened 2015-09-22
serde_macros incompatible with `extern crate serde` not in toplevel module


whitequark:
To reproduce:
```rust
#![plugin(serde_macros)]
#![feature(plugin, custom_derive)]

mod x {
    extern crate serde;
    #[derive(Serialize)] struct S;
}

fn main() {}
```

erickt:
Unfortunately I'm not sure if there's much we can do in this circumstance. I'm not sure if it's possible to resolve the location of the serde crate, so the most reliable location to find it is in the toplevel crate namespace.

whitequark:
Hm, good point. This might be worth bringing up whenever custom derivers are being stabilized.

erickt:
It's worse than that, I think we need name resolution to happen in order to find where the serde crate was imported, which may not get exposed to compiler plugins for a long time :(

whitequark:
What if you did something like expanding it to...
```rust
mod internal {
  extern crate serde;
  impl serde::whatever for dunno {}
}
use internal::*;
```
but cleaner? Basically, hygiene for crate imports.

grncdr:
hm, am I correct in understanding that `#[derive(Serialize)]` essentially won't work anywhere outside of the top-level of a crate? I was trying to derive these traits in a child module (e.g. separate file) and it seemed impossible to convince the compiler that the implementation was there.

dtolnay:
@grncdr 

> hm, am I correct in understanding that `#[derive(Serialize)]` essentially won't work anywhere outside of the top-level of a crate? I was trying to derive these traits in a child module (e.g. separate file) and it seemed impossible to convince the compiler that the implementation was there.

It should work as long as `extern crate serde` is at the top level of your crate. The reason is the generated code expects to be able to use absolute paths like `::serde::ser::Serialize`.

The following works for me:

```rust
#![feature(plugin, custom_derive)]
#![plugin(serde_macros)]

extern crate serde;

mod m {
    #[derive(Serialize)]
    pub struct Success;
}

fn main() {
    fn assert<T: serde::Serialize>(_: T) {}
    assert(m::Success);
}
```

grncdr:
Thanks @dtolnay

dtolnay:
In line with @whitequark's idea, I started implementing this as:

```rust
mod m {
    struct A { ... }
    #[derive(Serialize)]
    struct B { a: A }
}

// -->

mod m {
    struct A { ... }
    struct B { a: A }
[...]

PR #298, opened 2016-04-24
fix(codegen): Support `extern crate serde` not in toplevel module

dtolnay:
Addresses #159.

The generated code looks like:

```rust
#[allow(non_upper_case_globals, unused_attributes, unused_qualifications)]
const _IMPL_DESERIALIZE_FOR_MyStruct: () = {
    extern crate serde as _serde;
    #[automatically_derived]
    impl _serde::de::Deserialize for MyStruct {
        // ...
    }
};
```

### DIFF
--- a/serde_tests/tests/test.rs
+++ b/serde_tests/tests/test.rs
@@ -1,6 +1,4 @@
 #![cfg_attr(feature = "nightly", feature(plugin))]
 #![cfg_attr(feature = "nightly", plugin(clippy))]
 
-extern crate serde;
-
 include!(concat!(env!("OUT_DIR"), "/test.rs"));
--- a/README.md
+++ b/README.md
@@ -46,7 +46,6 @@ serde_macros = "*"
 #![feature(custom_derive, plugin)]
 #![plugin(serde_macros)]
 
-extern crate serde;
 extern crate serde_json;
 
 #[derive(Serialize, Deserialize, Debug)]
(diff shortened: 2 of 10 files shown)


serde_tests/tests/test.rs
```diff
@@ -1,6 +1,4 @@
 #![cfg_attr(feature = "nightly", feature(plugin))]
 #![cfg_attr(feature = "nightly", plugin(clippy))]
 
-extern crate serde;
```
oli-obk: why was this removed? Did it cause an error?

oli-obk:
Does this break any existing code? You changed quite a few tests' imports.

serde_tests/tests/test.rs
dtolnay: No error, just to demonstrate that things work even without "extern crate serde" at the top level. I kept the equivalent line in [bench.rs](https://github.com/dtolnay/serde/blob/5eb1d7dd4b9d5aa19599be10771fb020a39c46a3/serde_tests/benches/bench.rs#L6) so the build will still uses both approaches.

dtolnay:
This does not break any existing code.

dtolnay:
This makes `extern crate serde` optional unless you are using it explicitly in your own code, so I have removed it from the example in the readme.

erickt:
@dtolnay: Once again, great work here. I'm working on merging this patch in by hand.

dtolnay:
@erickt I fixed the merge conflict myself.

dtolnay:
Ping @erickt - looks like you said this was ready to merge. Once you merge it, I will resolve the merge conflict in https://github.com/serde-rs/serde/pull/293#issuecomment-214442429.

erickt:
Thanks @dtolnay! I've been swamped with work recently, and unfortunately that's resulted in lots of my projects suffering. Thanks for keeping up with this.

Issue #163, opened 2015-09-28
CBOR support


pyfisch:
I have written a [CBOR](https://github.com/pyfisch/cbor) serialization and deserialization crate for serde. The conforms to the CBOR spec, is tested and documented. Is it possible to move it to the serde-rs organization? I would also really like to hear some code review.

Serde supports annotations to values to assign a special meaning. For example the tag 32 is used to annotate URIs. To support these tags within serde a method for `serde::de::Visitor` to for tags would be useful. The signature could be something like `fn visit_tag<D>(&mut self, tag: u64,  _deserializer: &mut D) -> Result<Self::Value, D::Error>` and default to ignoring the tag and just visit the inner value. There could be a similiar method for `ser::Serializer`.

erickt:
@pyfisch: I'd be fine if you want to have CBOR be a part of serde-rs. Have you and @burntsushi figured out if you can consolidate your projects?

On tags, I was struggling as well to figure out a good way to express them when I ported https://github.com/BurntSushi/rust-cbor over to serde. I did a terrible hack where I inspected the struct name and did something special with that, but I wasn't particularly happy with that approach. Having a `visit_tag` is interesting, but I do worry a little that it would be specific just to CBOR. Would you need an equivalent construct on the Serializer?

pyfisch:
I already compared my code to the implementation by @burntsushi. The code is very similiar but rust-cbor uses an intermediate syntax while serde-cbor directly deserializes the values. We did not yet discuss if we can cosolidate projects.

I would need `visit_tag` both for deserializer and serializer, but it would probably easier to start with just deserialization. Tags are not only used in CBOR, also [BSON](http://bsonspec.org/spec.html) has tags for binary data with values ranging from 0 to 255. Also [messagepack](https://github.com/msgpack/msgpack/blob/master/spec.md#extension-type) has extension types ranging from -128 to 127 while the negative codes are reserved. There are probably some more formats using a similiar extension mechanism. If the `visit_tag` method would take a string describing the format like "cbor", or "bson", and supported different numeric and string types as tags it could be generic over many formats.

BurntSushi:
I have clearly dropped the ball on using `serde` (in all my crates, not just `cbor`). The worst part is that I don't see myself having cycles to dive into it in the immediate future. Therefore, I shouldn't stand in the way of progress.

I have two suggestions:

[...]

PR #301, opened 2016-05-01
WIP: Parsing and emitting tagged values

pyfisch:
Some serialization formats like [CBOR](http://cbor.io/) or [BSON](http://bsonspec.org/) allow the definition of "tagged values" or "subtypes" to extend the format or to add additional type information to values. They can only supported with some help from serde.

This patch should add support to deserialize and serialize these tagged values. The default behavior will be to discard the tag and only use the value so they are completely optional to use.

For now the patch only supports serialization but I would like to get early feedback about the inclusion in serde.

See also #163 and pyfisch/cbor#3

### DIFF
--- a/serde/src/ser/mod.rs
+++ b/serde/src/ser/mod.rs
@@ -30,7 +30,7 @@ pub trait Serialize {
 ///////////////////////////////////////////////////////////////////////////////
 
 /// A trait that describes a type that can serialize a stream of values into the underlying format.
-pub trait Serializer {
+pub trait Serializer: Sized {
     /// The error type that can be returned if some error occurs during serialization.
     type Error: Error;
 
@@ -326,6 +326,23 @@ pub trait Serializer {
         where V: Serialize,
     {
         self.serialize_struct_elt(key, value)
+    }
+    
+    /// Serializes a tagged value.
+    ///
+    /// By default, the tag is discarded and the value is serialized as-is.
+    /// 
+    /// The tags are provided as a pair of format (like `cbor`) and a numeric
+    /// tag. It is possible to supply different tags for different formats.
+    /// The serialization library will select an appropriate tag if available
+    /// and otherwise will serialize the raw value.
+    #[inline]
+    fn serialize_tagged_value<V>(&mut self,
+                                 _tags: &[(&'static str, u64)],
+                                 value: V) -> Result<(), Self::Error>
+        where V: Serialize,
+    {
+        value.serialize(self)
     }
 }
 


serde/src/ser/mod.rs
```diff
@@ -327,6 +327,23 @@ pub trait Serializer {
...
         self.serialize_struct_elt(key, value)
     }
+    
+    /// Serializes a tagged value.
+    ///
+    /// By default, the tag is discarded and the value is serialized as-is.
+    /// 
+    /// The tags are provided as a pair of format (like `cbor`) and a numeric
+    /// tag. It is possible to supply different tags for different formats.
+    /// The serialization library will select an approbiate tag if available
```
oli-obk: Typo: Appropriate

oli-obk:
This means that types need to know about the formats they serialize to... This would break the current separation of serializers and serializees. I'm assuming you want the same type to yield different tag values depending on the serializer? Otherwise this could maybe be merged with some kind of general enum discriminant mapping. Then your tagged types would simply be an enum whose discriminant is converted to a specific type.

Another possibility might be to allow specialization to create different impls depending on the serializer, but I have not put any thought into whether that is feasible.

pyfisch:
The tag values are defined by the format. Some formats do not support tags at all and formats that support them will have different values for the same thing. For this reason I need to yield different tag values depending on the serializer.

In CBOR every application can define custom tags so any solution using an enum with a fixed list of types is unusable. Currently there are tags for URLs (tag 32) and UUIDs (tag 37) for example. Such types are useful for multiple formats. But one day a user might define a tag 6877 he uses in an application to mark color names in french language. The user must be able to serialize this specific tag. There is a list of [IANA registered tags](https://www.iana.org/assignments/cbor-tags/cbor-tags.xhtml).

I have not yet worked with specialization and don't know if how it could be used in this case.

dtolnay:
Would it be possible to generalize this to formats in which tags are not u64? [YAML tags are strings](http://www.yaml.org/spec/1.2/spec.html#id2761694).

pyfisch:
Probably I should generalize it. But I don't know how to do that. Because if I support two different formats (`u64` and `String`) I should also support `i8` (used by BSON but can be mapped to `u64`) and probably other formats.

oli-obk:
Well... If we can all agree that aggregate types make no sense here, we have a rather short list:

```rust
enum Tag<'a> {
    U64(u64), U32(u32), U16(u16), U8(u8),
    I64(i64), I32(i32), I16(i16), I8(i8),
    String(&'a str),
    Bytes(&'a [u8]),
    F64(f64), F32(f32),
}
```

Then the second tuple field of the tag arguments can simply be that enum, and everyone can decide what they want to do with it.

erickt:
I've thought about trying to implement something like this for cbor, because I had to do that annoying hack to get tags to work. I'm not crazy about that mapping from serializer to tag though, especially since it would result in linear scan of tag types. Have you considered other options?

I suppose one option could be a central registry of tag names that both the serializers and serializees use, but then that'd make it hard for people to create a custom type. The other option could be to just use a string, but that might make things non-portable.

However, tags are to a certain extent a non-portable construct. If I have, say, a `DateTime` object that I want to serialize to cbor and yaml (assuming they have custom), I would have to modify the impl of `DateTime` to support each of these backends to add in the right tag names for each serializer, so we might not actually need something super generic. Specialization might be the right approach here once it's stabilized.

pyfisch:
@oli-obk Also floats make no real sense since they are inexact and for this reason no good keys.

@erickt as I said before people will want to use custom types so this should be easy. An alternative to the linear scan could be a callback. A closure is passed to `serialize_tagged_value` that takes a single argument describing the used serialization format.

PR #302, opened 2016-05-01
Enable use in no_std environments

cmr:
These changes are fairly invasive to imports and uses of non-libcore types,
but allow for some or none of the freestanding crates (core, rustc_unicode,
alloc, collections) to be supported by serde.


cmr:
Note that this plain old inlines some code from `libstd` (or other rust crate) for some stuff, to avoid changing behavior in any of these configurations. Alas, behavior will still change for plain `no_std` without `collections`, due to lacking a `String` type for error handling. Otherwise, this is a fairly stock `no_std` conversion.

cmr:
One change that snuck into the non-`nightly` portion is that I inlined the definition of `encode_utf8` instead of doing an allocation, for `serialize_char`.

serde/src/de/impls.rs
```diff
@@ -1,30 +1,63 @@
...
+
+#[cfg(all(feature = "collections", not(feature = "std")))]
+use collections::{
     BinaryHeap,
     BTreeMap,
     BTreeSet,
     LinkedList,
+    VecDeque,
+    Vec,
+	String,
```
oli-obk: tab instead of spaces

cmr: That's what I get for lazy use of `:set paste`... will whack it with `rustfmt`.

serde/src/de/impls.rs
```diff
@@ -82,15 +141,39 @@ impl Visitor for BoolVisitor {
...
+        use rustc_unicode::str::UnicodeStr;
         match s.trim() {
             "true" => Ok(true),
             "false" => Ok(false),
             _ => Err(Error::invalid_type(Type::Bool)),
         }
     }
+
+    #[cfg(all(not(feature = "std"), not(feature = "rustc_unicode")))]
+    fn visit_str<E>(&mut self, s: &str) -> Result<bool, E>
```
oli-obk: why not just use this version? Is there an advantage to using the std version when it's there?

serde/src/de/impls.rs
```diff
@@ -151,13 +234,33 @@ impl<T> Visitor for PrimitiveVisitor<T>
     impl_deserialize_num_method!(f64, visit_f64, from_f64, Type::F64);
 
     #[inline]
+	#[cfg(feature = "std")]
```
oli-obk: tab instead of spaces

serde/src/de/impls.rs
cmr: less code size/duplication... at least that's the idea.

serde/src/de/value.rs
```diff
@@ -34,29 +59,60 @@ pub enum Error {
...
 
     /// EOF while deserializing a value.
     EndOfStream,
 
     /// Unknown variant in enum.
+    #[cfg(any(feature = "std", feature = "collections"))]
     UnknownVariant(String),
+    /// Unknown variant in enum.
+    #[cfg(all(not(feature = "std"), not(feature = "collections")))]
+    UnknownVariant(&'static str),
```
oli-obk: remove the `&'static str` field as it's always the same without allocations

cmr: Couldn't custom implementations use different values?

cmr:
Guh. Is there a way to `#[cfg]` based on Rust version? `1.5.0` didn't have a stable `libcore`, which this depends on (otherwise there is not a good way to not duplicate every import from `std`)

oli-obk:
Nope. I think the consensus is that someone who calls `cargo update` can also update rustc.

cmr:
So should the minimum of `1.5.0` for serde be bumped?

cmr:
I think the errors with the macros are not my fault.

oli-obk:
> I think the errors with the macros are not my fault.

nightly is broken right now

cmr:
Timeline for getting this merged? I need to decide if I should vendor this dep or wait a bit and work on other stuff.

serde/src/de/impls.rs
```diff
@@ -70,6 +103,32 @@ impl Deserialize for () {
 
 ///////////////////////////////////////////////////////////////////////////////
 
+#[cfg(all(not(feature = "std"), not(feature = "rustc_unicode")))]
```
erickt: Is this not available in one of the Unicode modules? If not, can you pull this Unicode stuff into a separate private module?

serde/src/de/impls.rs
```diff
@@ -91,6 +151,29 @@ impl Visitor for BoolVisitor {
             _ => Err(Error::invalid_type(Type::Bool)),
         }
     }
+
+    #[cfg(all(not(feature = "std"), feature = "rustc_unicode"))]
+    fn visit_str<E>(&mut self, s: &str) -> Result<bool, E>
```
erickt: This doesn't seem worthwhile if we need to bundle in our own trim implementation.

serde/src/de/impls.rs
```diff
@@ -151,13 +234,33 @@ impl<T> Visitor for PrimitiveVisitor<T>
...
             Err(Error::invalid_type(Type::Str))
         })
     }
+
+    #[cfg(all(not(feature = "std"), feature = "rustc_unicode"))]
+    fn visit_str<E>(&mut self, s: &str) -> Result<T, E>
+        where E: Error,
+    {
+        use rustc_unicode::str::UnicodeStr;
+        str::FromStr::from_str(s.trim()).or_else(|_| {
```
erickt: Here too.

serde/src/de/impls.rs
oli-obk: The outermost function is available in `rustc_unicode`: http://manishearth.github.io/rust-internals-docs/rustc_unicode/property/fn.Pattern_White_Space.html

serde/src/de/impls.rs
```diff
@@ -954,14 +1081,27 @@ impl<T, E> Deserialize for Result<T, E> where T: Deserialize, E: Deserialize {
...
                     }
 
+                    #[cfg(all(not(feature = "std"), not(feature = "collections")))]
+                    fn visit_usize<E>(&mut self, value: usize) -> Result<Field, E> where E: Error {
+                        #[cfg(feature = "collections")]
+                        use collections::string::ToString;
+                        match value {
+                            0 => Ok(Field::Ok),
+                            1 => Ok(Field::Err),
+                            _ => Err(Error::unknown_field("some number (no to_string)")),
```
erickt: You can get rid of the "no to_string" here.

serde/src/de/impls.rs
```diff
@@ -954,14 +1081,27 @@ impl<T, E> Deserialize for Result<T, E> where T: Deserialize, E: Deserialize {
...
+                        }
+                    }
```
erickt: Nit: add a newline here.

serde/src/de/mod.rs
```diff
@@ -555,12 +566,20 @@ pub trait Visitor {
...
+    /// `visit_char` deserializes a `char` into a `Value`.
+    #[inline]
+    #[cfg(feature = "nightly")]
     fn visit_char<E>(&mut self, v: char) -> Result<Self::Value, E>
         where E: Error,
     {
-        // FIXME: this allocation is required in order to be compatible with stable rust, which
-        // doesn't support encoding a `char` into a stack buffer.
-        self.visit_string(v.to_string())
+        self.visit_str(::core::str::from_utf8(v.encode_utf8().as_slice()).unwrap())
```
erickt: Unless I'm missing something, isn't this the same as the non-nightly version?

serde/src/lib.rs
```diff
@@ -10,23 +10,91 @@
...
 pub mod bytes;
 pub mod de;
+#[cfg(feature = "std")]
 pub mod iter;
 pub mod ser;
+#[cfg(not(feature = "std"))]
+pub mod error;
+
+// FIXME: remove this when encode_utf8 is stabilized.
+const TAG_CONT: u8    = 0b1000_0000;
```
erickt: This should be in a separate internal module.

serde/src/de/impls.rs
erickt: Yeah, but I'd prefer one of the stable Unicode modules like from https://github.com/unicode-rs so it could be used by stable rust. I'd like to avoid an implementation for stable rust and for nightly rust as much as possible.

.travis.yml
```diff
@@ -3,7 +3,7 @@ rust:
 - stable
 - beta
 - nightly
-- 1.5.0
+- 1.6.0
```
erickt: This might require a major bump unfortunately. I never got around to figuring out a good support story for older compilers, but someone did ask me to support 1.5 in serde (or syntex? I can't remember) but either way I don't want to break people that might depend on this.

README.md
```diff
@@ -719,6 +719,21 @@ Field Annotations:
...
 | `#[serde(deserialize_with="$path")]`    | Call a function `fn<T, D>(&mut D) -> Result<T, D::Error> where D: Deserializer` to deserialize this value  |
 
+Using in `no_std` crates
+========================
+
+The core `serde` package defines a number of features to enable usage in a
+variety of freestanding environments. Enable any or none of the following
+features, and use `default-features = false` in your `Cargo.toml`:
+
+- `rustc_unicode` (implies `nightly`)
```
erickt: Is it possible to remove this dependency on rustc_unicode?

serde/src/de/mod.rs
cmr: The non-nightly one uses a bundled version.

serde/src/de/impls.rs
cmr: This being "trimming the string"?

serde/src/de/impls.rs
cmr: Doesn't look like any of those crates define this (nor should they, IMO).

serde/src/lib.rs
cmr: Will-do.

cmr:
@erickt updated with the various style fixes. Additionally, I modified it to support `1.5.0` in the default feature set. No major feature bump needed :)

cmr:
Additionally, wherever something was bundled in optionally, I made the bundling permanent. So the whitespace trimming and UTF8 encoding is done in the same on every configuration.

erickt:
@cmr: looks good! I'm going to consider this an unstable feature though, so it might break in a minor update.

dtolnay:
This PR never passed in Travis. I opened #314 to revert it because I do not feel comfortable fixing it myself. @cmr please take a look.

PR #308, opened 2016-05-07
Reduce dependence on type inference

### LINKED ISSUE #305: Regression in serde_codegen 0.7.4
Today I have upgraded one of my projects, updating all the Serde dependencies to the latest versions.  My project can be found here: https://github.com/benashford/rs-es/tree/upgrade_serde_dependencies the upgrade is currently in a branch `upgrade_serde_dependencies`.

In the project I have many structs which use `#[derive(Deserialize)`, these were all working and tested using the previous version of serde_codegen, but since 0.7.4 it seems to be generating incorrect code.  But only for two of the structs.

What the two structs have in common is a type parameter, with a trait bound of `Deserialize` (i.e. the struct could contain any arbitrary struct which is also deserializeable).

The specific structs are: [SearchHitsHitsResult](https://github.com/benashford/rs-es/blob/upgrade_serde_dependencies/src/operations/search/mod.rs#L738-L755) and [GetResult](https://github.com/benashford/rs-es/blob/upgrade_serde_dependencies/src/operations/get.rs#L121-L135)

The error is repeated for every field in each struct:

```
src/operations/get.rs:122:1: 122:30 error: mismatched types:
 expected `std::string::String`,
    found `T`
(expected struct `std::string::String`,
    found type parameter) [E0308]
src/operations/get.rs:122 #[derive(Debug, Deserialize)]
                          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/operations/get.rs:122:1: 122:30 note: in this expansion of #[derive_Deserialize] (defined in src/operations/get.rs)
src/operations/get.rs:122:1: 122:30 note: in this expansion of include!
src/operations/get.rs:122:1: 122:30 help: run `rustc --explain E0308` to see a detailed explanation
```

And also, for each field in both structs:

```
src/operations/get.rs:122:1: 1:1 error: mismatched types:
 expected `T`,
    found `operations::get::_IMPL_DESERIALIZE_FOR_GetResult::<impl serde::Deserialize for operations::get::GetResult<T>>::deserialize::__Field`
(expected type parameter,
    found enum `operations::get::_IMPL_DESERIALIZE_FOR_GetResult::<impl serde::Deserialize for operations::get::GetResult<T>>::deserialize::__Field`) [E0308]
(internal compiler error: unprintable span)
src/operations/get.rs:122:1: 1:1 note: in this expansion of #[derive_Deserialize] (defined in src/operations/get.rs)
src/operations/get.rs:122:1: 1:1 note: in this expansion of include!
src/operations/get.rs:122:1: 1:1 help: run `rustc --explain E0308` to see a detailed explanation
```

All of the above is with Rust: `rustc 1.10.0-nightly (8da2bcac5 2016-04-28)` but it can also be seen with the stable and beta Rust's too.

The output is quite long, so I won't list it all here, but for reference it can be seen on the Travis CI log of my project: https://travis-ci.org/benashford/rs-es/builds/127847754

### LINKED ISSUE #307: Deserialize_with generates code that does not compile
```rust
#[derive(Deserialize)]
struct S<T> {
    t: T,
    #[serde(deserialize_with="dw")]
    d: i32,
}
```

The implementation of `deserialize_with` turns this into a field deserializer:

```rust
struct __SerdeDeserializeWithStruct<T> where T: Deserialize
    value: i32,
}
```

... which does not compile because the type parameter `T` is unused. It incorrectly uses all of the generic type parameters without filtering for ones relevant to the specific field.

Filtering is hard so a quick fix would be:

```rust
struct __SerdeDeserializeWithStruct<T> where T: Deserialize
    value: i32,
    phantom: PhantomData<S<T>>,
}
```

dtolnay:
Addresses both #305 and #307.

To fix #305, this PR adds turbofish to SeqVisitor and MapVisitor calls where type inference can fail due to https://github.com/rust-lang/rust/issues/33449.

To fix #307, this PR adds a PhantomData in the deserialize_with implementation as explained in that ticket.

I did both in one PR because they touch the same code.

### DIFF
--- a/serde_codegen/src/attr.rs
+++ b/serde_codegen/src/attr.rs
@@ -180,7 +180,7 @@ pub struct FieldAttrs {
     skip_serializing_field_if: Option<P<ast::Expr>>,
     default_expr_if_missing: Option<P<ast::Expr>>,
     serialize_with: Option<P<ast::Expr>>,
-    deserialize_with: P<ast::Expr>,
+    deserialize_with: Option<ast::Path>,
 }
 
 impl FieldAttrs {
@@ -197,8 +197,6 @@ impl FieldAttrs {
             None => { cx.span_bug(field.span, "struct field has no name?") }
         };
 
-        let identity = quote_expr!(cx, |x| x);
-
         let mut field_attrs = FieldAttrs {
             name: Name::new(field_ident),
             skip_serializing_field: false,
@@ -206,7 +204,7 @@ impl FieldAttrs {
             skip_serializing_field_if: None,
             default_expr_if_missing: None,
             serialize_with: None,
-            deserialize_with: identity,
+            deserialize_with: None,
         };
 
         for meta_items in field.attrs.iter().filter_map(get_serde_meta_items) {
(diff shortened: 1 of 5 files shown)


dtolnay:
The build failure is clippy.

PR #310, opened 2016-05-08
Clarify serialize_with and deserialize_with documentation

dtolnay:
The field type `T` does not have to be a generic type parameter of the serialize_with or deserialize_with function.

### DIFF
--- a/README.md
+++ b/README.md
@@ -705,18 +705,18 @@ Variant Annotations:
 
 Field Annotations:
 
-| Annotation                              | Function                                                                                                   |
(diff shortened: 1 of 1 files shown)


oli-obk:
There are a lot of unrelated whitespace-changes, are they necessary?

dtolnay:
I don't know whether they are necessary but they align the rightmost vertical line that delimits the table. I added characters to the longest line in the table so the other rows need spaces to keep the delimiters aligned.

oli-obk:
Ah, github displays them quite weirdly due to the long lines.

lgtm

PR #311, opened 2016-05-09
Field with deserialize_with should not implement Deserialize

dtolnay:
Addresses #259.

The solution is, if a field is missing:

- If there is a default expression, use the default expression (this includes `skip_deserializing`).
- If the field has deserialize_with, fail with `Visitor::Error::missing_field(name)`.
- Otherwise call `visitor.missing_field(name)` to possibly handle the field.

This means `deserialize_with` no longer goes through the visitor's `missing_field`, instead failing directly unless there is a default. Thus fields with `deserialize_with` do not require Deserialize as a bound. The missing field behavior can be controlled by setting a default for the field.

### DIFF
--- a/serde_codegen/src/attr.rs
+++ b/serde_codegen/src/attr.rs
@@ -308,22 +308,6 @@ impl FieldAttrs {
         &self.name
     }
 
-    /// Predicate for using a field's default value
-    pub fn expr_is_missing(&self) -> P<ast::Expr> {
-        match self.default_expr_if_missing {
-            Some(ref expr) => expr.clone(),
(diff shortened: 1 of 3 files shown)


dtolnay:

dtolnay:
Rebased, per https://github.com/serde-rs/serde/pull/315#issuecomment-218154729.

oli-obk:
took me a while to grok it, but the implementation is sound.

I'm fine with the design, and can't think of another one until specialization stabilizes (so `missing_field` can handle fields that don't implement `Deserialize`)

Issue #312, opened 2016-05-09
Deserialize_with generates bad code if there is a type parameter called D


dtolnay:
Both of these generate code that fails to compile:

```rust
#[derive(Deserialize)]
struct S<D> {
    #[serde(deserialize_with="de")]
    d: D
}
```

```rust
#[derive(Deserialize)]
struct S<D> {
    d: D,
    #[serde(deserialize_with="de")]
    i: i32,
}
```

```
error: type parameter `D` shadows another type parameter of the same name [E0194]
                                     fn deserialize<D>(__d:
                                                    ^
```

dtolnay:
Fixed in #315, can be closed.

oli-obk:
if you add `fixes #id` to the PR, or a commit, the issues will be closed automatically when the PR is merged

dtolnay:
Neat, I will do that in the future. Please also close #90 #141 #158.

Issue #313, opened 2016-05-09
`--features nightly-testing` failing


dtolnay:
```
error: unresolved import `core::nonzero::NonZero`. Could not find `nonzero` in `core` [E0432]
use core::nonzero::{NonZero, Zeroable};
                    ^~~~~~~
error: unresolved import `core::nonzero::Zeroable`. Could not find `nonzero` in `core` [E0432]
use core::nonzero::{NonZero, Zeroable};
                             ^~~~~~~~
error: unresolved import `core::num::Zero`. Could not find `num` in `core` [E0432]
use core::num::Zero;
    ^~~~~~~~~~~~~~~
error: unresolved import `core::num`. There is no `num` in `core` [E0432]
use core::num;
    ^~~~~~~~~
error: unresolved import `core::nonzero::NonZero`. Could not find `nonzero` in `core` [E0432]
use core::nonzero::{NonZero, Zeroable};
                    ^~~~~~~
error: unresolved import `core::nonzero::Zeroable`. Could not find `nonzero` in `core` [E0432]
use core::nonzero::{NonZero, Zeroable};
                             ^~~~~~~~
error: `Zeroable` is not a trait [E0404]
impl<T> Deserialize for NonZero<T> where T: Deserialize + PartialEq + Zeroable + Zero {
                                                                      ^~~~~~~~
error: `Zero` is not a trait [E0404]
impl<T> Deserialize for NonZero<T> where T: Deserialize + PartialEq + Zeroable + Zero {
                                                                                 ^~~~
error: failed to resolve. Use of undeclared type or module `num` [E0433]
     where A: Serialize + Clone + iter::Step + num::One,
                                               ^~~~~~~~
error: `Zeroable` is not a trait [E0404]
impl<T> Serialize for NonZero<T> where T: Serialize + Zeroable {
                                                      ^~~~~~~~
```

dtolnay:
Caused by #302 (see its last Travis build: https://travis-ci.org/serde-rs/serde/builds/127547985).

oli-obk:
was fixed by #314

PR #315, opened 2016-05-09
Prefix type parameters and lifetimes with double underscore

dtolnay:
Addresses #312. We already do this almost everywhere.

### DIFF
--- a/serde_codegen/src/attr.rs
+++ b/serde_codegen/src/attr.rs
@@ -563,37 +563,37 @@ fn wrap_serialize_with(cx: &ExtCtxt,
 
     quote_expr!(cx, {
         trait __SerdeSerializeWith {
-            fn __serde_serialize_with<S>(&self, serializer: &mut S) -> Result<(), S::Error>
-                where S: _serde::ser::Serializer;
+            fn __serde_serialize_with<__S>(&self, serializer: &mut __S) -> Result<(), __S::Error>
+                where __S: _serde::ser::Serializer;
         }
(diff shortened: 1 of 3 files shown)


dtolnay:

oli-obk:
up to now the "style" of PRs was to rebase them instead of merging them. I think we should stay with that style.

dtolnay:
Okay I rebased this PR and will rebase my other ones.

We can discuss this in a separate ticket if you are interested, but just to raise some points in favor of merge commits:

1. A large part of the point of PRs is code review. As PRs change, GitHub [keeps track](https://github.com/blog/2123-more-code-review-tools#pick-up-where-you-left-off) of what part each reviewer has seen. Every rebase screws that up and forces reviewers to start from the beginning.
2. Many times a PR has already been reviewed fully but is not ready to merge for a variety of reasons. If there are only merge commits since the time that it was reviewed, it is probably okay to merge. On the other hand if there is a single rebased commit, you have no idea. Better review the whole thing again to make sure nothing important has changed.
3. The "Files changed" view is no different whether there are merge commits or not. It shows you the diff compared to the base branch. So if that is your workflow, it is not hindered by merge commits.
4. If we care about preserving a relatively simple commit history, PRs that include merge commits can be rebased and squashed right at the end, after the review is finished.

oli-obk:
I find your arguments very convincing, but I don't want to change the style without consulting @erickt, so lets open a new issue for now

PR #320, opened 2016-05-11
Do not depend on multiple syntex versions

dtolnay:
serde_codegen has syntex 0.32 under dependencies, but syntex 0.31 under build-dependencies...


oli-obk:
we should probably find a way to lint against such things

oli-obk:
@homu r+

oli-obk:
haha what? homu is great

PR #329, opened 2016-05-14
Add changelog

### LINKED ISSUE #306: Add a changelog
It would be very helpful if the project included a changelog document, or if the releases on the releases page were given release notes. Right now there isn't an easy way to tell what has changed in each version of serde (and related crates.)

dtolnay:
Fixes #306.

[Rendered](https://github.com/dtolnay/serde/blob/changelog/CHANGELOG.md)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,25 +1,31 @@
 ## v0.7.5 (May 12, 2016)
 
 - Fixes a codegen regression from 0.7.4 that generates invalid code for some
-  generic structs (#308)
-- Visiting the same struct field or map key twice is now an error (#293)
+  generic structs ([#308](https://github.com/serde-rs/serde/issues/308))
+- Visiting the same struct field or map key twice is now an error
+  ([#293](https://github.com/serde-rs/serde/issues/293))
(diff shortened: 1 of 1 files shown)


oli-obk:
what's the advantage of a textual changelog over https://github.com/serde-rs/serde/releases ?

dtolnay:
I had a somewhat negative experience with the Releases page yesterday. They are sorted by date tagged instead of date committed so we get 0.6.13 -> 0.6.14 -> 0.6.15-> 0.6.12 -> 0.6.11. Change history is not tracked in git, unlike for the wiki. We cannot accept PRs that modify the release notes.

I agree that we do not need to have both. We could have the releases page link to the changelog file.

oli-obk:
Well... ususally we don't get a wrong ordering. Also it makes total sense to sort by date tagged, so if there are bugfix releases, these are given by the date they were created, not in order of release numbers.

Can't we backdate tags manually?

dtolnay:
Good call. I will look into backdating the tags when I am back home next week.

dtolnay:
I backdated the 0.6.13, 0.6.14, 0.6.15 tags to fix the ordering.

PR #335, opened 2016-05-16
Support (de)serialize_with in tuples

### LINKED ISSUE #330: Tuple struct fields with deserialize_with still require Deserialize bound
The following complains about `Url` not implementing `Deserialize`

```rust
#[derive(Deserialize)]
pub struct SerializableUrl(#[serde(deserialize_with="DeserializeWith::deserialize_with")] Url)
```

dtolnay:
Fixes #330 and a number of related bugs around (de)serialize_with in newtype structs, tuple structs, newtype variants, and tuple variants.

The most interesting thing in this PR is a significantly condensed version of `wrap_serialize_with`. It now works a lot more like `wrap_deserialize_with`, with just a thin wrapper that provides a `serialize` method. Please confirm that the contortions in the original implementation were not necessary. I had to make this change to support serialize_with in newtype structs and newtype variants.

### DIFF
--- a/serde_codegen/src/attr.rs
+++ b/serde_codegen/src/attr.rs
@@ -177,31 +177,29 @@ pub struct FieldAttrs {
     name: Name,
     skip_serializing_field: bool,
     skip_deserializing_field: bool,
-    skip_serializing_field_if: Option<P<ast::Expr>>,
+    skip_serializing_if: Option<ast::Path>,
     default_expr_if_missing: Option<P<ast::Expr>>,
-    serialize_with: Option<P<ast::Expr>>,
+    serialize_with: Option<ast::Path>,
     deserialize_with: Option<ast::Path>,
 }
 
 impl FieldAttrs {
     /// Extract out the `#[serde(...)]` attributes from a struct field.
     pub fn from_field(cx: &ExtCtxt,
-                      container_ty: &P<ast::Ty>,
-                      generics: &ast::Generics,
-                      field: &ast::StructField,
-                      is_enum: bool) -> Result<Self, Error> {
+                      index: usize,
+                      field: &ast::StructField) -> Result<Self, Error> {
         let builder = AstBuilder::new();
 
         let field_ident = match field.ident {
             Some(ident) => ident,
-            None => { cx.span_bug(field.span, "struct field has no name?") }
+            None => builder.id(index.to_string()),
         };
 
         let mut field_attrs = FieldAttrs {
             name: Name::new(field_ident),
             skip_serializing_field: false,
             skip_deserializing_field: false,
-            skip_serializing_field_if: None,
+            skip_serializing_if: None,
             default_expr_if_missing: None,
             serialize_with: None,
             deserialize_with: None,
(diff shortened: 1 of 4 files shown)


serde_codegen/src/attr.rs
```diff
@@ -177,31 +177,29 @@ pub struct FieldAttrs {
     name: Name,
     skip_serializing_field: bool,
     skip_deserializing_field: bool,
-    skip_serializing_field_if: Option<P<ast::Expr>>,
+    skip_serializing_if: Option<ast::Path>,
```
oli-obk: iirc the idea was to allow arbitrary expressions in the future

serde_codegen/src/ser.rs
```diff
@@ -478,7 +516,7 @@ fn serialize_tuple_variant(
     structure_ty: P<ast::Ty>,
     fields: &[ast::StructField],
     field_names: Vec<Ident>,
```
oli-obk: Unrelated to this PR, but why do we have field names for tuple fields? We forward to `SeqVisitor` anyway afaik... is this leftover from an early serde?

dtolnay: The field names are `__field0` `__field1` etc and they come from the captures in the switch statement that matched the current enum variant. See [ser.rs#L401](https://github.com/serde-rs/serde/blob/16d3e96b77fe8d50a2f9d9716ec231fc04251bce/serde_codegen/src/ser.rs#L401)

serde_codegen/src/attr.rs
dtolnay: I see. Given that it's not on our roadmap (no issue for it) and I am not convinced that it would be a valuable thing to support, I do no think we should keep around a lot of convoluted dead code in the hope that some day it will be useful.

dtolnay:
I rebased to fix the merge conflict with #338. We both added something to the bottom of serde_tests/tests/test_gen.rs.

dtolnay:
The travis failure is clippy doesn't build.

oli-obk:
I think I grokked most of the code now. 

> I am not convinced that it would be a valuable thing to support, I do no think we should keep around a lot of convoluted dead code in the hope that some day it will be useful.

agreed on both points. We should probably add cfail tests some day to test for behavior we don't want.

@homu r+

oli-obk:
hmmm right... well... it looks like clippy is going to work again soon, I'll postpone instead of merging manually

dtolnay:
> We should probably add cfail tests some day to test for behavior we don't want.

I filed #341 to add cfail tests.

The clippy PR we need to watch is https://github.com/Manishearth/rust-clippy/pull/944, going to be published as [0.0.69](https://crates.io/crates/clippy/0.0.69).

mcarton:
@dtolnay done

dtolnay:
@homu retry

oli-obk:
@homu retry force

oli-obk:
@homu r+

oli-obk:
huh

dtolnay:
Merging because I retriggered the Travis build and it passed.

PR #343, opened 2016-05-24
Better documentation for de::Error::invalid_length

### LINKED ISSUE #342: Usage of de::Error::invalid_length(len: usize)
Is Serde trait method [`de::Error::invalid_length(len: usize)`](https://serde-rs.github.io/serde/serde/de/trait.Error.html#method.invalid_length) supposed to be called with the number of elements found or the number of elements expected?

Suppose I try to deserialize a sequence that must contain at least one element but may contain many more. If the sequence is found to be empty should I emit `de::Error::invalid_length(0)` or should I use a custom error because the length is not fixed?

The documentation should be updated to be more precise here.

pyfisch:
Closes #342

### DIFF
--- a/serde/src/de/mod.rs
+++ b/serde/src/de/mod.rs
@@ -39,6 +39,9 @@ pub trait Error: Sized + error::Error {
     }
 
     /// Raised when a fixed sized sequence or map was passed in the wrong amount of arguments.
+    ///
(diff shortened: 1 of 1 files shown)


oli-obk:
thanks!

@homu r+

PR #344, opened 2016-05-24
use iterators instead of a custom `MapVisitor` type

oli-obk:
This can be done to `SeqVisitor`, too.

I was mostly wondering about the design. It's shorter (`167 additions and 418 deletions`), but it might be unnecessarily complex, due to the closure indirection.

The interesting part is in `serde/src/ser/mod.rs`, `serde/src/ser/impls.rs` and `serde_tests/tests/token.rs`. (the rest is just codegen)

I'll implement this for json to see a real world example

### DIFF
--- a/serde/src/ser/impls.rs
+++ b/serde/src/ser/impls.rs
@@ -65,7 +65,6 @@ use super::{
     Serialize,
     Serializer,
     SeqVisitor,
-    MapVisitor,
 };
 
 ///////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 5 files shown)


oli-obk:
hmmm... doesn't work, because the `Serializer` serializing the map might not be the one serializing the map keys/values. E.g. json uses a special `MapKeySerializer` to forbid all non-string-serializing keys.

To do this in a non-dynamic-dispatch way, we'd probably need HKT. Or we redesign the entire system. Instead of having a single `Serialize` trait, we could have gazillions of `SerializeAsX` traits (one for every method of `Serializer`). Then we can probably create really unreadable generic errors that will prevent the use of maps with non-string-keys in json.

erickt:
> Then we can probably create really unreadable generic errors that will prevent the use of maps with non-string-keys in json.

Hehe. Yeah, I'm not fond of our situation, but I couldn't figure out a better option. I'm glad you're at least starting to explore passing closures around. Back when I was building the core serde model, closures were still in a bit of a weird state, so I didn't really do much research on if things would be better with a more strategic closure or two.

Hm. Perhaps rather than the closure doing `move |s: &mut S| k.serialize(s),` what if instead we did `move |s: &mut S| s.serialize_key(k)`? Maybe the json serializer could then intercept this call and pass it to a sub-serializer that'll error on non-string keys?

Also, `SerializeAsX` seems a bit scary to me :)

dtolnay:
Closing in favor of https://github.com/serde-rs/serde/pull/437.

PR #346, opened 2016-05-30
Rustup too rustc 1.11.0-nightly (7746a334d 2016-05-28)

Manishearth:
cc @erickt

### DIFF
--- a/serde/Cargo.toml
+++ b/serde/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "serde"
-version = "0.7.5"
+version = "0.7.6"
 authors = ["Erick Tryzelaar <erick.tryzelaar@gmail.com>"]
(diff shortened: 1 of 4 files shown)


Manishearth:
Okay, I'm not familiar enough with aster traits to fix this

Manishearth:
Blocked on a publish of https://github.com/serde-rs/quasi/pull/42

erickt:
@homu: r+ 379c9e7

alexcrichton:
Changes like this which update syntex, a public dependency of serde_codegen (e.g. it reexports types from syntex) are actually a breaking change for the serde crate. Could this be signaled with a major version update or holding off the upgrade? This unfortunately breaks any other crate which is depending on serde_codegen as it may get a mismatch of syntex versions otherwise.

PR #351, opened 2016-06-01
also publish the `.in` file used by the build script

oli-obk:


oli-obk:
the publish worked, but is there anything else we need to upload? license files maybe?

dtolnay:
I don't think we need to include the license files. Maybe if we were using a custom license, but `license = "MIT/Apache-2.0"` is obvious enough.

oli-obk:
@homu r+

Issue #336, opened 2016-05-17
Cannot generate Serialize impl for recursive generic enum


alexbool:
Consider this example:
```rust
#[derive(Serialize, Deserialize)]
pub enum RecursiveTree<D> {                                       
    Split {                                              
        lt: Box<RecursiveTree<D>>,                       
        gt: Box<RecursiveTree<D>>,                       
    },                                                   
    Leaf {                                               
        data: D,                                         
    },                                                   
} 
```
on nightly rust using serde compiler plugin.

This fails with
```
src/model.rs:35:35: 35:66 error: overflow evaluating the requirement `_: std::marker::Sized` [E0275]
src/model.rs:35 #[cfg_attr(feature = "serialize", derive(Serialize, Deserialize))]
                                                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/model.rs:35:35: 35:66 note: in this expansion of #[derive_Serialize] (defined in src/model.rs)
src/model.rs:35:35: 35:66 help: run `rustc --explain E0275` to see a detailed explanation
src/model.rs:35:35: 35:66 note: consider adding a `#![recursion_limit="128"]` attribute to your crate
src/model.rs:35:35: 35:66 note: required because of the requirements on the impl of `serde::Serialize` for `model::RecursiveTree<_>`
src/model.rs:35:35: 35:66 note: required because of the requirements on the impl of `serde::Serialize` for `Box<model::RecursiveTree<_>>`
src/model.rs:35:35: 35:66 note: required because of the requirements on the impl of `serde::Serialize` for `model::RecursiveTree<_>`
src/model.rs:35:35: 35:66 note: required because of the requirements on the impl of `serde::Serialize` for `Box<model::RecursiveTree<_>>`
src/model.rs:35:35: 35:66 note: required because of the requirements on the impl of `serde::Serialize` for `model::RecursiveTree<_>`
src/model.rs:35:35: 35:66 note: required because of the requirements on the impl of `serde::Serialize` for `Box<model::RecursiveTree<_>>`
src/model.rs:35:35: 35:66 note: required because of the requirements on the impl of `serde::Serialize` for `model::RecursiveTree<_>`
src/model.rs:35:35: 35:66 note: required because of the requirements on the impl of `serde::Serialize` for `Box<model::RecursiveTree<_>>`
src/model.rs:35:35: 35:66 note: required because of the requirements on the impl of `serde::Serialize` for `model::RecursiveTree<_>`
[...]

PR #352, opened 2016-06-04
Attribute for handwritten where clauses

dtolnay:
Addresses (2) and (3) in https://github.com/serde-rs/serde/issues/336#issuecomment-220378916.

- If there is a `#[serde(bound="...")]` attribute on the type, use the union of that and the actual type's `where` clause as the `where` clause for the impl and do not attempt to generate any additional `where` clauses whatsoever.
- If there is a `#[serde(bound="...")]` attribute on a field, use that and do not attempt to generate any additional `where` clauses for the field.

The `bound` attribute behaves similar to `rename` in that you can specify a single attribute that applies to both ser and de, or individual ones.

```
#[serde(bound="D: Serialize + Deserialize")]

#[serde(bound(serialize="D: Serialize", deserialize="D: Deserialize"))]
```

EDIT: now addresses (4) from https://github.com/serde-rs/serde/issues/336#issuecomment-220378916 as well.

- If a field contains direct recursion, do not generate any bounds based on that field except from `bound` attributes.

### DIFF
--- a/serde_codegen/src/attr.rs
+++ b/serde_codegen/src/attr.rs
@@ -137,12 +137,12 @@ impl ContainerAttrs {
         self.deny_unknown_fields
     }
 
-    pub fn ser_where(&self) -> Option<&Vec<ast::WherePredicate>> {
-        self.ser_where.as_ref()
+    pub fn ser_where(&self) -> Option<&[ast::WherePredicate]> {
+        self.ser_where.as_ref().map(Vec::as_slice)
     }
 
-    pub fn de_where(&self) -> Option<&Vec<ast::WherePredicate>> {
-        self.de_where.as_ref()
+    pub fn de_where(&self) -> Option<&[ast::WherePredicate]> {
+        self.de_where.as_ref().map(Vec::as_slice)
     }
 }
 
@@ -361,12 +361,12 @@ impl FieldAttrs {
         self.deserialize_with.as_ref()
     }
 
-    pub fn ser_where(&self) -> Option<&Vec<ast::WherePredicate>> {
-        self.ser_where.as_ref()
+    pub fn ser_where(&self) -> Option<&[ast::WherePredicate]> {
+        self.ser_where.as_ref().map(Vec::as_slice)
     }
 
-    pub fn de_where(&self) -> Option<&Vec<ast::WherePredicate>> {
-        self.de_where.as_ref()
+    pub fn de_where(&self) -> Option<&[ast::WherePredicate]> {
+        self.de_where.as_ref().map(Vec::as_slice)
     }
 }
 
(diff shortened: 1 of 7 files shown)


dtolnay:
I failed a handful of clippy lints, working on them now.

mcarton:
FYI, clippy 0.0.72 was just published, fixing the false positives in this PR :smile:

dtolnay:
@oli-obk @erickt this is ready for review.

serde_codegen/src/lib.rs
```diff
@@ -2,6 +2,7 @@
 #![cfg_attr(feature = "nightly-testing", feature(plugin))]
 #![cfg_attr(feature = "nightly-testing", allow(too_many_arguments))]
 #![cfg_attr(feature = "nightly-testing", allow(used_underscore_binding))]
+#![cfg_attr(feature = "nightly-testing", allow(useless_let_if_seq))]
```
dtolnay: Due to https://github.com/Manishearth/rust-clippy/issues/985.

mcarton: Should be fixed in Clippy 0.0.73 just published. Thanks again for the report.

dtolnay: Thanks Martin, that solved it.

serde_codegen/src/attr.rs
```diff
@@ -316,53 +360,81 @@ impl FieldAttrs {
...
-    fields: &'a [ast::StructField],
-) -> Result<Vec<(&'a ast::StructField, FieldAttrs)>, Error> {
+    fields: &[ast::StructField],
+) -> Result<Vec<(ast::StructField, FieldAttrs)>, Error> {
     fields.iter()
         .enumerate()
         .map(|(i, field)| {
             let attrs = try!(FieldAttrs::from_field(cx, i, field));
-            Ok((field, attrs))
+            Ok((field.clone(), attrs))
```
oli-obk: why was this changed from a reference to a value?

dtolnay: Because in all_fields_with_attrs the `Vec<ast::StructField>` does not live long enough to return references into it.

oli-obk:
Implementation lgtm.

Design wise I think it's good, too. We can always add more magic later if it seems helpful/necessary, but the current version gives the user a lot of control to fix whatever issues they have with codegen.

Discovering this feature is hard though... But I'm not sure what to do about it... It would be great if we had some attribute to attach to generated ast-nodes that would cause rustc to emit some custom link/info if an error occurs inside that generated node. Let's see how often people complain, and adding a bound is the solution.

@homu r+

alexbool:
It would be great to bump and publish a version

dtolnay:
@alexbool https://crates.io/crates/serde-codegen/0.7.8

alexbool:
Thanks!

PR #354, opened 2016-06-06
Fix attributes canceling each other

### LINKED ISSUE #353: `rename` attributes overwrite each other
```rust
#![feature(custom_derive, plugin)]
#![plugin(serde_macros)]

extern crate serde_json;

#[derive(Serialize, Deserialize, Default)]
struct S {
    #[serde(rename(serialize="s1", deserialize="d1"))]
    v1: (),
    #[serde(rename(serialize="s2"))]
    #[serde(rename(deserialize="d2"))]
    v2: (),
}

fn main() {
    let s: S = Default::default();
    println!("{}", serde_json::to_string(&s).unwrap());
}
```

Expected: `{"s1":null,"s2":null}`
Actual: `{"s1":null,"v2":null}`

dtolnay:
Fixes #353.

### DIFF
--- a/serde_codegen/src/attr.rs
+++ b/serde_codegen/src/attr.rs
@@ -82,7 +82,6 @@ impl ContainerAttrs {
                     // Parse `#[serde(rename="foo")]`
                     ast::MetaItemKind::NameValue(ref name, ref lit) if name == &"rename" => {
                         let s = try!(get_str_from_lit(cx, name, lit));
-
                         container_attrs.name.serialize_name = Some(s.clone());
                         container_attrs.name.deserialize_name = Some(s);
                     }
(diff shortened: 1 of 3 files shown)


serde_codegen/src/attr.rs
```diff
@@ -82,17 +82,19 @@ impl ContainerAttrs {
...
                     }
 
                     // Parse `#[serde(rename(serialize="foo", deserialize="bar"))]`
                     ast::MetaItemKind::List(ref name, ref meta_items) if name == &"rename" => {
                         let (ser_name, de_name) = try!(get_renames(cx, meta_items));
-
-                        container_attrs.name.serialize_name = ser_name;
-                        container_attrs.name.deserialize_name = de_name;
+                        if ser_name.is_some() {
+                            container_attrs.name.serialize_name = ser_name;
```
oli-obk: I think this should also check whether this is overwriting a previous value and emit an error (or at least a warning) if it is.

oli-obk:
Ok... it really is time for cfail tests... I'm on it ^^

serde_codegen/src/attr.rs
dtolnay: Good call. Bedtime for me but I filed #355 and I will try to follow up this week.

oli-obk:
@homu r+

Issue #358, opened 2016-06-08
Remove public dependency on Syntex


dtolnay:
Currently Syntex is a public dependency because the argument to `serde_codegen::register` is `&mut syntex::Registry`:

```rust
// build.rs
pub fn main() {
    let out_dir = env::var_os("OUT_DIR").unwrap();
    let src = Path::new("src/main.rs.in");
    let dst = Path::new(&out_dir).join("main.rs");

    let mut registry = syntex::Registry::new();
    serde_codegen::register(&mut registry);
    registry.expand("", &src, &dst).unwrap();
}
```

Let's see if we can do something more similar to this instead:

```rust
// build.rs
pub fn main() {
    let out_dir = env::var_os("OUT_DIR").unwrap();
    let src = Path::new("src/main.rs.in");
    let dst = Path::new(&out_dir).join("main.rs");

    // no more public Syntex dependency
    serde_codegen::expand(&src, &dst).unwrap();
}
```

The downside would be if someone wants to apply more than one compiler plugin, their code is parsed multiple times:

```rust
serde_codegen::expand(&src, &tmp1).unwrap();
other_plugin::expand(&tmp1, &tmp2).unwrap();
third_plugin::expand(&tmp2, &dst).unwrap();
```

jimmycuadra:
This seems like a worthwhile compromise to me. I'm betting that the number of projects that use more than one compiler plugin on stable is pretty small. (My main project actually does, but it's nightly-only, so this wouldn't affect me.)

dtolnay:
The new `serde_codegen::expand` is in, but I am leaving this open until we remove `serde_codegen::register` or decide not to remove it.

dtolnay:
The new `serde_codegen::expand` is in, but I am leaving this open until we either remove `serde_codegen::register` in a breaking release or decide to keep it.

erickt:
As I mentioned in in the other thread, I'm all for getting rid of `serde_codegen::register`.

PR #362, opened 2016-06-09
Add serde_codegen::expand to avoid public Syntex dependency

dtolnay:
Required for #358. We can remove `serde_codegen::register` in the next breaking release.

This allows Syntex users to avoid being broken by Serde bumping its Syntex dependency.

### DIFF
--- a/serde_codegen/src/lib.rs
+++ b/serde_codegen/src/lib.rs
@@ -22,6 +22,11 @@ extern crate syntax;
 #[cfg(not(feature = "with-syntex"))]
 extern crate rustc_plugin;
 
+#[cfg(feature = "with-syntex")]
+use std::io;
+#[cfg(feature = "with-syntex")]
+use std::path::Path;
+
 #[cfg(not(feature = "with-syntex"))]
 use syntax::feature_gate::AttributeType;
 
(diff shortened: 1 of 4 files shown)


serde_codegen/src/lib.rs
```diff
@@ -32,6 +37,13 @@ include!(concat!(env!("OUT_DIR"), "/lib.rs"));
 include!("lib.rs.in");
 
 #[cfg(feature = "with-syntex")]
+pub fn expand(src: &Path, dst: &Path) -> io::Result<()> {
```
oli-obk: I think the arguments should be generic over `AsRef<Path>`.

dtolnay: Good idea, although [Syntex::expand accepts &Path](https://github.com/serde-rs/syntex/blob/bf7b812bedab8f1f02ad7e0cd237c48f30dd9e1d/syntex/src/lib.rs#L126). I made the change here and filed https://github.com/serde-rs/syntex/issues/64 to change it in Syntex.

oli-obk:
@homu r+

dtolnay:
The serde-syntex-example test is not going to work because it depends on the released serde_codegen, not the relative path. I will add a cargo override in the examples directory so the code remains copy-pastable but still builds with the code we want.

oli-obk:
@homu r+

PR #367, opened 2016-06-10
Simplify implementation of #[serde(default=...)]

dtolnay:


oli-obk:
lgtm

@homu r+

serde_codegen/src/attr.rs
```diff
@@ -212,13 +212,24 @@ pub struct FieldAttrs {
...
 
+/// Represents the default to use for a field when deserializing.
+#[derive(Debug, PartialEq)]
+pub enum FieldDefault {
+    /// Field must always be specified because it does not have a default.
+    None,
+    /// The default is given by `std::default::Default::default()`.
+    Default,
+    /// The default is given by this function.
+    Path(ast::Path),
```
alexbool: It would be also great to support literals and constants as well as functions, so you can actually write
```rust
#[serde(default="0.1")]
pub field: f32,
```

oli-obk: The only situation that would be problematic is when the field is of function type, but I don't think we can do anything useful there anyway.

dtolnay: I filed #368 to track your idea.

alexbool: Thanks!

PR #375, opened 2016-06-11
Add decimal (d128) support.

nubis:
Hi guys, this pull request adds support for serializing and deserializing arbitrary precision decimal numbers. This was needed to support a general 'Number' format in serde_json, but it may come in handy for others. 

A few things to notice about this change:
- I made ToPrimitive a public trait, as it is very useful. For example, my serde_json proposal will use it to keep the internal representation of numbers as d128 when serializing and deserializing.
- Both d128 from the decimal module and the d128! macro are re-exported.
- It bumps the version to 0.8.0 as it introduces a 'big' dependency (decimal).
- I'm overly relying on string representations to convert to/from d128 types. But looking at their code they do it too, I think looking forward we can use their type conversions once they have more.

I know this is a rather big change for a first time contributor like myself, perhaps this changes can stay on a branch until some more eyes have a look at them and we gather consensus. I'm more than happy to point my own project's dependencies to a branch in this repo, I just don't want it to remain a fork.

Looking forward to your comments. (I'm following up with a pull request for serde_json 0.8.0 soon)


erickt:
Thanks! I only briefly skimmed this since I'm on the road, but my first concern is that I don't know much of anything about the public decimal crate. It hasn't let gotten a lot of traction, and I'm not sure if it's c library is portable. Serde now supports being run with just libcore, and potentially inside a kernel, so it'd be a shame to lose that ability. 

I'd love for some abstraction that supports BigNum types. Have you thought if there's any way to support them without adding a dependency on that crate?

dtolnay:
Ditto Erick. Looks like there is currently exactly one crate on crates.io with a dependency on the decimal crate, and all of the d128 code on their master branch is commented out......

dtolnay:
Side note in case anyone else is curious - this is how much precision is available in d128:

```rust
assert!(d128!(1) != d128!(1.0000000000000000000000000000000006));
assert!(d128!(1) == d128!(1.0000000000000000000000000000000005));

assert!(d128!(2).pow(d128!(20413)) != d128!(2).pow(d128!(20414)));
assert!(d128!(2).pow(d128!(20414)) == d128!(2).pow(d128!(20415)));
```

nubis:
I fully agree with your comments, think of this as a proof-of-concept.

I probably need to give you guys some background so you understand how this half-cooked pull request came to be:
I'm (or was) fairly new to rust, I started building a simple bitcoin trading robot about 2 months ago. I ended up "having" to build an http client, an http integration test library, then this patch for serde_json which turned out to need a patch to serde itself. When it got to the point that I was going to have to build a bignum/decimal library I decided to use the one available instead. I'm in way over my head, that's why this took so long.

- I'm happy to share my work on this, in case anyone else needs it. It can be left on an experimental state, I'll make sure to keep my fork synced.
- I'd hate for my fork to diverge too much, I'll let you know the moment I start receiving pull requests on my fork.
- Having a branch here instead of a fork on my user could help this get some more visibility to address all issues.
- I think as rust gains popularity deserializing numbers precisely in JSON APIs will need to be supported.
- I fully agree decimal is still green, as a lot of other things in rust land. (at least for someone coming from ruby and haskell). I don't know who wrote the decimal crate, it looked relatively well maintained as compared to other things, at least a few months ago.
- Someone's got to use it, It's not going to get validated on its own.
- I can do this all over with a more respected decimal crate, it would be ideal for rust to provide decimal support in the standard lib FWIW.
- I do not think is serde's job to also build the bignum support, but someone's got to do it.. Perhaps I (we?) can contact the maintainer and adopt the crate?

I'm not in a hurry or anything, I'm really happy of just having this out here and receiving your feedback as it gives me enough to move forward on my project until we figure out a better approach. 🍻

nubis:
@homu I will, but as we were talking with the guys, maybe we don't want to merge it just yet.

oli-obk:
> boxing a string that then gets serialized/deserialized further using into/to_string, 

That's not a great solution, because `bincode` and similar start out with a n-bit representation that they can turn into a rust numeric type.


### Recap

**TLDR**: Let's do a clean restart of number serialization fitting everyone's needs. Skip forward to *Proposal*

There is the set `A` of number representations that the serialization formats use. Then there's the set `B` of number representations in rust. `B` also contains arbitrary user types (`BigNum`, `d128`, ...).

Ideally we'd have a function `f(A) -> B` and `f'(B) -> A`. Obviously that makes little sense, as it would require serde to know about all possible types. So instead we have a set `S` that contains all basic rust number types and  `f(A) -> S`, `f'(S)->A` as well as `g(B) -> S`, `g'(S) -> B`.

Right now we are discussing what `B` should contain and how to represent it. The problem I'm seeing is that we are discussing the memory representation instead of the abstract representation. There will never be a perfect memory representation that fits all needs (e.g. `BigNum` can represent all numbers, but requires memory allocation). We'll always have to do tradeoffs. The tradeoff decision shouldn't really be made by `serde` or an implementor, but by the final user. We could of course get there through a huge `cfg!` madness, but I have a proposal (no concrete design, just some thoughts):

### Proposal

We add a new trait `NumericSerialization` that is implemented for the regular rustc types. Users can implement this trait for their own types. Also there's a type `NumericTypeInfo` that contains various things that are useful to know.

```rust
struct NumericTypeInfo {
    name: &'static str,
    // if you have to hold more than 2^64 bits, then I have bad news for you.
    // None for infinitely sized types
    min_bits: Option<u64>,
    min_value: Option<&'static NumericSerialization>,
    // can a value of this type be signed
    signed: bool,
    // can there only be integral values? (only zeros after the comma)
    integral: bool,
}

/// `Display` bound for textual representations
trait NumericSerialization: core::fmt::Display {
    /// yields a value of the above struct
    /// FIXME: use associated constant once that's stable
    fn type_info() -> &'static NumericTypeInfo;
    /// yields the number of bits required to hold this value
    fn num_bits(&self) -> u64;
    /// whether the value is signed
    fn is_signed(&self) -> bool;
    /// whether the value is integral
    fn is_integral(&self) -> bool;
    /// yield an f64 representation of the value if that can be done losslessly
    fn as_f64(&self) -> Option<f64>;
    /// yield an u64 representation of the value if that can be done losslessly
    fn as_u64(&self) -> Option<u64>;
    /// yield an i64 representation of the value if that can be done losslessly
    fn as_i64(&self) -> Option<i64>;
    /// Iterate over the bits starting with the least significant one
    /// This function uses a closure until we get `impl Trait` return types
    /// FIXME: add more bounds to the `Iterator` to allow things like `.rev()`
    fn iter_bits<F: FnOnce(&mut Iterator<Item=bool>)>(&self, f: F);
}
```

With this information any numeric type can offer enough information for a `Serializer` to decide how to proceed. It's important to use static dispatch when passing this value (no other way anyway, it's not object safe, let's keep it this way). Static dispatch helps optimizing out the entire thing.

Deserialization needs the design in reverse. The trait `NumericDeserialization` is again implemented for concrete types. 

```rust
trait NumericDeserialization {
    /// FIXME: use associated constant once that's stable
    fn type_info() -> &'static NumericTypeInfo;
    fn from_str(&str) -> Self;
    fn from_bits(&mut Iterator<Item=bool>) -> Self;
    fn from_u64(u64) -> Self;
    fn from_f64(f64) -> Self;
    fn from_i64(i64) -> Self;
}
```

The `Deserializer` trait would then have a method `fn deserialize_numeric<V>(&mut self, visitor: V) where V::Value: NumericDeserialization`.

I'm sure I have forgotten some corner case.

nubis:
Yes! I think this is a good start. Sounds fun to design and implement too.

erickt:
@oli-obk: Nice! That's a along the lines of what I had in mind, but you've fleshed it out even more. Want to copy this into the never-used https://github.com/serde-rs/rfcs, then it'll be a little easier to comment on it?

oli-obk:
Will do

dtolnay:
Closing in favor of https://github.com/serde-rs/rfcs/pull/1.

PR #377, opened 2016-06-11
Remove some low-risk travis builds

dtolnay:


dtolnay:
This PR takes us from 69 minutes of total build time to 36 minutes, and 26 minutes of wall time to 10 minutes. I will follow up with another PR to parallelize within individual builds because Travis gives us 2 cores and I suspect we are not using them.

dtolnay:
I implemented parallel builds in https://github.com/dtolnay/serde/commit/8c1f0250f278aee9bf42e4bb137b1310d757f752 but only managed to speed up the build by about a minute, so not worth pursuing right now.

oli-obk:
lgtm

@homu r+

Issue #359, opened 2016-06-08
Derive Deserialize to Struct with Lifetime 


craftytrickster:
I have been using serde to deserialize json and it has been very easy to use and helpful.

However, I have a question regarding deserializing to a struct with a lifetime. Take the following for example:

```rust
#[derive(Deserialize)]
struct Person<'a> {
    first_name: &'a str,
    last_name: &'a str
}
```

When using owned Strings, it works fine, but if I try to compile the above, I get the following error message:

```
error: the trait bound `&'a str: _IMPL_DESERIALIZE_FOR_Person::_serde::Deserialize` is not satisfied [E0277]
src/main.rs:19 #[derive(Deserialize)]
```
As long as this struct does not outlive the original input string being deserialized, I do not see why having a str reference is problematic. 

Is there something I am doing wrong here, or is it not supported? If it is not supported, are there any plans in the future? Perhaps I can help.

dtolnay:
This is not supported and there are currently no plans to support it.

How would it work in this case?

```json
{"first_name": "escaped\"quote"}
```

The struct needs to contain a reference to the string `escaped"quote`. There is no such string in the input for it to refer to.

craftytrickster:
Let's say I have the following input:
`let input_string_slice = "{\"first_name\":\"escaped\\\" quote\"}";` 

In order to deserialize I make the following call:
`let person = serde_json::from_str(input_string_slice).unwrap();` 

Why can't the person's first_name field just be a &str that refers to `&input_string_slice[15..30]`? After all, we already know that input_string_slice exists, since it is used in the `from_str` call. Owned Strings (like in the following code) deserialize completely fine.
```rust
#[derive(Deserialize)]
struct Person {
    first_name: String
}
```

I assume that the first_name value is able to be turned into an owned String by copying the chars from that same 15..30 range with something like ``&input_string_slice[15..30].to_string()`. 

Since we already know that range, I would just like to make a slice reference for the below, to avoid needing to allocate new string objects:

```
#[derive(Deserialize)]
struct Person<'a> {
    first_name: &'a str
}
```

dtolnay:
Because `&input_string_slice[15..30]` is this string:

```
escaped\" quote
```

And the correct deserialized string that the struct should contain after deserializing from JSON is this string:

```
escaped" quote
```

And those are different strings.

dtolnay:
I filed #360 to improve the error message for this case.

craftytrickster:
Thank you for the explanation.

PR #378, opened 2016-06-12
Better error when deriving Deserialize for struct containing &str

### LINKED ISSUE #360: Better error when deriving Deserialize for struct containing &str
The current error looks like:

```
src/main.rs:6:32: 6:32 error: the trait bound `&'a str: _IMPL_DESERIALIZE_FOR_S::_serde::Deserialize` is not satisfied [E0277]
src/main.rs:6 #[derive(Serialize, Deserialize)]
                                             ^
```

The codegen for Deserialize should detect specifically the case of a struct (or tuple or variant) containing &str, and fail with a reasonable message saying to try String instead. Trying to deserialize &str is an easy mistake especially because Serialize will work fine.

dtolnay:
Fixes #360. The error looks like this:

```rust
struct Test<'a> {
    s: &'a str,
}
```

```
src/main.rs:6:5: 6:15 error: Serde does not support deserializing fields of type &str; consider using String instead
src/main.rs:6     s: &'a str,
                  ^~~~~~~~~~
```

### DIFF
--- a/serde_codegen/src/de.rs
+++ b/serde_codegen/src/de.rs
@@ -379,6 +379,7 @@ fn deserialize_tuple(
 
     let nfields = fields.len();
     let fields_with_attrs = try!(attr::fields_with_attrs(cx, fields));
+    try!(check_no_str(cx, &fields_with_attrs));
 
     let visit_newtype_struct = if !is_enum && nfields == 1 {
         Some(try!(deserialize_newtype_struct(
@@ -576,6 +577,7 @@ fn deserialize_struct(
     };
 
     let fields_with_attrs = try!(attr::fields_with_attrs(cx, fields));
+    try!(check_no_str(cx, &fields_with_attrs));
 
     let visit_seq_expr = try!(deserialize_seq(
         cx,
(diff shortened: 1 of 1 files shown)


oli-obk:
Oh right. We can just fail the codegen.

The `does not support` sounds to me like it `could support, but doesn't`. Maybe state that 

> Deserializing to `&str` is not possible without leaking memory; ...

jmesmon:
> Deserializing to &str is not possible without leaking memory

Seems like a round-about way of saying "we'd have to allocate a String and leak it" (it even took me a bit to understand that's what was meant). Also note that for deserialization in general (ignoring serde), deserializing to &str will be possible for some formats (those that don't use escape characters in their strings).

oli-obk:
> deserializing to &str will be possible for some serialization formats (those that don't use escape characters in their strings)

The `&str` still needs to be stored $somewhere. How could such a deserialization work without having a heap allocated string to begin with and passing down lifetimes through the entire system?

jmesmon:
> How could such a deserialization work without having a heap allocated string to begin with and passing down lifetimes through the entire system?

It wouldn't. One would need to propagate the lifetime (probably just 1) of the thing being deserialized-from down to the outputs of the deserialization process.

dtolnay:
#359 was asking for deserializing `&'a str` tied to the lifetime of the input data. I think "does not support" is accurate given that it is something Serde could support for some subset of formats.

oli-obk:
> It wouldn't. One would need to propagate the lifetime (probably just 1) of the thing being deserialized-from down to the outputs of the deserialization process.

> I think "does not support" is accurate given that it is something Serde could support for some subset of formats.

that makes sense

@homu r+

Issue #148, opened 2015-08-28
serde_codegen does not generate code for derive in cfg_attr


jmesmon:
For example, this doesn't work:

```
#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
struct Foo {
 a: u32,
};
```

It appears that the codegen simply doesn't know what to do about cfg_attr.

I suppose it's possible that not enough info is available when codegen occurs to actually resolve this. If so, feel free to close.

dtolnay:
@jmesmon I just tried this and it seems to work. Can you try again and report back? Here is the code I used:

```rust
#![feature(plugin, custom_derive)]
#![plugin(serde_macros)]

#[cfg(feature = "serde")]
extern crate serde;

#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
struct Foo {
    a: u32,
}

#[cfg(feature = "serde")]
fn main() {
    println!("with serde");

    fn assert<T: serde::Serialize + serde::Deserialize>(_: T) {}
    assert(Foo { a: 0 });
}

#[cfg(not(feature = "serde"))]
fn main() {
    println!("without serde");
}
```

jmesmon:
@dtolnay you're using `serde_macros` (plugin). This bug is talking about `serde_codegen` (generates code with a build script that you then `include!`).

I'm still able to reproduce this with `serde_codegen`.

Here's a project that reproduces: https://github.com/jmesmon/serde-readme

dtolnay:
> I suppose it's possible that not enough info is available when codegen occurs to actually resolve this. If so, feel free to close.

I suspect this is the case, but we can fix it anyway, at least for the common cases. We need to transform this:

```rust
#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
struct Foo { ... }
```
... into this:

```rust
#[cfg(feature = "serde")]
impl Serialize for Foo { ... }
#[cfg(feature = "serde")]
impl Deserialize for Foo { ... }
struct Foo { ... }
```

dtolnay:
For those not watching the Syntex repo - check out the PR at https://github.com/serde-rs/syntex/pull/71.

erickt:
bad homu, this isn't implemented in serde yet.

dtolnay:
Actually it is. :smile: That's all it took.

nox:
I have a similar problem with:

```rust
#[cfg(target_os = "macos")]
pub type NativeFontHandle = CGFont;

[...]

PR #382, opened 2016-06-12
Add test for expansion inside of cfg_attr

dtolnay:
Test for #148. Requires https://github.com/serde-rs/syntex/pull/71 + Syntex release.

### DIFF
--- a/serde_tests/tests/test_macros.rs
+++ b/serde_tests/tests/test_macros.rs
@@ -14,8 +14,19 @@ trait Trait {
 #[deny(unused_variables)]
 enum Void {}
 
-#[derive(Debug, PartialEq, Serialize, Deserialize)]
+// Test that `derive` expansion works inside of `cfg_attr`. The `all()` is
+// equivalent to true. This struct is used below in a way that requires it to
+// implement all four traits.
+#[cfg_attr(all(), derive(Debug, PartialEq, Serialize, Deserialize))]
 struct NamedUnit;
+
+// Test that `derive` expansion pays attention to the `cfg_attr`. The `any()`
+// is equivalent to false.
+#[allow(dead_code)]
+struct NotSerializable;
+#[allow(dead_code)]
+#[cfg_attr(any(), derive(Serialize))]
+struct AlsoNotSerializable(NotSerializable);
 
 #[derive(Debug, PartialEq, Serialize)]
 struct SerNamedTuple<'a, 'b, A: 'a, B: 'b, C>(&'a A, &'b mut B, C);


erickt:
r=me once serde-rs/syntex#71 lands.

dtolnay:
I marked this WIP until we get a Syntex release to depend on. As far as I know nobody is waiting urgently for this so we can just wait for the next libsyntax break.

erickt:
@homu: r+ 675caa8

erickt:
@dtolnay: syntex is updated, but this is failing on stable and beta with things like:

```
/Users/erickt/projects/rust/serde/serde/serde_tests/target/debug/build/serde_tests-0dc644810d6621ba/out/test.rs:16260:9: 16260:22 error: the trait bound `test_macros::NamedUnit: std::cmp::PartialEq` is not satisfied [E0277]
/Users/erickt/projects/rust/serde/serde/serde_tests/target/debug/build/serde_tests-0dc644810d6621ba/out/test.rs:16260         assert_tokens(&NamedUnit, vec!(Token :: UnitStruct ( "NamedUnit" )));
```

Do you know what's up? Maybe the bug hasn't been fixed in syntex yet.

dtolnay:
That's so strange. It works on the nightly compiler using Syntex, but not on the beta compiler using Syntex. What would be the difference between those?

The generated code looks correct in both cases:

```rust
#[cfg_attr(all(), derive(PartialEq, Debug))]
struct NamedUnit;
#[cfg(all())]
const _IMPL_DESERIALIZE_FOR_NamedUnit: () = { /* ... */ };
#[cfg(all())]
const _IMPL_SERIALIZE_FOR_NamedUnit: () = { /* ... */ };
```

dtolnay:
Interesting, [`cargo expand`](https://github.com/dtolnay/cargo-expand) on beta shows `#[derive(PartialEq, Debug)]` but on nightly it shows them expanded. That must have something to do with it. Now to figure out why...

dtolnay:
```rust
#[cfg_attr(all(), cfg_attr(all(), derive(Debug)))]
struct S;

fn main() {
    println!("{:?}", S);
}
```

This works on nightly but not stable/beta. [Playground link](https://play.rust-lang.org/?code=%23%5Bcfg_attr(all()%2C%20cfg_attr(all()%2C%20derive(Debug)))%5D%0Astruct%20S%3B%0A%0Afn%20main()%20%7B%0A%20%20%20%20println!(%22%7B%3A%3F%7D%22%2C%20S)%3B%0A%7D&version=nightly&backtrace=0)

I am going to write this off as a bug that needed to be fixed in both Syntex and rustc. We can add the test later when we support only versions that contain the rustc side of the fix.

PR #383, opened 2016-06-13
more cfail tests

### LINKED ISSUE #341: Add compile-fail tests to test expected compilation failures
For example, referring to `self` within a `serialize_with` attribute.

oli-obk:
I tried adding cfail tests to `serde_tests`, but mixing syntex-`serde_codegen` and `serde_macros` doesn't end well. We can't use `serde_codegen` as-is for testing without building our own test harness out of asserting that `expand` runs successful. Luckily there are already compile-fail tests in `serde_macros`. Let's simply continue adding them there.

fixes #341


dtolnay:
@homu r+

dtolnay:
Neat, I didn't notice that there already was one. I think testing only serde_macros is fine.

Issue #355, opened 2016-06-06
Duplicate attributes should emit an error


dtolnay:
```rust
struct S {
    #[serde(rename(serialize="x"))]
    #[serde(rename(serialize="y"))] // error
    #[serde(rename(deserialize="y"))] // ok
    #[serde(rename="y")] // error
    z: Z,
}
```

dtolnay:
Remember to add a compile-fail test for this (requires #341).

dtolnay:
Oliver added serde_macros/tests/compile-fail/duplicate_attributes.rs_ in #383 - just needs to be enabled when this is implemented.

PR #385, opened 2016-06-15
Error on duplicate attributes

dtolnay:
Initial attempt at fixing #355. Unfortunately the errors are reported multiple times because we have multiple calls into the attribute-reading code. I am working on fixing it.


dtolnay:
I fixed the duplicate messages by centralizing all attribute-reading right at the beginning. This has the other benefit of removing `Result` almost everywhere because attribute-reading was the only thing that could fail.

I think it will be easiest to review the two commits separately.

oli-obk:
> This has the other benefit of removing Result almost everywhere because attribute-reading was the only thing that could fail.

yay!

What do you think about moving the entire attribute parsing to the early gathering?

dtolnay:
> What do you think about moving the entire attribute parsing to the early gathering?

What do you mean? I think I did this.

serde_codegen/src/ser.rs
```diff
@@ -586,7 +525,6 @@ fn serialize_struct_variant(
         .with_fields(
             fields.iter().map(|field| {
                 builder.struct_field(field.ident.expect("struct has unnamed fields"))
-                    .with_attrs(field.attrs.iter().cloned())
```
dtolnay: This line is weird. It was added [here](https://github.com/serde-rs/serde/commit/de89f95f3162a2f1404ed63940dc26ffa1bfbaf0#diff-3ffa97289153ae55d850a4b1effa6225R505). Let me know if you can figure out why. I didn't want to keep the `Vec<ast::Attribute>` just to use here. cc @erickt

oli-obk:
> What do you mean? I think I did this.

You are right. I got confused with the order of things. `attr` should probably be a sub-module of `item` now.

dtolnay:
The parsing could be but the other modules still reference the types like `attr::ContainerAttrs`, so it would need to be a pub mod. The possibilities are:

1. Types at the top level, parsing as a module under `item`. I think it will be annoying later when making changes for these to be in separate files.
3. Types and parsing at the top level (current).
2. Types and parsing as a module under `item`. I don't think there is any advantage of this over (2).

serde_codegen/src/attr.rs
```diff
@@ -234,107 +290,99 @@ impl FieldAttrs {
...
 
                     // Parse `#[serde(rename(serialize="foo", deserialize="bar"))]`
                     ast::MetaItemKind::List(ref name, ref meta_items) if name == &"rename" => {
-                        let (ser_name, de_name) = try!(get_renames(cx, meta_items));
-                        if ser_name.is_some() {
-                            field_attrs.name.serialize_name = ser_name;
-                        }
-                        if de_name.is_some() {
-                            field_attrs.name.deserialize_name = de_name;
+                        if let Ok((ser, de)) = get_renames(cx, meta_items) {
```
erickt: Aren't you silently ignoring any error from `get_renames` here? Is this on purpose?

serde_codegen/src/ser.rs
erickt: To be honest I can't remember why I did this. This might have been from some attempt to unify variant structs with regular structs in order to generically handle fields, but if removing this still passes all the unit tests then it's probably not needed, so I think it'll be okay to remove it. And if it does break things, then shame on me for not writing a unit test for this case :)

serde_macros/tests/compile-fail/duplicate_attributes.rs
```diff
@@ -0,0 +1,23 @@
+#![feature(custom_attribute, custom_derive, plugin)]
+#![plugin(serde_macros)]
+
+#[derive(Serialize)]
+struct S {
+    #[serde(rename(serialize="x"))]
+    #[serde(rename(serialize="y"))] //~ ERROR: duplicate serde attribute `rename`
```
erickt: Awesome. Could you also add these cases?

* `#[serde(rename(serialize="x", serialize="x"))]`
*  `#[serde(rename="x", serialize="x"))]`
* ``#[serde(rename(serialize="x"), rename(serialize="x"))]`

serde_codegen/src/attr.rs
```diff
@@ -234,107 +290,99 @@ impl FieldAttrs {
...
         for meta_items in field.attrs.iter().filter_map(get_serde_meta_items) {
             for meta_item in meta_items {
+                let span = meta_item.span;
                 match meta_item.node {
                     // Parse `#[serde(rename="foo")]`
                     ast::MetaItemKind::NameValue(ref name, ref lit) if name == &"rename" => {
-                        let s = try!(get_str_from_lit(cx, name, lit));
-                        field_attrs.name.serialize_name = Some(s.clone());
-                        field_attrs.name.deserialize_name = Some(s);
+                        if let Ok(s) = get_str_from_lit(cx, name, lit) {
```
erickt: here too.

serde_codegen/src/attr.rs
```diff
@@ -234,107 +290,99 @@ impl FieldAttrs {
...
                     ast::MetaItemKind::Word(ref name) if name == &"default" => {
-                        field_attrs.default = FieldDefault::Default;
+                        default.set(span, FieldDefault::Default);
                     }
 
                     // Parse `#[serde(default="...")]`
                     ast::MetaItemKind::NameValue(ref name, ref lit) if name == &"default" => {
-                        let path = try!(parse_lit_into_path(cx, name, lit));
-                        field_attrs.default = FieldDefault::Path(path);
+                        if let Ok(path) = parse_lit_into_path(cx, name, lit) {
```
erickt: here too.

serde_codegen/src/attr.rs
```diff
@@ -234,107 +290,99 @@ impl FieldAttrs {
...
-                            field_attrs.default = FieldDefault::Default;
-                        }
+                        skip_deserializing.set_true(span);
                     }
 
                     // Parse `#[serde(skip_serializing_if="...")]`
                     ast::MetaItemKind::NameValue(ref name, ref lit) if name == &"skip_serializing_if" => {
-                        let path = try!(parse_lit_into_path(cx, name, lit));
-                        field_attrs.skip_serializing_if = Some(path);
+                        if let Ok(path) = parse_lit_into_path(cx, name, lit) {
```
erickt: here too.

serde_codegen/src/attr.rs
```diff
@@ -234,107 +290,99 @@ impl FieldAttrs {
...
+                            skip_serializing_if.set(span, path);
+                        }
                     }
 
                     // Parse `#[serde(serialize_with="...")]`
                     ast::MetaItemKind::NameValue(ref name, ref lit) if name == &"serialize_with" => {
-                        let path = try!(parse_lit_into_path(cx, name, lit));
-                        field_attrs.serialize_with = Some(path);
+                        if let Ok(path) = parse_lit_into_path(cx, name, lit) {
```
erickt: here too.

serde_codegen/src/attr.rs
```diff
@@ -234,107 +290,99 @@ impl FieldAttrs {
...
+                            serialize_with.set(span, path);
+                        }
                     }
 
                     // Parse `#[serde(deserialize_with="...")]`
                     ast::MetaItemKind::NameValue(ref name, ref lit) if name == &"deserialize_with" => {
-                        let path = try!(parse_lit_into_path(cx, name, lit));
-                        field_attrs.deserialize_with = Some(path);
+                        if let Ok(path) = parse_lit_into_path(cx, name, lit) {
```
erickt: and here.

serde_codegen/src/attr.rs
```diff
@@ -234,107 +290,99 @@ impl FieldAttrs {
...
+                            deserialize_with.set(span, path);
+                        }
                     }
 
                     // Parse `#[serde(bound="D: Serialize")]`
                     ast::MetaItemKind::NameValue(ref name, ref lit) if name == &"bound" => {
-                        let where_predicates = try!(parse_lit_into_where(cx, name, lit));
-                        field_attrs.ser_bound = Some(where_predicates.clone());
-                        field_attrs.de_bound = Some(where_predicates);
+                        if let Ok(where_predicates) = parse_lit_into_where(cx, name, lit) {
```
erickt: here too.

serde_codegen/src/attr.rs
```diff
@@ -234,107 +290,99 @@ impl FieldAttrs {
...
 
                     // Parse `#[serde(bound(serialize="D: Serialize", deserialize="D: Deserialize"))]`
                     ast::MetaItemKind::List(ref name, ref meta_items) if name == &"bound" => {
-                        let (ser_bound, de_bound) = try!(get_where_predicates(cx, meta_items));
-                        if ser_bound.is_some() {
-                            field_attrs.ser_bound = ser_bound;
-                        }
-                        if de_bound.is_some() {
-                            field_attrs.de_bound = de_bound;
+                        if let Ok((ser, de)) = get_where_predicates(cx, meta_items) {
```
erickt: here too.

serde_codegen/src/attr.rs
```diff
@@ -68,55 +119,51 @@ pub struct ContainerAttrs {
...
         for meta_items in item.attrs().iter().filter_map(get_serde_meta_items) {
             for meta_item in meta_items {
+                let span = meta_item.span;
                 match meta_item.node {
                     // Parse `#[serde(rename="foo")]`
                     ast::MetaItemKind::NameValue(ref name, ref lit) if name == &"rename" => {
-                        let s = try!(get_str_from_lit(cx, name, lit));
-                        container_attrs.name.serialize_name = Some(s.clone());
-                        container_attrs.name.deserialize_name = Some(s);
+                        if let Ok(s) = get_str_from_lit(cx, name, lit) {
```
erickt: here too.

erickt:
<3 this is awesome. I always wanted to do this. Beyond resolving the apparently ignored error in the `if let Ok(...) = ...`, this looks great to me.

serde_codegen/src/attr.rs
dtolnay: Yes I am. This is on purpose. The span_err has already been emitted inside get_ser_and_de. Ignoring the error here means that we can simultaneously show errors for all of the bad attributes in a file, rather than showing them one at a time. Since there were errors, compilation will fail after the macro expansion phase.

dtolnay:
@erickt let me know if you are satisfied by my justification of why the errors should be ignored. I will add a comment explaining it, and I will add the test cases you pointed out.

oli-obk:
whoa, github now sends emails when new commits are added <3

erickt:
🎉

PR #392, opened 2016-06-17
fix fixed size array deserialization

### LINKED ISSUE #391: Fixed size array deserialization
Why does deserializer trait delegate fixed size array deserialization to [generic](https://github.com/serde-rs/serde/blob/b5c0406afe9ad1e20616fdb18a2e3f9d7dd40d1e/serde/src/de/mod.rs#L363) method?

oli-obk:
fixes #391 

@erickt @dtolnay can we make this breaking change assuming that anyone relying on this behaviour already wrote buggy code?

### DIFF
--- a/serde/src/de/mod.rs
+++ b/serde/src/de/mod.rs
@@ -365,7 +365,7 @@ pub trait Deserializer {
                                        visitor: V) -> Result<V::Value, Self::Error>
         where V: Visitor,
     {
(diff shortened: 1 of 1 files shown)


dtolnay:
I would prefer to fix this in bincode and schedule this PR for 0.8.0 with a big mention in the release notes.

PR #395, opened 2016-06-19
Fix identity_op Clippy lint

dtolnay:
Fixes https://github.com/Manishearth/rust-clippy/issues/969.


oli-obk:
wouldn't it be better to fix the span so it knows that it's an expanded span?

dtolnay:
Thanks, I will try to fix it the right way.

PR #397, opened 2016-06-20
Factor attr parsing into serde_item crate

### LINKED ISSUE #396: Public Codegen API
I have a scneario where I want to be able to know what `serde` will do with a struct field, but there isn't a public API for querying `serde_codegen`. Whether or not it's even a good idea to produce one is a good question, my scenario may be an edge-case that isn't really worth supporting.

I have some serialisation codegen that I want to be 'compatible' with `serde` attributes (so if a user serialises their struct field as `renamed`, I want to know so I can serialise my generated field as `renamed` too.) Currently, I just do some basic checks on `serde` attributes, but don't mark them as used. This is crummy because it doesn't cater for more complex renaming, and could fall out of step with `serde` itself.

The easiest way to make this work would be to just make `serde_codegen::attr` public, so a caller could access the exact same metadata as `serde` itself.

dtolnay:
Fixes #396. @KodrAus [let me know whether this fits the bill.](https://github.com/dtolnay/serde/tree/5c6a0e12e95974e3c131386e8e0528b6b9cfa6fa/serde_item/src)

a few other changes to make the API a little more presentable:

- Rename attr::{ContainerAttrs,VariantAttrs,FieldAttrs} to remove the "Attrs" (I see you worked on the corresponding [clippy lint](https://github.com/Manishearth/rust-clippy/issues/904)).
- Rename attr::Container* to attr::Item to correspond with item::Item and ast::Item. The others already had a correspondence (attr::Variant/item::Variant/ast::Variant, attr::Field/item::Field/ast::Field). Also a unit struct isn't much of a "container."
- Change item::Item::from_ast to return a meaningful error enum instead of printing a message that was hard to generalize to other uses.
- Add item::Variant.span for consistency because Item and Field already had span.
- Remove the "ident" field from attr::Name because we can just fold it into the other two fields.
- Remove attr::Name::(de)serialize_name_expr because it wasn't using the right AstBuilder in the first place.
- Rename the attr:: constructors from_item/from_variant/from_field to from_ast to line up with the item:: constructors; the signatures match.
- Remove attr's dependency on aster because we were only using it for two very simple things.

### DIFF
--- a/serde_item/src/error.rs
+++ b/serde_item/src/error.rs
@@ -1,0 +1,19 @@
+use std::error;
+use std::fmt;
+
+#[derive(Debug)]
+pub enum Error {
+    UnexpectedItemKind,
+}
+
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "expected a struct or enum")
+    }
+}
+
+impl error::Error for Error {
+    fn description(&self) -> &str {
+        "expected a struct or enum"
+    }
+}
--- a/serde_codegen/Cargo.toml
+++ b/serde_codegen/Cargo.toml
@@ -14,7 +14,14 @@ include = ["Cargo.toml", "build.rs", "src/**/*.rs", "src/lib.rs.in"]
 default = ["with-syntex"]
 nightly = ["quasi_macros"]
 nightly-testing = ["clippy"]
-with-syntex = ["quasi/with-syntex", "quasi_codegen", "quasi_codegen/with-syntex", "syntex", "syntex_syntax"]
+with-syntex = [
+    "quasi/with-syntex",
+    "quasi_codegen",
+    "quasi_codegen/with-syntex",
+    "serde_item/with-syntex",
+    "syntex",
+    "syntex_syntax",
+]
 
 [build-dependencies]
 quasi_codegen = { version = "^0.12.0", optional = true }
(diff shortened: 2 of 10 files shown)


KodrAus:
Looks perfect!

dtolnay:
@oli-obk a few other changes to make the API a little more presentable:

- Rename attr::{ContainerAttrs,VariantAttrs,FieldAttrs} to remove the "Attrs" (I see you worked on the corresponding [clippy lint](https://github.com/Manishearth/rust-clippy/issues/904)).
- Rename attr::Container* to attr::Item to correspond with item::Item and ast::Item. The others already had a correspondence (attr::Variant/item::Variant/ast::Variant, attr::Field/item::Field/ast::Field). Also a unit struct isn't much of a "container."
- Change item::Item::from_ast to return a meaningful error enum instead of printing a message that was hard to generalize to other uses.
- Add item::Variant.span for consistency because Item and Field already had span.
- Remove the "ident" field from attr::Name because we can just fold it into the other two fields.
- Remove attr::Name::(de)serialize_name_expr because it wasn't using the right AstBuilder in the first place.
- Rename the attr:: constructors from_item/from_variant/from_field to from_ast to line up with the item:: constructors; the signatures match.
- Remove attr's dependency on aster because we were only using it for two very simple things.

serde_item/src/error.rs
```diff
@@ -0,0 +1,19 @@
+use std::error;
+use std::fmt;
+
+#[derive(Debug)]
+pub enum Error {
+    UnexpectedItemKind,
```
oli-obk: :+1:

oli-obk:
Wonderful. Thanks for the super detailed clear text explanation. I copied it into the main message to make sure it goes into the merge commit.

@homu r+

PR #403, opened 2016-06-22
Bump aster and syntex in serde_codegen

KodrAus:
I think `serde_codegen` needs to be updated to use the latest versions of `syntex` and `aster`.

Include fix for: https://github.com/serde-rs/aster/issues/88

Not sure about `quasi`?


dtolnay:
Quasi doesn't work yet. https://github.com/serde-rs/quasi/pull/45

KodrAus:
Ah no worries. I'll can this PR then, since it'll probably be quicker for you to update the right dependencies when everything's ready. 👍

PR #407, opened 2016-06-23
Rename serde_item to serde_codegen_internals

### LINKED ISSUE #399: Rename serde_item
I'm not sure if the name serde_item is really capturing that crate's functionality, so I'd like us to consider some other names. Off the top of my head could be `serde_codegen_ast`, or even `serde_ast`. @dtolnay  / @oli-obk, do either of you have a better suggestion?

dtolnay:
Fixes #399.


erickt:
I'm still not quite sold on the name `serde_internals`, since that sounds like it's an internal library for the `serde` crate, rather than the `serde_codegen` crate. It's wordy, but `serde_codegen_internals` might be more appropriate. Or something else that ties it to the plugin side of serde.

dtolnay:
Erick and I discussed in IRC. I don't want to see people using `serde_codegen_internals::Field` etc in their code so let's move everything out of the top level into `::ast`. Then encourage `use serde_codegen_internals::ast as serde_ast`.

KodrAus:
If possible can I please get a ping when `serde_codegen_internals` shows up on crates.io and `serde_item` is unlisted?

dtolnay:
@KodrAus don't worry about it, I will make a PR in elasticsearch-rs.

PR #408, opened 2016-06-23
Add support for tagged values.

pyfisch:
Some serialization formats allow the addition of "tags"
to values to add additional information. These formats
include CBOR, BSON and YAML.

This commit introduces serialization and deserialization
of tags. It introduces a trait Tagger that resolves
the tag for a specific format. It supports integer,
binary, and string tags and is extensible to cover
more yet unknown formats.

By default tags are discarded at deserialization
except when the user has implemented visit_tagged_value
function.

To serialize tags the user has to call
serialize_tagged_value.

Tagged values may be of any type.

Thanks to @dtolnay, @oli-obk and @erickt for their comments on tags.

Supersedes #301

### DIFF
--- a/serde/src/ser/mod.rs
+++ b/serde/src/ser/mod.rs
@@ -41,7 +41,7 @@ pub trait Serialize {
 ///////////////////////////////////////////////////////////////////////////////
 
 /// A trait that describes a type that can serialize a stream of values into the underlying format.
-pub trait Serializer {
+pub trait Serializer: Sized {
     /// The error type that can be returned if some error occurs during serialization.
     type Error: Error;
 
@@ -345,6 +345,18 @@ pub trait Serializer {
         where V: Serialize,
     {
         self.serialize_struct_elt(key, value)
+    }
+
+    /// Serializes a tagged value.
+    ///
+    /// By default, the tag is discarded and the value is serialized as-is.
+    #[inline]
+    fn serialize_tagged_value<T, V>(&mut self,
+                                    _tag: T,
+                                    value: V) -> Result<(), Self::Error>
+        where T: ::Tagger, V: Serialize,
+    {
+        value.serialize(self)
     }
 }
 
--- a/serde/src/tagger.rs
+++ b/serde/src/tagger.rs
@@ -1,0 +1,32 @@
+/// Provides the correct tag for a given format.
+pub trait Tagger {
+    /// Returns a `u8` tag if available.
+    fn u8_tag(&self, _format: &'static str) -> Option<u8> { None }
+    
+    /// Returns a `u16` tag if available.
+    fn u16_tag(&self, _format: &'static str) -> Option<u16> { None }
+    
+    /// Returns a `u32` tag if available.
+    fn u32_tag(&self, _format: &'static str) -> Option<u32> { None }
+    
+    /// Returns a `u64` tag if available.
+    fn u64_tag(&self, _format: &'static str) -> Option<u64> { None }
+    
+    /// Returns an `i8` tag if available.
+    fn i8_tag(&self, _format: &'static str) -> Option<i8> { None }
+    
+    /// Returns an `i16` tag if available.
+    fn i16_tag(&self, _format: &'static str) -> Option<i16> { None }
+    
+    /// Returns an `i32` tag if available.
+    fn i32_tag(&self, _format: &'static str) -> Option<i32> { None }
+    
+    /// Returns an `i64` tag if available.
+    fn i64_tag(&self, _format: &'static str) -> Option<i64> { None }
+    
+    /// Provides a string tag if available.
+    fn string_tag(&self, _format: &'static str) -> Option<&str> { None }
+    
+    /// Provides a binary tag if available.
+    fn bytes_tag(&self, _format: &'static str) -> Option<&[u8]> { None }
+}
--- a/serde/src/de/mod.rs
+++ b/serde/src/de/mod.rs
@@ -663,6 +663,13 @@ pub trait Visitor {
     {
         self.visit_bytes(&v)
     }
+    
+    /// `visit_tagged_value` deserializes a tagged value.
+    fn visit_tagged_value<T, D>(&mut self, _tag: T, deserializer: &mut D) -> Result<Self::Value, D::Error>
+        where T: ::Tagger, D: Deserializer 
+    {
+        Deserialize::deserialize(deserializer)
+    }
 }
 
 ///////////////////////////////////////////////////////////////////////////////
--- a/serde/src/lib.rs
+++ b/serde/src/lib.rs
@@ -39,6 +39,7 @@ mod core {
 
 pub use ser::{Serialize, Serializer};
 pub use de::{Deserialize, Deserializer, Error};
+pub use tagger::Tagger;
 
 #[cfg(not(feature = "std"))]
 macro_rules! format {
@@ -52,4 +53,5 @@ pub mod iter;
 pub mod ser;
 #[cfg(not(feature = "std"))]
 pub mod error;
+mod tagger;
 mod utils;


serde/src/ser/mod.rs
```diff
@@ -41,7 +41,7 @@ pub trait Serialize {
 ///////////////////////////////////////////////////////////////////////////////
 
 /// A trait that describes a type that can serialize a stream of values into the underlying format.
-pub trait Serializer {
+pub trait Serializer: Sized {
```
oli-obk: is that necessary?

serde/src/tagger.rs
```diff
@@ -0,0 +1,23 @@
+/// Provides the correct tag for a given format.
+pub trait Tagger {
```
oli-obk: I wonder if this could somehow be made generic over [`Visitor`](https://serde-rs.github.io/serde/serde/de/trait.Visitor.html), so the implementor simply calls a method on the passed `Visitor`.

serde/src/ser/mod.rs
pyfisch: The compiler says so (and is usually right :grimacing:).
```
   Compiling serde v0.7.7 (file:///serde/serde)
/serde/serde/src/ser/mod.rs:359:15: 359:24 error: the trait bound `Self: std::marker::Sized` is not satisfied [E0277]
/serde/serde/src/ser/mod.rs:359         value.serialize(self)
                                                                         ^~~~~~~~~
/serde/serde/src/ser/mod.rs:359:15: 359:24 help: run `rustc --explain E0277` to see a detailed explanation
/serde/serde/src/ser/mod.rs:359:15: 359:24 help: consider adding a `where Self: std::marker::Sized` bound
error: Could not compile `serde`.
```

oli-obk: well... if you look at all the other default method impls, none call methods on the value. I'm aware of the motivation, that formats not caring about tags don't need to implement a new method. But maybe we need something along the lines of `MapVisitor` and `SeqVisitor`.

pyfisch: Yes adding a `TaggedVisitor` is possible. In this case the `Sized` bound can be removed.

pyfisch:
I have added support for tagged values [in a branch of my CBOR crate](https://github.com/pyfisch/cbor/tree/tagged_values)

serde/src/tagger.rs
pyfisch: Can you please explain this in more detail?

oli-obk: Right now you have a method for most non aggregates. What I'm wondering if it would work to pass something like a `TagDeserializer` to `visit_tagged_value` and then use the regular deserialization framework to deserialize to your target type.

So [this function](https://github.com/pyfisch/cbor/commit/96f67b5c01e264c749d4df4ff25e1005dcce87c7#diff-dd26c2766e65aca43e4228e64ebd0cf8R312) would look like that:

```rust
fn visit_tagged_value<T, D>(&mut self, tag: T, deserializer: &mut D) -> Result<Self::Value, D::Error>
                    where T: de::Deserializer, D: de::Deserializer 
{
    let val = try!(de::Deserialize::deserialize(deserializer));
    let tag = try!(de::Deserialize::deserialize(tagger));
    Ok(Value::Tag(tag, Box::new(val)))
}
```

I haven't looked at serialization closely enough yet.

pyfisch: One needs to keep in mind that many applications support multiple serialization formats. They may have colliding tags. For this reason the serialization format needs to be checked. One example is a deserializer that encounters an integer tag 0. In BSON it describes a generic binary value while in CBOR it is used for a datetime  value. The function you propose does not handle this ambiguity.

In serialization this problem is even worse: you have a tag 42 for RGBA colors used in CBOR but it must never be used with BSON.

serde/src/ser/mod.rs
```diff
@@ -346,6 +346,18 @@ pub trait Serializer {
...
         self.serialize_struct_elt(key, value)
     }
+
+    /// Serializes a tagged value.
+    ///
+    /// By default, the tag is discarded and the value is serialized as-is.
+    #[inline]
+    fn serialize_tagged_value<T, V>(&mut self,
+                                 _tag: T,
+                                 value: V) -> Result<(), Self::Error>
```
erickt: Nit: can you update the indent here?

serde/src/tagger.rs
```diff
@@ -0,0 +1,23 @@
...
+    /// Returns a `u8` tag if available.
+    fn u8_tag(&self, _format: &'static str) -> Option<u8> { None }
+    /// Returns a `u16` tag if available.
```
erickt: Nit, can you add newlines between each function?

erickt:
How do you handle tagging higher level types, like a datetime value? Would the idea be that the serializer calls `Tagger::string_tag`, which returns `Some("DateTime")`, and it's the responsibility of the serializer to interpret those tags, or error out if it doesn't know how to handle it?

Another thing to consider is that specialization is coming down the pipeline. How could that impact this design? It would be interesting if there was a design that eventually could enable serde to have a registry of generic tag names that then could be used by the different serialization formats without having to do a runtime check.

pyfisch:
I'll take the [rust-url serialization](https://github.com/servo/rust-url/blob/dc963b72e5bc4f3911b8616b25a2140e77db2e89/src/lib.rs#L1216-1235) as an example. Currently the URL is serialized as a string:

```rust
impl serde::Serialize for Url {
    fn serialize<S>(&self, serializer: &mut S) -> Result<(), S::Error> where S: serde::Serializer {
        format!("{}", self).serialize(serializer)
    }
}
```

Since CBOR recommends an optional tagging of URIs with tag 32 this information could be added to the implementation of `Serialize`. As you propose to use a registry for common types there could be a "serde-registry" format mapping types to string tags. It would be up to the formats serializer to map the generic tag to a format specific tag or to ignore it. The introduction of a serde registry is not strictly necessary, but it adds a bit of portability. Now a format the `rust-url` authors are unaware of can still tag URL values correctly if it knows the mapping from "serde-url" to an internal tag.
```rust
impl serde::Serialize for Url {
    fn serialize<S>(&self, serializer: &mut S) -> Result<(), S::Error> where S: serde::Serializer {
        struct Tagger;
        impl serde::Tagger for Tagger {
            fn u64_tag(&self, format: &'static str) -> Option<u64> {
                if format == "cbor" {
                    // URI; see RFC7049 Section 2.4.4.3
                    return Some(32);
                }
                None
            }
            
            fn string_tag(&self, format: &'static str) -> Option<&str> {
                if format == "serde-registry" {
                    return Some("serde-url");
                }
                None
            }
        }
        serializer.serialize_tagged_value(Tagger, self.as_str())
    }
}
```

pyfisch:
@dtolnay Why is this "WIP"? What needs to be changed?

oli-obk:
So I've been thinking some about this. And I still dislike the runtime decision about the tag type.

For serialization I played with impl specialization, and (I think) [it works](https://play.rust-lang.org/?gist=efe3997b10bd842a6b2a93daefc05935&version=nightly&backtrace=0)

Output for serializing with `DefaultSerializer`:

```
Tagged::serialize
S::serialize_tag (default)
DefaultSerializer::i32: 99
```

Output for serializing with `MySerializer` (that wants tags):

```
Tagged::serialize
MySerializer::serialize_tag
Tag: MySerializer::i32: 42
MySerializer::i32: 99
```

The impl specialization magic happens here:

```rust
struct Tagged(i32);

trait SerializeTag<S: Serialize>: Serializer {
    fn serialize_tag(&mut self, value: &S) -> Result<(), ()>;
}

impl SerializeTag<Tagged> for MySerializer {
    fn serialize_tag(&mut self, value: &Tagged) -> Result<(), ()> {
        println!("MySerializer::serialize_tag");
        self.serialize_tagged(42, value.0)
    }
}

impl<S: Serializer> SerializeTag<Tagged> for S {
    default fn serialize_tag(&mut self, value: &Tagged) -> Result<(), ()> {
        println!("S::serialize_tag (default)");
        value.0.serialize(self)
    }
}

impl Serialize for Tagged {
    fn serialize<S>(&self, serializer: &mut S) -> Result<(), ()>
        where S: Serializer {
        println!("Tagged::serialize");
        serializer.serialize_tag(self)
    }
}
```

I haven't put any thought into deserialization yet, but I feel confident we can rig something.

Note: this design does **not** touch the `Serialize` trait at all, and adds **zero** implementation work (I considered both, but found this "workaround" to be suitable)

oli-obk:
Ok, there's an issue with this. It works fine if the `SerializeTag` trait exists in the same crate as the type to be serialized and the `Serializer` that requires tags. I can't get it to work yet, if the `Serializer` comes from a different crate, and the local crate defines the `SerializeTag` trait and impls it. Might be a specialization-bug though

oli-obk:
Ah, nevermind, it's just horrible error reporting. If the `Serializer` is generic, make sure to specify ALL bounds that are on the `impl Serializer`.

oli-obk:
I implemented the serialization part in serde on a branch based on yours (+ I merged master into it) https://github.com/oli-obk/rust-serde/commits/tagged_values

oli-obk:
I got deserialization to mostly work. it's now failing on something really weird. I think syntex is eating my code.

dtolnay:
Closing in favor of #455 which seems more flexible and more promising. We can reopen later if it turns out otherwise.

PR #412, opened 2016-06-29
Move Token De/Serializer to serde_test crate

### LINKED ISSUE #411: Publish a serde_test crate containing the Token De/Serializer.
[This Serializer and Deserializer](https://github.com/serde-rs/serde/blob/6ab508a93c8163a12315729f00fce4c539dfebd5/serde_tests/tests/token.rs) would be really useful for writing tests in other crates that implement Serialize/Deserialize for their types.

dtolnay:
Fixes #411.


dtolnay:
@oli-obk it would be best to review the commits individually - there are a few changes to make the API more presentable, especially around the Error variant and the assert_* functions.

This is required for linked_hash_map to drop its dependency on serde_json: https://github.com/dtolnay/linked-hash-map/commit/dabca618762ed31b50bfa5ea92b5bbfc305d57c5?diff=split which is required for fixing https://github.com/serde-rs/json/issues/91.

oli-obk:
Just out of curiosity. Is there a concrete motivation for this? (not a blocker at all, just curious)

reviewing now

serde_test/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "serde_test"
+version = "0.1.0"
```
oli-obk: Since people are actually meant to use this crate, maybe we should sync versions with serde. Otherwise this becomes a piston-like mess

dtolnay: Good call. Done.

serde_test/src/error.rs
```diff
@@ -0,0 +1,75 @@
...
+    }
+}
+
+impl error::Error for Error {
+    fn description(&self) -> &str {
+        "Serde Error"
+    }
+
+    fn cause(&self) -> Option<&error::Error> {
+        None
```
oli-obk: this was this way before the refactoring, but I wonder if it should forward to `de::value::Error` in case of `ValueError`?

serde_test/src/assert.rs
```diff
@@ -22,13 +22,14 @@ pub fn assert_ser_tokens<T>(value: &T, tokens: &[Token])
     assert_eq!(ser.next_token(), None);
 }
 
-// Expect an error deserializing tokens into a T
+// Expect an error serializing T
```
oli-obk: make this a doc comment

serde_test/src/error.rs
dtolnay: I got rid of ValueError in a later commit. Nothing was using ValueError.

serde_macros/Cargo.toml
```diff
@@ -25,6 +25,7 @@ compiletest_rs = "^0.2.0"
 fnv = "1.0"
 rustc-serialize = "^0.3.16"
 serde = { version = "^0.7.11", path = "../serde" }
+serde_test = { version = "0.1", path = "../serde_test" }
```
oli-obk: What's the reason behind this?

also, the version needs to be adjusted

serde_test/src/assert.rs
dtolnay: Done.

serde_macros/Cargo.toml
dtolnay: Required by https://github.com/serde-rs/serde/blob/6ab508a93c8163a12315729f00fce4c539dfebd5/serde_macros/tests/test.rs#L7

oli-obk:
lgtm

For a public crate, the docs are very sparse. Also I'm not sure if the docs will be included in the automatically generated ones hosted by @erickt .

Both issues can be addressed later.

PR #418, opened 2016-07-03
Record expansion info

### LINKED ISSUE #350: serde_macros spans need add expansion info
clippy can't detect whether the code was expanded or not without that info

cc https://github.com/Manishearth/rust-clippy/issues/969

dtolnay:
Fixes #350 and fixes https://github.com/Manishearth/rust-clippy/issues/969.

### DIFF
--- a/serde_codegen/src/de.rs
+++ b/serde_codegen/src/de.rs
@@ -8,6 +8,7 @@ use syntax::ptr::P;
 use syntax::tokenstream::TokenTree;
 
 use bound;
+use span;
 use internals::ast::{Body, Field, Item, Style, Variant};
 use internals::{attr, Error};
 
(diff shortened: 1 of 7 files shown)


dtolnay:
@oli-obk this fixes it for `#[plugin(serde_macros)]` but not for syntex. Maybe we need both this and #395?

oli-obk:
The syntex-expanded code can't be properly linted. As [you said](https://github.com/Manishearth/rust-clippy/issues/969#issuecomment-225494037), we already attach `#[automatically_derived]`. Clippy should additionally have something that honors that attribute.

The main issue is fixed by this PR: expanded code lints trigger on the `#[derive(Deserialize)]` attribute instead of the actual location (which doesn't exist for `serde_macros`). But when using syntex, the lint triggers on the expanded code, so users will see the actual location and issue.

This needs a run-pass regression test. The one from the clippy issue + a deny should work

```rust
#![feature(custom_derive, plugin)]
#![plugin(serde_macros)]
extern crate serde_json;

#![deny(identity_op)]

#[derive(Serialize)]
struct A { b: u8 }

fn main() {}
```

r=me with a regression test

PR #421, opened 2016-07-05
Invalid-length when enum seq is too short

dtolnay:
Fixes https://github.com/serde-rs/json/issues/96.

### DIFF
--- a/serde_codegen/src/de.rs
+++ b/serde_codegen/src/de.rs
@@ -395,6 +395,7 @@ fn deserialize_seq(
     fields: &[Field],
     is_struct: bool,
 ) -> P<ast::Expr> {
+    let mut index_in_seq = 0usize;
     let let_values: Vec<_> = fields.iter()
         .enumerate()
         .map(|(i, field)| {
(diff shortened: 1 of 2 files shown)


serde_codegen/src/de.rs
```diff
@@ -420,14 +421,16 @@ fn deserialize_seq(
                         })
                     }
                 };
-                quote_stmt!(cx,
+                let assign = quote_stmt!(cx,
                     let $name = match $visit {
                         Some(value) => { value },
                         None => {
-                            return Err(_serde::de::Error::end_of_stream());
+                            return Err(_serde::de::Error::invalid_length($index_in_seq));
```
oli-obk: any reason you are not using `$i`?

dtolnay: `$i` is the index in the struct, not index in the input sequence. The second unit test tests this. With `$i` it would return `InvalidLength(2)` instead of `InvalidLength(1)`.

oli-obk: ah, that makes sense

oli-obk:
@homu r+

PR #422, opened 2016-07-05
Close the map before checking for missing fields

dtolnay:
I am working on JSON parser performance. One huge improvement (30-40%) is making the `peek()` function as cheap as possible and doing pessimistic error handling such that line/column is only computed if an error happens.

For context: the `while` loop exits when `peek()` returns a closing brace `}`. `$extract_values` checks for any missing keys. `visitor.end()` consumes the closing brace.

The current `peek()` updates the iterator position so currently "missing field" errors correctly point to the closing brace. The faster `peek()` does not update position. In order for "missing field" errors to point to the closing brace instead of the previous character, we need to consume the closing brace before checking for missing keys.

### DIFF
--- a/serde_codegen/src/de.rs
+++ b/serde_codegen/src/de.rs
@@ -1093,9 +1093,9 @@ fn deserialize_map(
             }
         }
 
-        $extract_values
-
         try!(visitor.end());
+
+        $extract_values
 
         Ok($result)
     })


oli-obk:
xml and bincode are fine with this change as I read the code. The `MapVisitor` docs don't specify anything here. This is where a serde crater tool would really be handy :)

I don't think there can be a format that expects a missing field error to have been triggered before the call to `end` (or never get triggered), but since it's breaking json's error reporting, it might break some other format's errors.

Not sure if we should hold off on this hypothetical error location breakage. Probably not.

dtolnay:
I don't think this should be considered a breaking change because it has been perfectly legitimate all along to write custom Deserialize implementations this way - the API does not specify.

PR #426, opened 2016-07-07
impl Display for de::Type

### LINKED ISSUE #425: impl Display for de::Type
`serde_json` is using `{:?}` to print it.

dtolnay:
Fixes #425.


oli-obk:
You could also implement the debug impl to forward to the display impl, but if you think that's not helpful, this PR lgtm.

dtolnay:
I subscribe to [this philosophy from the docs](https://doc.rust-lang.org/std/fmt/trait.Debug.html):

> Generally speaking, you should just `derive` a `Debug` implementation.

PR #433, opened 2016-07-11
add envy

softprops:
I recently published a new crate for deserializing structs from `std::env::vars()`. Wanted to see if I could get it listed here


dtolnay:
Nice!

Issue #219, opened 2016-01-21
Serializer and Deserializer should be pure interface


lcy03406:
For now the Serializer and Deserializer traits contain many functions, but many of them have default implementations. A problem is, the default implementations make too many assumption about the serializing protocol, probably based on json. Though them may be convenient for implementing a json serializer, or other text protocols, but most of them do not suit for a binary protocol. The point is, Serializer and Deserializer are so big, that when implementing a concrete serializer/deserializer it is quite possible to accidentally miss some functions. So it is better to remove the default implementations,    and make the Serializer and Deserializer traits a 'checklist', the compiler can help checking the implemention. In addition, the current default implementations can be moved to some JsonlikeSerializer or CommonTextSerializer sub-traits.

oli-obk:
I agree. This "magic" has caused me some headache debugging, so I ended up replacing all of them manually.

JohnHeitmann:
+1. I think magic layers are more appropriate for Deserialize/Serialize sugar, which are types that many serde users need to interact with. The -er types are relatively rarely implemented, and harder for an author to get complete ambient test coverage of via their own client code, so sacrificing sugar for correctness feels appropriate.

erickt:
I can get behind this. One downside though with making this always pure is that any new hint we add would trigger a major release, or we just add a default implementation for that method. Would everyone be okay with this?

jwilm:
@erickt adding a default implementation when adding a hint seems preferable to a major release. It's entirely backwards compatible this way, and formats which need the hinting can start using it immediately. Since this is all about getting rid of default impls, issues should be added to remove the default impl come the next major version.

PR #437, opened 2016-07-12
Revamp Map/Seq serialization

### LINKED ISSUE #386: Unable to serialize sequence without knowing all elements.
In the logging library I'm working on, I would like to serialize map elements, provided by user API calls. However at any given time, I have only one `key: value` pair.

So I can't efficiently serialize a map:

```
    fn serialize_map<V>(&mut self, visitor: V) -> Result<(), Self::Error> where V: MapVisitor;
    fn serialize_map_elt<K, V>(&mut self, key: K, value: V) -> Result<(), Self::Error> where K: Serialize, V: Serialize;
```

As I have to build a Visitor

If the API would be returning some kind of `Visitor` that one can call to serialize field after field, then both my and currently-supported use cases would work. Would it be worse in some aspect that I'm missing?

oli-obk:
I took the liberty of also simplifying some of serde codegen (mainly the parts where we do a lot of effort to create types that are going to get optimized out). I think all that was necessary to fit the visitor pattern. The new system as imagined by @dpc is much easier to implement, although it requires additional helper objects

fixes #386

cc @dpc

# Serializers compatible with the new API

* [x] serde_json https://github.com/serde-rs/json/pull/116
* [x] bincode https://github.com/TyOverby/bincode/pull/83
* [x] serde_yaml https://github.com/dtolnay/serde-yaml/pull/22
* [x] toml https://github.com/alexcrichton/toml-rs/pull/104

### DIFF
--- a/serde/src/ser/impls.rs
+++ b/serde/src/ser/impls.rs
@@ -64,8 +64,8 @@ use super::{
     Error,
     Serialize,
     Serializer,
-    SeqVisitor,
-    MapVisitor,
+    SeqSerializer,
+    MapSerializer,
 };
 
 ///////////////////////////////////////////////////////////////////////////////
@@ -133,26 +133,6 @@ impl<T> Serialize for Option<T> where T: Serialize {
     }
 }
 
-impl<T> SeqVisitor for Option<T> where T: Serialize {
-    #[inline]
-    fn visit<S>(&mut self, serializer: &mut S) -> Result<Option<()>, S::Error>
-        where S: Serializer,
-    {
-        match self.take() {
-            Some(value) => {
-                try!(serializer.serialize_seq_elt(value));
-                Ok(Some(()))
-            }
-            None => Ok(None),
-        }
-    }
-
-    #[inline]
-    fn len(&self) -> Option<usize> {
-        Some(if self.is_some() { 1 } else { 0 })
-    }
-}
-
 ///////////////////////////////////////////////////////////////////////////////
 
 impl<T> Serialize for PhantomData<T> {
@@ -164,69 +144,6 @@ impl<T> Serialize for PhantomData<T> {
     }
 }
 
-///////////////////////////////////////////////////////////////////////////////
-
-/// A `serde::Visitor` for sequence iterators.
-///
-/// # Examples
-///
-/// ```
-/// use serde::{Serialize, Serializer};
-/// use serde::ser::impls::SeqIteratorVisitor;
-///
-/// struct Seq(Vec<u32>);
-///
-/// impl Serialize for Seq {
-///     fn serialize<S>(&self, ser: &mut S) -> Result<(), S::Error>
-///         where S: Serializer,
-///     {
-///         ser.serialize_seq(SeqIteratorVisitor::new(
-///             self.0.iter(),
-///             Some(self.0.len()),
-///         ))
-///     }
-/// }
-/// ```
-pub struct SeqIteratorVisitor<Iter> {
-    iter: Iter,
-    len: Option<usize>,
-}
-
-impl<T, Iter> SeqIteratorVisitor<Iter>
-    where Iter: Iterator<Item=T>
-{
-    /// Construct a new `SeqIteratorVisitor<Iter>`.
-    #[inline]
-    pub fn new(iter: Iter, len: Option<usize>) -> SeqIteratorVisitor<Iter> {
-        SeqIteratorVisitor {
-            iter: iter,
-            len: len,
-        }
-    }
-}
-
-impl<T, Iter> SeqVisitor for SeqIteratorVisitor<Iter>
-    where T: Serialize,
-          Iter: Iterator<Item=T>,
-{
-    #[inline]
-    fn visit<S>(&mut self, serializer: &mut S) -> Result<Option<()>, S::Error>
-        where S: Serializer,
-    {
-        match self.iter.next() {
-            Some(value) => {
-                try!(serializer.serialize_seq_elt(value));
-                Ok(Some(()))
-            }
-            None => Ok(None),
-        }
-    }
-
-    #[inline]
-    fn len(&self) -> Option<usize> {
-        self.len
-    }
-}
 
 ///////////////////////////////////////////////////////////////////////////////
 
@@ -237,7 +154,12 @@ impl<T> Serialize for [T]
     fn serialize<S>(&self, serializer: &mut S) -> Result<(), S::Error>
         where S: Serializer,
     {
-        serializer.serialize_seq(SeqIteratorVisitor::new(self.iter(), Some(self.len())))
+
+        let mut seq_serializer = try!(serializer.serialize_seq());
+        for e in self.iter() {
+            try!(seq_serializer.serialize_elt(e));
+        }
+        Ok(())
     }
 }
 
@@ -250,8 +172,11 @@ macro_rules! array_impls {
             fn serialize<S>(&self, serializer: &mut S) -> Result<(), S::Error>
                 where S: Serializer,
             {
-                let visitor = SeqIteratorVisitor::new(self.iter(), Some($len));
-                serializer.serialize_fixed_size_array(visitor)
+                let mut seq_serializer = try!(serializer.serialize_seq());
+                for e in self.iter() {
+                    try!(seq_serializer.serialize_elt(e));
+                }
+                Ok(())
             }
         }
     }
@@ -301,7 +226,11 @@ impl<T> Serialize for BinaryHeap<T>
     fn serialize<S>(&self, serializer: &mut S) -> Result<(), S::Error>
         where S: Serializer,
     {
-        serializer.serialize_seq(SeqIteratorVisitor::new(self.iter(), Some(self.len())))
+        let mut seq_serializer = try!(serializer.serialize_seq());
+        for e in self.iter() {
+            try!(seq_serializer.serialize_elt(e));
+        }
+        Ok(())
     }
 }
 
@@ -313,7 +242,11 @@ impl<T> Serialize for BTreeSet<T>
     fn serialize<S>(&self, serializer: &mut S) -> Result<(), S::Error>
         where S: Serializer,
     {
-        serializer.serialize_seq(SeqIteratorVisitor::new(self.iter(), Some(self.len())))
+        let mut seq_serializer = try!(serializer.serialize_seq());
+        for e in self.iter() {
+            try!(seq_serializer.serialize_elt(e));
+        }
+        Ok(())
     }
 }
 
@@ -325,7 +258,10 @@ impl<T> Serialize for EnumSet<T>
     fn serialize<S>(&self, serializer: &mut S) -> Result<(), S::Error>
         where S: Serializer,
     {
-        serializer.serialize_seq(SeqIteratorVisitor::new(self.iter(), Some(self.len())))
+        let mut seq_serializer = try!(serializer.serialize_seq());
+        for e in self.iter() {
+            try!(seq_serializer.serialize_elt(e));
+        }
     }
 }
 
@@ -338,7 +274,11 @@ impl<T, H> Serialize for HashSet<T, H>
     fn serialize<S>(&self, serializer: &mut S) -> Result<(), S::Error>
         where S: Serializer,
     {
-        serializer.serialize_seq(SeqIteratorVisitor::new(self.iter(), Some(self.len())))
+        let mut seq_serializer = try!(serializer.serialize_seq());
+        for e in self.iter() {
+            try!(seq_serializer.serialize_elt(e));
+        }
+        Ok(())
     }
 }
 
@@ -350,7 +290,11 @@ impl<T> Serialize for LinkedList<T>
     fn serialize<S>(&self, serializer: &mut S) -> Result<(), S::Error>
         where S: Serializer,
     {
-        serializer.serialize_seq(SeqIteratorVisitor::new(self.iter(), Some(self.len())))
+        let mut seq_serializer = try!(serializer.serialize_seq());
+        for e in self.iter() {
+            try!(seq_serializer.serialize_elt(e));
+        }
+        Ok(())
     }
 }
 
@@ -361,10 +305,13 @@ impl<A> Serialize for ops::Range<A>
 {
     #[inline]
     fn serialize<S>(&self, serializer: &mut S) -> Result<(), S::Error>
-        where S: Serializer,
+        where S: Serialize,
     {
-        let len = iter::Step::steps_between(&self.start, &self.end, &A::one());
-        serializer.serialize_seq(SeqIteratorVisitor::new(self.clone(), len))
+        let mut seq_serializer = try!(serializer.serialize_seq());
+        for e in self.iter() {
+            try!(seq_serializer.serialize_elt(e));
+        }
+        Ok(())
     }
 }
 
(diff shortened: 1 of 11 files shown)


serde/src/ser/impls.rs
```diff
@@ -363,8 +305,11 @@ impl<A> Serialize for ops::Range<A>
     fn serialize<S>(&self, serializer: &mut S) -> Result<(), S::Error>
         where S: Serializer,
     {
-        let len = iter::Step::steps_between(&self.start, &self.end, &A::one());
-        serializer.serialize_seq(SeqIteratorVisitor::new(self.clone(), len))
+        let mut seq_serializer = try!(serializer.serialize_seq(Some(self.len())));
```
dtolnay: Are you sure this is the same `len` as the original code?

```rust
iter::Step::steps_between(&self.start, &self.end, &A::one())
```

serde/src/ser/mod.rs
```diff
@@ -202,53 +206,60 @@ pub trait Serializer {
...
+    fn serialize_seq_end(&mut self, len: Option<usize>) -> Result<(), Self::Error>;
+
     /// Serializes a tuple.
     ///
     /// By default this serializes a tuple as a sequence.
     #[inline]
-    fn serialize_tuple<V>(&mut self, visitor: V) -> Result<(), Self::Error>
-        where V: SeqVisitor,
+    fn serialize_tuple(&mut self, len: usize) -> Result<(), Self::Error>
     {
```
dtolnay: Please put this brace on the same line as `fn serialize_tuple`

serde/src/ser/mod.rs
```diff
@@ -202,53 +206,60 @@ pub trait Serializer {
...
     }
 
     /// Serializes a tuple element.
     ///
     /// By default, tuples are serialized as a sequence.
     #[inline]
     fn serialize_tuple_elt<T>(&mut self, value: T) -> Result<(), Self::Error>
-        where T: Serialize
-    {
+        where T: Serialize {
```
dtolnay: I liked this before.

Looks like you meant to move the brace of the previous function, not this one.

dpc: BTW. Is serde using `cargo fmt`?

dtolnay: No :cry: I filed #438 to try it out.

dpc:
Oh, so you went with a route of getting everything into `Serializer` and adding "start sequence", "serialize element" * x, "end sequence"? Looks OK to me (the API, as I didn't thoroughly review the code) - the inverstion of control is gone.

Just for learning purposes: did you hit any blockers with the breaking up API into `Serializer` `MapSerializer and `SeqSerializer`? Any other noteworthy Rust limitation (or just problems with breaking it up)?

dtolnay:
I am so excited for this.

I only made it partway through but I like the approach. I will review more thoroughly tonight or tomorrow.

This may make tuples and structs a lot faster because they no longer have to do `match self.index` when serializing each element. On the other hand some `Serializer`s will need to do a lot more work bookkeeping (and maybe more memory allocation) in order to associate the seq_elt and seq_end calls with the corresponding start call. We may need to provide a way to thread some state through these calls without allocation. For example instead of this:

```rust
try!(serializer.serialize_seq($len));
for e in self.iter() {
    try!(serializer.serialize_seq_elt(e));
}
serializer.serialize_seq_end(Some(self.len()))
```

... something like this:

```rust
let mut seq_state = try!(serializer.serialize_seq($len));
for e in self.iter() {
    try!(serializer.serialize_seq_elt(&mut seq_state, e));
}
serializer.serialize_seq_end(seq_state)
```

... or equivalently:

```rust
let mut seq_ser = try!(serializer.serialize_seq($len));
for e in self.iter() {
    try!(seq_ser.elt(e));
}
seq_ser.end()
```

I think the last one is more similar to some of the proposals in #386.

I would like to have working commits against serde_json, bincode, serde_yaml and maybe one more before merging this, just so we validate it against fairly different use cases.

dpc:
@dtolnay : Yes, so:

```
let mut seq_ser = try!(serializer.serialize_seq(Some(self.len())));
for e in self.iter() {
    try!(seq_ser.elt(e));
}
seq_ser.end()
```
Has a problem of how will `seq_ser` access original `serializer`.

What I've mentioned in #386 , where`serializer.serialize_seq` takes `serializer` by value and wraps it, and then returns it back in `seq_ser.end()` step. This way only once sequence can be serialized at once, and API becomes harder to use wrong.

But I guess the problem is that `serializer` needs to be passed by value everywhere then and returned back to the caller. So:

```
fn serialize<S>(&self, serializer: &mut S) -> Result<(), S::Error>
         where S: Serializer;
```

would have to become:

```
fn serialize<S>(&self, serializer: S) -> Result<S, S::Error>
         where S: Serializer;
```
or something like that.

Would that hinder the API too much? Does this introduce additional copying?

Nice alternative where `seq_ser` is actually borrowing `serializer` for the time of it's existance does not work well with traits, as it requires HKT(?).

What about:

```
let mut seq_ser = try!(serializer.serialize_seq(Some(self.len())));
for e in self.iter() {
    try!(seq_ser.elt(e, serializer));
}
seq_ser.end(serializer)
```

This way `seq_ser` does not have to keep reference to `serializer` as it will be provided with it on every `SeqSer::elt(...)`. API can still be misused:


```
let mut seq_ser1 = try!(serializer.serialize_seq(Some(self.len())));
let mut seq_ser2 = try!(serializer.serialize_seq(Some(self.len())));
try!(seq_ser1.elt(e, serializer));
try!(seq_ser2.elt(e, serializer));
seq_ser2.end(serializer);
seq_ser1.end(serializer);
```

and user can forget about calling `SeqSer::end`, but it would still be improvement, I guess.

dtolnay:
It looks like he actually implemented exactly your "what about" approach but then got rid of it in [this commit](https://github.com/serde-rs/serde/pull/437/commits/fb575225bc7e4a536a2b6756acc75e26c54cc82f#diff-d10a4862a2935aa58736fc9c43538657).

dtolnay:
I tried a few variations of having separate SeqSerializer and MapSerializer rather than having all the methods on the Serializer trait, and they were all confusing or limited. I am satisfied with the approach in this PR. I still think it will be beneficial to some serializers to have associated types in which they can track state - we can default them to `()`.

```rust
// I don't know which of these really need to be distinct types...
type SeqState = ();
type MapState = ();
type StructState = ();
type TupleState = ();
type StructVariantState = ();
type TupleVariantState = ();
type FixedSizeArrayState = ();
```

```rust
let mut seq_state = try!(serializer.serialize_seq(Some(self.len())));
for e in self.iter() {
    try!(serializer.serialize_seq_elt(&mut seq_state, e));
}
serializer.serialize_seq_end(seq_state)
```

Also I noticed that YAML does something weird where it calls the visitor using a Serializer that is different from the one on which serialize_tuple_variant/serialize_struct_variant were called: [yaml/src/ser.rs#L168-L171](https://github.com/dtolnay/serde-yaml/blob/71771bf6f15596e5bc12ea2e3193b6f18ef3deed/yaml/src/ser.rs#L168-L171). We need to figure out how that maps to the new API.

oli-obk:
> ... or equivalently:

actually that was my first design, but it got very tedious to keep it in sync with my very frequent changes to adapt to situations I hadn't considered. We can add this without a problem, but during the design phase I'd rather stay with manual control.

@dpc: borrowing from the `Serializer` is not a problem:

```rust
fn serialize_seq<'a>(&'a mut self) -> Result<SeqSerializer<'a, Self>, Self::Error>
```

The final `end` call then consumes the object. In debug mode we can add an additional `Drop` impl that panics, so tests and debug runs will hit that if the code path is tested and `end` is forgotten.

If we ever need the `Serializer` back in the same scope, we can have `end` return it.

Also I'll add a clippy lint that warns about mis-use. These checks combined should keep everyone from mis-using the API.

dtolnay:
@oli-obk:

> ```rust
fn serialize_seq<'a>(&'a mut self) -> Result<SeqSerializer<'a, Self>, Self::Error>
```

I don't think that works if the Serializer has an explicit lifetime as in [serde-yaml](https://github.com/dtolnay/serde-yaml/blob/71771bf6f15596e5bc12ea2e3193b6f18ef3deed/yaml/src/ser.rs#L24) because the trait can't require the Serializer's lifetime to outlive `serialize_seq`'s `'a`.

**Comment by dtolnay** [comment]
I can take care of trying this out on serde-yaml. I filed https://github.com/dtolnay/serde-yaml/issues/21 to track that.

**Comment by oli-obk** [comment]
> I still think it will be beneficial to some serializers to have associated types in which they can track state - we can default them to ().

Jup, json-value can profit from it and `toml-rs` needs it, otherwise the `toml-rs` serializer needs an internal stack of temporary stuff.

I'll add it now

**Comment by oli-obk** [comment]
> we can default them to ().

sadly not on stable rust

```
error: associated type defaults are unstable (see issue #29661)
help: add #![feature(associated_type_defaults)] to the crate attributes to enable
```

**Comment by oli-obk** [comment]
toml works now, adjusting the others is trivial (just add the `MapState` and `SeqState` associated types as `()` and add the `state: ()` arg to the `*_end` methods

File: serde/src/ser/impls.rs (hunk 1 above)
Comment (oli-obk) [inline serde/src/ser/impls.rs]: copy paste error

**Comment by oli-obk** [comment]
> I don't know which of these really need to be distinct types...

We can't make them distinct types, because `serialize_struct` forwards to `serialize_map`. We can of course require some conversion methods, but that gets messy... I implemented the simplest version with just a `MapState` and a `SeqState` (it was enough for toml, and while json-value might profit from something else, I don't think it's worth it, as json-value already has a stack of values that we can re-use (and that's what I actually did))

**Comment by dtolnay** [comment]
I think json-value using a stack of state is an indication of a deficiency in the Serializer API. Even if it is a little more work, I would like to have it use MapState/SeqState just to make sure those resolve the deficiency.

File: serde/src/ser/mod.rs [hunk 4]
```diff
@@ -261,123 +278,126 @@ pub trait Serializer {
...
 
-    /// Serializes a map element (key-value pair).
-    fn serialize_map_elt<K, V>(&mut self, key: K, value: V) -> Result<(), Self::Error>
-        where K: Serialize,
-              V: Serialize;
+    /// Serialize a map.
+    fn serialize_map(&mut self, len: Option<usize>) -> Result<Self::MapState, Self::Error>;
+
+    /// Serialize a map element
+    fn serialize_map_elt<K, V>(&mut self, key: K, value: V) -> Result<(), Self::Error> where K: Serialize, V: Serialize;
```
dtolnay: This may need access to `&mut state`. Same for `serialize_seq_elt`.

serde/src/ser/mod.rs
```diff
@@ -261,123 +278,126 @@ pub trait Serializer {
...
+
+    /// Finishes serializing a map
+    fn serialize_map_end(&mut self, len: Option<usize>, state: Self::MapState) -> Result<(), Self::Error>;
```
dtolnay: Let's not pass `len` a second time. If they need it, stick it in `state`.

serde/src/ser/mod.rs
```diff
@@ -261,123 +278,126 @@ pub trait Serializer {
...
-    {
+    /// By default, structs are serialized as a map with the field name as the key.
+    fn serialize_struct_elt<K, V>(&mut self, key: K, value: V) -> Result<(), Self::Error> where K: Serialize, V: Serialize {
         self.serialize_map_elt(key, value)
     }
 
+    /// Finishes serializing a struct
+    ///
+    /// By default, structs are serialized as a map with the field name as the key.
+    fn serialize_struct_end(&mut self,
```
dtolnay: `_name` and `len` can both go in `state` instead of being passed again.

serde/src/ser/mod.rs
```diff
@@ -202,53 +212,60 @@ pub trait Serializer {
...
+        self.serialize_seq_end(Some(len), state)
+    }
+
     /// Serializes a fixed-size array.
     ///
     /// By default this serializes an array as a sequence.
     #[inline]
-    fn serialize_fixed_size_array<V>(&mut self, visitor: V) -> Result<(), Self::Error>
-        where V: SeqVisitor,
+    fn serialize_fixed_size_array(&mut self, size: usize) -> Result<Self::SeqState, Self::Error>
```
dtolnay: Let's remove this method. [No Serializer on GitHub uses it.](https://github.com/search?l=rust&q=serialize_fixed_size_array&type=Code&utf8=%E2%9C%93) Why would anything treat it differently from `serialize_seq(Some(size))`?

serde/src/ser/mod.rs
```diff
@@ -261,123 +278,126 @@ pub trait Serializer {
...
     ///
-    /// By default, struct elements are serialized as a map element with the field name as the key.
-    #[inline]
-    fn serialize_struct_elt<V>(&mut self,
-                               key: &'static str,
-                               value: V) -> Result<(), Self::Error>
-        where V: Serialize,
-    {
+    /// By default, structs are serialized as a map with the field name as the key.
+    fn serialize_struct_elt<K, V>(&mut self, key: K, value: V) -> Result<(), Self::Error> where K: Serialize, V: Serialize {
```
dtolnay: Why did you turn `key` from `&'static str` into `K`?

Separate question: do we need it to be `'static` or is `&str` good enough?

serde/src/ser/mod.rs
```diff
@@ -261,123 +278,126 @@ pub trait Serializer {
...
     ///
-    /// By default, struct variant elements are serialized as a struct element.
-    #[inline]
-    fn serialize_struct_variant_elt<V>(&mut self,
-                                       key: &'static str,
-                                       value: V) -> Result<(), Self::Error>
-        where V: Serialize,
-    {
+    /// By default, structs are serialized as a map with the field name as the key.
+    fn serialize_struct_variant_elt<K, V>(&mut self, key: K, value: V) -> Result<(), Self::Error> where K: Serialize, V: Serialize {
```
dtolnay: Same questions as on `serialize_struct_elt`.

dtolnay:
I opened https://github.com/oli-obk/rust-serde/pull/1 with a Serializer API that I believe is more flexible and more consistent. It also addresses https://github.com/serde-rs/serde/issues/219.

dtolnay:
`serde_codegen` and `serde_test` need to be updated for my changes. I will start doing it in the morning if you do not get to it.

oli-obk:
all serializers have been brought up to date.

One more experiment would be to check if we now can get rid of the stack in the serde-value serializer

erickt:
This is so cool. You all are awesome. We never quite got to the whole "the type system enforces the serde protocol" so might as well not constrain ourselves to having all these visitors.

Have you also considered how this approach might affect the deserialization API?

erickt:
I'm a +1 on this (once the tests pass), and definitely curious if this has any performance implications as well. Great work.

dtolnay:
I read through the whole PR one more time and this looks good to me. Nicely done @oli-obk.

@erickt I think Visitors still make sense in the Deserializer API to allow the Deserializer to push values out to the Deserialize, for example `serde_json::Value`.

dpc:
I've just looked and `Visitor` pattern in `de` seems OK to me. It's still "push", as per my original comment about `ser::Vistior` (which was "pull").

Nice. I'm happy about the result.

dtolnay:
I merged this along with 855f3d9 to remind us that master is no longer 0.7. I made an 0.7.x branch in case we need any more 0.7 releases.

KodrAus:
A bit late to the party, but looking through this it looks like it'll eliminate a _lot_ of code in my manual struct serialisation :+1:

I'm guessing it'll still be a little while before this hits crates.io since codegen will need updating?

dtolnay:
Codegen has been updated in this PR. I would expect sometime next week - there are a few other fixes we want in 0.8.

dtolnay:
For those subscribed to this issue - I filed #447 to track the release of these changes.

Issue #438, opened 2016-07-12
Consider using rustfmt


dtolnay:

dtolnay:
Overall I am very happy with the formatting in #439. There are a few bugs to iron out and maybe we need to develop new idioms around chained aster builders but I think rustfmt will be a net benefit.

oli-obk:
Should we add a rustfmt target to travis? Doesn't need to fail the build, but just so we know the state of things?

oli-obk:
note that we fail the build in clippy if rustfmt fails, but there have been some voice speaking against that.

dtolnay:
I don't think that would be valuable. The tradeoff is scaring away new contributors. If someone makes a PR, I would prefer that they remember it as a reasonable discussion about meaningful things, rather than as a struggle to satisfy a grumpy buildbot. Even if the overall build is green, seeing a red task can be anxiety-inducing and uncertain about whether you are responsible for fixing it. Instead, if the code looks mostly fine then I won't notice or care anyway, and somebody will run rustfmt every month or so and it takes 10 seconds to clean everything up so there isn't really anything to gain from enforcing it.

oli-obk:
Sounds sensible!

PR #439, opened 2016-07-13
Use rustfmt

dtolnay:
Just to see what this would look like. Let's wait for #437 to merge before moving forward with this so that we avoid merge conflicts. This PR is much easier to rebase.

I have not read through this carefully but I am making a PR so we can discuss the formatting and file rustfmt issues as necessary. I already filed https://github.com/rust-lang-nursery/rustfmt/issues/1094.

One thing that got much worse is nested aster builders. Before:

```rust
    let wrapper_ty = builder.path()
        .segment("__SerializeWith")
            .with_generics(wrapper_generics.clone())
            .build()
        .build();
```

After:

```rust
    let wrapper_ty = builder.path()
        .segment("__SerializeWith")
        .with_generics(wrapper_generics.clone())
        .build()
        .build();
```

Ref #438.

### DIFF
--- a/serde/src/bytes.rs
+++ b/serde/src/bytes.rs
@@ -1,6 +1,6 @@
 //! Helper module to enable serializing bytes more efficiently
 
-use core::{ops, fmt, char, iter, slice};
+use core::{char, fmt, iter, ops, slice};
 use core::fmt::Write;
 
 use ser;
@@ -11,7 +11,7 @@ pub use self::bytebuf::{ByteBuf, ByteBufVisitor};
 #[cfg(feature = "collections")]
 use collections::Vec;
 
-///////////////////////////////////////////////////////////////////////////////
+/// ////////////////////////////////////////////////////////////////////////////
 
 /// `Bytes` wraps a `&[u8]` in order to serialize into a byte array.
 #[derive(Clone, Copy, Eq, Hash, PartialEq, PartialOrd, Ord)]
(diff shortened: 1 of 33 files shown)


testing/tests/test_macros.rs
```diff
@@ -39,91 +34,81 @@ struct DeNamedMap<A, B, C> {
...
-        e: &'a mut D,
-        //f: <D as Trait>::Type,
+        // d: C::Type,
+        e: &'a mut D, // f: <D as Trait>::Type,
     },
 }
 
 #[derive(Debug, PartialEq, Serialize, Deserialize)]
-enum DeEnum<B, C: /* Trait */, D> /* where D: Trait */ {
+enum DeEnum<B, C /* : /* Trait */, D> {
```
dtolnay: Well this is a bug.

dtolnay: Wait what? How does the old one compile?

```rust
enum DeEnum<B, C: /* Trait */, D> /* where D: Trait */ {
```

without comments:

```rust
enum DeEnum<B, C: , D> {
```

dtolnay: Filed https://github.com/rust-lang-nursery/rustfmt/issues/1095

dtolnay: Also filed https://github.com/rust-lang-nursery/rustfmt/issues/1096 related to this line.

I bet all these comments are remnants of an erickt debugging session so we need to clean them up anyway.

serde/src/bytes.rs
```diff
@@ -11,7 +11,7 @@ pub use self::bytebuf::{ByteBuf, ByteBufVisitor};
 #[cfg(feature = "collections")]
 use collections::Vec;
 
-///////////////////////////////////////////////////////////////////////////////
+/// ////////////////////////////////////////////////////////////////////////////
```
jwilm: This seems undesirable; the long string of `/` will show up in the docs.

dtolnay: As I noted in the comment, I already filed a rustfmt issue about this.

jwilm: Ah, sorry. I didn't actually jump to the rustfmt issue links.

erickt:
@dtolnay: Yeah I thought about those aster chains being a problem for rustfmt. One way of handling this we could change the aster api to use closures instead of doing my silly chaining. So your example before could look like:

```rust
    let wrapper_ty = builder.path()
        .segment("__SerializeWith", |b| {
            b.with_generics(wrapper_generics.clone()).build()
        })
        .build();
```

Which could look much nicer. Or we just lift these segments out into a variable like:

```rust
    let segment = builder.segment("__SerializeWith")
        .with_generics(wrapper_generics.clone())
        .build();
    let wrapper_ty = builder.path()
        .with_segment(segment)
        .build();
```

Or even just create the variable in place with:

```rust
    let wrapper_ty = builder.path()
        .with_segment(
            builder.segment("__SerializeWith")
                .with_generics(wrapper_generics.clone())
                .build()
        )
        .build();
```

PR #442, opened 2016-07-15
Use cargo override instead of relative paths

dtolnay:
This makes it possible to use `cargo clone` + `cargo build`.

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,0 +1,7 @@
+paths = [
+  "serde",
+  "serde_codegen",
+  "serde_codegen_internals",
+  "serde_macros",
+  "serde_test",
+]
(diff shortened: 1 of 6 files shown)


dtolnay:
The build worked except for Clippy.

oli-obk:
I have never used cargo clone or path overrides, but if it still works when I do `git pull` I'm fine with this.

dtolnay:
The point is you can build all these crates from within the serde repo (they will depend on other crates within the same clone) or by themselves (they will depend on released versions from crates.io).

Here is the relevant output from Travis that shows things still work as expected:

```
$ (cd serde_macros && travis-cargo --only nightly test -- --features nightly-testing)
[...]
   Compiling serde_codegen_internals v0.3.0 (file:///home/travis/build/serde-rs/serde/serde_codegen_internals)
[...]
   Compiling serde_codegen v0.7.14 (file:///home/travis/build/serde-rs/serde/serde_codegen)
```
```

PR #446, opened 2016-07-17
Rename feature nightly-testing to unstable-testing

### LINKED ISSUE #404: Rename feature "nightly-testing" to "unstable-testing"
From what I understand, the community has started to use the feature `unstable*` instead of `nightly*` for things that may depend on the nightly compiler. If this is the case, when we do 0.8, we should consider renaming this feature.

dtolnay:
Closes #404.


dtolnay:
The failure seems like a Cargo bug... it is using the features of the released 0.8.0-rc1 instead of the features in the local override.

PR #448, opened 2016-07-18
add Hjson

laktak:


README.md
```diff
@@ -358,7 +358,7 @@ impl serde::Serialize for i32 {
 As you can see it's pretty simple. More complex types like `BTreeMap` need to
 pass a
 [MapVisitor](http://serde-rs.github.io/serde/serde/serde/ser/trait.MapVisitor.html)
-to the
```
oli-obk: please leave those spaces in there, I know they are unfortunate, but otherwise the next line is very long.

oli-obk: ah, looking at the rendered version, this doesn't seem to be an issue

oli-obk:
thanks! great to have more (de)serializers here.

README.md
laktak: OK, sorry. That was my editor and I have `git diff` set to ignore space.

Issue #447, opened 2016-07-18
Serde 0.8.0


dtolnay:
Issue for tracking and discussion related to the upcoming 0.8.0 release.

Let's all do a thorough read through the entire public API and see if there is anything we do not want to live with for the next several months.

The big ticket items remaining are:

- [x] Various bounds-related codegen issues which are all secretly the same issue #436 #441 #443. This is next on my list.
- [x] Remove default implementations from the Deserializer trait as per #219. This is a lot of tedious work because we have 17 implementations in the codebase and Deserializer has 31 methods.
- [x] Look at existing Deserializer implementations across GitHub and see whether there are any common helpers that would be good for us to own, like we do with [all of these](https://serde-rs.github.io/serde/serde/de/trait.Deserializer.html#implementors).
- [x] Decide what to do about #444. It may or may not require a breaking change - if not, it can wait.

Aside from that, now is a good time for documentation improvements and code cleanup as many developers will be looking at docs as they migrate from 0.7.

dtolnay:
The big ticket items remaining are:

- Various bounds-related codegen issues which are all secretly the same issue #436 #441 #443. This is next on my list.
- Remove default implementations from the Deserializer trait as per #219. This is a lot of tedious work because we have 17 implementations in the codebase and Deserializer has 31 methods. Let's also look at existing Deserializer implementations across GitHub and see whether there are any common helpers that would be good for us to own, like we do with [all of these](https://serde-rs.github.io/serde/serde/de/trait.Deserializer.html#implementors).
- Decide what to do about #444. It may or may not require a breaking change - if not, it can wait.

Aside from that, now is a good time for documentation improvements and code cleanup as many developers will be looking at docs as they migrate from 0.7.

oli-obk:
> This is a lot of tedious work because we have 17 implementations in the codebase and Deserializer has 31 methods. Let's also look at existing Deserializer implementations across GitHub and see whether there are any common helpers 

is it possible to specify a macro that takes method names as arguments and produces the default methods?

dtolnay:
> is it possible to specify a macro that takes method names as arguments and produces the default methods?

Yes, for internal use but we need to be very careful about what we make public. The reason for removing the defaults was because it is hard to know what makes a good default for a "typical" Deserializer.

dtolnay:
Here is what adoption looks like so far (not counting the 100+ google APIs which are really one project):

```
     54 ^0.7
     34 ^0.7.0
     21 ^0.6
[...]

PR #449, opened 2016-07-18
add more extensive docs

oli-obk:
cc #447


serde/src/ser/impls.rs
```diff
@@ -1,4 +1,8 @@
-//! Implementations for all of Rust's builtin types.
+//! Implementations for all of Rust's builtin types. Tuples implement the `Serialize` trait if they
+//! have less than 16 fields. Arrays implement the `Serialize` trait if their length is 32 or less.
```
dtolnay: At most 16 fields: I see an `impl<T0, T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12, T13, T14, T15>`.

PR #452, opened 2016-07-20
Pure interface for Deserializer

### LINKED ISSUE #219: Serializer and Deserializer should be pure interface

oli-obk:
fixes #219

This also makes the `de_forward_to_deserialize` macro public, which forwards every given method to the `deserialize` method. This is mostly interesting for `json`, `yaml`, `xml` and friends


serde/src/de/mod.rs
```diff
@@ -245,211 +307,147 @@ pub trait Deserializer {
     /// This method hints that the `Deserialize` type is expecting a `bool` value.
     #[inline]
```
dtolnay: Remove the `#[inline]` from all of these functions. They do nothing.

serde/src/de/mod.rs
```diff
@@ -10,6 +10,68 @@ use collections::{String, Vec};
...
+        de_forward_to_deserialize!{named: deserialize_newtype_struct}
+    };
+    (func: deserialize_tuple) => {
+        de_forward_to_deserialize!{tup_fn: deserialize_tuple}
+    };
+    (func: deserialize_fixed_size_array) => {
+        de_forward_to_deserialize!{tup_fn: deserialize_fixed_size_array}
+    };
+    (func: deserialize_tuple_struct) => {
+        fn deserialize_tuple_struct<__V>(&mut self, _: &str, _: usize, visitor: __V) -> Result<__V::Value, Self::Error>
```
dtolnay: Add `#[inline]` to all of these functions.

serde/src/de/mod.rs
```diff
@@ -245,211 +307,147 @@ pub trait Deserializer {
...
     #[inline]
     fn deserialize_bool<V>(&mut self, visitor: V) -> Result<V::Value, Self::Error>
-        where V: Visitor,
-    {
-        self.deserialize(visitor)
-    }
+        where V: Visitor;
 
     /// This method hints that the `Deserialize` type is expecting an `usize` value.
+    /// A reasonable default is to forward to `deserialize_u64` and cast the yielded value.
```
dtolnay: I don't think "and cast the yielded value" accurately captures what is going on here. The Visitor has an associated type `Value: Deserialize` and it has a method `visit_u64` which gives back `Value`. If `deserialize_usize` receives a visitor whose `Value` is `usize`, and `deserialize_usize` forwards to `deserialize_u64`, and `deserialize_u64` calls `visit_u64` on the Visitor, really it would be the `Visitor` that is responsible for any casting to go from u64 to usize.

I would stick with "A reasonable default is to forward to \`deserialize_u64\`." Similar in the other methods.

serde/src/de/mod.rs
```diff
@@ -245,211 +307,147 @@ pub trait Deserializer {
...
-        self.deserialize(visitor)
-    }
+        where V: Visitor;
 
     /// This method hints that the `Deserialize` type is expecting a fixed size array. This allows
     /// deserializers to parse arrays that aren't tagged as arrays.
-    ///
-    /// By default, this deserializes arrays from a sequence.
     #[inline]
     fn deserialize_fixed_size_array<V>(&mut self,
```
dtolnay: Let's call this `deserialize_seq_fixed_size`.

In the Serializer we renamed `serialize_fixed_size_array` to `serialize_seq_fixed_size` to make it a little less confusing that you need to use it with `serialize_seq_elt` and `serialize_seq_end` (instead of some `serialize_fixed_size_array_elt` and `serialize_fixed_size_array_end`) and `SeqState` (instead of `FixedSizeArrayState`).

That reasoning does not apply here, but it would be nice for the name to be consistent across Serializer and Deserializer.

dtolnay:
LGTM. Ready to merge when Travis says so.

oli-obk:
travis had a weird bug with the new macro, but I can't reproduce locally...

oli-obk:
stable, 1.8 and 1.9 must have some different macro resolution rules, because beta works fine.

nightly fails on something that I can't reproduce, but travis and I have the same nightly.

I'll investigate tomorrow, but if you want to get this through fast, feel free to grab it.

Issue #436, opened 2016-07-11
Only emit bounds that contain type parameters


dtolnay:
Currently we generate something like:

```rust
struct S {
    i: i32,
}

impl Serialize for S where i32: Serialize { /* ... */ }
```

PR #451, opened 2016-07-20
Mention Empty Bounds for Recursion in Readme

killercup:
Fixes #427

PR #453, opened 2016-07-20
Mention Empty Field Bounds for Recursion

killercup:
Please excuse my stupidity. This corrects #451

### DIFF
--- a/README.md
+++ b/README.md
@@ -762,15 +762,15 @@ how types are serialized. Here are the supported annotations:
 
 Container Annotations:
 
-| Annotation                                  | Function                                                                                                                                                                                      |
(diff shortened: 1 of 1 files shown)


dtolnay:
I would rather explain why this sometimes works rather than simply saying to set it to `""`, because that is not a magic bullet for recursion and will not work in many cases.

dtolnay:
Also it will be made unnecessary sometime in the next few days for 0.8.0 once I fix #436.

dtolnay:
Closing because this is no longer necessary as of #456. I will revert the other commit.

killercup:
Thanks, @dtolnay! :)

PR #459, opened 2016-07-22
PhantomData<T> does not require bounds on T

dtolnay:


serde_codegen/src/bound.rs
```diff
@@ -79,6 +79,13 @@ pub fn with_bound<F>(
     }
     impl visit::Visitor for FindTyParams {
         fn visit_path(&mut self, path: &ast::Path, _id: ast::NodeId) {
+            if let Some(seg) = path.segments.last() {
+                if seg.identifier.name.as_str() == "PhantomData" {
```
oli-obk: While it's not great to implement a custom type called Phantom data, i think this should check the rest of the path, too.

dtolnay: I don't think we always get the full path here... but I will try it.

oli-obk: Oh..we're in the ast, i keep forgetting... Well, it's good enough :)

oli-obk:
@homu r+

PR #462, opened 2016-07-23
Fix typos in docstring of ser::Error.

jimmycuadra:
Just a minor documentation fix for grammar and and a copy/paste error.


oli-obk:
Good catch. Thanks

@homu r+

PR #465, opened 2016-07-26
Remove underscore from parameter names in deserializer traits

dtolnay:
This makes the generated rustdoc easier to read.


oli-obk:
Yea.. This is an issue with Rust. I've seen some threads talking about it...

PR #468, opened 2016-07-27
Split serialize_map_elt

sfackler:
Like what's been done on the deserialization side with MapVisitor, this
allows some weirder uses of Serde to handle the key and value in
separate steps.

r? @dtolnay


sfackler:
Should we do the same for serialize_struct_elt?

erickt:
@homu: r+

@sfackler: can you think of any cases where would we would want to serialize struct field names separately from the field value?

sfackler:
@erickt Nope, it's more of a symmetry question there.

Issue #470, opened 2016-07-29
Is there a version that (still) compiles with nightly-2016-07-07 ?


azerupi:
I'm trying to build a webapp with Serde and Diesel, both using nightly for codegen. But diesel does only support `nightly-2016-07-07` and I can't seem to build Serde for that nightly.

In my `Cargo.toml` I have:

```toml
serde = "=0.7.14"
serde_json = "0.7.4"
serde_macros = "=0.7.14"
```

I have tried `0.8.0`, `0.7.15`, `0.7.14`, `0.7.13` and `0.7.12` but `quasi` always fails to build.

I've tried to find the cause of this and I have come up with a plausible hypothesis:

Even though I strictly rely on `serde_macros = "=0.7.14"` it itself relies on any `serde_codegen = "0.7.x"` so it downloads the latest compatible version, which is `0.7.15`. That in turns relies on the latest version of quasi which I guess uses compiler APIs that are not yet available in `nightly-2016-07-07` 

Is this possible? And what can we do about it?

oli-obk:
I think there's a way to override the dependency: http://doc.crates.io/specifying-dependencies.html#overriding-dependencies

azerupi:
That didn't work unfortunately :cry: `[replace]` can't be used without forking and cheating because, quoting the docs:

>Note that when a crate is overridden the copy it's overridden with must have both the same name and version, but it can come from a different source (e.g. git or a local path).

One way I know of to specify a good version for `serde_codegen` is to manually modify the lock file, but that is a terrible idea...

I think it would be useful to be able to restrain dependency constraints even further in cargo. Something like:

```toml
[dependencies]
serde = "=0.7.14"
serde_json = "0.7.4"
serde_macros = "=0.7.14"

[serde_macros.dependencies]
# serde_macros depends on "0.7.x" constrain it to "<=0.7.14"
serde_codegen = "<=0.7.14"
```

But for now, I think it's an interesting flaw in the Serde ecosystem that old builds that built with a specific version of the compiler can end up not building anymore with that same version a couple of weeks later.

dtolnay:
Yeah this is not supported well. Let us know if you have ideas how we could make this better.

For now you should be able to use [this other way of overriding dependencies](http://doc.crates.io/specifying-dependencies.html#overriding-with-local-dependencies). I just did this successfully today for running benchmarks against 0.7.0 so even older than yours.

Basically clone the relevant dependencies (probably just serde) and create a `.cargo/config` file containing:

[...]

PR #472, opened 2016-07-29
Use exact dependencies on other serde crates

dtolnay:
@azerupi @oli-obk @erickt do you think this would fix the problem in #470? Are there other problems it could cause?

When compiling an old serde_macros this prevents it from depending on a new serde_codegen which may be on an incompatible quasi version.

### DIFF
--- a/serde_macros/Cargo.toml
+++ b/serde_macros/Cargo.toml
@@ -18,15 +18,15 @@ unstable-testing = ["clippy", "serde/unstable-testing", "serde_codegen/unstable-
 
 [dependencies]
 clippy = { version = "^0.*", optional = true }
-serde_codegen = { version = "0.8.0", default-features = false, features = ["unstable"] }
+serde_codegen = { version = "=0.8.0", default-features = false, features = ["unstable"] }
 
 [dev-dependencies]
 clippy = "^0.*"
 compiletest_rs = "^0.2.0"
 fnv = "1.0"
 rustc-serialize = "^0.3.16"
-serde = "0.8.0"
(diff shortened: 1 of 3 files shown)


serde_macros/Cargo.toml
```diff
@@ -18,15 +18,15 @@ unstable-testing = ["clippy", "serde/unstable-testing", "serde_codegen/unstable-
...
+serde_codegen = { version = "=0.8.0", default-features = false, features = ["unstable"] }
 
 [dev-dependencies]
 clippy = "^0.*"
 compiletest_rs = "^0.2.0"
 fnv = "1.0"
 rustc-serialize = "^0.3.16"
-serde = "0.8.0"
-serde_test = "0.8.0"
+serde = "=0.8.0"
```
azerupi: Does `serde` need to be constrained?

dtolnay: No I guess not - I just changed every dependency of a serde crate on a different serde crate. We should be okay constraining just serde_macros->serde_codegen and serde_codegen->serde_codegen_internals.

oli-obk:
I think this is the correct solution. Serde_macros is so finnicky, technically it should require major version bumps of serde_codegen whenever it's adjusted for a new nightly

dtolnay:
@oli-obk opinion on constraining all or just those two dependencies? I think constraining just serde_macros->serde_codegen and serde_codegen->serde_codegen_internals is less likely to cause trouble.

oli-obk:
Yea, there's no point in restricting the serde version

PR #479, opened 2016-08-04
Remove unnecessary nesting from generated impls

dtolnay:
We were using ast::Expr surrounded by `{ }` in a few places that should have been ast::Block. This PR decreases the indentation to make the generated code easier to read when it needs to be debugged.


oli-obk:
I was surprised this works for function bodies... But the retokenization erases all info... I guess that's ok

dtolnay:
It works but not with a comma after the `where` clause, for some reason.

```rust
// does not work
fn deserialize<__D>(deserializer: &mut __D) -> ::std::result::Result<$ty, __D::Error>
    where __D: _serde::de::Deserializer,
$body
```

Issue #486, opened 2016-08-09
Stack overflow compiling serde_codegen 0.8 on OSX nightly


alexcrichton:
I've been seeing some [failures](https://travis-ci.org/alexcrichton/backtrace-rs/jobs/150823832) in compiling `serde_codegen` on Travis OSX builders using the nightly compiler:

```
   Compiling serde_codegen v0.8.0

error: failed to run custom build command for `serde_codegen v0.8.0`

process didn't exit successfully: `/Users/travis/build/alexcrichton/backtrace-rs/target/debug/build/serde_codegen-bb1ef080fae9a50a/build-script-build` (signal: 6, SIGABRT: process abort signal)

--- stderr

thread 'main' has overflowed its stack

fatal runtime error: stack overflow
```

I wonder, is the build script for serde-codegen doing anything that might cause this? This started regressing for me about the time MIR codegen landed, and there's a few known stack-space issues with MIR, but for now could an explicitly large stack be used perhaps?

alexbool:
Hmm, I am using serde 0.8 with nightly and MIR on OS X for a while and have never run into this.

oli-obk:
@alexcrichton the serde-codegen build script parses the entire crate with quasi and then expands the quasi macros. Increasing the stack size would be easy (pack everything in a thread and set the RUST_MIN_STACK env var), but I have no way of testing it. I could simply pack everything in a thread and you can experiment with the RUST_MIN_STACK env var yourself. Once that works, I can add code that sets the RUST_MIN_STACK var to a value you have found to work (as long as it isn't already set).

@alexbool I'm not sure whether Rust chooses the main thread's stack size depending on some system information, maybe it differs between your setup and @alexcrichton's.

The difference might also be because of some compiler options. debug mode + no optimizations will obviously use more stack space than release mode + optimizations.

alexcrichton:
Yeah unfortunately I've been unable to reproduce locally as well, it's only reproducing on Travis. It may be the case that https://github.com/rust-lang/rust/pull/35409 fixes this, but in the meantime using `RUST_MIN_STACK` works for me!

oli-obk:
Wait... It works without the changes or it should work with them?

alexcrichton:
Oh it doesn't work yet, I was gonna wait for a merge so I can test out

oli-obk:
oh, good, that would have been weird.

[...]

PR #488, opened 2016-08-09
use a thread in the build script of serde_codegen to allow env vars to control the stack size

oli-obk:
cc #486

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,7 +1,0 @@
-paths = [
-  "serde",
-  "serde_codegen",
-  "serde_codegen_internals",
-  "serde_macros",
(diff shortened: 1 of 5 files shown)


oli-obk:
the .cargo/config overwrites don't seem to work as well as the path dependencies

dtolnay:
LGTM. The .cargo/config worked until #472 - I guess we go back to a path dependency for the serde_macros->serde_codegen dependency, or fix it in Cargo.

oli-obk:
skeptic has some issues with finding libs on nightly... I don't have time for it now, so I'm merging as is, since everything else passed

PR #490, opened 2016-08-12
Adjust doc(html_root_url)

bluss:
The URL does not need to contain the crate name. With this URL rustdoc can generate correct external doc links.


oli-obk:
Where would it generate the wrong doc links? If serde is a dependency?

bluss:
Yes

oli-obk:
ah, it was correct before moving to the serde-rs project

PR #501, opened 2016-08-18
Condense readme and link to serde.rs

dtolnay:
There is [a lot of work left](https://github.com/serde-rs/serde-rs.github.io/issues) but I think http://serde.rs has hit the point where it is a more useful resource than the current readme.

[Rendered](https://github.com/serde-rs/serde/blob/332d59f362a3b0744eef91d750cc3d361390f03d/README.md)


oli-obk:
so pretty :heart:

PR #498, opened 2016-08-17
Update syntex to 0.41

dtolnay:
~~Blocked on https://github.com/serde-rs/aster/pull/101 and https://github.com/serde-rs/quasi/pull/57.~~

PR #504, opened 2016-08-18
Disable clippy until Manishearth/rust-clippy#1174 is fixed

dtolnay:
I don't like that Clippy is delaying us from [supporting nightly](https://github.com/serde-rs/serde/pull/498).


dtolnay:
This won't build because of aster :frowning: I will combine it with the syntex update.

PR #509, opened 2016-08-19
Fix codegen with lifetimes but no type parameters

### LINKED ISSUE #507: Deriving Deserialize for structs with Cow<'a, str> fails
Consider the following code:
```rust
#[derive(Deserialize)]
struct S<'a>(Cow<'a, str>);
```
This fails with a cryptic error message:
```
error[E0207]: the lifetime parameter `'a` is not constrained by the impl trait, self type, or predicates
 --> <quote expansion>:1:2
  |
1 | <'a>
  |  ^^ unconstrained lifetime parameter
```
I believe `Deserialize` is implemented for `Cow<'a, str>`, so I think this should work.

dtolnay:
Fixes #507.

### DIFF
--- a/serde_codegen/src/de.rs
+++ b/serde_codegen/src/de.rs
@@ -192,7 +192,7 @@ fn deserialize_visitor(
     builder: &aster::AstBuilder,
     generics: &ast::Generics,
 ) -> (P<ast::Item>, P<ast::Ty>, P<ast::Expr>) {
-    if generics.ty_params.is_empty() {
+    if generics.lifetimes.is_empty() && generics.ty_params.is_empty() {
         (
             builder.item().unit_struct("__Visitor"),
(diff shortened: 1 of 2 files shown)


oli-obk:
@homu r+

I'm signing out now for the next two weeks. I'll be back on the 5th.

dtolnay:
The special case is still required, at least for now:

```
error: empty tuple structs and enum variants are unstable, use unit structs and enum variants instead (see issue #35626)
  --> tests/../../testing/tests/test_de.rs:35:28
   |
35 | #[derive(PartialEq, Debug, Deserialize)]
   |                            ^^^^^^^^^^^
   |
   = help: add #![feature(relaxed_adts)] to the crate attributes to enable
```

dtolnay:
Enjoy your vacation! I will try not to break things too badly.

PR #524, opened 2016-08-24
Remove skeptic

dtolnay:
Now that the readme is a single trivial example it should be easy to keep up to date. Skeptic with a dependency on serde_json makes it impossible to test breaking changes in Travis (came up in #523).


dtolnay:
I filed https://github.com/serde-rs/serde-rs.github.io/issues/16 to implement skeptic for the https://serde.rs examples.

PR #525, opened 2016-08-24
Expose forward_to_deserialize macro

### LINKED ISSUE #522: Expose forward_to_deserialize macro to help Deserializer implementations
[This](https://github.com/serde-rs/serde/blob/d690ffda8d38fe261abf2bea804bce28b57ee6fa/serde/src/de/value.rs#L181-L195) turns out to be a common pattern and currently lots of crates containing Deserializers are copying the same macro.

cc @nox

dtolnay:
Fixes #522.

```rust
impl Deserializer for MyDeserializer {
    fn deserialize<V>(&mut self, visitor: V) -> Result<V::Value, Self::Error>
        where V: Visitor
    {
        /* ... */
    }

    forward_to_deserialize! {
        bool usize u8 u16 u32 u64 isize i8 i16 i32 i64 f32 f64 char str string
        unit option seq seq_fixed_size bytes map unit_struct newtype_struct
        tuple_struct struct struct_field tuple enum ignored_any
    }
}
```

cc @nox


dtolnay:
I don't have time now but I need to fix the no_std usage. I don't want to use just `Result<__V::Value, Self::Error>` because people commonly do `type Result<T> = result::Result<T, MyError>`.

nox:
Could there be a way to customise the forwarding? I would like to delegate everything to an inner deserializer.

dtolnay:
Maybe a different macro?

```rust
forward_to_deserializer! {
    self.integers {
        usize u8 u16 u32 u64 isize i8 i16 i32 i64
    }
    self.floats {
        f32 f64
    }
}
```

Can you say more about your use case?

dtolnay:
Updated to work with no_std.

dtolnay:
I filed #531 to follow up on the other macro.

dtolnay:
Fixed the merge conflict.

oli-obk:
> Could there be a way to customise the forwarding? I would like to delegate everything to an inner deserializer.

@nox: Could you elaborate on the details of you use case?

dtolnay:
I copied his explanation from IRC into #531:

> So for my deserializer forwarder,
> my use case is that I want to wrap MapDeserializer in a newtype,
> and implement Deserializer on the newtype, forwarding to inner deserializer.

oli-obk:
Ah, completely different issue, since this one always forwards to `deserialize` and #531 forwards to the inner `Deserializer`'s `deserialize_*` method

@homu r+

PR #530, opened 2016-08-29
Macros 1.1

dtolnay:
RFC: [1681-macros-1.1.md](https://github.com/rust-lang/rfcs/blob/master/text/1681-macros-1.1.md)
Implementation: https://github.com/rust-lang/rust/pull/35957


dtolnay:
Some limitations:

- `#[derive(Deserialize)]` always panics. I have not debugged this yet.

- `#[derive(Serialize)]` for ordinary structs always panics. Newtype structs, unit structs, and enums not containing ordinary struct variants all seem to work.

Both panics are this:

```
thread 'rustc' panicked at 'fold_mac disabled by default', src/libsyntax/fold.rs:171
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

dtolnay:
Also, AFAICT the implementation in https://github.com/rust-lang/rust/pull/35957 does not allow using `#[serde(...)]` attributes together with `#[derive(Serialize, Deserialize)]`. The reason is that it expects `#[derive(Serialize)]` and `#[derive(Deserialize)]` to each strip away `#[serde(...)]` attributes, otherwise it fails on the unrecognized attribute. Ideally in the case of `#[derive(Serialize, Deserialize)]` whichever one runs first would leave the attributes and the second one would strip them, but there is no way to tell that both are present.

serde_codegen/src/lib.rs
```diff
@@ -36,51 +36,67 @@ include!(concat!(env!("OUT_DIR"), "/lib.rs"));
 include!("lib.rs.in");
 
 #[cfg(feature = "with-syntex")]
-pub fn expand<S, D>(src: S, dst: D) -> Result<(), syntex::Error>
```
dtolnay: Sorry for the confusing diff - this is just splitting `expand` into `expand` and `expand_str` with common code in `syntex_registry`.

serde_derive/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+documentation = "https://serde.rs/codegen.html"
+keywords = ["serde", "serialization"]
+include = ["Cargo.toml", "src/**/*.rs"]
+
+[lib]
+name = "serde_derive"
+rustc-macro = true
+
+[dependencies]
+serde_codegen = { version = "=0.8.5", path = "../serde_codegen" }
```
erickt: Why pin to a particular serde_codegen? Shouldn't it be forward compatible?

dtolnay: Same reason we pin `serde_macros`. See https://github.com/serde-rs/serde/issues/470#issuecomment-236307729. It allows you to use an old nightly with an old serde_derive without pulling in a new serde_codegen that no longer compiles on the old nightly.

dtolnay: I guess what I'm saying is: no it's not forward compatible :smiling_imp:

erickt:
Looks good to me!

dtolnay:
I added a test suite but we need to fix https://github.com/rust-lang/rust/issues/36211 before it will pass.

tikue:
Trying to use this, I get

    error: the `#[rustc_macro_derive]` attribute is only usable with crates of the `rustc-macro` crate type

dtolnay:
@tikue you need a version of Cargo that supports rustc-macro crates. My Cargo PR merged a few hours ago so it will be in the next nightly. If you need it now you can build cargo yourself from the master branch. It should be as simple as cloning and running `cargo install`.

tikue:
@dtolnay cool, thanks!

PR #536, opened 2016-09-02
Fix "variable does not need to be mutable" warning

### LINKED ISSUE #534: `struct S {}`: variable does not need to be mutable
```rust
#![feature(plugin, custom_derive)]
#![plugin(serde_macros)]

#[derive(Serialize)]
struct S {}

fn main() {
    let _ = S {};
}
```

```
warning: variable does not need to be mutable, #[warn(unused_mut)] on by default
 --> src/main.rs:4:10
  |
4 | #[derive(Serialize)]
  |          ^^^^^^^^^^
```

dtolnay:
Fixes #534.

cc @EliDupree

### DIFF
--- a/serde_codegen/src/ser.rs
+++ b/serde_codegen/src/ser.rs
@@ -4,6 +4,7 @@ use syntax::ast::{self, Ident, MetaItem};
 use syntax::codemap::Span;
 use syntax::ext::base::{Annotatable, ExtCtxt};
 use syntax::ptr::P;
+use syntax::tokenstream::TokenTree;
 
 use bound;
 use span;
@@ -226,9 +227,10 @@ fn serialize_tuple_struct(
 
     let type_name = name_expr(builder, item_attrs.name());
     let len = serialize_stmts.len();
+    let let_mut = mut_if_nonzero(cx, len);
 
     quote_block!(cx, {
-        let mut state = try!(_serializer.serialize_tuple_struct($type_name, $len));
+        let $let_mut state = try!(_serializer.serialize_tuple_struct($type_name, $len));
         $serialize_stmts
         _serializer.serialize_tuple_struct_end(state)
     }).unwrap()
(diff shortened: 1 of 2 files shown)


serde_codegen/src/ser.rs
```diff
@@ -226,9 +227,10 @@ fn serialize_tuple_struct(
 
     let type_name = name_expr(builder, item_attrs.name());
     let len = serialize_stmts.len();
+    let let_mut = mut_if_nonzero(cx, len);
 
     quote_block!(cx, {
-        let mut state = try!(_serializer.serialize_tuple_struct($type_name, $len));
+        let $let_mut state = try!(_serializer.serialize_tuple_struct($type_name, $len));
         $serialize_stmts
```
dtolnay: @nox suggests rolling this into `$serialize_stmts` and generating:

```rust
let state = try!(...);
let mut state = state;
```

where the initial call is never mut and `$serialize_stmts` includes "let mut state = state" if necessary.

dtolnay: @nox I think I would rather move in the opposite direction, of making our generated code more straightforward where that can be accomplished easily. See as examples #479 and #538. There are a few reasons that is valuable:

1. The most straightforward code will give us (and others) the best experience when the generated code needs to be debugged.
2. The generated code is valuable as a template for implementing Serialize and Deserialize manually with minor modifications.
3. Convoluted generated code gives people the impression that Serde is hard or Serde codegen is magic and impossible to understand, scaring people away from implementing Serialize and Deserialize manually where necessary.

I'm not saying we need to make the generated code beautiful at the expense of maintainability of serde_codegen, but there is a middle ground and I don't think rolling `let mut state = state` into `$serialize_stmts` is the right tradeoff in this case. Thanks for the idea though!

serde_codegen/src/ser.rs
```diff
@@ -253,8 +255,12 @@ fn serialize_struct(
     );
 
     let type_name = name_expr(builder, item_attrs.name());
-    let len = fields.iter()
+
+    let serialized_fields: Vec<_> = fields.iter()
```
oli-obk: we could eliminate the allocation by folding into a tuple, not sure if that makes any difference in readability or performance, so feel free to merge as is.

nox: We could also make a `Peekable` iterator and check if `fields.peek().is_none()` is true, no?

oli-obk: I like it, much cleaner!

dtolnay: Done. Great idea.

oli-obk:
@homu r+

PR #538, opened 2016-09-03
Remove `if !false { ... }` from generated serialization code

dtolnay:
I don't think this negatively affects maintainability of the code in serde_codegen and I think there is some value in keeping our generated code relatively clear so that people can use it as a template when implementing Serialize manually with minor modifications.


oli-obk:
great! this has always annoyed me in the generated code.

I'd say the codegen code is even clearer now

@homu r+

PR #546, opened 2016-09-07
Result needs to be fully qualified

dtolnay:
Without this:

```
error[E0244]: wrong number of type arguments
  --> src/api/accounts.rs:19:10
   |
19 | #[derive(Serialize, Deserialize, Debug)]
   |          ^^^^^^^^^^ expected 1 type arguments, found 2
```


oli-obk:
@homu r+

PR #548, opened 2016-09-12
No more syntex for serde_derive

dtolnay:
- [x] decide whether this is a good idea
- [x] speed up compile time by more than 10x
- [x] implement Serialize
- [x] implement Deserialize
- [x] way less code than before
- [x] workaround for https://github.com/rust-lang/rust/issues/36211
- [x] test suite passes with serde_derive
- [x] runtime performance as fast as serde_macros
- [x] implement \x and \u escape sequences in attributes (https://github.com/dtolnay/syn/issues/9)
- [x] sort out error handling in serde_codegen_internals/src/attr.rs
- [x] backward compatibility considerations for the "with-syntex" feature
- [x] [cargo expand](https://github.com/dtolnay/cargo-expand) the test suite and confirm it is identical before and after
- [x] travis build passes

Follow-up tasks:
- [ ] factor the attribute stripping into a fold (#561)
- [ ] clean up workarounds (#562)


dtolnay:
Running `cargo test` in serde_derive now passes! It didn't before due to rust-lang/rust#36211.

dtolnay:
For some reason serde_codegen by itself builds three times as fast this way, which I didn't expect. I guess quasi_macros is slow?

```
$ git checkout master
$ cargo build --no-default-features --features unstable
$ cargo clean -p serde_codegen
$ cargo build --no-default-features --features unstable  # 12 SECONDS
$ git checkout syn
$ cargo build --no-default-features --features with-syn
$ cargo clean -p serde_codegen
$ cargo build --no-default-features --features with-syn  # 4 SECONDS

**Comment by oli-obk** [comment]
> I guess quasi_macros is slow?

fewer generics?

**Comment by dtolnay** [comment]
I did some testing with rusoto in https://github.com/rusoto/rusoto/issues/364#issuecomment-246577174 to see how performance of this PR compares with the current serde_macros (runtime performance, after they have been compiled). They come out about even which is great.

**Comment by dtolnay** [comment]
There is a lot of work left here but let's take a moment to celebrate 9a86e68 the first syntex-free green build.

**Comment by dtolnay** [comment]
@oli-obk @erickt this is ready for review. There were a few places that I uncovered strange behavior in the old codegen (just unusually placed semicolons / parentheses / braces) that I went out of my way to preserve. I left comments and I can clean that up in a separate PR. I confirmed that this PR expands the test suite byte-for-byte identical to master.

I filed #561 as a follow-up task. I don't think that needs to block this PR.

**Comment by alexcrichton** [comment]
Awesome progress @dtolnay!

File: serde_codegen_internals/src/ctxt.rs [hunk 1]
```diff
@@ -0,0 +1,43 @@
+use std::fmt::Display;
+use std::cell::RefCell;
+
+#[derive(Default)]
+pub struct Ctxt {
```
oli-obk: I'm not very happy with the `Ctxt` structure and it's usage. I'd much rather have regular `Result` type error reporting, but we can leave that to a future refactoring. Maybe the code gets horribly cluttered with `Result`

serde_codegen/src/lib.rs
```diff
@@ -3,10 +3,10 @@
 #![cfg_attr(feature = "clippy", allow(too_many_arguments))]
 #![cfg_attr(feature = "clippy", allow(used_underscore_binding))]
 #![cfg_attr(not(feature = "with-syntex"), feature(rustc_private, plugin))]
-#![cfg_attr(not(feature = "with-syntex"), plugin(quasi_macros))]
 
-extern crate aster;
-extern crate quasi;
+// The `quote!` macro requires deep recursion.
+#![recursion_limit = "192"]
```
erickt: I agree.  I love how've much simpler everything looks.

serde_codegen/src/ser.rs
```diff
@@ -353,26 +278,23 @@ fn serialize_variant(
                 &variant.fields[0],
             );
 
-            quote_arm!(cx,
-                $type_ident::$variant_ident(ref __simple_value) => $block
-            )
+            quote! {
+                // The braces are unnecessary but quasi used to put them in. We
```
erickt: If I recall correctly, quote_block required the extra braces, so shouldn't be a problem to remove them later.

serde_codegen_internals/src/ctxt.rs
dtolnay: I filed #563 to follow up in a future refactoring.

serde_codegen/src/de.rs
```diff
@@ -179,133 +126,132 @@ fn deserialize_body(
...
+            })
             .collect();
 
+        let ty_param_idents = if ty_param_idents.is_empty() {
+            None
+        } else {
+            Some(quote!(::<#(ty_param_idents),*>))
+        };
+
+        let phantom_exprs = (0 .. num_phantoms).map(|_| quote!(::std::marker::PhantomData));
```
dtolnay: Done.

PR #566, opened 2016-09-28
Delete serde_macros

### LINKED ISSUE #565: Delete serde_macros
This requires moving the compile-tests over to the serde_derive crate.

dtolnay:
Fixes #565.


dtolnay:
Unfortunately compiletest can't do multi-line error messages, so I had to do this:

```rust
#[derive(Serialize)] //~ ERROR: custom derive attribute panicked
#[serde(abc="xyz")] // ERROR: unknown serde container attribute `abc`
struct A {
    x: u32,
}
```

We are only checking that the thing fails to compile because "custom derive attribute panicked", not the specific message. I left the messages there so that it is clear what is being tested.

oli-obk:
> Unfortunately compiletest can't do multi-line error messages

What do you mean by multi-line error messages?
Multiple ones per line? Or error messages with a span going over multiple lines?

dtolnay:
Here is what the message looks like:

```rust
error: custom derive attribute panicked
 --> src/main.rs:9:10
  |
9 | #[derive(Serialize)]
  |          ^^^^^^^^^
  |
  = help: message: unknown serde container attribute `abc`
```

Compiletest only sees the error at the top, not the "help". I think related to https://github.com/rust-lang/rust/issues/31380.

oli-obk:
Compiletest only sees a help when you ask for a help in the code. So adding a //~^ HELP below the error line should work

dtolnay:
Good call, I didn't know about that. I pushed a fix.

PR #579, opened 2016-10-12
Deserialize big arrays using a loop

dtolnay:
This improves compile time of serde by more than 30%. See https://github.com/rust-lang/rust/issues/37106.

### DIFF
--- a/serde/src/de/impls.rs
+++ b/serde/src/de/impls.rs
@@ -524,14 +524,14 @@ impl<A> ArrayVisitor<A> {
     }
 }
 
-impl<T> Visitor for ArrayVisitor<[T; 0]> where T: Deserialize + Default {
+impl<T> Visitor for ArrayVisitor<[T; 0]> where T: Deserialize {
     type Value = [T; 0];
 
     #[inline]
     fn visit_unit<E>(&mut self) -> Result<[T; 0], E>
         where E: Error,
     {
-        Ok([T::default(); 0])
+        Ok([])
     }
 
     #[inline]
@@ -539,12 +539,12 @@ impl<T> Visitor for ArrayVisitor<[T; 0]> where T: Deserialize + Default {
         where V: SeqVisitor,
     {
         try!(visitor.end());
-        Ok([T::default(); 0])
+        Ok([])
     }
 }
 
 impl<T> Deserialize for [T; 0]
-    where T: Deserialize + Default
+    where T: Deserialize
 {
     fn deserialize<D>(deserializer: &mut D) -> Result<[T; 0], D::Error>
         where D: Deserializer,
--- a/serde/src/de/impls.rs
+++ b/serde/src/de/impls.rs
@@ -610,81 +610,22 @@ macro_rules! big_array_impls {
                 fn visit_seq<V>(&mut self, mut visitor: V) -> Result<[T; $len], V::Error>
                     where V: SeqVisitor,
                 {
-                    use std::{mem, ptr};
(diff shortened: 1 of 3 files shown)


dtolnay:
- [ ] handle no_std
- [ ] document why this workaround exists
- [ ] unit test all of the failure cases
- [ ] check whether the slowness is already fixed upstream

serde/src/de/impls.rs
```diff
@@ -612,38 +612,148 @@ array_impls! {
...
+                type Value = [T; $len];
+
+                #[inline]
+                fn visit_seq<V>(&mut self, mut visitor: V) -> Result<[T; $len], V::Error>
+                    where V: SeqVisitor,
+                {
+                    use std::{mem, ptr};
+                    use std::ops::{Deref, DerefMut};
+
+                    enum PartialArray<T> {
```
oli-obk: can't we just use [arrayvec](https://crates.io/crates/arrayvec)?

serde/src/de/impls.rs
```diff
@@ -612,38 +612,148 @@ array_impls! {
...
-    ArrayVisitor31, 31 => (a, b, c, d, e, f, g, h, i, j, k, l, m, n, o, p, q, r, s, t, u, v, w, x,
-                           y, z, aa, ab, ac, ad, ae),
-    ArrayVisitor32, 32 => (a, b, c, d, e, f, g, h, i, j, k, l, m, n, o, p, q, r, s, t, u, v, w, x,
-                           y, z, aa, ab, ac, ad, ae, af),
+}
+
+macro_rules! big_array_impls {
+    ($($visitor:ident [$len:expr])+) => {
+        $(
+            struct $visitor<T> {
```
oli-obk: I think we can get around the 50 visitor names by making this 

```rust
struct ArrayVisitor<T, ARR> {
    marker: PhantomData<T>,
    arr_marker: PhantomData<ARR>,
}
```

and then

```rust
impl<T> Visitor for ArrayVisitor<T, [T; $len]> where T: Deserialize { ...
```

oli-obk:
Depending how good LLVM is, we could do all impls with the loops. LLVM should unroll the short ones.

serde/src/de/impls.rs
dtolnay: Done. With the additional dependencies the speedup is only 20% instead of 30% but I am okay with that because the code is simpler and safer.

serde/src/de/impls.rs
dtolnay: Good idea. I updated to use `ArrayVisitor<[T; $len]>` across all of them from 0 to 32.

dtolnay:
I snuck #582 and #583 in here while we're rewriting array deserialization anyway.

- [T; 0] should use deserialize_seq_fixed_size not deserialize_seq
- [T; 0] should not require T: Default in order to deserialize

dtolnay:
I'd like to wait and see whether https://github.com/rust-lang/rust/pull/37152 speeds things up enough that we don't need this fix.

dtolnay:
I tried out the fix in https://github.com/rust-lang/rust/pull/37152 and I am seeing the following times:

- master: 5.3 seconds
- the unsafe implementation: 4.3 seconds
- arrayvec implementation including dependencies: 4.9 seconds

This is compared to 10-11 seconds with the most recent nightly, so it seems like the problem is about to be fixed in Rust. I think we can write off the last 1 second in exchange for keeping the code sane and free of dependencies.

dtolnay:
Moved the parts we care about to #584.

PR #584, opened 2016-10-14
Array and tuple deserialization cleanup

### LINKED ISSUE #582: [T; 0] should use deserialize_seq_fixed_size not deserialize_seq

### LINKED ISSUE #583: [T; 0] should not require T: Default in order to deserialize

dtolnay:
- Fixes #582
- Fixes #583
- Combines the ArrayVisitor* types into one ArrayVisitor like you suggested in #579; can't do the same for the TupleVisitor*s because those are public
- Removes noise from the `array_impls!` and `tuple_impls!` macros

### DIFF
--- a/serde/src/de/impls.rs
+++ b/serde/src/de/impls.rs
@@ -512,27 +512,26 @@ seq_impl!(
 
 ///////////////////////////////////////////////////////////////////////////////
 
-struct ArrayVisitor0<T> {
-    marker: PhantomData<T>,
+struct ArrayVisitor<A> {
+    marker: PhantomData<A>,
 }
 
-impl<T> ArrayVisitor0<T> {
-    /// Construct a `ArrayVisitor0<T>`.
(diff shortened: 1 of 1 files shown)


serde/src/de/impls.rs
```diff
@@ -512,65 +512,51 @@ seq_impl!(
...
 
 impl<T> Deserialize for [T; 0]
-    where T: Deserialize + Default
+    where T: Deserialize
 {
     fn deserialize<D>(deserializer: &mut D) -> Result<[T; 0], D::Error>
         where D: Deserializer,
     {
-        deserializer.deserialize_seq(ArrayVisitor0::new())
+        deserializer.deserialize_seq_fixed_size(0, ArrayVisitor::<[T; 0]>::new())
```
oli-obk: Technically a breaking change... Let's call it a bugfix

PR #591, opened 2016-10-18
Not safe to share Spans from one ParseSess to another

### LINKED ISSUE #574: Code generation randomly produces code which does not compile
Code generation sometimes fails with corrupted code producing type errors. A workaround is to slightly change the code and retry (adding or removing blank lines works). I don't know if deterministic (so once it builds, it always builds correctly) a or random yet (so any build can randomly fail).

I'm using serde_codegen version 0.8.11

UPDATE: it does look deterministic and it appears to have something to do with the optional trailing commas. I'm using ```rustfmt``` so the comma is added automatically (and I am not keen on changing this policy):

```
	#[serde(skip_serializing_if="Option::is_none")]
	pub sequences: Option<Vec<Sequence>>, // removing this comma fixes it
}
// also adding an extra blank line fixes it
```



```
[...]/out/reference/template_types.rs:239:82: 239:95 error: mismatched types [E0308]
[...]/out/reference/template_types.rs:239                                                _serde::de::Error::invalid_length("SpeedOfTime"));

[...]/out/reference/template_types.rs:239:82: 239:95 help: run `rustc --explain E0308` to see a detailed explanation
[...]/out/reference/template_types.rs:239:82: 239:95 note: expected type `usize`
[...]/out/reference/template_types.rs:239:82: 239:95 note:    found type `&'static str`

```                                                                                                                                                                                                                                
```rust
                        let __field4 =
                            match try!(visitor . visit :: < Option < Int > > (
                                        )) {


                                Some(value) => {
                                    value
                                }
                                None => {
                                    try!(visitor . end (  ));
                                    return Err(

                                               _serde::de::Error::invalid_length(4usize));
                                }
                            };
                        let __field5 =
                            match try!(visitor . visit :: < Option < Int > > (
                                        )) {


                                Some(value) => {
                                    value
                                }
                                None => {
                                    try!(visitor . end (  ));
                                    return Err(

                                               _serde::de::Error::invalid_length("SpeedOfTime"));
						                                                        ^^^^^^^^^^^^^^^^ <- error in line 239
                                }
                            };
                        let __field6 =
                            match try!(visitor . visit :: < Option < Int > > (
                                        )) {
                                Some(value) =>
                                {
                                    value
                                }
                                None => {
                                    try!(visitor . end (  ));
                                    return Err(_serde::de::Error::invalid_length(6usize));
                                }
                            };
```

this is the snippet from the .rs.in . "SpeedOfTime" is a field of another struct?

```rust
#[derive(Default,Serialize,Deserialize)]
pub struct Event {
	#[serde(rename="Name")]
	#[serde(skip_serializing_if="Option::is_none")]
	pub name: Option<String>,

	#[serde(rename="Format")]
	#[serde(skip_serializing_if="Option::is_none")]
	pub format: Option<String>,

	#[serde(rename="EventDescription")]
	#[serde(skip_serializing_if="Option::is_none")]
	pub event_description: Option<String>,

	#[serde(rename="Colour")]
	#[serde(skip_serializing_if="Option::is_none")]
	pub colour: Option<String>,

	#[serde(rename="Period")]
	#[serde(skip_serializing_if="Option::is_none")]
	pub period: Option<Int>,

	#[serde(rename="Length")]
	#[serde(skip_serializing_if="Option::is_none")]
	pub length: Option<Int>,

	#[serde(rename="Repeat")]
	#[serde(skip_serializing_if="Option::is_none")]
	pub repeat: Option<Int>,

	#[serde(rename="Start")]
	#[serde(skip_serializing_if="Option::is_none")]
	pub start: Option<Int>,

	#[serde(rename="StartOffsetSeconds")]
	#[serde(skip_serializing_if="Option::is_none")]
	pub start_offset_seconds: Option<Int>,

	#[serde(rename="DoubleXP")]
	#[serde(skip_serializing_if="Option::is_none")]
	pub double_xp: Option<Bool>,

	#[serde(rename="Sequences")]
	#[serde(skip_serializing_if="Option::is_none")]
	pub sequences: Option<Vec<Sequence>>,
}
```
Cargo.toml:
```
[package]
name = "event_list_editor_cli"
version = "0.1.0"
authors = ["Nicola Orrù <nicola_orru@scee.net>"]
build = "build.rs"

[build-dependencies]
syntex = "*"
serde_codegen = "*"

[dependencies]
num = "*"
getopts = "*"
csv = "*"
log = "*"
log4rs = "*"
rand = "*"
chrono="*"
bitflags="*"
custom_derive="*"
enum_derive="*"
enum_primitive="*"
serde="*"
serde_json="*"
```
build.rs
```rust
extern crate syntex;
extern crate serde_codegen;

use std::env;
use std::path::Path;
use std::fs;

fn main() {
	let out_dir = env::var_os("OUT_DIR").unwrap();

	let src = Path::new("src/reference/template_types.rs.in");
	let dst = Path::new(&out_dir).join("reference/template_types.rs");

	fs::create_dir_all(dst.parent().unwrap()).unwrap_or(());

	serde_codegen::expand(&src, &dst).unwrap();
}
```

### LINKED ISSUE #590: BTreeMap types generate broken visit_str functions
The code generated by `serde_codegen::expand()` for deserializing types that contain `BTreeMap` contains code that is both broken and baffling.  It appears to be the result of an index/offset/pointer error in the codegen itself.  

This is a weird one, so let me know if I can provide any additional info.

This struct:
```
#[derive(Debug, Deserialize)]
pub struct Service {
    pub documentation: Option<String>,
    pub examples: Option<BTreeMap<String, String>>,
    pub metadata: Metadata,
    pub operations: BTreeMap<String, Operation>,
    #[serde(deserialize_with="ShapesMap::deserialize_shapes_map")]
    pub shapes: BTreeMap<String, Shape>,
    pub version: String,
}
```

when fed through `serde_codegen::expand()` produces (among otherwise functional and correct code) this `visit_str` function:

```
                            fn visit_str<__E>(&mut self, value: &str)
                             -> ::std::result::Result<__Field, __E> where
                             __E: _serde::de::Error {
                                match value {
                                    "documentation" => {
                                        Ok(__Field::__field0)
                                    }
                                    "examples" => { Ok(__Field::__field1) }
                                    "metadata" => { Ok(__Field::__field2) }
                                    "AWS Marketplace Commerce Analytics" => {
                                        Ok(__Field::__field3)
                                    }
                                    "shapes" => { Ok(__Field::__field4) }
                                    "version" => { Ok(__Field::__field5) }
                                    _ => Ok(__Field::__ignore),
                                }
                            }
```

Note how the matcher arm that should match the `operations` field instead matches the string "AWS Marketplace Commerce Analytics". 

The *only* place the string "AWS Marketplace Commerce Analytics" appears in our entire codebase is  in a different function in the same source file, the relevant portion of which is:

```
impl Service {
    pub fn service_type_name(&self) -> &str {
        match &self.metadata.service_full_name[..] {
            [...snip...]
            "AWS Marketplace Commerce Analytics" => "MarketplaceCommerceAnalytics",
            [...snip...]
```

Interestingly, if I insert a comment in the `Service` struct we're feeding to the codegen,  a completely different match arm in the same method is replaced with the wrong string:

```
                             fn visit_str<__E>(&mut self, value: &str)
                             -> ::std::result::Result<__Field, __E> where
                             __E: _serde::de::Error {
                                match value {
                                    "documentation" => {
                                        Ok(__Field::__field0)
                                    }
                                    "Iot" => { Ok(__Field::__field1) }
                                    "metadata" => { Ok(__Field::__field2) }
                                    "operations" => { Ok(__Field::__field3) }
                                    "shapes" => { Ok(__Field::__field4) }
                                    "version" => { Ok(__Field::__field5) }
                                    _ => Ok(__Field::__ignore),
                                }
                            }
```
In this case, "examples" is replaced by "Iot" (which also appears as a string in the matcher of the `service_type_name` function above.

The source file that we're feeding to `serde_codegen` is https://github.com/rusoto/rusoto/blob/master/codegen/src/botocore.in.rs

The issue can be reliably reproduced by creating a new Cargo project with an external dependency on any Rusoto feature (except S3), like:
```
[dependencies]
rusoto = {version = "0.18.0", features = ["sqs"]}
```

Like I said, this is a weird one.  I'll be happy to help debug.

dtolnay:
Fixes #590 and probably fixes #574.

Spans in the AST returned by `parse_item_from_source_str` and other parsing
functions contain byte offsets into the source code they were parsed from. The
pretty printer uses these Spans [here][1] to preserve the representation of
literals when parsing and printing back out unmodified.

In this bug, the byte offset of a string in the input to
`parse_item_from_source_str` coincidentally matched the byte offset of a totally
different string in the input to `parse_crate_from_file` called [here][2] by
Syntex. The Span from the former triggered the pretty printer to write out the
content of the latter.

By using the same ParseSess, Spans from the two `parse_*` calls never collide.

[1]: https://github.com/rust-lang/rust/blob/1.12.0/src/libsyntax/print/pprust.rs#L628
[2]: https://github.com/serde-rs/syntex/blob/v0.45.0/syntex/src/registry.rs#L134

### DIFF
--- a/serde_codegen/src/lib.rs
+++ b/serde_codegen/src/lib.rs
@@ -169,8 +169,8 @@ macro_rules! shim {
             use syntax::parse;
             let name = stringify!($name).to_string();
             let cfg = Vec::new();
-            let sess = parse::ParseSess::new();
-            let impl_item = parse::parse_item_from_source_str(name, expanded, cfg, &sess);
+            let sess = cx.parse_sess;
+            let impl_item = parse::parse_item_from_source_str(name, expanded, cfg, sess);
             push(::syntax::ext::base::Annotatable::Item(impl_item.unwrap().unwrap()));
         }
     };


dtolnay:
@erickt @oli-obk I found two other suspicious ParseSess::new() calls [here](https://github.com/serde-rs/aster/blob/v0.30.0/src/mac.rs#L93) and [here](https://github.com/serde-rs/syntex/blob/v0.45.0/syntex/src/registry.rs#L51). Let me know if you agree those need to be fixed. I also found [this](https://github.com/serde-rs/syntex/blob/v0.45.0/syntex/src/registry.rs#L62) but I think it is okay because the parsed item immediately gets respanned.

oli-obk:
Isn't the issue only if the newly created `ParseSess` is used to parse from a `str`? So only the second one is bad. But I might be wrong.

dtolnay:
The first one passes the bad ParseSess to `to_tokens` [here](https://github.com/serde-rs/aster/blob/v0.30.0/src/mac.rs#L95-L96). The `to_tokens` can use the ParseSess to parse strings, for example [here](https://github.com/serde-rs/quasi/blob/v0.22.0/quasi/src/lib.rs#L144).

erickt:
@dtolnay: r+ on this change. Do you think it'd just be best for us to replace all spans with dummy spans in these cases? Or if we should thread `ExtCtxt`s through these cases to use the same `ParseSess` objects?

dtolnay:
@erickt as far as I understand the issue, in theory respanning to DUMMY_SP is sufficient to fix any problems of not using the right ParseSess.

In practice [`fold.rs`](https://github.com/rust-lang/rust/blob/master/src/libsyntax/fold.rs) has a ton of holes around not handling spans correctly. I have hit 9 so far with syn (I have a unit test that parses a file with syn, prints it back out, parses both the output and the original using syntex, respans them to dummy spans, and asserts they are the same). I still need to push those fixes upstream.

dtolnay:
I released serde_codegen 0.8.14 with this fix and I filed https://github.com/serde-rs/aster/issues/116 and https://github.com/serde-rs/syntex/issues/100 to follow up on the other occurrences.

norru:
Fixes #574

PR #592, opened 2016-10-18
Use post-expansion crate to let other custom derives see serde attrs

### LINKED ISSUE #585: Serde field attributes not written to TokenStream
Previously, I've piggybacked on `serde_codegen_internals` to get whether or not a field will be serialised, and if so, what it will be called.

With the new macros 1.1 stuff this doesn't work because the `serde` field attributes are being removed from the token stream output if my code runs after `Serialize`:

```rust
//No serde attributes for ElasticType
#[derive(Serialize, ElasticType)]
struct MyType { ... }
```

```rust
//All the serde attributes for ElasticType
#[derive(ElasticType, Serialize)]
struct MyType { ... }
```

Wouldn't this also affect attributes visible to `Deserialize` if you also derive `Serialize` first?

dtolnay:
Fixes #585. cc @KodrAus

Not ready to merge yet because first I want to benchmark its impact on compile time in something like rusoto.


dtolnay:
I confirmed that this has no measurable impact on compile time. In Rusoto `cargo expand --no-default-features --features 'unstable all'` expands rusoto_codegen and rusoto in 43 seconds with or without this change.

@oli-obk please also review https://github.com/dtolnay/post-expansion as though it were part of this PR.

KodrAus:
@dtolnay I took this for a spin and it all worked as advertised :+1:

oli-obk:
I'm a little worried about the way the builtin derives are detected. Adding a new builtin derive will be a breaking change, because then either the attributes aren't stripped or we go into an infinite loop

oli-obk:
Couldn't we find a way to mark attributes as used by attaching a `#[used]` before every used attribute Rustc's lint could then ignore these attributes in the unkown attr and unused attr lint?

dtolnay:
> I'm a little worried about the way the builtin derives are detected. Adding a new builtin derive will be a breaking change, because then either the attributes aren't stripped or we go into an infinite loop

Can you give an example where a new builtin derive would cause attributes not to be stripped or an infinite loop?

The reason builtin derives are relevant is because:

```rust
#[derive(Serialize, ElasticType)]
#[serde(rename = "T")]
struct S;
```

should be expanded to:

```rust
impl Serialize for S { /* ... */ }

#[derive(ElasticType)]
#[serde(rename = "T")]
#[derive(PostExpansion_serde)]
#[post_expansion(strip = "serde")]
struct S;
```

but:

```rust
#[derive(Serialize, Debug)]
#[serde(rename = "T")]
struct S;
```

as an optimization can be expanded directly to:

```rust
impl Serialize for S { /* ... */ }

#[derive(Debug)]
struct S;
```

So attributes can be stripped immediately if the only remaining derives are guaranteed not to read those attributes (i.e. are builtin derives). This optimization accounts for the no performance difference in Rusoto. If the compiler adds a new builtin derive, *we want it to be treated like a custom derive* because we don't know whether it reads attributes (probably doesn't read `serde` attributes, but technically we need to check). I think the current code handles this. Attributes are stripped either way and there is no infinite loop. We can add a new builtin derive to post-expansion in a nonbreaking change because it is an optimization but not a change in behavior.

> Couldn't we find a way to mark attributes as used by attaching a `#[used]` before every used attribute Rustc's lint could then ignore these attributes in the unkown attr and unused attr lint?

I think that would be harder. Either you need a &mut AST everywhere or your entire proc macro needs to be structured as a great big fold in order to be able to add #[used] as attributes are used. Here one call takes care of updating the AST.

oli-obk:
I reread the code, and I see now that the edge case I thought of can't happen. The code could use a few more detail comments, because some of the code snippets don't have an obvious reason. Other than that, as always, great work!

dtolnay:
Quick update on why I haven't merged this yet - I am tracking down a case where code that worked before this change now requires adding `#[feature(custom_derive)]`. Everything happens through proc_macro so it doesn't seem like custom_derive should be necessary anywhere. I am working on minimizing the failing case and tracking down the problem.

dtolnay:
Never mind, I had screwed up the `[replace]` section of my Cargo.toml and was pulling in both an old and a new serde_derive, one of which did not have the PostExpansion derive.

dtolnay:
@KodrAus I typically don't like to release late at night and immediately go to bed for 8 hours, so I will release this for you tomorrow morning.

KodrAus:
@dtolnay no worries :+1: I'm currently targetting the branch so am not blocked.

I'll keep an eye out for the release tomorrow. Thanks for the awesome work!

dtolnay:
Released as serde_derive 0.8.15.

PR #594, opened 2016-10-19
No longer need `e!`

dtolnay:
This compiles all the way back to Rust 1.7.0, and 1.6.0 fails for other reasons so I think we are clear here. The referenced issue https://github.com/rust-lang/rust/issues/19630 is still open so they must have fixed it some other way.


oli-obk:
Your attention to detail is remarkable!

@homu r+

bluss:
Don't need `e!()` inside other macros -- the two lines were inside `try!()`, so unrelated to the linked issue.

PR #602, opened 2016-10-29
Serializeable objects

withoutboats:
This PR introduces a `SerializeTo` trait, which is similar to `Serialize`, except that it is object-safe. The `Serializer` trait now takes `SerializeTo<Self>` as a bound anywhere it used to use `Serialize`, and a blanket impl of `SerializeTo<S: Serializer>` is provided for all types which implement `Serialize`. This enables the construction of serializeable trait objects.

**This is a breaking a change.**

Most serializers will probably be able to transform to using this API trivially. However, some serializers take advantage of the current definition in a way which is not compatible with the new definition in this PR. In particular, the feature by which serializers can restrict the values serializable in a certain position by delegating to a separate serializer is no longer supported.

I have adapted the json crate (which currently uses this feature) to the new version in a separate PR. See that PR for background on the rest of this PR comment. [json#164](https://github.com/serde-rs/json/pull/164)

There are some serious trade offs here:
1. This allows trait objects to be serialized - for example, supporting the serialization of heterogeneous trait collections, as some schemas require.
2. This is a breaking change.
3. The fix to the JSON serializer requires specialization. This is an unstable feature.
4. The fix to the JSON serializer _changes which values can be serialized by keys in a JSON object_ (in my opinion, this is an improvement, but your mileage may vary). This is discussed at greater length on the other PR.

I think, because this would make the JSON serializer unstable, it would make sense not to accept this PR until specialization is stabilized.

### DIFF
--- a/serde/Cargo.toml
+++ b/serde/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "serde"
-version = "0.8.16"
+version = "0.9.0"
 authors = ["Erick Tryzelaar <erick.tryzelaar@gmail.com>"]
 license = "MIT/Apache-2.0"
 description = "A generic serialization/deserialization framework"
--- a/serde/src/lib.rs
+++ b/serde/src/lib.rs
@@ -37,7 +37,7 @@ mod core {
     pub use self::core::nonzero;
 }
 
-pub use ser::{Serialize, Serializer};
+pub use ser::{Serialize, Serializer, SerializeTo};
 pub use de::{Deserialize, Deserializer, Error};
 
 #[cfg(not(feature = "std"))]
--- a/serde/src/ser/impls.rs
+++ b/serde/src/ser/impls.rs
@@ -727,7 +727,7 @@ impl<T> Serialize for NonZero<T> where T: Serialize + Zeroable {
 
 
 ///////////////////////////////////////////////////////////////////////////////
-impl<S, T> SerializeTo<S> for T where S: Serializer, T: Serialize {
+impl<S, T: ?Sized> SerializeTo<S> for T where S: Serializer, T: Serialize {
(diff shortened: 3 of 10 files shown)


withoutboats:
If we want to go through with this, I'm happy to open PRs to other serde libs to handle this breaking change for them also.

withoutboats:
I've written up a gist to describe the pattern behind this change, which I'm going to expand and make into a blog post probably. Posting the draft here in case it helps make sense of this PR.

https://gist.github.com/withoutboats/96095e45c1f9ff45d1550eae00087a88

dtolnay:
I would like to benchmark this against erased-serde in a realistic use case. It seems like the only reason so far to do this is "it's probably faster."

withoutboats:
@dtolnay That seems very reasonable, but I'm not sure the best way to devise the benchmark. What do you think?

dtolnay:
I would shoot for two benchmarks to start with:

- An absolute worst case for erased-serde. Basically whatever contrived case you can come up with that demonstrates the biggest possible performance difference between this and erased-serde. That way we have a quantitative upper bound on the value we can hope to get out of this change.
- Some reasonable case, like serializing a heterogeneous collection of some nonprimitive but not overly complicated types to JSON. Maybe a vec of 65,536 objects spanning 8 distinct types, each about 64 bytes in size to give 8 MiB of JSON.

withoutboats:
Those seem like two reasonable benchmarks, though I would characterize the latter as close to best case for erased_serde than the "normal" case; most JSON objects I deal with are much larger than 64 bytes, and the performance improvement in this change grows with the size of the individual objects, not the number of them.

dtolnay:
Let's reopen when somebody is ready to dedicate time to this and we have a clearer commitment to specialization.

PR #604, opened 2016-11-02
Do not use syntex or syntax at all in serde_codegen with-syn

nox:


nox:
I'm not sure this makes sense, but serde_derive still manages to build so I'm confused why it needs syntex or syntax at all.

dtolnay:
This is for backward compatibility. Previously "with-syntex" meant depend on syntex but not libsyntax, and "not(with-syntex)" meant depend on libsyntax and not syntex so that behavior needs to be preserved until 0.9. Otherwise adding serde_derive into a project that depends on serde_codegen through some other dependency could break that other dependency.

PR #635, opened 2016-12-01
Enable deserializing an empty `String` from a unit value

oli-obk:


oli-obk:
failure is due to nightly not being able to update it's crates.io database due to some ssl failure...

oli-obk:
see https://github.com/rust-lang/cargo/issues/3340 for details

dtolnay:
Seems reasonable. I added a test in 5e47c87ba0f0a44d4d632331246e64251946e3c4.

PR #655, opened 2016-12-25
Allow strings in place of other types for map keys

bjorn3:
Just works for types implementing FromStr.

See serde-rs/json#177

### DIFF
--- a/serde/src/de/impls.rs
+++ b/serde/src/de/impls.rs
@@ -743,7 +743,13 @@ macro_rules! map_impl {
             {
                 let mut values = $with_capacity;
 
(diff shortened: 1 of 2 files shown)


serde/src/de/impls.rs
```diff
@@ -743,7 +743,13 @@ macro_rules! map_impl {
             {
                 let mut values = $with_capacity;
 
-                while let Some((key, value)) = try!($visitor.visit()) {
+                while let Some((key, value)) = try!(
+                    $visitor.visit()
```
oli-obk: I don't think all deserializers can cope with running visit twice.

PR #670, opened 2017-01-08
Stop exporting serde::de::Error as serde::Error

### LINKED ISSUE #669: Stop exporting serde::de::Error as serde::Error
This dates back about 2 years to 3fac47e01cbb10c2b6c519f354e9ce1f3792e567, when `serde::de::Error` was the only Error and `serde::ser::Error` did not exist.

Now `serde::Error` is just confusing.

dtolnay:
Fixes #669.

This dates back about 2 years to 3fac47e01cbb10c2b6c519f354e9ce1f3792e567, when `serde::de::Error` was the only Error and `serde::ser::Error` did not exist.

Now `serde::Error` is just confusing.

### DIFF
--- a/serde/src/lib.rs
+++ b/serde/src/lib.rs
@@ -38,7 +38,7 @@ mod core {
 }
 
 pub use ser::{Serialize, Serializer};
-pub use de::{Deserialize, Deserializer, Error};
+pub use de::{Deserialize, Deserializer};
 
 #[cfg(not(feature = "std"))]
 macro_rules! format {


oli-obk:
We can't deprecate reexports, right? Otherwise my vote would be for deprecation for now

dtolnay:
We can't deprecate reexports, I think this is https://github.com/rust-lang/rust/issues/15702. We could keep it with `#[doc(hidden)]` and a comment but what is the advantage? We would just be breaking a potentially larger number of users later when it is eventually removed.

oli-obk:
Right. Let's just get rid of it.

Rust needs some feature to at least help ppl after a breaking change to
tell them what to do...

Are these PRs supposed to get merged or do you want to batch them
before 0.9?

dtolnay:
Yesterday I disabled publish in ad34c14c8c62e0a0c1aa7ed94751934adeed229a and merged two other breaking changes to master. We can spin off a branch from the v0.8.21 tag if we need more 0.8 releases.

PR #671, opened 2017-01-08
Stop generating visit_usize and visit_bytes for FieldVisitor

### LINKED ISSUE #667: Stop generating visit_usize and visit_bytes for FieldVisitor
As far as I know, not a single format uses these. The visit_usize case is better served by going through visit_seq like Bincode does. The visit_bytes case causes problems with supporting no_std: https://github.com/serde-rs/serde/issues/666#issuecomment-271109830.

We can reconsider later if someone comes up with a compelling use case.

dtolnay:
Fixes #667.

### DIFF
--- a/testing/tests/test_de.rs
+++ b/testing/tests/test_de.rs
@@ -723,20 +723,6 @@ declare_tests! {
             Token::EnumMapEnd,
         ],
     }
-    test_enum_unit_usize {
-        Enum::Unit => &[
-            Token::EnumStart("Enum"),
-            Token::Usize(0),
-            Token::Unit,
-        ],
-    }
-    test_enum_unit_bytes {
-        Enum::Unit => &[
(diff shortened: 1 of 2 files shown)


testing/tests/test_de.rs
```diff
@@ -723,20 +723,6 @@ declare_tests! {
...
-            Token::Unit,
-        ],
-    }
-    test_enum_unit_bytes {
-        Enum::Unit => &[
-            Token::EnumStart("Enum"),
-            Token::Bytes(b"Unit"),
-            Token::Unit,
-        ],
-    }
```
oli-obk: Could you move these tests to the failure tests instead of removing them?

dtolnay: Good call, done. I like how easy it is to add failure tests!

Issue #677, opened 2017-01-11
How to handle impossible enum


dtolnay:
```rust
#[derive(Deserialize)]
enum Impossible {}
```

Should this fail at compile time or generate a Deserialize implementation that always fails at runtime?

oli-obk:
I think we need the impl, because we might have `Vec<Impossible>` or something similar that can be deserialized, but the `Deserialize` impl requires a `Deserialize` impl for the generic type. The same goes for serialization

oli-obk:
alternatively we need to patch all `Vec`, `Option`, ... impls with specialization to allow for the case where we deserialize a `None::<NotDeserializable>`.

dtolnay:
That makes sense. I added some tests of this case in #678.

PR #678, opened 2017-01-11
Handle various degenerate cases

### LINKED ISSUE #676: Empty braced struct generates dead code
```rust
#[derive(Deserialize)]
#[serde(deny_unknown_fields)]
pub struct Config {
}
```

The generated code (cleaned up) looks like:

```rust
impl Deserialize for Config {
    fn deserialize<D>(deserializer: &mut D) -> Result<Config, D::Error>
        where D: Deserializer
    {
        enum Field {}

        impl Deserialize for Field {
            /* ... */
        }

        struct Visitor<D: Deserializer>(PhantomData<D>);

        impl<D: Deserializer> Visitor for Visitor<D> {
            type Value = Config;
            fn visit_seq<V>(&mut self, mut visitor: V) -> Result<Config, V::Error>
                where V: SeqVisitor
            {
                try!(visitor.end());
                Ok(Config{})
            }
            fn visit_map<V>(&mut self, mut visitor: V) -> Result<Config, V::Error>
                where V: MapVisitor
            {
                // ERROR: irrefutable while-let pattern
                while let Some(key) = try!(visitor.visit_key::<Field>()) {
                    match key {}
                }
                try!(visitor.end());
                Ok(Config{})
            }
        }

        const FIELDS: &'static [&'static str] = &[];
        deserializer.deserialize_struct("Config", FIELDS, Visitor::<D>(PhantomData))
    }
}
```

This is an error on Rust 1.16. We need to omit the Field enum and the while loop in the empty braced struct case.

dtolnay:
Fixes #676 and adds tests for #677.

### DIFF
--- a/serde_codegen/src/de.rs
+++ b/serde_codegen/src/de.rs
@@ -509,14 +509,6 @@ fn deserialize_item_enum(
         const VARIANTS: &'static [&'static str] = &[ #(#variant_names),* ];
     };
 
-    let ignored_arm = if item_attrs.deny_unknown_fields() {
-        None
-    } else {
-        Some(quote! {
-            __Field::__ignore => { Err(_serde::de::Error::end_of_stream()) }
-        })
-    };
-
     // Match arms to extract a variant from a string
     let mut variant_arms = vec![];
     for (i, variant) in variants.iter().filter(|variant| !variant.attrs.skip_deserializing()).enumerate() {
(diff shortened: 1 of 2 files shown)


serde_codegen/src/de.rs
```diff
@@ -536,7 +528,20 @@ fn deserialize_item_enum(
         };
         variant_arms.push(arm);
     }
-    variant_arms.extend(ignored_arm.into_iter());
+
+    let match_variant = if variant_arms.is_empty() {
+        // This is an empty enum like `enum Impossible {}` or an enum in which
+        // all variants have `#[serde(skip_deserializing)]`.
+        quote! {
+            visitor.visit_variant::<__Field>().map(|impossible| match impossible {})
```
oli-obk: also add the `// FIXME: Replace once we drop support for Rust 1.15:` comment

dtolnay: Added in b01c23b5ee3ba96c286b22fb8f958ea6b2576fde. This case is a little different because `visit_variant` returns `Result<__Field>` rather than `Result<Option<__Field>>`.

Issue #571, opened 2016-10-02
impl<T> Serialize for Iterator<T> ? 


Cobrand:
it would be nice to be able to do something like this : 

```
let v = vec![1,7,3,5,9];
let string = serde_json::to_string(v.iter().filter(|value| value > 3))
```

Don't you think ?

If no one is against it I will implement this by next week (and open a PR as needed)

dtolnay:
Rust will allow you to write `impl<T> Serialize for Iterator<Item = T>` but it will only apply to Iterator trait objects (in practice, only `&Iterator<Item = T>` and `Box<Iterator<Item = T>>`) not the iterator in your example.

```
let string = serde_json::to_string(&v.iter().filter(|&&value| value > 3));
                         ^^^^^^^^^ trait `std::iter::Filter<std::slice::Iter<'_, {integer}>, [closure@test.rs:24:32: 24:51]>: serde::Serialize` not satisfied
```

The correct impl would be `impl<T, I> Serialize for I where I: Iterator<Item = T>, T: Serialize` but it conflicts with at least one existing impl:

```
117 | impl Serialize for String {
    | - first implementation here
...
128 | impl<T, I> Serialize for I where I: Iterator<Item = T>, T: Serialize {
    | ^ conflicting implementation for `std::string::String`
```

I am keeping this open because there may be something we can do to make this better (especially when specialization is stable) but it probably won't be as simple as `impl<T> Serialize for Iterator<T>`.

Cobrand:
You are right, it was just an idea I didn't think about the actual impl too much. I didn't realize that it would conflict with others impls, too. I guess this will be on hold until specialization hits stable.

jmesmon:
One could possibly provide a wrapper type to do this serialization:

```
struct SerializeIterator<T: Serialize, I: Iterator<Item = T>>(a: I);
impl<T, I> Serialize for SerializeIterator<T, I> where I: Iterator<Item = T>, T: Serialize { ... }
```

(though choosing a more convenient name or using a extension trait for construction might be preferable)

This isn't perfect because it makes handling structures containing iterators that aren't top level (and it isn't very composable if one requires other trait impls), but it might still be useful to have.

bluss:
[...]

PR #679, opened 2017-01-12
add wrapper to serde that allows serializing iterators as sequences

oli-obk:
addresses #571


oli-obk:
If you prefer I'll refactor everything to `Serialize` by value and all the impls to impls for references.

dtolnay:
I like it and I think RefCell\<Option\<I>> is fine. My only feedback is IteratorSerializer seems like the wrong name - Serializer means something else and this is not one. How about just serde::ser::Iterator?

PR #680, opened 2017-01-12
make `ser::iterator` more general by taking `IntoIterator`

oli-obk:
When I left the comment in the issue I noticed that the code wasn't as general as it could be. Alternatively the `IntoIterator` could be forwarded all the way into the `Serialize` function, as to not call `into_iter` before it is needed. Any preferences?

### DIFF
--- a/serde/src/ser/impls.rs
+++ b/serde/src/ser/impls.rs
@@ -226,7 +226,7 @@ array_impls!(32);
 
 #[cfg(feature = "unstable")]
 impl<'a, I> Serialize for Iterator<I>
-    where I: iter::Iterator, <I as iter::Iterator>::Item: Serialize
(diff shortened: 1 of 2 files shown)


dtolnay:
I like this approach but the return type should be `Iterator<I::IntoIter>`.

dtolnay:
I have to go for a couple hours but feel free to merge this once you get the types right. I think you want:

```rust
impl<I> Serialize for Iterator<I>
    where I: iter::Iterator,
          I::Item: Serialize

pub struct Iterator<I>(RefCell<Option<I>>); // no bounds

pub fn iterator<I>(iter: I) -> Iterator<I::IntoIter>
    where I: IntoIterator,
          I::Item: Serialize
```

Issue #444, opened 2016-07-17
Fields marked skip_deserializing are ignored even if deny_unknown_fields is set


sfackler:
I have [this struct definition](https://github.com/sfackler/log4rs/blob/6eea8767b5ed5c014d06685d78a6df8c54147a92/src/file/serde.rs.in#L1-L14). It is configured to derive `Deserialize` and deny unknown fields. I have a field  marked to be skipped, but [the generated implementation](https://github.com/sfackler/log4rs/blob/6eea8767b5ed5c014d06685d78a6df8c54147a92/src/file/serde.rs#L9-L397) appears to [explicitly ignore](https://github.com/sfackler/log4rs/blob/6eea8767b5ed5c014d06685d78a6df8c54147a92/src/file/serde.rs#L347-L350) the skipped field.

I would expect the implementation would treat that field as not even existing, in which case we would see an error if `_p` is referenced in the input being deserialized.

dtolnay:
@oli-obk @erickt opinion on this?

It seems like there are two different use cases for `skip_deserializing`.

1. A field of the struct is not present in the serialized form.
2. A field of the struct is present in the serialized form but should not be deserialized.

If we make this change, there would be no way to handle the second type together with `deny_unknown_fields`.

Maybe we need a second attribute to differentiate the two cases.

dtolnay:
As an initial proposal to iterate on, how about the following change:

- Keep the current behavior of `skip_deserializing` instead of what @sfackler suggests. **"Skip deserializing"** means the field may (or may not) be present in the serialized form, even when using `deny_unknown_fields`.
- Add an attribute `#[serde(not_serialized)]` as the equivalent of the current `#[serde(skip_serializing, skip_deserializing)]` but which behaves the way @sfackler wants when using `deny_unknown_fields`. If a field is **"not serialized"**, it is an error for it to show up when deserializing.

@sfackler please confirm that this would fit your use case.

sfackler:
`#[serde(not_serialized)]` would do what I'm looking for, though it seems confusingly similar to the existing modifiers. In what contexts would someone be looking for the "A field of the struct is present in the serialized form but should not be deserialized." use case?

oli-obk:
> In what contexts would someone be looking for the "A field of the struct is present in the serialized form but should not be deserialized." use case?

[...]

PR #681, opened 2017-01-12
fix various serde(rename) bugs

### LINKED ISSUE #559: `deserialize_struct` is called with real field names instead of deserializable names
``` rust
#[derive(Deserialize)]
pub struct S {
    #[serde(rename = "A")]
    a: u8,
    #[serde(skip_deserializing)]
    b: u8,
}
```

I was surprised to see this in the generated code:

``` rust
const FIELDS: &'static [&'static str] = &["a", "b"];
deserializer.deserialize_struct("S", FIELDS, __Visitor)
```

If I had been asked to guess, I would have guessed it would be `["A"]` instead of `["a", "b"]`. Unclear whether this affects anything because I don't know what uses those field names.

oli-obk:
fixes #559 

writing tests for this is hard, because we never have a string available, just a generic `K: Deserialize`, we could create a custom Deserializer that allows exactly one string deserialization and it has to match a given string. It's somewhat roundabout, but that's the best I can think of.

### DIFF
--- a/serde_codegen/src/de.rs
+++ b/serde_codegen/src/de.rs
@@ -503,7 +503,7 @@ fn deserialize_item_enum(
         true,
     );
 
-    let variant_names = variants.iter().map(|variant| variant.ident.to_string());
+    let variant_names = variants.iter().map(|variant| variant.attrs.name().deserialize_name());
 
     let variants_stmt = quote! {
         const VARIANTS: &'static [&'static str] = &[ #(#variant_names),* ];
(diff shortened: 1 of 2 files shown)


dtolnay:
[Here](https://users.rust-lang.org/t/how-to-tell-if-a-struct-contains-a-key/7406/7?u=dtolnay) is a way to tell what fields are being passed to a Deserializer. We could use something like that in a test and assert on the field names all at once.

Can you also check that `#[serde(skip_deserializing)]` fields are not included in the slice of field names? Skipped fields should act like they don't exist from the point of view of the Deserializer. (This will be true after we fix #444, anyway.)

dtolnay:
I filed #683 to follow up on tests and so I can fix merge conflicts in the other PR before I go to bed.

PR #682, opened 2017-01-13
Treat skipped fields as unknown

### LINKED ISSUE #444: Fields marked skip_deserializing are ignored even if deny_unknown_fields is set

dtolnay:
Fixes #444.

Previously, skipped fields were included in the __Field enum. Now they are not, so the variants of the __Field enum may be non-consecutive like __field0, __field2 if the second field is skipped. Skipped fields will error with Error::unknown_field if using deny_unknown_fields and will go through the usual __Field::__ignore variant otherwise.

### DIFF
--- a/serde_codegen/src/de.rs
+++ b/serde_codegen/src/de.rs
@@ -702,20 +702,11 @@ fn deserialize_struct_visitor(
     fields: &[Field],
     item_attrs: &attr::Item,
 ) -> (Tokens, Tokens, Tokens) {
-<<<<<<< HEAD
     let field_names_idents = fields.iter()
         .enumerate()
         .filter(|&(_, field)| !field.attrs.skip_deserializing())
         .map(|(i, field)| (field.attrs.name().deserialize_name(), field_i(i)))
-||||||| merged common ancestors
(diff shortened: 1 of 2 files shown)


oli-obk:
Can you add a test for a struct with all fields skipped? It looks like the code does all the right checks, but it never hurts to check

dtolnay:
Added.

PR #684, opened 2017-01-13
Remove SeqVisitor::end() and MapVisitor::end()

### LINKED ISSUE #656: Simplify contract around SeqVisitor::end() and MapVisitor::end()
These are really easy to screw up when implementing visit_seq and visit_map. If you forget to call them, things may work in some formats and fail in really confusing ways in other formats.

Hopefully we can find a way for the type system to enforce correct behavior here.

dtolnay:
Fixes #656.

I need to think more about this and also update serde_json before merging.


dtolnay:
I am pretty happy with this approach. It was an easy upgrade in serde_json: https://github.com/serde-rs/json/pull/185.

Issue #651, opened 2016-12-24
Enforce correct use of Serialize and Deserialize traits


dtolnay:
Currently the responsibilities of these traits are:

- **Serialize** - make exactly one primitive Serializer call or exactly one chain of begin/elt/end complex Serializer calls
- **Deserialize** - make exactly one Deserializer call

But the type system does not enforce this in either case - you can make zero or more than one call. The result will be a correctly-written Serializer generating invalid output or a correctly-written Deserializer accepting invalid input / rejecting valid input.

Let's see if we can foolproof these.

oli-obk:
If we change the `Serialize` trait to take a `SerializerRef` object that is consumed instead of taking references (and internally has a reference to the actual Serializer) we might be able to enforce it through ownership

dtolnay:
Yep ownership can help enforce <=1 call. For deserialization I would also like to enforce >=1 call i.e. forbid a Deserialize implementation that just returns Ok(Self) without invoking the Deserializer. And for serialization I would like to enforce >=1 call in order to return Ok (should be fine to Err immediately without serializing anything).

dtolnay:
<=1 call turned out to be quite easy on the Deserializer side - just change every method to take `self` instead of `&mut self` and implement Deserializer for `&'a mut JsonDeserializer` instead of for `JsonDeserializer`.

More complicated on the Serializer side because of how we do seq / seq_elt / seq_end. They could shuffle ownership of the Serializer into the SeqState object but that raises the complexity for Serializer implementers...

This commit https://github.com/servo/rust-url/pull/252/commits/ad8d172b7e220a752003d259d507ba7eb9d9d23d is an example of what I would like to prevent.

dtolnay:
Here is a Serializer API that can only be used correctly. Still need to benchmark and mitigate any performance impact.

```rust
pub trait Serializer {
    /// Everybody set Ok = ().
    type Ok;
    type Error: Error;
    type SerializeSeq: SerializeSeq<Ok = Self::Ok, Error = Self::Error>;
    fn serialize_u8(self, v: u8) -> Result<Self::Ok, Self::Error>;
    fn serialize_seq(self, len: Option<usize>) -> Result<Self::SerializeSeq, Self::Error>;
}

pub trait SerializeSeq {
    type Ok;
    type Error: Error;
    fn serialize_elt<T: Serialize>(&mut self, value: T) -> Result<(), Self::Error>;
[...]

PR #686, opened 2017-01-14
Enforce correct use of Deserialize trait

dtolnay:
This ensures Deserialize::deserialize calls Deserializer::deserialize_\* at most once.

Half of #651.

### DIFF
--- a/serde_test/src/de.rs
+++ b/serde_test/src/de.rs
@@ -3,11 +3,13 @@ use std::iter;
 use serde::de::{
     self,
     Deserialize,
+    EnumVisitor,
     MapVisitor,
     SeqVisitor,
     VariantVisitor,
     Visitor,
 };
+use serde::de::value::ValueDeserializer;
 
 use error::Error;
 use token::Token;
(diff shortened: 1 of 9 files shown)


dtolnay:
JSON update: https://github.com/serde-rs/json/commit/1bf409f543cfe74c0d8b406c8ec2594294b42ad0

serde_test/src/de.rs
```diff
@@ -496,7 +498,7 @@ impl<'a, I> SeqVisitor for DeserializerSeqVisitor<'a, I>
             Some(&Token::SeqSep) => {
                 self.de.tokens.next();
                 self.len = self.len.map(|len| len - 1);
-                Ok(Some(try!(Deserialize::deserialize(self.de))))
+                Deserialize::deserialize(&mut *self.de).map(Some)
```
oli-obk: What's with the `&mut *`? A reborrow issue?

dtolnay: Yes. Without it:

```
error[E0507]: cannot move out of borrowed content
   --> src/de.rs:501:42
    |
501 |                 Deserialize::deserialize(self.de).map(Some)
    |                                          ^^^^ cannot move out of borrowed content
```

Previously `Deserialize::deserialize` took the argument by mut reference and `self.de` is a mut reference so it was fine. Now `Deserialize::deserialize` takes the argument by value so Rust sees this as moving the reference out of `self`.

dtolnay:
I filed #688 to follow up with compile-fail tests.

PR #691, opened 2017-01-16
abstract over formats that deal with bytes

### LINKED ISSUE #644: Abstract over formats that deal with bytes
Currently there is no way to abstract over different data formats that convert Serialize -> bytes and bytes -> Deserialize. This is a common API for Serde data formats (but not all of them). It may be useful to have something like:

```rust
trait ByteSerializer: Serializer {
    fn to_writer<W, T: ?Sized>(writer: W, value: &T) -> Result<(), Self::Error>
        where W: Write, T: Serialize;

    fn to_string<T: ?Sized>(value: &T) -> Result<String, Self::Error>
        where T: Serialize
    { /* default implementation */ }
    fn to_bytes<T: ?Sized>(value: &T) -> Result<Vec<u8>, Self::Error>
        where T: Serialize
    { /* default implementation */ }
}

trait ByteDeserializer: Deserializer {
    fn from_reader<R, T>(reader: R) -> Result<T, Self::Error>
        where R: Read, T: Deserialize;

    fn from_str<T>(s: &str) -> Result<T, Self::Error>
        where T: Deserialize
    { /* default implementation */ }
    fn from_bytes<T>(v: &[u8]) -> Result<T, Self::Error>
        where T: Deserialize
    { /* default implementation */ }
}
```

oli-obk:
fixes #644

### DIFF
--- a/serde/src/de/mod.rs
+++ b/serde/src/de/mod.rs
@@ -1,7 +1,7 @@
 //! Generic deserialization framework.
 
 #[cfg(feature = "std")]
-use std::error;
+use std::{error, io};
 #[cfg(not(feature = "std"))]
 use error;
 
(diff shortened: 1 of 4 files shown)


dtolnay:
I need to go but I will finish reviewing this afternoon. Looks great so far. I probably would not have remembered to implement these for the various ValueDeserializers but it makes sense.

dtolnay:
I know these traits came from my comment but I tried writing some code against this and I don't think it is the right approach. I commented in #644, let's continue discussion there.

oli-obk:
I played around some more. It's very messy and hard to use correctly. I'll park this and revisit again later.

Issue #668, opened 2017-01-07
Serde 0.9.0


dtolnay:
I would like to release Serde 0.9.0 some time in the week of 1/15 - 1/21. This is 6 months after the release of 0.8.0 and 2 weeks before the stabilization of Macros 1.1.

Here are the open issues I intend to include in 0.9: https://github.com/serde-rs/serde/milestone/16

Possibly controversially, I would like to raise the minimum supported Rust version to 1.15.0 and drop support for Syntex. Otherwise we will be supporting that mess for a long time.

In between the release and Rust 1.15, we can help the major players update (bincode, msgpack, toml, servo, rusoto) and get the ecosystem in a good place in time for any increase in adoption brought on by Rust 1.15.

Not included in this release are any changes related to specialization, bignums, or object-safety. We can continue working on those in time for a 1.0 release later this year.

oli-obk:
Can't we release after 0.15 so noone needs to use nightly even in the transition period?

dtolnay:
Mainly I am contrasting these two perceptions of Serde:

- "At one point for a period of a week or two, the latest Serde release required a not-yet-stable feature but now that feature is stable and Serde will always work on stable."
- "I tried Serde for the first time and got errors like `the trait serde::ser::Serialize is not implemented for S` even though I did everything right according to the docs and put `#[derive(Serialize)]` on the thing, everything sucks."

The two major non-technical obstacles we need to overcome before 1.0 are "Serde is unstable" and "Serde is confusing." I think given "Serde is unstable now but will be stable in 2 weeks" it is easy to believe "Serde will continue to be stable," whereas given "Serde is busted now but will be less busted in 2 weeks" it is harder to convince someone "Serde will continue being less busted." In that sense the second perception does more lasting damage to our reputation.

Another aspect is support burden. If we are anticipating an increase in adoption, I would rather spend support hours on the interesting new use cases rather than mechanics of why two libraries don't work together in ways that give useless error messages. I think doing the release early will shift the brokenness in the direction that has fewer new users, which is good.

Existing users are not forced to use nightly, they just stay on 0.8 until ready.

I haven't made up my mind yet which way is better, but that is my thought process. Let me know if you have a different point of view.

dtolnay:
Reasonable suggestion from @sfackler:

[...]

PR #696, opened 2017-01-17
use serde_derive in our unit tests

oli-obk:


oli-obk:
oh... I could have just run the tests from `serde_derive`, cool, that makes this much easier

dtolnay:
We'll clean all this stuff up after dropping syntex. I plan to support serde_codegen for 0.9.0 and then immediately abandon it https://github.com/serde-rs/serde/issues/668#issuecomment-271234652.

PR #699, opened 2017-01-18
Remove visit_usize from Duration impl

dtolnay:
This is equivalent to #671 but for this hardcoded Deserialize impl.


dtolnay:
Per #700, we actually want to keep visit_bytes so I added that back. The build failure on nightly is because of clippy.

dtolnay:
I fixed the conflict.

PR #700, opened 2017-01-18
Restore visit_usize and visit_bytes for identifying variants

dtolnay:
I removed this in #671 but it is still used by Bincode [here](https://github.com/TyOverby/bincode/blob/54205283375e6a3d79e6506225f8e15a7cab7520/src/serde/reader.rs#L519-L520).

### DIFF
--- a/serde_codegen/src/de.rs
+++ b/serde_codegen/src/de.rs
@@ -657,6 +657,25 @@ fn deserialize_field_visitor(
         Some(quote!(__ignore,))
     };
 
+    let visit_usize = if is_variant {
+        let variant_indices = 0usize..;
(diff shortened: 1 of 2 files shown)


oli-obk:
I solved this in https://github.com/serde-rs/serde/pull/701/files#diff-4240a2637a13f212fb9d53f7137d828eR762 by looking into the variants arrays given to the `deserialize_enum` method. But either is fine with me. Merge this if you think it's good to support deserialization by index.

dtolnay:
You convinced me in IRC that visit_bytes is important for performance reasons so let me add that back in as well...

dtolnay:
Updated. The build failure on nightly is because of clippy.

dtolnay:
I fixed the conflict.

PR #701, opened 2017-01-18
Expose number of variants when serializing enum

### LINKED ISSUE #663: Expose number of variants when serializing enum
From IRC:

> **\<cmr>** Hm, this seems problematic. My serializer can't keep unbounded state, but I want to encode enum variant tags with the minimum amount of bytes necessary, using the same representation Rust does ideally.
> **\<cmr>** The obvious approach of using a stack of recent total-number-of-variants is out, but I don't see how else it's possible.
> **\<cmr>** (in fact, afaict, the serializer \*never even has access\* to that information, although the deserializer does.)
> **\<cmr>** in the form of the len of the variants arg to deserialize_enum

cc @cmr

oli-obk:
fixes #663 

drive-by fix for deserializing enums with skipped variants

### DIFF
--- a/serde/src/ser/impls.rs
+++ b/serde/src/ser/impls.rs
@@ -74,6 +74,7 @@ use super::{
     SerializeStruct,
     SerializeTuple,
     Serializer,
+    SerializeEnum,
 };
 
 #[cfg(feature = "unstable")]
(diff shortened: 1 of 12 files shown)


dtolnay:
See https://github.com/serde-rs/serde/issues/663#issuecomment-273652015. I changed my mind about the importance of supporting this use case. Basically I failed to come up with an approach that I prefer more than the one you have (so nicely done :smile:) but looking at the tradeoff this involves for both Serialize and Serializer implementations, I don't think it makes sense for us to make this change.

We definitely need the drive-by fixes though. Good catch on that! Can you extract it out?

dtolnay:
I pulled out the drive-by fix into #711.

PR #703, opened 2017-01-19
Stateful deserialization

### LINKED ISSUE #650: Stateful deserialization
In many places the current Serde API has reasonable support for deserializing in a stateful way - where the deserialization depends on some data structure that previously exists. As a concrete example the following code deserializes a JSON array into a preallocated slice. Something similar could be useful in a time series database that wants to reuse a single buffer rather than allocating a new Vec for each array in the input.

```rust
extern crate serde;
extern crate serde_json;

use serde::de::{Deserialize, Deserializer, Visitor, SeqVisitor};

fn main() {
    let j = "[10, 20, 30]";
    let mut vec = vec![0; 5];

    deserialize_into_slice(&mut vec, j).unwrap();

    println!("{:?}", vec); // [10, 20, 30, 0, 0]
}

fn deserialize_into_slice<T>(out: &mut [T], j: &str) -> Result<(), serde_json::Error>
    where T: Deserialize
{
    struct IntoSlice<'a, T: 'a>(&'a mut [T]);

    impl<'a, T> Visitor for IntoSlice<'a, T>
        where T: Deserialize
    {
        type Value = ();

        fn visit_seq<V>(&mut self, mut visitor: V) -> Result<Self::Value, V::Error>
            where V: SeqVisitor
        {
            let mut iter = self.0.iter_mut();
            while let Some(elem) = visitor.visit()? {
                iter.next().map(|spot| *spot = elem);
            }
            visitor.end()
        }
    }

    let mut deserializer = serde_json::Deserializer::new(j.bytes().map(Ok));
    deserializer.deserialize_seq(IntoSlice(out))
}
```

In other places the support is a lot worse. Some libraries ([serde-transcode](https://github.com/sfackler/serde-transcode), [erased-serde](https://github.com/dtolnay/erased-serde)) are able to work around the limitations by using a thread_local stack to pass state into Deserialize::deserialize but this is confusing and unsafe.

Let's see if we can come up with an API that supports stateful use case in a better way. It may even be possible to prototype a solution as a helper crate which uses thread_local behind the scenes.

CC @sfackler who has been hoping for progress on this.

dtolnay:
I went with the name `DeserializeSeed` for the trait but we can discuss. It contains the word "Deserialize" to suggest that it is analogous to the Deserialize trait. I want to avoid anything ending in "-er" like "Initializer" because those sound more similar to Deserializer. I used the word "Seed" rather than "State" to highlight the unique property of this trait which is we can start with some piece of data up front, whereas "state" places more emphasis on mutating the data along the way which is something Deserialize can do just as well.

Fixes #650.

### DIFF
--- a/serde/src/de/mod.rs
+++ b/serde/src/de/mod.rs
@@ -9,6 +9,7 @@ use error;
 use collections::{String, Vec};
 
 use core::fmt;
+use core::marker::PhantomData;
 
 ///////////////////////////////////////////////////////////////////////////////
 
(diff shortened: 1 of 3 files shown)


dtolnay:
The build failure looks like clippy.

dtolnay:
Damn, did not mean to merge this. I reopened #650 and will follow up on your comments in another PR.

serde/src/de/mod.rs
```diff
@@ -230,6 +231,149 @@ pub trait Deserialize: Sized {
...
+    /// The type produced by using this seed.
+    type Value;
+
+    /// Equivalent to the more common `Deserialize::deserialize` method, except
+    /// with some initial piece of data (the seed) passed in.
+    fn deserialize<D>(self, deserializer: D) -> Result<Self::Value, D::Error>
+        where D: Deserializer;
+}
+
+impl<T> DeserializeSeed for PhantomData<T>
```
dtolnay: PhantomData is a bit magical in that it has a type parameter but you can use it like a unit struct.

```rust
pub struct PhantomData<T: ?Sized>;

pub struct NoSeed<T: ?Sized>(PhantomData<T>);
```

So you can call `f(PhantomData)` but you cannot call `f(NoSeed)`. It would need to be `f(NoSeed::new())` or `f(NoSeed(PhantomData))` or something unfortunate like that.

The workaround for this would be what we did in [serde-stateful-deserialize](https://github.com/sfackler/serde-stateful-deserialize) which is:

```rust
pub use std::marker::PhantomData as NoSeed;
```

This reexports both the type and the unit-like way of referring to it, so it lets you do `f(NoSeed)`. The downside is there is no way to override PhantomData's documentation so in our rustdoc there would be a type `serde::de::NoSeed` with all of PhantomData's documentation and impls which is confusing.

Do you think NoSeed::new() would work better?

dtolnay: Oh it could also be `DeserializeSeed::none()`. How about that?

serde/src/de/mod.rs
```diff
@@ -641,10 +785,26 @@ pub trait SeqVisitor {
...
+    /// `Deserialize` implementations should typically use `SeqVisitor::visit`
+    /// instead.
+    fn visit_seed<T>(&mut self, seed: T) -> Result<Option<T::Value>, Self::Error>
+        where T: DeserializeSeed;
+
+    /// This returns `Ok(Some(value))` for the next value in the sequence, or
+    /// `Ok(None)` if there are no more remaining items.
+    ///
+    /// This method exists as a convenience for `Deserialize` implementations.
+    /// `SeqVisitor` implementations should not need to override the default
```
dtolnay: I can't think of any reason.

The three places to put this function are:

- Free function like serde::ser::visit_seq_element(seq_visitor),
- Separate trait where we impl<T> SeqVisitorExt for T where T: SeqVisitor,
- Here.

I think of the three, this is the best tradeoff.

- I don't like the first option because if you write a method that takes in a SeqVisitor, your preference is going to be to call methods on the SeqVisitor. I would prefer for Deserialize implementations to deal with the non-seed method wherever possible and putting them here at least puts them on equal footing with the seed methods.
- I don't like the second option because Deserialize will need to `use` an extra trait before they can use the methods and again they will prefer not doing that.

Deserializer implementors will be much more likely to be reading our rustdoc than Deserialize implementors, so a doc comment here that says you don't need to override the behavior counteracts most of the disadvantage.

serde/src/de/mod.rs
```diff
@@ -678,49 +845,140 @@ pub trait MapVisitor {
...
     type Error: Error;
 
-    /// This returns a `Ok(Some((key, value)))` for the next (key-value) pair in the map, or
-    /// `Ok(None)` if there are no more remaining items.
+    /// This returns `Ok(Some(key))` for the next key in the map, or `Ok(None)`
+    /// if there are no more remaining entries.
+    ///
+    /// `Deserialize` implementations should typically use
+    /// `MapVisitor::visit_key` or `MapVisitor::visit` instead.
+    fn visit_key_seed<K>(&mut self, seed: K) -> Result<Option<K::Value>, Self::Error>
```
dtolnay: Me too. But the updates in serde/src/de/value.rs show it is the same amount of code for implementors. I will be putting together a comprehensive guide for writing Deserializers over the next two weeks and I will explain this thoroughly.

serde/src/de/mod.rs
dtolnay: I couldn't get DeserializeSeed::none() to work :crying_cat_face:. Here is what I tried:

```rust
pub trait DeserializeSeed { // no more Sized bound
    type Value;

        where D: Deserializer,
              Self: Sized; // trait object hack
}

impl<T> DeserializeSeed<Value = T> {
    pub fn none() -> PhantomData<T> {
        PhantomData
    }
}
```

This compiles but fails when you try to use it:

```
   |
11 | let vec: Vec<u32> = f(DeserializeSeed::none())?;
   |                       ^^^^^^^^^^^^^^^^^^^^^ missing associated type `Value` value
```

In theory it could infer that `f` takes `D: DeserializeSeed` and returns `D::Value`, `none` returns `PhantomData<T>` which implements `DeserializeSeed<Value = T>`, and the return from `f` is `Vec<u32>` so that means we need `DeserializeSeed::<Vec<u32>>::none` but I guess it doesn't work that way.

oli-obk: You could create a `NoSeed` constant that produces a `NoSeed` value: https://is.gd/xKxNk2

serde/src/de/mod.rs
oli-obk: Please add some note to the docs that this method should NOT be overridden and is only here for convenience.

serde/src/de/mod.rs
dtolnay: But we need the type parameter because we need:

```rust
impl<T> DeserializeSeed for ??? {
    type Value = T;
    /* ... */
}
```

Otherwise I could just use an ordinary unit struct as NoSeed.

oli-obk: oh right... no generic constants...

PR #706, opened 2017-01-20
Standardize behavior of missing fields

### LINKED ISSUE #702: Reconsider MapVisitor::missing_field
Out of all the Serde deserializers on GitHub, there are exactly four distinct behaviors for missing_field:

- Unconditionally an error
- For options treat it as None, otherwise error
- Treat it as unit
- [Whatever serde_xml is doing](https://github.com/serde-rs/xml/blob/f9330b703c7405ea8f54e1adfa972cb869924e38/src/de/value.rs#L358-L382)

Ideally I would like to standardize on one of these (my preference is None/error) and remove the ability for deserializers to modify the behavior. This would improve consistency and predictability across the ecosystem.

@oli-obk can you explain the various cases that serde_xml needs to handle in missing_field and whether you can think of any alternative ways to accomplish this?

dtolnay:
Fixes #702. Missing fields are treated as None.

The expectation for handwritten Deserialize implementations is that you know whether the field is optional:

```rust
struct S {
    optional: Option<u8>,
    required: u8,
}

// visit_map:
let mut optional: Option<u8> = None;
let mut required: Option<u8> = None;
while let Some(key) = try!(visitor.visit_key::<Field>()) {
    /* ... */
}
Ok(S {
    optional: optional,
    required: required.ok_or_else(|| Error::missing_field("required"))?,
})
```

### DIFF
--- a/serde/src/de/impls.rs
+++ b/serde/src/de/impls.rs
@@ -1061,11 +1061,11 @@ impl Deserialize for Duration {
                 }
                 let secs = match secs {
                     Some(secs) => secs,
-                    None => try!(visitor.missing_field("secs")),
+                    None => return Err(<V::Error as Error>::missing_field("secs")),
                 };
                 let nanos = match nanos {
                     Some(nanos) => nanos,
-                    None => try!(visitor.missing_field("nanos")),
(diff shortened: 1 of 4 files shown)


dtolnay:
I merged this into #710.

PR #710, opened 2017-01-21
Errors redesign

### LINKED ISSUE #528: Deserialization errors like "Error: Invalid type. Expected `I64`" are hard for users to understand
Thank you for your help parsing mixed strings and structs! Here's a second issue to think over. :-)

Here, I'm parsing a 1,300-line `docker-compose.yml` file which uses bare string notation, and which therefore sometimes accidentally uses an integer value like `1` where it should have used a string:

``` yaml
version: 2
services:
  foo:
    environment:
      NOKOGIRI_USE_SYSTEM_LIBRARIES: 1
      REDIS_URL: redis://redis:6379
```

I deserialize this with the following declaration:

``` rust
    #[serde(default, skip_serializing_if = "BTreeMap::is_empty",
            deserialize_with = "deserialize_map_or_key_value_list")]
    pub environment: BTreeMap<String, String>,
```

The function `deserialize_map_or_key_value_list` contains a `visit_map` function which calls `visitor.visit_value::<String>()` to read a string.

When I run this program over the 1,300 line input file, it gives me the following error:

```
target/debug/examples/normalize < docker-compose.yml > test.yml
Error: Invalid type. Expected `I64`
```

There are several problems here:
1. The error message should be the "other way around", and include both the expected type and the one we actually got.
2. The error message should ideally include a line number so the user knows where in the enormous file the error occurs.
3. The error message should include the actual I64 value.

This would ideally give us something like:

```
<input>:869: Expected String, got I64 value `1`.
```

This would make my program much more user friendly. :-) I can easily provide a slightly better workaround in this particular case, but I'd love to find a more general solution.

### LINKED ISSUE #675: Remove de::Error::end_of_stream
Currently this is used in the following cases:

- Tuple or fixed-size-array being deserialized and there are not enough values in the SeqVisitor. This is better served by `invalid_value`. We may be nowhere near the end of the input stream.
- Char being deserialized from an empty string. Again this is definitely `invalid_value`.
- A Value deserializer being misused...? This will be rendered impossible by #651.
- Generated code for an enum. As in... it's there in the generated code. But it is unreachable. Interesting.

The `de::Error` constructors are designed to be used from the Visitor of a Deserialize type. I struggle to think of any valid reason that a Visitor would think it is in an end_of_stream situation.

Note that Deserializers construct their own errors however they want without going through the `de::Error` constructors, so those can keep returning errors that are equivalent to end of stream. For example JSON uses EOFWhileParsingValue for this purpose.

Overall I would like to focus on improving the messages from invalid_type / invalid_value which I am doing as part of #528. I don't think end_of_stream means anything.

### LINKED ISSUE #702: Reconsider MapVisitor::missing_field

### LINKED ISSUE #709: Remove Visitor::visit_unit_struct
This is a bizarrely placed API. For tuple structs we go deserialize_tuple_struct -> visit_seq, for braced structs we go deserialize_struct -> visit_map. Neither of these involves passing the &'static str name of the struct back into the Visitor. It would be more consistent to remove visit_unit_struct and have unit structs go deserialize_unit_struct -> visit_unit.

dtolnay:
Fixes #528 and fixes #675 and fixes #709 and fixes #702..

Essentially we rely on every Visitor to tell us what type of data it is expecting. See the doc comments in serde/src/de/mod.rs.

In general my goal is to centralize error message building in Serde rather than leaving it to each individual format to reimplement lousy messages. This is a step in that direction and I have a design for https://github.com/serde-rs/json/issues/173 which I will implement in another PR.

### DIFF
--- a/serde/Cargo.toml
+++ b/serde/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "serde"
-version = "0.9.0-rc1"
+version = "0.9.0-rc2"
 authors = ["Erick Tryzelaar <erick.tryzelaar@gmail.com>"]
 license = "MIT/Apache-2.0"
 description = "A generic serialization/deserialization framework"
--- a/serde/src/bytes.rs
+++ b/serde/src/bytes.rs
@@ -181,6 +181,10 @@ mod bytebuf {
     impl de::Visitor for ByteBufVisitor {
         type Value = ByteBuf;
 
+        fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+            formatter.write_str("byte array")
+        }
+
         #[inline]
         fn visit_unit<E>(self) -> Result<ByteBuf, E>
             where E: de::Error,
--- a/serde/src/de/impls.rs
+++ b/serde/src/de/impls.rs
@@ -32,6 +32,7 @@ use collections::enum_set::{CLike, EnumSet};
 #[cfg(all(feature = "unstable", feature = "collections"))]
 use collections::borrow::ToOwned;
 
+use core::fmt;
 use core::hash::{Hash, BuildHasher};
 use core::marker::PhantomData;
 #[cfg(feature = "std")]
@@ -69,7 +70,7 @@ use de::{
     Error,
     MapVisitor,
     SeqVisitor,
-    Type,
+    Unexpected,
     VariantVisitor,
     Visitor,
 };
@@ -82,6 +83,10 @@ pub struct UnitVisitor;
 
 impl Visitor for UnitVisitor {
     type Value = ();
+
+    fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+        formatter.write_str("unit")
+    }
 
     fn visit_unit<E>(self) -> Result<(), E>
         where E: Error,
@@ -112,6 +117,10 @@ pub struct BoolVisitor;
 impl Visitor for BoolVisitor {
     type Value = bool;
 
+    fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+        formatter.write_str("a boolean")
+    }
+
     fn visit_bool<E>(self, v: bool) -> Result<bool, E>
         where E: Error,
     {
@@ -124,7 +133,7 @@ impl Visitor for BoolVisitor {
         match s.trim_matches(::utils::Pattern_White_Space) {
             "true" => Ok(true),
             "false" => Ok(false),
-            _ => Err(Error::invalid_type(Type::Bool)),
+            _ => Err(Error::invalid_type(Unexpected::Str(s), &self)),
         }
     }
 }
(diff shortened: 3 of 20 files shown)


dtolnay:
@oli-obk this is stable enough to review when you have time. I need to follow up on some TODOs in serde/src/de/impls.rs related to invalid_length which I will take care of tomorrow.

dtolnay:
Some of the error messages in the tests are wonky because of the bug in #711.

> "unknown field \`b\`, expected \`a\` or \`b\`"
> "unknown variant \`Skipped\`, expected one of \`Skipped\`, \`Unit\`, \`Simple\`, \`Seq\`, \`Map\`"

I will merge that fix into this branch and update the tests.

dtolnay:
I merged the MapVisitor::missing_field change from #706 into this branch so I have something to tag and work against for serde_json.

oli-obk:
awesome redesign. This will also greatly improve the errors reported on no-std builds.

Arnavion:
I have 0.8 code like:

```rust
impl Deserialize for Foo {
    fn deserialize(...) {
        struct Visitor;
        impl Visitor for Visitor {
            fn visit_str(&mut self, v: &str) {
                some_operation.map_err(|err| Error::invalid_value(err.description()))
            }
        }
    }
```

Where does the underlying error description go in 0.9?

If it goes in the `Unexpected` parameter, that means I would need to use `Unexpected::Other` instead of `Unexpected::Str`:

```rust
fn visit_str(self, v: &str) {
    some_operation.map_err(|err| {
        Error::invalid_value(Unexpected::Other(&format!("{} - {}", s, err.description())), &self);
    })
}
```

If it goes in the `Expected`, then I can use `Unexpected::Str` for the first parameter but the second parameter becomes harder:

```rust
fn expecting(&self, f) {
    f.write_str("a string that can be deserialized into a Foo")
}

fn visit_str(self, v: &str) {
    some_operation.map_err(|err| {
        // can't reuse `expecting` directly since we don't have a Formatter
        let expected = format!("a string that can be deserialized into a Foo - {}", err.description());
        Error::invalid_value(Unexpected::Str(s), &&*expected);
    })
}
```

unless I have some other method that returns the string and use it in both `expecting` and `visit_str`, or I stash the description as a property of `Visitor` and read it in `expecting`.

Maybe `invalid_value` etc could take a third `reason: Option<Display>` parameter?

dtolnay:
@Arnavion I would prefer to use Error::custom for this.

```rust
Error::custom(format!("invalid Foo {:?}: {}", s, err.description()))
```

Compare:

> invalid value: string "xyz", expected a string that can be deserialized into a Foo - string must start with "F"

> invalid Foo "xyz": string must start with "F"

What do you think?

Arnavion:
Yes, I suppose. It's a shame that `expecting` goes to waste.

PR #711, opened 2017-01-21
Filter the slice of expected field and variant names

dtolnay:
This is the drive-by fix from #701.


dtolnay:
This will be tested much more thoroughly by the new error messages in #710.

PR #728, opened 2017-01-31
Introduce ser::Impossible (fixes #694)

### LINKED ISSUE #694: Provide an Impossible compound type serializer
For formats that cannot serialize one of the compound types. JSON and #691 both have needed to do this. Maybe something like:

```rust
enum Void {}
pub struct Impossible<Ok, Error>(Void, PhantomData<Ok>, PhantomData<Error>);

impl<Ok, Error> ser::SerializeSeq for Impossible<Ok, Error> {
    type Ok = Ok;
    type Error = Error;
    fn serialize_element<T: ser::Serialize>(&mut self, _: T) -> Result<(), Error> {
        match self.0 {}
    }
    fn end(self) -> Result<Ok, Error> {
        match self.0 {}
    }
}

// and the other ones
```

Then the Serializer can just do:

```rust
type SerializeSeq = Impossible<(), Error>;
```

And these can be the defaults once we have associated type defaults.

nox:


serde/src/ser/impossible.rs
```diff
@@ -0,0 +1,168 @@
...
+///                      len: Option<usize>)
+///                      -> Result<Self::SerializeSeq, Error> {
+///         // Given ImpossibleSerializer cannot be instantiated, the only
+///         // thing we can do here is to return an error.
+///         Err(...)
+///     }
+///
+///     ...
+/// }
+///
```
dtolnay: Missing the closing ```.

serde/src/lib.rs
```diff
@@ -86,6 +86,8 @@ mod core {
     pub use actual_core::nonzero;
 }
 
+extern crate void;
```
dtolnay: I don't think we are getting enough bang for our buck out of this dependency. The entire functionality we need is provided equally well by:

```rust
enum Void {}
```

```rust
match self.void {}
```

If this were part of serde's public API then I would be all for using the void crate as a well-established way of signaling impossibility, but here it is just a private implementation detail...

serde/Cargo.toml
```diff
@@ -18,14 +18,15 @@ travis-ci = { repository = "serde-rs/serde" }
 [features]
 default = ["std"]
 
-std = []
+std = ["void/std"]
```
nox: Removed the dependency.

serde/src/lib.rs
nox: Removed the dependency.

serde/src/ser/impossible.rs
nox: Added it.

serde/src/ser/impossible.rs
```diff
@@ -0,0 +1,166 @@
...
+///     fn serialize_seq(self, len: Option<usize>) -> Result<(), Error> {
+///         // Given ImpossibleSerializer cannot be instantiated, the only
+///         // thing we can do here is to return an error.
+///         Err(...)
+///     }
+///
+///     ...
+/// }
+///
+pub struct ImpossibleSerializer<Ok, E> {
```
nox: Renamed it.

serde/src/ser/mod.rs
```diff
@@ -104,6 +104,7 @@ use core::cell::RefCell;
 use core::fmt::Display;
 
 mod impls;
+pub mod impossible;
```
nox: Reexported it from `serde::ser`.

dtolnay:
Thanks! I released this in serde 0.9.4 and upgraded serde_json in https://github.com/serde-rs/json/commit/cdf063aefaa9fa3927b4259a94bc16fa53d84232.

PR #731, opened 2017-01-31
No tuple structs with private fields in public API

### LINKED ISSUE #730: No tuple structs with private fields in public API
@nox made the observation in IRC that public tuple structs with private fields leaks the arity of the tuple struct and blocks us from changing the (private) internal representation in certain ways.

```rust
pub struct StrDeserializer<'a, E>(&'a str, PhantomData<E>);
```

```rust
let d: StrDeserializer<E> = /* ... */;
match d {
    // compiles now, but breaks if we change arity
    StrDeserializer(_, _) => { /* ... */ }
}
```

Let's replace these with braced structs to be safe.

dtolnay:
Fixes #730.

### DIFF
--- a/serde/src/de/value.rs
+++ b/serde/src/de/value.rs
@@ -51,7 +51,9 @@ use bytes;
 
 /// This represents all the possible errors that can occur using the `ValueDeserializer`.
 #[derive(Clone, Debug, PartialEq)]
-pub struct Error(ErrorImpl);
+pub struct Error {
+    err: ErrorImpl,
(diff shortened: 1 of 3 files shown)


dtolnay:
@oli-obk what do you think of fixing this now? Of the few people on 0.9 I bet none of them are pattern-matching on these.

oli-obk:
yes let's do this. I'll have the lint checking this done today, too, then we can depend on it not happening again.

PR #735, opened 2017-01-31
Allow running travis build locally

dtolnay:
This lets me run `./travis.sh` locally and it completes all of the builds against all of the trains in 2 minutes 25 seconds on my machine, which is faster than waiting for Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,9 +13,5 @@ matrix:
   include:
     - rust: nightly
       env: CLIPPY=true
-  allow_failures:
-    - rust: nightly
-      env: CLIPPY=true
-  fast_finish: true
 
 script: ./travis.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,12 +18,4 @@ matrix:
       env: CLIPPY=true
   fast_finish: true
 
-before_script:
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -13,12 +13,16 @@ matrix:
...
+    - rust: nightly
+      env: CLIPPY=true
+  fast_finish: true
 
 before_script:
   - pip install 'travis-cargo<0.2' --user
   - export PATH=$HOME/.local/bin:$PATH
 
-script: sh travis.sh
+script: ./travis.sh
```
dtolnay: You were missing `chmod +x` before.

.travis.yml
```diff
@@ -13,13 +13,9 @@ matrix:
   include:
     - rust: nightly
       env: CLIPPY=true
+  allow_failures:
```
oli-obk: hmm... if we do it this way, we will never notice if clippy complains

oli-obk:
lgtm. just wondering about the clippy thing. I chose the "only error if clippy install fails" pattern on purpose.

dtolnay:
You're right, I changed it back to what you had.

Issue #553, opened 2016-09-23
Helper library of useful de/serialize_with functions


dtolnay:
For example the ones from https://github.com/serde-rs/serde/issues/550 and https://github.com/serde-rs/serde-rs.github.io/issues/22. These don't need to be in serde itself but if there are ones that people ask for over and over, we can stick those in a helper crate and provide a better out-of-the-box experience compared to "implement your own serialize_with" or "paste this code into your project."

dtolnay:
Another one: include struct name when serializing struct #554.

dtolnay:
Another deserialize_with: a generic comma-separated list of FromStr values.

dtolnay:
Another deserialize_with: a generic comma-separated list of FromStr values #581.

dtolnay:
#415 / #636, tagged enums.

dtolnay:
#661 - as_base64 for AsRef<[u8]> and from_base64 for From<&[u8]>

oli-obk:
Add a way for deserializing numbers from strings containing numbers (which is a feature that serde 0.9 had implicitly)

albel727:
I suggest providing helper functions that would work for anything implementing TryFrom or FromStr traits.

dtolnay:
Deserializing map and set types while treating duplicate entries as error, and treating duplicate entries as first-one-wins (rather than the current default of last-one-wins). #916

alexreg:
> Deserializing map and set types while treating duplicate entries as error, and treating duplicate entries as first-one-wins (rather than the current default of last-one-wins). #916

How can you treat duplicate entries as both errors and first-one-wins? 🤔

dtolnay:
Two functions.

clarcharr:
It'd be nice if we could have the same level of configuration available for structs, where instead of erroring on duplicates, we could customise them as first-one-wins or last-one-wins.

dtolnay:
Have you ever wanted that in a project? Can you explain the use case in more detail?

dtolnay:
Serialize using Display and deserialize using FromStr, as in https://github.com/serde-rs/json/issues/329.

spease:
I'm not sure if it would make sense to use Display for serialization, since Display is "for user-facing output" and so may drop data or not be interchangeable with FromStr.

std::string::ToString might be more appropriate for serialization, but I don't know if this is how it's intended to be used either.

dtolnay:
#994 -- represent a field as nested JSON string.

[...]

PR #736, opened 2017-01-31
Introduce collect_seq and collect_map

nox:
These are two convenience functions to serialize iterators.

### DIFF
--- a/serde/src/ser/mod.rs
+++ b/serde/src/ser/mod.rs
@@ -625,6 +625,24 @@ pub trait Serializer: Sized {
         }
         serializer.end()
     }
+
+    /// Collect an iterator as a map.
+    ///
+    /// The default implementation serializes each pair yielded by the iterator
+    /// using `Self::SerializeMap`. Implementors should not need to override
+    /// this method.
+    fn collect_map<K, V, I>(self, iter: I) -> Result<Self::Ok, Self::Error>
+        where K: Serialize,
+              V: Serialize,
+              I: IntoIterator<Item = (K, V)>,
+    {
+        let iter = iter.into_iter();
+        let mut serializer = try!(self.serialize_map(iterator_len_hint(&iter)));
+        for (key, value) in iter {
+            try!(serializer.serialize_entry(&key, &value));
+        }
+        serializer.end()
+    }
 }
 
 /// Returned from `Serializer::serialize_seq` and
--- a/serde/src/ser/mod.rs
+++ b/serde/src/ser/mod.rs
@@ -98,9 +98,6 @@ use std::error;
 #[cfg(not(feature = "std"))]
 use error;
 
-#[cfg(feature = "unstable")]
-use core::cell::RefCell;
-
 use core::fmt::Display;
 use core::iter::IntoIterator;
 
@@ -837,30 +834,6 @@ pub trait SerializeStructVariant {
 
     /// Finish serializing a struct variant.
     fn end(self) -> Result<Self::Ok, Self::Error>;
-}
-
-/// A wrapper type for iterators that implements `Serialize` for iterators whose
-/// items implement `Serialize`. Don't use multiple times. Create new versions
-/// of this with the `serde::ser::iterator` function every time you want to
-/// serialize an iterator.
-#[cfg(feature = "unstable")]
-pub struct Iterator<I>
-    where <I as IntoIterator>::Item: Serialize,
-          I: IntoIterator
-{
-    data: RefCell<Option<I>>,
-}
-
-/// Create a wrapper type that can be passed to any function expecting a
-/// `Serialize` and will serialize the given iterator as a sequence.
-#[cfg(feature = "unstable")]
-pub fn iterator<I>(iter: I) -> Iterator<I>
-    where <I as IntoIterator>::Item: Serialize,
-          I: IntoIterator
-{
-    Iterator {
-        data: RefCell::new(Some(iter)),
-    }
 }
 
 fn iterator_len_hint<I: Iterator>(iter: &I) -> Option<usize> {
--- a/serde/src/ser/mod.rs
+++ b/serde/src/ser/mod.rs
@@ -616,7 +616,7 @@ pub trait Serializer: Sized {
               <I as IntoIterator>::Item: Serialize,
     {
         let iter = iter.into_iter();
-        let mut serializer = try!(self.serialize_seq(iterator_len_hint(&iter)));
+        let mut serializer = try!(self.serialize_seq(iter.len_hint()));
         for item in iter {
             try!(serializer.serialize_element(&item));
         }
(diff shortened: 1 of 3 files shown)


nox:
@sfackler on IRC taught me that there is a wrapper behind the feature `unstable` to serialise iterators through a wrapper putting it in a `DOMRefCell`. I posit that we should not do that, because that doesn't help for struct fields and friends, as it is an opaque thing that will consume the iterator on serialisation.

nox:
Silly me, it uses `IntoIterator` too so it doesn't consume it, but given there is no function bound to this behaviour, it can't be used in `#[serde(serialize_with)]`.

If we think the wrapper is useful, we could keep it but simplify it using `mitochondria::MoveCell`. :P

sfackler:
If we were going to do this it'd probably be worth also having one that serializes an iterator of tuples as a map.

nox:
@sfackler Should I rename this one to `serialize_seq`, and make another one `serialize_map`? We could also make `serialize_fixed_size_seq` relying on `TrustedLen` once it's stable too.

dtolnay:
Can you share some example code of how you imagine people using this?

As I see it, there are two primary use cases for these functions:

- Helpers when implementing Serialize for a type. I don't think the tradeoff of including these functions for this purpose makes sense. Just use the loop way and it is 5 straightforward lines and those 5 lines match a pattern that is common to all other compound type serialization in Serde (tuples, tuple structs, structs, struct variants, tuple variants). In my opinion the typing you save by using these helpers does not outweigh the people stumbling over this when browsing our rustdoc, the confusion of mixing up ser::serialize_seq and Serializer::serialize_seq and needing to think about which one to use where.

- As a `serialize_with` function in cases where the type is not from your crate and the original author does not want to add a Serialize implementation. In this case I would prefer to include these in #553 which is a library dedicated to this purpose, rather than in the main serde crate.

nox:
> Can you share some example code of how you imagine people using this?

Same way as the unstable `iterator` thing, except that there is no "this must be serialised only once" as on the `Iterator` struct. Tuples, tuple structs and structs are similar, but there is no legit reason to support them from an iterator, after all tuples, tuple structs and structs cannot be iterated.

As for the fact that it is just a 5 lines loop, the `Pair` trait with its two implementations is certainly less trivial to repeat all the time.

IMO this is core enough that it should be in `serde`.

dtolnay:
Sorry for the silence on my end - I have been busy with #739. I will consider this again tomorrow, including your suggestion from IRC of possibly making these Serializer trait methods with default implementations.

sfackler:
I like the default method approach personally. It's a nice convenience similar to https://doc.rust-lang.org/std/fmt/struct.DebugList.html#method.entries

dtolnay:
It's a little different in that case because DebugList is a struct and not a trait. But sure, let's do it. Any thoughts on naming? Serializer::serialize_iter_as_seq and serialize_iter_as_map aren't perfect. It would be nice to reference [`Extend`](https://doc.rust-lang.org/std/iter/trait.Extend.html) somehow because that is one of the most common methods taking an IntoIterator, but that would fit better as SerializeSeq::extend or something.

nox:
Rather than `Extend`, if you want to reuse iterator terminology, I would opt for `collect_seq` and `collect_map` methods on `Serializer`.

dtolnay:
Good idea, let's go with that. Thanks!

dtolnay:
Also I would prefer not to make the Pair changes. It makes sense for MapDeserializer because people pass around variables of type MapDeserializer\<I, E\> so additional redundant type parameters are a nuisance, but here let's just use K, V type parameters on collect_map.

dtolnay:
Do you expect any serializers to want to override these default implementations? If so, please add a comment that explains why someone would want to override. If not, please add a comment that serializers should not need to override. I don't want serializer implementors to wonder whether the default implementation is doing the wrong thing for their format.

nox:
I would expect serde_json to override it to not even have to allocate an intermediate `String`, instead writing directly on its output the string escaped appropriately.

nox:
@dtolnay I added the comments, I don't expect implementors to have to override this.

serde/src/ser/mod.rs
```diff
@@ -607,6 +608,39 @@ pub trait Serializer {
         variant: &'static str,
         len: usize,
     ) -> Result<Self::SerializeStructVariant, Self::Error>;
+
+    /// Collect an iterator as a sequence.
+    ///
+    /// The default implementation serializes each item yielded by the iterator
+    /// using `Self::SerializeSeq`. Implementors should not need to override
+    /// this method.
+    fn collect_seq<I>(self, iter: I) -> Result<Self::Ok, Self::Error>
```
oli-obk: we can immediately use this in https://github.com/nox/serde/blob/a0a4d782957d351b608d08aa892e7e86317b5182/serde/src/ser/impls.rs#L224

Can we use the same [size-trick](https://github.com/nox/serde/blob/a0a4d782957d351b608d08aa892e7e86317b5182/serde/src/ser/impls.rs#L236) so we don't pass `None`?

Is there a way to use specialization to improve this impl for `ExactSizeIterator`? If so, can this function be written in a way so that is possible (by adding a `feature=unstable` specialization, which gets checked on nightly travis)?

nox: Why would you even trust that size trick? I don't understand what tells you that this is actually correct to do.

oli-obk: > size_hint() is primarily intended to be used for optimizations such as reserving space for the elements of the iterator, but must not be trusted to e.g. omit bounds checks in unsafe code. An incorrect implementation of size_hint() should not lead to memory safety violations.

Quoting the standard, since we don't guarantee that the `Option` in `serialize_seq` is correct, we can depend on the `size_hint` to be just as correct.

All correct `Serialize` impls will benefit from this, by being compatible with formats like `bincode`.

nox: Ok, then I don't understand the second question, if we can just check that the two hints are equal and use that, why specialise for `ExactSizeIterator` at all?

nox:
@oli-obk @dtolnay Btw, when do we kill the `Iterator` wrapper? In my opinion it is misguided.

oli-obk:
> Btw, when do we kill the Iterator wrapper? In my opinion it is misguided.

with your changes there's no need for it. Feel free to remove it in this PR.

serde/src/ser/mod.rs
oli-obk: uhm... because I confused myself? ;) ignore it.

nox: @oli-obk I wonder if there is something to simplify around `serialize_seq!` in `ser::impls` though.

oli-obk: well... we could replace it with `collect_seq`, but I wonder if that optimizes the same way. Without benchmarks that's not a change we should do.

nox: I introduced a trait `LenHint` that is specialised for `I: ExactSizeIterator`.

oli-obk:
@dtolnay: you wanna take another look?

dtolnay:
> well... we could replace it with `collect_seq`, but I wonder if that optimizes the same way. Without benchmarks that's not a change we should do.

There is no difference in performance based on my benchmarks so I removed the old serialize_seq and serialize_map macros in 219abd2e00e689dd13fa4b4f9bc609fc0ec37c5d.

PR #739, opened 2017-02-03
Untagged and internally tagged enums

### LINKED ISSUE #415: Is it possible to handle tagged enums?
``` rust
enum Motion {
  Column(i32),
  Line(i32),
}

enum Command {
  Move(Motion),
  Delete(Motion),
  Input(String),
}

fn main() {
  let cmd = Command::Move(Motion::Column(1));
  let cmd_str = serde_json::to_string(&cmd).unwrap();
  println!("result: {}", cmd_str);
}
```

this will get result like this:

``` json
{
  "Move": {
    "Column": 1
  }
}
```

Is it possible to make it be:

``` json
{
  "type": "Move",
  "content": {
    "type": "Column",
    "content": 1
  }
}
```

dtolnay:
Fixes #415 and fixes https://github.com/serde-rs/json/issues/67.

The attribute for untagged enums is:

```rust
#[serde(untagged)]
enum E {
    /* variants */
}
```

During serialization, the variant is serialized as though it were not part of an enum. Struct variants are serialized as structs, tuple variants are serialized as tuples, unit variants are serialized as unit, and newtype variants are serialized as the value.

During deserialization, we buffer the content of the Deserializer and try it against each variant in order, returning the first one that deserializes successfully from that content.

The attribute for internally tagged enums is:

```rust
#[serde(tag = "type")]
enum E {
    /* variants */
}
```

The enum must contain only unit, newtype and struct variants. Tuple variants are not supported. During serialization unit variants are serialized as structs with a single field: `{"type": "T"}`, struct variants are serialized with `"type": "T"` inserted as the first field, and newtype variants must contain either a struct or a map and are serialized with `"type": "T"` inserted as the first field.

During deserialization, we buffer the content of the Deserializer, pull out the tag, and deserialize the rest of the content against the correct variant.

The use case from #415 looks like:

```rust
#[serde(tag = "type")]
enum Motion {
    Column { content: i32 },
    Line { content: i32 },
}

#[serde(tag = "type")]
enum Command {
    Move { content: Motion },
    Delete { content: Motion },
    Input { content: String },
}
```

And the use case from https://github.com/serde-rs/json/issues/67 looks like:

```rust
#[serde(tag = "type")]
enum Message {
    #[serde(rename = "ok")]
    Ok { response: Response },
    #[serde(rename = "error")]
    Error { message: String },
}
```

cc @sfackler who asked about this in IRC and @alexcrichton who is removing this functionality from toml.

### DIFF
--- a/serde/src/de/content.rs
+++ b/serde/src/de/content.rs
@@ -537,11 +537,17 @@ impl<E> Deserializer for Content<E>
             Content::Newtype(v) => visitor.visit_newtype_struct(*v),
             Content::Seq(v) => {
                 let seq = v.into_iter();
-                visitor.visit_seq(de::value::SeqDeserializer::new(seq))
+                let mut seq_visitor = de::value::SeqDeserializer::new(seq);
+                let value = try!(visitor.visit_seq(&mut seq_visitor));
+                try!(seq_visitor.end());
+                Ok(value)
             },
             Content::Map(v) => {
                 let map = v.into_iter();
-                visitor.visit_map(de::value::MapDeserializer::new(map))
+                let mut map_visitor = de::value::MapDeserializer::new(map);
+                let value = try!(visitor.visit_map(&mut map_visitor));
+                try!(map_visitor.end());
+                Ok(value)
             },
             Content::Bytes(v) => visitor.visit_byte_buf(v),
         }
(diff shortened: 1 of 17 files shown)


alexcrichton:
Awesome, thanks @dtolnay!

sfackler:
Yay!

sfackler:
Out of curiosity, why is `Content` not public API?

dtolnay:
I would prefer to be able to iterate on the implementation for a few releases and only have to think about backward compatibility for the attributes rather than the entire mechanism. This is quite a large PR and I think we can reach consensus on the attributes and the semantics significantly sooner than we could reach consensus on the entire `Content` API. The focus of this PR should be whether these two attributes solve the right problem in the right way, and not on designing a large API that we feel comfortable supporting for the rest of the year.

Once this merges I will file a ticket to follow up on polishing and exposing some of the pieces.

sfackler:
Cool, sounds good.

serde/src/de/content.rs
```diff
@@ -0,0 +1,732 @@
+use core::fmt;
```
oli-obk: hide the entire module with `![doc(hidden)]` or at least write a doc comment stating that nothing in this module should be used outside of generated code. The PR comments contains all of that already, maybe just place them here?

serde/src/de/content.rs
```diff
@@ -0,0 +1,732 @@
...
+    Unit,
+    Newtype(Box<Content<E>>),
+    Seq(Vec<Content<E>>),
+    Map(Vec<(Content<E>, Content<E>)>),
+}
+
+impl<E> Deserialize for Content<E> {
+    fn deserialize<D: Deserializer>(deserializer: D) -> Result<Self, D::Error> {
+        // Untagged and internally tagged enums are only supported in
+        // self-describing formats.
```
oli-obk: We could find some way to support other formats through some rewinding API. This would also have uses for streaming.

dtolnay: I filed #742 to follow up on this.

serde/src/de/content.rs
dtolnay: I added a comment in 066c9a15cfc0a37111ae615ce57b44f7de3b6b1e and filed #741 to follow up on making it public.

PR #740, opened 2017-02-03
Remove the EnumSet impls

dtolnay:
In the most recent nightly this gives a strongly worded warning:

> **warning:** use of deprecated item: long since replaced


oli-obk:
yay!

PR #749, opened 2017-02-04
Add deserialization helpers for map and sequence types

Marwes:
The motivating case for this is that https://github.com/gluon-lang/languageserver-types has a type which contains a `HashMap<Url, SomeType>` but `Url` can't be deserialized directly with serde 0.9, it is instead provided by functions and wrappers in [a separate crate](https://github.com/servo/rust-url/tree/master/url_serde).

#553 mentions that this should perhaps be in another crate but since this PR simply reuses the normal sequence and map deserializers I'd argue it is a good idea to just provide them from `serde` directly.

 https://github.com/gluon-lang/languageserver-types/pull/7/commits/88e2565f6b1873b59dc1493c00d1c7166acbb766#diff-b4aea3e418ccdb71239b96952d9cddb6R335
#553

### DIFF
--- a/serde/src/de/impls.rs
+++ b/serde/src/de/impls.rs
@@ -711,87 +711,93 @@ tuple_impls! {
 macro_rules! map_impl {
     (
         $ty:ty,
-        $visitor_ty:ident < $($typaram:ident : $bound1:ident $(+ $bound2:ident)*),* >,
-        $visitor:ident,
-        $ctor:expr,
+        < $($typaram:ident : $bound1:ident $(+ $bound2:ident)*),* >,
         $with_capacity:expr
     ) => {
-        /// A visitor that produces a map.
-        pub struct $visitor_ty<$($typaram),*> {
-            marker: PhantomData<$ty>,
-        }
-
-        impl<$($typaram),*> $visitor_ty<$($typaram),*>
-            where $($typaram: $bound1 $(+ $bound2)*),*
-        {
-            /// Construct a `MapVisitor*<T>`.
-            pub fn new() -> Self {
-                $visitor_ty {
-                    marker: PhantomData,
-                }
-            }
-        }
-
-        impl<$($typaram),*> Visitor for $visitor_ty<$($typaram),*>
-            where $($typaram: $bound1 $(+ $bound2)*),*
-        {
-            type Value = $ty;
-
-            fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
-                formatter.write_str("a map")
-            }
-
-            #[inline]
-            fn visit_unit<E>(self) -> Result<$ty, E>
-                where E: Error,
-            {
-                Ok($ctor)
-            }
-
-            #[inline]
-            fn visit_map<Visitor>(self, mut $visitor: Visitor) -> Result<$ty, Visitor::Error>
-                where Visitor: MapVisitor,
-            {
-                let mut values = $with_capacity;
-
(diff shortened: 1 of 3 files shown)


oli-obk:
can you add some test cases which use the functions with something other than `|v| v` (and a type that doesn't impl `Deserialize`)

Marwes:
@oli-obk Done. Also noticed that I made a copy-paste mistake in https://github.com/serde-rs/serde/pull/749/commits/e250e21ea59cd3b317a155ef581b889ad5f4c57a which wasn't caught by the tests.

serde/src/de/impls.rs
```diff
@@ -396,125 +396,108 @@ impl<T> Deserialize for PhantomData<T> {
...
-        $visitor_ty:ident < $($typaram:ident : $bound1:ident $(+ $bound2:ident)*),* >,
-        $visitor:ident,
-        $ctor:expr,
-        $with_capacity:expr,
-        $insert:expr
+        $ty:ident< $($typaram:ident : $bound1:ident $(+ $bound2:ident)*),* >,
+        $with_capacity:expr
     ) => {
-        /// A visitor that produces a sequence.
-        pub struct $visitor_ty<$($typaram),*> {
```
dtolnay: [This is a breaking change.](https://github.com/search?l=Rust&q=%22impls%3A%3AVecVisitor%22&ref=searchresults&type=Code&utf8=%E2%9C%93)

serde/src/de/impls.rs
```diff
@@ -710,88 +693,91 @@ tuple_impls! {
...
     (
-        $ty:ty,
-        $visitor_ty:ident < $($typaram:ident : $bound1:ident $(+ $bound2:ident)*),* >,
-        $visitor:ident,
-        $ctor:expr,
+        $ty:ident < $($typaram:ident : $bound1:ident $(+ $bound2:ident)*),* >,
         $with_capacity:expr
     ) => {
-        /// A visitor that produces a map.
-        pub struct $visitor_ty<$($typaram),*> {
```
dtolnay: [This is a breaking change.](https://github.com/search?l=Rust&q=%22impls%3A%3ABTreeMapVisitor%22&ref=searchresults&type=Code&utf8=%E2%9C%93)

dtolnay:
@Marwes let me know what you think of the alternative plan and whether you would like to try implementing it.

Marwes:
> I think I would prefer to handle this using attributes:

That was actually my original idea but I rejected it because I thought it would incur to much complexity for a fairly niche use case. I think I have an idea how to implement it using an attribute but I am unsure how to debug the generated code (generated from test_annotations.rs). I remembered reading this post https://quodlibetor.github.io/posts/debugging-rusts-new-custom-derive-system/ but trying to compile the code generated by `cargo expand` gives a ton of name resolution errors. I assume there is a better way of debugging the generated code?

serde/src/de/impls.rs
Marwes: Didn't realize that these types were public :/.

dtolnay:
What I have been doing is handwrite what I would like the generated code to look like in a separate tiny crate that I can compile, then iterate on on the macro code until the output from cargo expand matches my handwritten one. As you saw, the generated code that gets printed is a lossy form of what is actually generated, because of hygiene and other issues, so compiling it typically doesn't work.

dtolnay:
I opened #753 to track the preferred approach.

PR #762, opened 2017-02-13
Use rustfmt

### LINKED ISSUE #438: Consider using rustfmt

dtolnay:
I think this is way better than last time we tried it.

Fixes #438.


oli-obk:
I dislike some changes, but I'd rather have rustfmt than no rustfmt. Let's do this. We can add rustfmt.toml later to finetune things if the final defaults are not to our liking.

dtolnay:
Sounds good. I don't have strong opinions on the formatting as long as a computer can do it for me - so I will leave the rustfmt.toml up to you.

Issue #758, opened 2017-02-12
haskell serialized formats could be better supported


oli-obk:
haskell serializes enums to json by encoding a `t` field for the variant name and a `c` field for the content. We can represent this with

```rust
#[derive(Serialize, Deserialize)]
#[serde(tag = "t")]
enum MyType {
    Variant { c: ActualContent },
    Variant2 { c: (A, B, C) },
}
```

This works fine, but is a little verbose. Not sure if this use case is too niche.

dtolnay:
This representation also came up in https://github.com/serde-rs/serde/issues/415#issue-162867209 so I don't think it is too niche. How about something like this?

```rust
#[derive(Serialize, Deserialize)]
#[serde(tag = "t", content = "c")]
enum MyType {
    Variant(ActualContent),
    Variant2(A, B, C),
}
```

oli-obk:
Sounds simple enough. We need to be aware of edge cases though. like what happens with empty variants (don't serialize the content field, deserialize without content field and with "unit" content field).

elliottslaughter:
I took a first pass on this in #771, but got stuck at the part where I need to serialize the internal contents of the `"c": ...` portion. If you have any advice, I'd appreciate suggestions in the other thread.

dtolnay:
Serialization has been implemented in #771. I would like an implementation of deserialization as well before publishing this.

dtolnay:
I released this in [0.9.8](https://github.com/serde-rs/serde/releases/tag/v0.9.8).

PR #771, opened 2017-02-18
WIP: Initial work on supporting #758

elliottslaughter:
I've been working on #758 but got stuck partway through the implementation, so I'm uploading what I've got for the moment. **Do not pull this, it's a work in progress and won't build anyway.**

See the comments in `serialize_internally_content_tagged_variant` for where I'm stuck. The basic issue is that I want to take a enum tuple variant like:

```
#[serde(tag = "t", content = "c")]
enum A {
  B(C, D)
}
```

And split out something like:

```
{"t": "B", "c": ["value_of_C", "value_of_D"]}
```

I know how to spit out the `"t": "B"` via `serialize_field`, and the interior of `["value_of_C", "value_of_D"]` via serialize_tuple_variant`. But I can't figure out a way to spit out a `"c":` to prefix the internal content with, because it seems that all the methods for serializing fields expect to do the field name and value in one shot, and expect the value to be a valid type. In this case, we're kind of faking the type of the interior field because of course no such type actually exists.

I'd appreciate any advice on how to approach this.

### DIFF
--- a/serde_derive/src/ser.rs
+++ b/serde_derive/src/ser.rs
@@ -230,30 +230,38 @@ fn serialize_variant(type_ident: &syn::Ident,
         }
     } else {
         // variant wasn't skipped
+        let field_values;
         let case = match variant.style {
             Style::Unit => {
+                field_values = vec![];
                 quote! {
                     #type_ident::#variant_ident
                 }
             }
             Style::Newtype => {
+                let field_name = Ident::new("__simple_value");
+                field_values = vec![field_name.clone()];
                 quote! {
-                    #type_ident::#variant_ident(ref __simple_value)
+                    #type_ident::#variant_ident(ref #field_name)
                 }
             }
             Style::Tuple => {
-                let field_names = (0..variant.fields.len())
-                    .map(|i| Ident::new(format!("__field{}", i)));
+                let field_names: Vec<_> = (0..variant.fields.len())
+                    .map(|i| Ident::new(format!("__field{}", i)))
+                    .collect();
+                field_values = field_names.clone();
                 quote! {
                     #type_ident::#variant_ident(#(ref #field_names),*)
                 }
             }
             Style::Struct => {
-                let fields = variant.fields
+                let field_names: Vec<_> = variant.fields
                     .iter()
-                    .map(|f| f.ident.clone().expect("struct variant has unnamed fields"));
+                    .map(|f| f.ident.clone().expect("struct variant has unnamed fields"))
+                    .collect();
+                field_values = field_names.clone();
                 quote! {
-                    #type_ident::#variant_ident { #(ref #fields),* }
+                    #type_ident::#variant_ident { #(ref #field_names),* }
                 }
             }
         };
@@ -274,6 +282,15 @@ fn serialize_variant(type_ident: &syn::Ident,
                                                     variant,
                                                     item_attrs,
                                                     tag)
+            }
+            attr::EnumTag::Adjacent(ref tag, ref content) => {
+                serialize_adjacently_tagged_variant(generics,
+                                                    ty,
+                                                    variant,
+                                                    item_attrs,
+                                                    &field_values,
+                                                    tag,
+                                                    content)
             }
             attr::EnumTag::None => serialize_untagged_variant(generics, ty, variant, item_attrs),
         };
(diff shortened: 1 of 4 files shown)


dtolnay:
I don't have time to review this until tomorrow but I would expect the output to look (conceptually) something like this. The key observation is that adjacently tagged enums are equivalent to a struct containing an untagged enum, so we can generate one impl for the struct and one impl for the untagged enum.

```rust
#[serde(tag = "t", content = "c")]
enum E {
    A,
    B(C, D),
}

impl Serialize for A {
        where S: Serializer
    {
        // The generics may be tricky to get right but it will be
        // very much like SerializeWith so use that as a model.
        struct Inner<'a>(&'a A);

        impl<'a> Serialize for Inner<'a> {
                where S: Serializer
            {
                match *self {
                    E::A => unreachable!(),
                    E::B(ref c, ref d) => {
                        serializer.serialize_tuple etc
                    }
                }
            }
        }

        match *self {
            E::A => {
                let mut state = serializer.serialize_struct("E", 1)?;
                state.serialize_field("t", "A")?;
                state.end()
            }
            E::B(..) => {
                let mut state = serializer.serialize_struct("E", 2)?;
                state.serialize_field("t", "B")?;
                state.serialize_field("c", &Inner(self))?;
                state.end()
            }
        }
    }
}
```

elliottslaughter:
I pushed another version. It's still not quite doing the right thing (you can't say `self.0.0` without a match clause), but I wanted to ask about some errors I'm seeing. To produce readable error messages, I copy-pasted the generated code into a separate test.

Here's what I'm seeing, and I can't make much heads or tails of any of it:

```
error[E0191]: the value of the associated type `Ok` (from the trait `serde::ser::SerializeTuple`) must be specified
  --> tests/content.rs:44:42
   |
44 |                                     try!(_serde::ser::SerializeTuple::serialize_field(&mut __tuple, &v1)) ;
   |                                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ missing associated type `Ok` value

error[E0191]: the value of the associated type `Error` (from the trait `serde::ser::SerializeTuple`) must be specified
  --> tests/content.rs:44:42
   |
44 |                                     try!(_serde::ser::SerializeTuple::serialize_field(&mut __tuple, &v1)) ;
   |                                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ missing associated type `Error` value

error: no associated item named `serialize_field` found for type `serde::ser::SerializeTuple` in the current scope
  --> tests/content.rs:44:42
   |
44 |                                     try!(_serde::ser::SerializeTuple::serialize_field(&mut __tuple, &v1)) ;
   |                                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error[E0038]: the trait `serde::ser::SerializeTuple` cannot be made into an object
  --> tests/content.rs:44:42
   |
44 |                                     try!(_serde::ser::SerializeTuple::serialize_field(&mut __tuple, &v1)) ;
   |                                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `serde::ser::SerializeTuple` cannot be made into an object
   |
   = note: method `serialize_element` has generic type parameters
```

For reference here's the copy-and-paste of the generated code, with some slight modifications by hand to get it to look more like what I want:

```
#[macro_use]
extern crate serde_derive;

extern crate serde;

struct S(i32);

struct T(i32);

// #[derive(Serialize, Debug)]
// #[serde(tag = "t", content = "c")]
enum E {
    A,
    B(S),
    C(S, T),
}

#[allow(non_upper_case_globals, unused_attributes, unused_qualifications)]
const _IMPL_SERIALIZE_FOR_E : () = {
    extern crate serde as _serde ;
    #[automatically_derived]
    impl _serde::Serialize for E {
        fn serialize<__S >(& self, _serializer : __S) -> _serde::export::Result<__S::Ok, __S::Error >
            where __S : _serde::Serializer {
            match * self {
                E::A => {
                    let mut __struct = try!(_serde::Serializer::serialize_struct(_serializer, "E", 1)) ;
                    try!(_serde::ser::SerializeStruct::serialize_field(& mut __struct, "t", "A")) ;
                    _serde::ser::SerializeStruct::end(__struct)
                }
                E::B(ref __simple_value) => _serde::ser::private::serialize_tagged_newtype(_serializer, "E", "B", "t", "B", __simple_value,),
                E::C(ref __field0, ref __field1) => {
                    struct Inner<'a >(& 'a E) ;
                    impl<'a>_serde::ser::Serialize for Inner<'a> {
                        fn serialize<S>(&self, __inner_serializer : S) -> Result<S::Ok, S::Error >
                            where S : _serde::ser::Serializer {
                            match self.0 {
                                &E::A => unreachable!(),
                                &E::B(ref v1) => unreachable!(),
                                &E::C(ref v1, ref v2) => {
                                    let mut __tuple = try!(_serde::ser::Serializer::serialize_tuple(__inner_serializer, 2usize)) ;
                                    try!(_serde::ser::SerializeTuple::serialize_field(&mut __tuple, &v1)) ;
                                    try!(_serde::ser::SerializeTuple::serialize_field(&mut __tuple, &v2)) ;
                                    _serde::ser::SerializeTuple::end(__tuple)
                                }
                            }
                        }
                    }
                    let mut __struct = try!(_serde::Serializer::serialize_struct(_serializer, "E", 2)) ;
                    try!(_serde::ser::SerializeStruct::serialize_field(& mut __struct, "t", "C")) ;
                    try!(_serde::ser::SerializeStruct::serialize_field(& mut __struct, "c", &Inner(&self))) ;
                    _serde::ser::SerializeStruct::end(__struct)
                }
            }
        }
    }
};
```

dtolnay:
I don't know what the message means but SerializeTuple does not have a serialize_field method.

serde_derive/src/ser.rs
```diff
@@ -275,6 +275,16 @@ fn serialize_variant(type_ident: &syn::Ident,
                                                     item_attrs,
                                                     tag)
             }
+            attr::EnumTag::InternalContent(ref tag, ref content) => {
+                serialize_internally_content_tagged_variant(type_ident.as_ref(),
```
dtolnay: "Internally content tagged" doesn't makes sense to me. The internal vs external distinction has to do with the relationship between the tag and the content.

```rust
// externally tagged - tag is outside the content
{
  TAG: CONTENT
}

// internally tagged - tag is inside the content
{
  "type": TAG,
  CONTENT
}
```

By extension I would call this style adjacently tagged - the tag is adjacent to the content.

```rust
{
  "t": TAG,
  "c": CONTENT
}
```

serde_derive/src/ser.rs
```diff
@@ -409,6 +419,99 @@ fn serialize_internally_tagged_variant(type_ident: &str,
...
+                .build();
+
+            let field_count = variant.fields.len();
+            let field_offset = variant.fields.iter().enumerate().map(
+                |(i, _)| syn::Lit::Int(i as u64, syn::IntTy::Unsuffixed));
+
+            let x = quote!({
+                // Define a tuple struct for the interior along with
+                // the necessary serialization logic.
+                struct Inner #wrapper_generics #where_clause (&'__a #ty);
```
dtolnay: I know this is WIP but you're going to want to declare just one of these wrappers for the entire enum like I showed in my code, not one per variant. The generics / PhantomData / etc are going to be impossible to get right otherwise.

dtolnay: Oh I guess one wrapper struct per non-unit variant could work as well. That way we avoid having to use unreachable!() for the unit variant cases. Something like:

```rust
// for newtype variants:
struct Inner<'a, IMPL_GENERICS> {
    data: &'a FIELD_TY,
    _marker: PhantomData<ENUM_TY<TY_GENERICS>>,
}

// for tuple variants:
struct Inner<'a, IMPL_GENERICS> {
    data: (&'a FIELD_TY1, &'a FIELD_TY2),
    _marker: PhantomData<ENUM_TY<TY_GENERICS>>,
}

// for struct variants:
struct Inner<'a, IMPL_GENERICS> {
    FIELD_NAME1: &'a FIELD_TY1,
    FIELD_NAME2: &'a FIELD_TY2,
    marker: PhantomData<ENUM_TY<TY_GENERICS>>,
}
```

dtolnay:
Thanks for working on this and sharing the code early! This is great work so far.

elliottslaughter:
Then is this documentation wrong?

https://docs.serde.rs/serde/trait.Serializer.html#tymethod.serialize_tuple

https://docs.serde.rs/serde/ser/trait.SerializeTuple.html

elliottslaughter:
Ok, it compiles the serializer at least. Is `test_suite/tests/test_macros.rs` the right place to add tests?

elliottslaughter:
Rebased. Some refactoring. Added tests. Serialization mostly works at this point.

Some snags:

 1. My new way of wrapping types doesn't work with nested newtype. I was hoping the nested serializer would just handle this, but apparently not. Suggestions welcome.

 2. I realize I'm not handling fields properly yet. I'm guessing the right approach is to delegate serialization of the inner struct to `serialize_struct_variant(StructVariant::Untagged ...)` but I'm still wrapping my head around how to drive it. I'd like to be sure this is the right way to go before I put in the effort.

elliottslaughter:
Oh, and the other obvious point: This doesn't work with non-tuple structs yet, mostly because I don't need those at all for Pandoc, but the approach should adapt reasonably well to that if desired.

dtolnay:
Thanks for the strong start on this! I merged your code along with https://github.com/serde-rs/serde/commit/a3fe03c3235dd9bdcaa935f6d5b42696ae05f96a which adds support for serializing adjacently tagged struct variants.

PR #778, opened 2017-02-20
Avoid some string allocations.

SimonSapin:


serde/src/macros.rs
```diff
@@ -188,3 +188,35 @@ macro_rules! forward_to_deserialize {
...
+            let remaining_len;
+            {
+                let mut remaining = &mut buffer[..];
+                write!(remaining, "{}", $value).unwrap();
+                remaining_len = remaining.len()
+            }
+            let written_len = buffer.len() - remaining_len;
+            let written = &buffer[..written_len];
+
+            // write! only provides std::fmt::Formatter to Display implementations,
```
dtolnay: I would strongly prefer to implement this safely if possible. Would it be possible to implement [`fmt::Write`](https://doc.rust-lang.org/std/fmt/trait.Write.html) and use that in the call to `write!`?

SimonSapin: Implementing `fmt::Write` how?

It’s possible to remove `unsafe` completely, at some performance cost:

```diff
diff --git a/serde/src/macros.rs b/serde/src/macros.rs
index 126c7ec..cc8775b 100644
--- a/serde/src/macros.rs
+++ b/serde/src/macros.rs
@@ -200,7 +200,7 @@ macro_rules! serialize_display_bounded_length {
     ($value: expr, $MAX_LEN: expr, $serializer: expr) => {
         {
             use std::io::Write;
-            let mut buffer: [u8; $MAX_LEN] = unsafe { ::std::mem::uninitialized() };
+            let mut buffer = [0; $MAX_LEN];
             let remaining_len;
             {
                 let mut remaining = &mut buffer[..];
@@ -209,13 +209,7 @@ macro_rules! serialize_display_bounded_length {
             }
             let written_len = buffer.len() - remaining_len;
             let written = &buffer[..written_len];
-
-            // write! only provides std::fmt::Formatter to Display implementations,
-            // which has methods write_str and write_char but no method to write arbitrary bytes.
-            // Therefore, `written` is well-formed in UTF-8.
-            let written_str = unsafe {
-                ::std::str::from_utf8_unchecked(written)
-            };
+            let written_str = ::std::str::from_utf8(written).unwrap();
             $serializer.serialize_str(written_str)
         }
     }

**Comment by dtolnay** [comment]
Benchmark:

```
test fast_safe   ... bench:         110 ns/iter (+/- 3)
test fast_unsafe ... bench:         103 ns/iter (+/- 3)
test string      ... bench:         189 ns/iter (+/- 6)
```

```rust
#![feature(test)]

extern crate test;
use test::Bencher;

use std::io::Write;
use std::mem;
use std::net::*;
use std::str;

macro_rules! addr {
    () => {
        Ipv4Addr::new(127, 0, 0, 1)
        //SocketAddrV6::new(Ipv6Addr::new(0, 0, 0, 0, 0, 0, 0, 1), 8080, 0, 0)
        //Ipv6Addr::new(0, 0, 0, 0, 0, 0xffff, 0xc00a, 0x2ff)
    }
}

#[bench]
fn string(b: &mut Bencher) {
    let v = addr!();
    b.iter(|| {
        test::black_box(v.to_string());
    });
}

#[bench]
fn fast_unsafe(b: &mut Bencher) {
    let v = addr!();
    b.iter(|| {
        /// "[1000:1002:1003:1004:1005:1006:1007:1008]:65000".len()
        const MAX_LEN: usize = 47;

        let mut buffer: [u8; MAX_LEN] = unsafe { mem::uninitialized() };
        let remaining_len;
        {
            let mut remaining = &mut buffer[..];
            write!(remaining, "{}", v).unwrap();
            remaining_len = remaining.len()
        }
        let written_len = buffer.len() - remaining_len;
        let written = &buffer[..written_len];

        // write! only provides std::fmt::Formatter to Display implementations,
        // which has methods write_str and write_char but no method to write arbitrary bytes.
        // Therefore, `written` is well-formed in UTF-8.
        let written_str = unsafe {
            str::from_utf8_unchecked(written)
        };
        test::black_box(written_str);
    });
}

#[bench]
fn fast_safe(b: &mut Bencher) {
    let v = addr!();
    b.iter(|| {
        /// "[1000:1002:1003:1004:1005:1006:1007:1008]:65000".len()
        const MAX_LEN: usize = 47;

        let mut buffer = [0; MAX_LEN];
        let remaining_len;
        {
            let mut remaining = &mut buffer[..];
            write!(remaining, "{}", v).unwrap();
            remaining_len = remaining.len()
        }
        let written_len = buffer.len() - remaining_len;
        let written = &buffer[..written_len];

        let written_str = str::from_utf8(written).unwrap();
        test::black_box(written_str);
    });
}
```

serde/src/macros.rs
dtolnay: You're right that doesn't help. Let's go with your way.

PR #780, opened 2017-02-20
Add support for `#[serde(default)]` on structs

Thomasdezeeuw:
This allows structs to use the default value for each field defined in
the struct’s `std::default::Default` implementation, rather then the
default value for the field’s type.

```
struct StructDefault {
    a: i32,
    b: String,
}

impl Default for StructDefault {
    fn default() -> StructDefault {
        StructDefault{
            a: 100,
            b: "default".to_string(),
        }
    }
}
```

The code above will now return `100` for field `a` and `”default”` for
`b`, rather then `0` and `””` respectively.

### DIFF
--- a/serde_codegen_internals/src/attr.rs
+++ b/serde_codegen_internals/src/attr.rs
@@ -90,6 +90,7 @@ impl Name {
 pub struct Item {
     name: Name,
     deny_unknown_fields: bool,
+    default: bool,
     ser_bound: Option<Vec<syn::WherePredicate>>,
     de_bound: Option<Vec<syn::WherePredicate>>,
     tag: EnumTag,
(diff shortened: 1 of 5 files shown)


Thomasdezeeuw:
This closes #780, forgot to add that to the commit message.

dtolnay:
Nicely done!

What would you expect it to do in these cases?

```rust
#[derive(Deserialize)]
#[serde(default)]
enum E {
    S { f: u8 },
}
```

```rust
#[derive(Deserialize)]
#[serde(default)]
struct N(u8);
```

```rust
#[derive(Deserialize)]
#[serde(default)]
struct T(u8, u8);
```

I think it would be reasonable to not support these. Please add a reasonable error message for this at compile time (see the various `cx.error` messages in serde_codegen_internals/src/attr.rs) and add compile-fail tests [here](https://github.com/serde-rs/serde/tree/master/test_suite/tests/compile-fail) to test those messages.

Thomasdezeeuw:
I added a check to allow the attribute to be only set on structs and I have added tests that fail to compile. However I'm not sure how to test them, I tried running `cargo test --features "unstable-testing"` however that fails all tests for me.

dtolnay:
For dumb reasons you need to build a separate crate of dependencies first:

```
cd $SERDE/test_suite/deps
cargo build
cd $SERDE/test_suite
cargo test --features unstable-testing
```

Thomasdezeeuw:
Thanks.

PR #787, opened 2017-02-23
Fix internally tagged enums by implementing ContentDeserializer::deserialize_enum

### LINKED ISSUE #775: Error when deserialising an enum within a tagged enum.
I've found a bug where a tagged enum can be serialised but not deserialised. Here's a testcase ([runnable gist here](https://gist.github.com/46bit/4f9a2daaa8a32cd27a5e7a39bf00154a)):

``` rust
#[derive(PartialEq, Clone, Debug, Serialize, Deserialize)]
#[serde(tag = "tag")]
pub enum A {
    A1 { field: B },
    A2,
}

#[derive(PartialEq, Clone, Debug, Serialize, Deserialize)]
pub enum B {
    B1,
    B2,
}

fn main() {
    // Construct an A containing a B.
    let a = A::A1 { field: B::B1 };
    println!("{:?}", a);

    // Successfully serialises to `{"tag":"A1","field":"B1"}`.
    let json = serde_json::to_string(&a).unwrap();
    println!("{}", json);

    // That correct, serde_json-generated JSON errors in the decode with
    // `Syntax(Message("invalid type: string \"B1\", expected enum B"), 0, 0)`
    let a_new: A = serde_json::from_str(json.as_str()).unwrap();
    println!("{:?}", a_new);
}
```

If you remove `#[serde(tag = "tag")]` (or, I think, change it to be untagged) then that format deserialises fine.

46bit:
These commits are to fix #775 for internally tagged enums. My testcases pass with this implementation but I'm not sure about the code correctness. Is there a particular test file I should add to?

Untagged enums need a similar fix.

### DIFF
--- a/serde/src/de/content.rs
+++ b/serde/src/de/content.rs
@@ -759,7 +759,7 @@ impl<E> de::VariantVisitor for VariantDeserializer<E> where E: de::Error {
 }
 
 struct SeqDeserializer<E> where E: de::Error {
-    iter: ::std::vec::IntoIter<Content>,
+    iter: <Vec<Content> as IntoIterator>::IntoIter,
     err: PhantomData<E>,
 }
 
--- a/serde/src/de/content.rs
+++ b/serde/src/de/content.rs
@@ -75,9 +75,7 @@ impl Content {
             Content::Char(c) => Unexpected::Char(c),
             Content::String(ref s) => Unexpected::Str(s),
             Content::Bytes(ref b) => Unexpected::Bytes(b),
-            // @TODO: Are these mappings correct?
-            Content::None => Unexpected::Option,
-            Content::Some(_) => Unexpected::Option,
+            Content::None | Content::Some(_) => Unexpected::Option,
             Content::Unit => Unexpected::Unit,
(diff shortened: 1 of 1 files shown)


dtolnay:
This PR is great. Thanks so much for working through this. I added two tiny commits to get the build to pass.

You or someone else can follow up in a separate PR for ContentRefDeserializer and tests before we close #775. For tests I would want a new function at the bottom of test_suite/tests/test_macros.rs dedicated to this case.

dtolnay:
More specifically, the test should have an internally tagged enum with one variant containing an externally (default) tagged enum with four variants, unit + newtype + tuple + struct. The test should use `assert_tokens` like the other tests in that file to make sure each case can be serialized and deserialized as expected.

46bit:
Cheers! I was wondering whether we should make `EnumDeserializer` et al type-parametric in whether they use `ContentDeserializer` or `ContentRefDeserializer` internally. This would save 200 lines of code duplication. What do you think?

dtolnay:
Sure, give it a try.

Issue #140, opened 2015-08-21
Add struct-level annotation for renaming fields


erickt:
Serde supports field level renaming with:

``` rust
struct Foo {
  #[serde(rename="baz")]
  bar: usize
}
```

But this can be inconvenient to use if you have a lot of fields you need to rename, such as going from rust's "snake-case" style into "camel-case". It would be nice to have struct level renaming support for this:

``` rust
#[serde(rename_camel_case)]
#[derive(Serialize, Deserialize)]
struct CamelCase {
  foo_bar_baz: usize
}

#[serde(rename_underscore_to_hypens)]
#[derive(Serialize, Deserialize)]
struct Hyphens {
  foo_bar_baz: usize
}

```

Would get serialized to `fooBarBaz`, and `foo-bar-baz`.

erickt:
This was implemented a while ago.

williamho:
does this feature still exist? I can't seem to find it in the docs.

scottpleb:
@williamho figuring out how to do this was surprisingly tricky, as the syntax apparently changed a few times, and it really doesn't seem to be documented anywhere. This is what worked for me:
```rust
#[derive(Debug, Deserialize)]
pub struct UploadResponse {
    fid: String,
    #[serde(rename="fileUrl")]
    file_url: String
}
```

dtolnay:
@scottpleb your specific use case is documented here: https://serde.rs/attr-rename.html

All the attributes are documented here: https://serde.rs/attributes.html

dtolnay:
I don't think this was ever implemented.

dtolnay:
@46bit mentioned possibly being interested in working on this.

Possibly the easiest implementation would be [right above this line](https://github.com/serde-rs/serde/blob/v0.9.8/serde_codegen_internals/src/ast.rs#L49), traverse the "body" and apply all of the renames in place. So if we have the item attrs asking for camelCase, we would change all of the names of the fields (except those that already have a rename) to be camelCase.

[...]

PR #788, opened 2017-02-23
`rename_all` container attribute to have all children fit a naming convention

46bit:
Hi! This implements the `rename_all` crystallised at the bottom of #140.

* On an enum it renames the variants.
* On a struct variant it renames the variant's fields.
* On a struct it renames the fields.

There are 5 rename rules defined, all of them to change the naming convention:

* `#[serde(rename_all = "PascalCase")]`
* `#[serde(rename_all = "camelCase")]`
* `#[serde(rename_all = "snake_case")]`
* `#[serde(rename_all = "SCREAMING_SNAKE_CASE")]`
* `#[serde(rename_all = "kebab-case")]`

The names very visibly convey the relevant convention. I used the [Inflector](https://github.com/whatisinternet/inflector) crate to apply the naming convention.


serde_codegen_internals/src/attr.rs
```diff
@@ -85,12 +87,57 @@ impl Name {
     }
 }
 
+#[derive(Debug, PartialEq)]
+pub enum RenameRule {
+    /// Don't apply a default rename rule.
+    None,
+    /// Rename direct children to "PascalCase" style, as typically used for enum variants.
+    PascalCase,
+    /// Rename direct children to "kebabCase" style.
```
dtolnay: this should say "camelCase"

serde_codegen_internals/src/attr.rs
```diff
@@ -297,21 +360,25 @@ impl Item {
...
+                    content: content,
+                }
             }
             (true, Some(_), Some(_)) => {
                 cx.error("untagged enum cannot have #[serde(tag = \"...\", content = \"...\")]");
                 EnumTag::External
             }
         };
-
+        // @DEBUG
```
dtolnay: Has this been debugged? It looks all right to me.

46bit: Derp. Sorry. This is old.

dtolnay:
This is very nicely done. The only major thing is as we discussed in IRC - let's figure out a way to avoid the heavyweight inflector dependency, whether that means grabbing the tiny amount of their code that we actually need, re-implementing some of it (each of these conversions is probably something like 10 lines), or making a PR to inflector to put some of the heavier functionality like pluralize and singularize behind a cfg.

46bit:
(Leaving here for a reminder.)

It might be best to implement very simple conversions, since we know from Rust conventions what the input formatting should be. `RenameRule::apply_to_variant` and `RenameRule::apply_to_field`.

whatisinternet:
@46bit I've added a little spike and PR for feature flagging Inflector https://github.com/whatisinternet/inflector/pull/44. It's not 100% final, but I think it's something I can get on board with. Let me know if there is anything more I can do to help out. I'd appreciate a code review if/when you have time.

46bit:
@whatisinternet Sounds great! I'd definitely prefer using your library to my own conversions and this will enable that :-)

I just wonder whether `without_full` is the best name long-term. Breaking each module into a separate feature is a bit excessive but maybe something like `lightweight` better conveys the spirit of this?

whatisinternet:
Agreed (names are hard). I've changed it to be `lightweight` which does feel better. WRT breaking each module out into a feature: It does feel a bit excessive at the moment. Let me know how the testing goes with the current feature gate, and we can reevaluate if needed.

alexbool:
Perhaps @dtolnay should make a release with this magnificent feature!

dtolnay:
I published this magnificent feature in Serde 0.9.10. Thanks for the reminder.

PR #789, opened 2017-02-26
Introduce Serializer::collect_str (fixes #786)

### LINKED ISSUE #786: Allow serialising format_args! directly
I've seen a fair number of `Serialize` implementations that involve calling `to_string` and I wonder if this extra allocation could be removed somehow. Take [this example](https://github.com/chronotope/chrono/blob/d036c0737059fe002463c10756553393bcbcc440/src/datetime.rs#L524) from chrono:

```
serializer.serialize_str(&format!("{:?}", self))
```

I wonder if we could perhaps modify this to:

```
serializer.serialize_str_format(format_args!("{:?}", self))
```

So that the serializer could call `write` itself directly to a buffer. For formats that don't require escaping characters, this could involve writing the data directly to the output stream, and for formats that do, this might involve writing to a temporary buffer (only allocated once) and then copying the data out with the escaped characters in it. Perhaps it could also default to calling `format` and then `serialize_str` if implementors choose to not specialise it.

Not sure how much of a performance gain this would be, but it seems worth investigating imo.

nox:


clarcharr:
For `no_std`, wouldn't it be best to have a reasonably sized buffer to cover cases where strings are small? Could we potentially also cover this in the `std` case?

dtolnay:
We can optimize this with a statically allocated buffer in a separate commit. This API looks reasonable to me. @nox could you also implement the serde_json side of this?

clarcharr:
My main concern is that the `no_std` case will just return a runtime error instead of doing anything reasonable. I'd rather return at least an error that points to this method than something cryptic.

nox:
@dtolnay Done.

nox:
@clarcharr I will change the error message tomorrow. Any suggestions?

clarcharr:
@nox Perhaps "collect_str is currently unimplemented for no_std builds" ?

nox:
I changed the error message and used `unwrap` for the result returned by `write!`, given it's documented that `Display::fmt` should never return spurious errors.

serde/src/ser/mod.rs
```diff
@@ -616,6 +620,29 @@ pub trait Serializer: Sized {
...
+    /// Collect a `Display` value as a string.
+    ///
+    /// The default implementation serializes the given value as a string with
+    /// `ToString::to_string`.
+    #[cfg(any(feature = "std", feature = "collections"))]
+    fn collect_str<T>(self, value: &T) -> Result<Self::Ok, Self::Error>
+        where T: Display,
+    {
+        let mut string = String::new();
+        write!(string, "{}", value).unwrap();
```
dtolnay: Any advantages of this over `value.to_string()`? I guess in case there is a specialized ToString impl to make this default implementation behave consistently with a serializer that overrides collect_str with a more efficient Display-based implementation? I think people deserve what they get if they implement ToString and Display in a way that is not consistent with each other.

ToString may be faster https://github.com/rust-lang/rust/issues/18404 (not that I expect this method to be used with T=str).

serde/src/ser/mod.rs
```diff
@@ -616,6 +620,29 @@ pub trait Serializer: Sized {
         }
         serializer.end()
     }
+
+    /// Collect a `Display` value as a string.
+    ///
+    /// The default implementation serializes the given value as a string with
+    /// `ToString::to_string`.
+    #[cfg(any(feature = "std", feature = "collections"))]
+    fn collect_str<T>(self, value: &T) -> Result<Self::Ok, Self::Error>
```
dtolnay: T can be ?Sized.

clarcharr:
Would it be reasonable to add a token to `serde_test::Token` for this?

dtolnay:
I don't think so, semantically it is no different from a string.

serde/src/ser/mod.rs
nox: I would like to use `ToString`, but unfortunately, I have to [hate](https://github.com/rust-lang/rust/issues/40103) it for now.

serde/src/ser/mod.rs
nox: Good catch.

shepmaster:
Very cool idea — I was just looking for something like this! Thanks!

PR #792, opened 2017-02-27
Fix documentation of serialize_tuple and SerializeTuple

elliottslaughter:
`SerializeTuple` supports `serialize_element`, not `serialize_field`.


dtolnay:
Thanks! I filed #793 to track progress on ignoring fewer of these doc examples, which would have caught this.

PR #794, opened 2017-02-28
Documentation for serde_test.

clarcharr:
Still WIP; don't merge yet.

My take at documenting `serde_test`. I was very confused how the tokens worked when I first got into it and think that it could use some documentation.

### DIFF
--- a/serde_test/src/assert.rs
+++ b/serde_test/src/assert.rs
@@ -7,6 +7,7 @@ use token::Token;
 
 use std::fmt::Debug;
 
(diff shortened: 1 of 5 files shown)


clarcharr:
I've finished this and want to add that I think that `EnumStart` should be removed. It doesn't seem to serve any purpose that isn't covered by the other tokens.

dtolnay:
Thanks! I am keeping #413 open until we get a chance to add usage examples to this documentation as well.

Issue #800, opened 2017-03-03
Implement Serialize/Deserialize for CStr


jonhoo:
`serde` currently ships with an `impl` of `Serialize` and `Deserialize` for [`str`](https://github.com/serde-rs/serde/blob/master/serde/src/ser/impls.rs#L80) and [`String`](https://github.com/serde-rs/serde/blob/master/serde/src/ser/impls.rs#L90), but not for the null-terminated C-like string types [`ffi::CStr`](https://doc.rust-lang.org/std/ffi/struct.CStr.html) and [`ffi::CString`](https://doc.rust-lang.org/std/ffi/struct.CString.html). Under the hood, these string implementations are very similar, and providing serialization implementations for the latter two shouldn't be too hard. Are there any particular reasons why `impl`s for these types are not provided?

dtolnay:
No reason. Would you be able to send a PR adding these two impls?

jonhoo:
@dtolnay I'll give it a shot. Should this be placed under its own flag, given that it's in `ffi`?

dtolnay:
No, both types are in std so they should be behind the `std` feature.

One open question is how you intend to handle non-UTF8 strings.

jonhoo:
I was thinking of serializing through `[u8]` rather than as strings, which should work fine, right?

dtolnay:
Sure, that works.

jonhoo:
@dtolnay patch in #801.

PR #801, opened 2017-03-03
impls for null-terminated FFI string types

jonhoo:
From #800: `serde` currently ships with an `impl` of `Serialize` and `Deserialize` for [`str`](https://github.com/serde-rs/serde/blob/master/serde/src/ser/impls.rs#L80) and [`String`](https://github.com/serde-rs/serde/blob/master/serde/src/ser/impls.rs#L90), but not for the null-terminated C-like string types [`ffi::CStr`](https://doc.rust-lang.org/std/ffi/struct.CStr.html) and [`ffi::CString`](https://doc.rust-lang.org/std/ffi/struct.CString.html). Under the hood, these string implementations are very similar, and providing serialization implementations for the latter two shouldn't be too hard.

This patch adds serialization and deserialization of those types through null-terminated `[u8]`s. I'm guessing I should also implement tests for this, but didn't find where those should go. Feedback welcome!

### DIFF
--- a/serde/src/de/impls.rs
+++ b/serde/src/de/impls.rs
@@ -25,6 +25,8 @@ use std::net;
 #[cfg(feature = "std")]
 use std::path;
 use core::str;
+#[cfg(feature = "std")]
+use std::ffi::{CStr, CString};
 
 #[cfg(feature = "std")]
 use std::rc::Rc;
@@ -290,6 +292,33 @@ impl Deserialize for String {
         where D: Deserializer
     {
         deserializer.deserialize_string(StringVisitor)
+    }
+}
+
+///////////////////////////////////////////////////////////////////////////////
+
+#[cfg(feature = "std")]
+impl Deserialize for Box<CStr> {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+        where D: Deserializer
+    {
+        use std::mem;
+        let s = try!(CString::deserialize(deserializer));
+        let slice = s.into_bytes_with_nul().into_boxed_slice();
+        Ok(unsafe { mem::transmute::<Box<[u8]>, Box<CStr>>(slice) })
+    }
+}
+
+
+#[cfg(feature = "std")]
+impl Deserialize for CString {
+    fn deserialize<D>(deserializer: D) -> Result<CString, D::Error>
+        where D: Deserializer
+    {
+        let mut v: Vec<u8> = try!(Deserialize::deserialize(deserializer));
+        v.pop(); // cut trailing NULL, because CString::new adds it
+        CString::new(v)
+            .map_err(|e| Error::custom(format!("unexpected NULL at byte {}", e.nul_position())))
     }
 }
 
--- a/serde/src/de/impls.rs
+++ b/serde/src/de/impls.rs
@@ -55,6 +55,7 @@ use de::{Deserialize, Deserializer, EnumVisitor, Error, MapVisitor, SeqVisitor, 
          VariantVisitor, Visitor};
 use de::from_primitive::FromPrimitive;
 
+#[cfg(feature = "std")]
 use bytes::ByteBuf;
 
 ///////////////////////////////////////////////////////////////////////////////
--- a/serde/src/de/impls.rs
+++ b/serde/src/de/impls.rs
@@ -298,19 +298,6 @@ impl Deserialize for String {
 ///////////////////////////////////////////////////////////////////////////////
 
 #[cfg(feature = "std")]
-impl Deserialize for Box<CStr> {
-    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
-        where D: Deserializer
-    {
-        use std::mem;
-        let s = try!(CString::deserialize(deserializer));
-        let slice = s.into_bytes_with_nul().into_boxed_slice();
-        Ok(unsafe { mem::transmute::<Box<[u8]>, Box<CStr>>(slice) })
-    }
-}
-
-
-#[cfg(feature = "std")]
 impl Deserialize for CString {
     fn deserialize<D>(deserializer: D) -> Result<CString, D::Error>
         where D: Deserializer
--- a/serde/src/de/impls.rs
+++ b/serde/src/de/impls.rs
@@ -26,7 +26,7 @@ use std::net;
 use std::path;
 use core::str;
 #[cfg(feature = "std")]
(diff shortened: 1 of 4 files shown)


serde/src/de/impls.rs
```diff
@@ -295,6 +297,33 @@ impl Deserialize for String {
 
 ///////////////////////////////////////////////////////////////////////////////
 
+#[cfg(feature = "std")]
+impl Deserialize for Box<CStr> {
```
dtolnay: Please remove this impl unless you can implement it safely. The internal representation of CStr is not stable and there are plans to change it in the future.

serde/src/de/impls.rs
```diff
@@ -295,6 +297,33 @@ impl Deserialize for String {
...
+    }
+}
+
+
+#[cfg(feature = "std")]
+impl Deserialize for CString {
+    fn deserialize<D>(deserializer: D) -> Result<CString, D::Error>
+        where D: Deserializer
+    {
+        let mut v: Vec<u8> = try!(Deserialize::deserialize(deserializer));
```
dtolnay: An empty vec would be an error right?

serde/src/ser/impls.rs
```diff
@@ -98,6 +100,28 @@ impl Serialize for String {
 
 ///////////////////////////////////////////////////////////////////////////////
 
+#[cfg(feature = "std")]
+impl Serialize for CStr {
+    #[inline]
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+        where S: Serializer
+    {
+        (self.to_bytes_with_nul()).serialize(serializer)
```
dtolnay: This does not require parentheses around self.to_bytes_with_nul().

dtolnay:
What are the reasons for including the null terminator in the serialized representation?

serde/src/ser/impls.rs
```diff
@@ -98,6 +100,28 @@ impl Serialize for String {
...
+    }
+}
+
+#[cfg(feature = "std")]
+impl Serialize for CString {
+    #[inline]
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+        where S: Serializer
+    {
+        (self.to_bytes_with_nul()).serialize(serializer)
```
dtolnay: Using [`serialize_bytes`](https://docs.serde.rs/serde/ser/trait.Serializer.html#tymethod.serialize_bytes) would allow most serializers to serialize this more efficiently than other types of slices.

serde/src/de/impls.rs
jonhoo: Ah, that's unfortunate. This is basically a direct translation of `String::into_boxed_str`, since the equivalent `CString::into_boxed_cstr` method does not exist. I think it's more common (and useful) to have the `impl` for `CStr`... Also allows working with, say `Arc<CStr>` (or [arccstr](https://github.com/jonhoo/arccstr)). Should I then also remove the `Serialize` `impl`?

jonhoo:
Good point about including the NULL bytes in the serialization. I'll remove that.

dtolnay:
Tests for serialization should go [here](https://github.com/serde-rs/serde/blob/d70636f4d4d5e324df8b6286a9aef99655dcee1f/test_suite/tests/test_ser.rs#L391), tests for deserialization should go [here](https://github.com/serde-rs/serde/blob/d70636f4d4d5e324df8b6286a9aef99655dcee1f/test_suite/tests/test_de.rs#L880), and tests for various deserialization error cases should go [here](https://github.com/serde-rs/serde/blob/d70636f4d4d5e324df8b6286a9aef99655dcee1f/test_suite/tests/test_de.rs#L997).

serde/src/de/impls.rs
dtolnay: No, keep the Serialize impl. It's fine to support serializing a type without supporting deserializing it.

jonhoo:
Ah, so in writing the tests I realized that using `serialize_bytes` means we also need a way of turning the decoded `Box<[u8]>` into ` Vec<u8>`. Unfortunately, there is no inverse function of `Vec::into_boxed_slice`, so there isn't an obvious way of doing that without doing an extra copy :/

jonhoo:
Unless there's a way to `deserialize_bytes` into ` Vec<u8>` instead of a `Box<[u8]>`.

dtolnay:
[`ByteBuf`](https://docs.serde.rs/serde/bytes/struct.ByteBuf.html) implements Deserialize and Into\<Vec\<u8>>. Would that help? Something like:

```rust
let v: Vec<u8> = try!(ByteBuf::deserialize(deserializer)).into();
```

jonhoo:
That works, though I'd have to import `ByteBuf` into `de/impl.rs`?

dtolnay:
Yep that's fine.

jonhoo:
Hmmm:
```rust
error[E0432]: unresolved import `serde::bytes::ByteBuf`
   --> /home/jon/dev/others/serde/serde/src/de/impls.rs:305:13
    |
305 |         use serde::bytes::ByteBuf;
    |             ^^^^^^^^^^^^^^^^^^^^^ Maybe a missing `extern crate serde;`?
```

jonhoo:
I guess `super::super::bytes::ByteBuf` works, but feels a bit hacky...

jonhoo:
So, it looks like we can also land deserialization for `CStr` once [`into_boxed_c_str`](https://doc.rust-lang.org/nightly/std/ffi/struct.CString.html#method.into_boxed_c_str) lands (it is currently nightly-only). May be worth doing that in a separate PR though.

jonhoo:
@dtolnay hmm -- why is `bytes::ByteBuf` only available on beta?

dtolnay:
You're missing #[cfg(feature = "std")] on the import.

jonhoo:
Okay, I think this is ready to merge @dtolnay. `Deserialize` for `CStr` can come later when `into_boxed_c_str` stabilizes.

clarcharr:
@jonhoo @dtolnay is there something open on serde's end to implement that once the method is stabilised? there's now a tracking issue for that feature at rust-lang/rust#40380

jonhoo:
@clarcharr Not as far as I know, but I'm watching the issue and the PR, so will submit a follow-up PR here once it lands.

dtolnay:
Thanks @clarcharr and @jonhoo, I filed #810 to track this on our end.

Issue #797, opened 2017-03-02
Consider re-exporting derives from serde


dtolnay:
```rust
// serde/src/lib.rs
#[allow(unused_imports)]
#[macro_use]
extern crate serde_derive;
pub use serde_derive::*;
```

Apparently this works in Rust 1.15...

```rust
#[macro_use] // now gives you derive(Serialize, Deserialize)
extern crate serde;
```

dtolnay:
Having to re-enabled std is obnoxious for libraries that want to use serde without derive, like linked-hash-map.

```toml
serde = { version = "0.9", default-features = false, features = ["std"] }
```

PR #803, opened 2017-03-05
WIP: Use prom-attire for parsing attributes

Nemo157:
This was mostly done as a test of `prom-attire`'s feature set, if you don't want to pull in another dependency for this purpose feel free to close this PR.

There's three main things that need doing before this would not be a WIP, publishing a release of `prom-attire` (mostly just need to write examples and documentation for that), checking that all the error cases work correctly and give nice errors (mostly around parsing things like bounds from a string since that's handled in `prom-attire` now), and there's a whole lot of unreachable patterns generated in `prom-attire` that I need to fixup.

There's also one more pattern I want to add support for in `prom-attire`, `#[serde(default)]` and `#[serde(default = "thing")]`. Currently it's done via 2 fields of different types that share the same attribute name and because of the parsing order in `prom-attire` this works (and might be the source of unreachable patterns). I have to think how to extend the valueless syntax from just supporting `bool` to supporting other types.

### DIFF
--- a/serde_codegen_internals/Cargo.toml
+++ b/serde_codegen_internals/Cargo.toml
@@ -13,6 +13,7 @@ include = ["Cargo.toml", "src/**/*.rs", "README.md", "LICENSE-APACHE", "LICENSE-
 
 [dependencies]
 syn = { version = "0.11", default-features = false, features = ["parsing"] }
+prom-attire = { git = "https://github.com/Nemo157/prom-attire-rs" }
 
 [badges]
 travis-ci = { repository = "serde-rs/serde" }
--- a/serde_codegen_internals/src/attr.rs
+++ b/serde_codegen_internals/src/attr.rs
@@ -1,8 +1,6 @@
+use std::str::FromStr;
 use Ctxt;
 use syn;
-use syn::MetaItem::{List, NameValue, Word};
-use syn::NestedMetaItem::{Literal, MetaItem};
-use std::str::FromStr;
 
 // This module handles parsing of `#[serde(...)]` attributes. The entrypoints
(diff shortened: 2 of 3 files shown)


Nemo157:
Oh, one other feature for `prom-attire` that would make this a little simpler; allow specifying a function for parsing so the new-types implementing `FromStr` aren't needed (also, maybe remove the requirement for `std::error::Error` from the errors of these).

nox:
Is that similar to `procedural-masquerade` and `proc-macro-hack`? AFAIK this trick requires Rust 1.1(;

Nemo157:
This does require 1.15, but there's no tricks involved, `prom-attire` is simply a Macros 1.1 derive macro for generating attribute parsers. As I understand it from the 0.9.0 release notes `serde_derive`/`serde_codegen_internals` have migrated to being Rust 1.15+ only now.

nox:
> As I understand it from the `serde` 0.9.0 release notes `serde_derive`/`serde_codegen_internals` have migrated to being Rust 1.15+ only now.

Yes, but `serde` still needs to work with older Rust versions.

nox:
https://github.com/serde-rs/serde/blob/master/.travis.yml#L7

nox:
See also https://github.com/serde-rs/serde/issues/797.

Nemo157:
This only touches `serde_codegen_internals`. I assume any sort of re-export of `serde_derive` in `serde` would have to be Rust 1.15+ only as `proc_macro` would not be available in earlier versions.

dtolnay:
Thanks! I am happy with our current approach but this seems really useful for smaller proc macro crates that want to handle basic string and boolean attributes.

PR #808, opened 2017-03-07
Implement deserialize_enum for U32Deserializer

dtolnay:
In serde_derive the visitors we generate for unit variants are able to deserialize from u32, so this should be supported for u32's value deserializer as well.

### DIFF
--- a/serde/src/de/value.rs
+++ b/serde/src/de/value.rs
@@ -183,12 +183,71 @@ primitive_deserializer!(i64, I64Deserializer, visit_i64);
 primitive_deserializer!(isize, IsizeDeserializer, visit_i64 as i64);
 primitive_deserializer!(u8, U8Deserializer, visit_u8);
 primitive_deserializer!(u16, U16Deserializer, visit_u16);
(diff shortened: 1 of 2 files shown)


oli-obk:
Huh? What's special about u32? Why not all other ints, too?

dtolnay:
It is special because u32 is the only integer type that automatically derived enum variants can be deserialized from, [here](https://github.com/serde-rs/serde/blob/v0.9.11/serde_derive/src/de.rs#L1056). It is special in bincode and other deserialize-variants-by-index formats too, [here](https://github.com/TyOverby/bincode/blob/0ed8863bcb8080911e64aab7e956f64c8ebaae36/src/serde/reader.rs#L224). I would prefer to be consistent by saying that Serde's data model represents variant indices as u32 and that any conversions are the responsibility of the user, for example `(my_u16 as u32).into_deserializer()`.

oli-obk:
OK. I agree with the consistency argument. I'm not sure about the u32 as the discriminant type. I feel we've had this argument before though

dtolnay:
I opened #809 to follow up.

Issue #737, opened 2017-02-01
Generated code is wrong if Ok or Err are redefined


dtolnay:
```rust
#[macro_use]
extern crate serde_derive;

struct Ok;
struct Err;

#[derive(Deserialize)]
struct S {
    x: ()
}
```

```rust
error: expected function, found `Ok`
 --> src/main.rs:7:10
  |
7 | #[derive(Deserialize)]
  |          ^^^^^^^^^^^
  |
note: defined here
 --> src/main.rs:4:1
  |
4 | struct Ok;
  | ^^^^^^^^^^

error: expected function, found `Err`
 --> src/main.rs:7:10
  |
7 | #[derive(Deserialize)]
  |          ^^^^^^^^^^^
  |
note: defined here
 --> src/main.rs:5:1
  |
5 | struct Err;
  | ^^^^^^^^^^^
```

dtolnay:
This happens if someone does `use hyper::*` which includes a reexport of `hyper::status::StatusCode::Ok` :crying_cat_face:

PR #817, opened 2017-03-14
Macro attributes to specify From and Into trait types for structs and enums

jbaublitz:
I have implemented `#[serde(types(from="type", into="type", from_as="type", into_as="type))]` to allow more complicated serialization and deserialization. The `from` type specifies the type from which the result will be converted after being passed through `Derializer`. The `into` type specifies the type that the value will be converted to before being passed through `Serializer`. `from_as` and `into_as` provide an additional safe cast as necessary between the `Serializer`/`Deserializer` and the `From`/`Into` traits.

### DIFF
--- a/serde_codegen_internals/src/attr.rs
+++ b/serde_codegen_internals/src/attr.rs
@@ -100,8 +100,6 @@ pub struct Item {
     tag: EnumTag,
     from_type: Option<syn::Ty>,
     into_type: Option<syn::Ty>,
-    from_as_type: Option<syn::Ty>,
-    into_as_type: Option<syn::Ty>,
 }
 
 /// Styles of representing an enum.
@@ -151,8 +149,7 @@ impl Item {
         let mut content = Attr::none(cx, "content");
         let mut from_type = Attr::none(cx, "from");
         let mut into_type = Attr::none(cx, "into");
-        let mut from_as_type = Attr::none(cx, "from_as");
-        let mut into_as_type = Attr::none(cx, "into_as");
+        let mut as_type = Attr::none(cx, "into_as");
 
         for meta_items in item.attrs.iter().filter_map(get_serde_meta_items) {
             for meta_item in meta_items {
@@ -279,13 +276,19 @@ impl Item {
                     }
 
                     // Parse `#[serde(types(from = "type", into = "type", as = "type"))]
-                    MetaItem(List(ref name, ref meta_items)) if name == "types" => {
-                        if let Ok((from_ty, into_ty, from_as_ty, into_as_ty)) =
-                                get_from_into_as(cx, name.as_ref(), meta_items) {
-                            from_type.set_opt(from_ty);
-                            into_type.set_opt(into_ty);
-                            from_as_type.set_opt(from_as_ty);
-                            into_as_type.set_opt(into_as_ty);
+                    MetaItem(NameValue(ref name, ref lit)) if name == "from" => {
+                        if let Ok(from_ty) = get_type(cx, name.as_ref(), lit) {
+                            from_type.set_opt(Some(from_ty));
+                        }
+                    }
+                    MetaItem(NameValue(ref name, ref lit)) if name == "into" => {
+                        if let Ok(into_ty) = get_type(cx, name.as_ref(), lit) {
+                            into_type.set_opt(Some(into_ty));
+                        }
+                    }
+                    MetaItem(NameValue(ref name, ref lit)) if name == "as" => {
+                        if let Ok(as_ty) = get_type(cx, name.as_ref(), lit) {
+                            as_type.set_opt(Some(as_ty));
                         }
                     }
 
@@ -360,8 +363,6 @@ impl Item {
             tag: tag,
             from_type: from_type.get(),
             into_type: into_type.get(),
-            from_as_type: from_as_type.get(),
-            into_as_type: into_as_type.get(),
         }
     }
 
@@ -399,14 +400,6 @@ impl Item {
 
     pub fn into_type(&self) -> Option<&syn::Ty> {
         self.into_type.as_ref()
-    }
-
-    pub fn from_as_type(&self) -> Option<&syn::Ty> {
-        self.from_as_type.as_ref()
-    }
-
-    pub fn into_as_type(&self) -> Option<&syn::Ty> {
-        self.into_as_type.as_ref()
     }
 }
 
@@ -785,62 +778,16 @@ fn get_ser_and_de<T, F>(cx: &Ctxt,
     Ok((ser_item.get(), de_item.get()))
 }
 
-type TypeConversions<T> = (Option<T>, Option<T>, Option<T>, Option<T>);
-
-fn get_from_into_as(cx: &Ctxt,
-                    attr_name: &str,
-                    items: &[syn::NestedMetaItem])
-                    -> Result<TypeConversions<syn::Ty>, ()> {
-    let mut from_type = None;
-    let mut into_type = None;
-    let mut from_as_type = None;
-    let mut into_as_type = None;
-
-    for item in items {
-        match *item {
-            MetaItem(NameValue(ref ident, ref lit)) if ident == "from" => {
-                if let Ok(ty) = parse_lit_into_ty(cx, attr_name, ident.as_ref(), &lit) {
-                    from_type = Some(ty);
-                } else {
-                    cx.error(format!("error parsing type for nested attribute {}", ident));
-                    return Err(());
-                }
-            }
-            MetaItem(NameValue(ref ident, ref lit)) if ident == "into" => {
-                if let Ok(ty) = parse_lit_into_ty(cx, attr_name, ident.as_ref(), &lit) {
-                    into_type = Some(ty);
-                } else {
-                    cx.error(format!("error parsing type for nested attribute {}", ident));
-                    return Err(());
-                }
-            }
-            MetaItem(NameValue(ref ident, ref lit)) if ident == "from_as" => {
-                if let Ok(ty) = parse_lit_into_ty(cx, attr_name, ident.as_ref(), &lit) {
-                    from_as_type = Some(ty);
-                } else {
-                    cx.error(format!("error parsing type for nested attribute {}", ident));
-                    return Err(());
-                }
-            }
-            MetaItem(NameValue(ref ident, ref lit)) if ident == "into_as" => {
-                if let Ok(ty) = parse_lit_into_ty(cx, attr_name, ident.as_ref(), &lit) {
-                    into_as_type = Some(ty);
-                } else {
-                    cx.error(format!("error parsing type for nested attribute {}", ident));
-                    return Err(());
-                }
-            }
-            MetaItem(NameValue(ref ident, _)) => {
-                cx.error(format!("error parsing nested attribute {}", ident));
-                return Err(());
-            }
-            _ => {
-                cx.error(format!("invalid input for type conversions"));
-                return Err(());
-            }
-        }
+fn get_type(cx: &Ctxt,
+            attr_name: &str,
+            lit: &syn::Lit)
+            -> Result<syn::Ty, ()> {
+    if let Ok(ty) = parse_lit_into_ty(cx, attr_name, &lit) {
+        Ok(ty)
+    } else {
+        cx.error(format!("error parsing type for attribute {}", attr_name));
+        return Err(());
     }
-    Ok((from_type, into_type, from_as_type, into_as_type))
 }
 
 fn get_renames(cx: &Ctxt, items: &[syn::NestedMetaItem]) -> Result<SerAndDe<String>, ()> {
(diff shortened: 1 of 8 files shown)


serde_codegen_internals/src/attr.rs
```diff
@@ -797,3 +894,20 @@ fn parse_lit_into_where(cx: &Ctxt,
...
 }
+
+fn parse_lit_into_ty(cx: &Ctxt,
+                     attr_name: &str,
+                     meta_item_name: &str,
+                     lit: &syn::Lit)
+                     -> Result<syn::Ty, ()> {
+    let string = try!(get_string_from_lit(cx, attr_name, meta_item_name, lit));
+    if string.is_empty() {
+        return Ok(syn::Ty::Tup(Vec::new()));
```
dtolnay: Please remove this special case. `from = ""` does not make sense and should be an error just like any other type that does not parse.

serde_codegen_internals/src/attr.rs
```diff
@@ -797,3 +894,20 @@ fn parse_lit_into_where(cx: &Ctxt,
...
+    }
+
+    if let Ok(ty) = syn::parse_type(&string) {
```
dtolnay: This should do what `parse_lit_into_path` does so that the error message ends up visible to the user.

test_suite/tests/test_annotations.rs
```diff
@@ -977,3 +977,83 @@ fn test_invalid_length_enum() {
         Error::Message("invalid length 1, expected tuple of 2 elements".to_owned()),
     );
 }
+
+impl Into<Option<u32>> for EnumToU32 {
```
dtolnay: This would be more readable if you declare the enum before starting to declare impls on it.

test_suite/tests/test_annotations.rs
```diff
@@ -977,3 +977,83 @@ fn test_invalid_length_enum() {
...
+    );
+    assert_ser_tokens::<StructFromEnum>(&StructFromEnum(None),
+        &[Token::Option(false)]
+    );
+    assert_de_tokens::<StructFromEnum>(&StructFromEnum(Some(2)),
+        &[Token::Option(true),
+          Token::U32(2)
+        ]
+    );
+}
```
dtolnay: Add tests for `as`.

serde_derive/src/ser.rs
```diff
@@ -61,28 +61,49 @@ fn needs_serialize_bound(attrs: &attr::Field) -> bool {
...
+            }
+            Body::Struct(Style::Unit, _) => serialize_unit_struct(&item.attrs),
         }
-        Body::Struct(Style::Unit, _) => serialize_unit_struct(&item.attrs),
+    }
+}
+
+fn serialize_into(into_type: &syn::Ty) -> Fragment {
+    quote_block! {
+        let cloned_val: #into_type = _serde::export::clone::Clone::clone(self).into();
```
dtolnay: This `into()` needs to be written in UFCS.

serde_derive/src/ser.rs
```diff
@@ -61,28 +61,49 @@ fn needs_serialize_bound(attrs: &attr::Field) -> bool {
...
+        cloned_val.serialize(_serializer)
```
dtolnay: UFCS

serde_derive/src/ser.rs
```diff
@@ -61,28 +61,49 @@ fn needs_serialize_bound(attrs: &attr::Field) -> bool {
...
+    }
+}
+
+fn serialize_into_as(into_type: &syn::Ty, into_as_type: &syn::Ty) -> Fragment {
+    quote_block! {
+        let cloned_val: #into_type = _serde::export::clone::Clone::clone(self).into();
```
dtolnay: UFCS

serde_derive/src/ser.rs
```diff
@@ -61,28 +61,49 @@ fn needs_serialize_bound(attrs: &attr::Field) -> bool {
...
+        let cloned_val_as = cloned_val as #into_as_type;
+        cloned_val_as.serialize(_serializer)
```
dtolnay: UFCS

serde_derive/src/ser.rs
```diff
@@ -61,28 +61,49 @@ fn needs_serialize_bound(attrs: &attr::Field) -> bool {
 }
 
 fn serialize_body(item: &Item, generics: &syn::Generics) -> Fragment {
-    match item.body {
-        Body::Enum(ref variants) => {
-            serialize_item_enum(&item.ident, generics, variants, &item.attrs)
+    if let Some(in_ty) = item.attrs.into_type() {
```
dtolnay: What happens if `into_as` is specified but `into` is not?

serde_codegen_internals/src/attr.rs
```diff
@@ -746,6 +785,64 @@ fn get_ser_and_de<T, F>(cx: &Ctxt,
     Ok((ser_item.get(), de_item.get()))
 }
 
+type TypeConversions<T> = (Option<T>, Option<T>, Option<T>, Option<T>);
+
+fn get_from_into_as(cx: &Ctxt,
```
dtolnay: Add compile-fail tests for all of these error messages in test_suite/tests/compile-fail.

serde_codegen_internals/src/attr.rs
```diff
@@ -746,6 +785,64 @@ fn get_ser_and_de<T, F>(cx: &Ctxt,
     Ok((ser_item.get(), de_item.get()))
 }
 
+type TypeConversions<T> = (Option<T>, Option<T>, Option<T>, Option<T>);
```
dtolnay: Make this a struct instead of a tuple so neither caller nor get_from_into_as can mix them up. Generally it is a bad idea to have tuples with many (more than 2) identical element types.

serde_codegen_internals/src/attr.rs
```diff
@@ -746,6 +785,64 @@ fn get_ser_and_de<T, F>(cx: &Ctxt,
...
+
+fn get_from_into_as(cx: &Ctxt,
+                    attr_name: &str,
+                    items: &[syn::NestedMetaItem])
+                    -> Result<TypeConversions<syn::Ty>, ()> {
```
dtolnay: Get rid of the type parameter on TypeConversions, it is only ever used with syn::Ty.

serde_codegen_internals/src/attr.rs
```diff
@@ -270,6 +278,17 @@ impl Item {
                         }
                     }
 
+                    // Parse `#[serde(types(from = "type", into = "type", as = "type"))]
```
dtolnay: Looks like this comment is out of date.

serde_codegen_internals/src/attr.rs
```diff
@@ -270,6 +278,17 @@ impl Item {
...
+                    MetaItem(List(ref name, ref meta_items)) if name == "types" => {
```
dtolnay: I would prefer not to have this `types(...)` layer. I think `#[serde(into = "...")]` is clear enough. What were the reasons for including it?

serde_derive/src/de.rs
```diff
@@ -85,37 +85,63 @@ fn requires_default(attrs: &attr::Field) -> bool {
...
+}
 
-            deserialize_struct(&item.ident,
-                               None,
-                               generics,
-                               fields,
-                               &item.attrs,
-                               None)
+fn deserialize_from(from_type: &syn::Ty) -> Fragment {
+    quote_block!({
```
dtolnay: Just use `quote_block { /* ... */ }` as you did in serialize_into, no need for the extra layer of brace.

serde_derive/src/de.rs
```diff
@@ -85,37 +85,63 @@ fn requires_default(attrs: &attr::Field) -> bool {
...
+        let de_val = <#from_type as _serde::Deserialize>::deserialize(deserializer);
+        match de_val {
+            Ok(from_in) => Ok(_serde::export::convert::From::from(from_in)),
```
dtolnay: Every path needs to be absolute. See https://github.com/serde-rs/serde/issues/737#issuecomment-276728419.

serde_derive/src/de.rs
```diff
@@ -85,37 +85,63 @@ fn requires_default(attrs: &attr::Field) -> bool {
...
-                              None,
-                              generics,
-                              fields,
-                              &item.attrs,
-                              None)
+fn deserialize_from_as(from_type: &syn::Ty, from_as_type: &syn::Ty) -> Fragment {
+    quote_block!({
+        let de_val = <#from_type as _serde::Deserialize>::deserialize(deserializer);
+        match de_val {
+            Ok(from_in) => Ok(_serde::export::convert::From::from(from_in as #from_as_type)),
```
dtolnay: `from_as` doesn't make sense to me. Can you think of any use cases for this?

dtolnay:
I think the point of `as` got lost somewhere along the way. The point was to support Deref types and C-like enums, neither of which requires doing an `into` conversion.

```rust
use std::ops::Deref;

// #[serde(as = "&T")]
pub struct Smart<T> {
    value: T,
}

impl<T> Deref for Smart<T> {
    type Target = T;
    fn deref(&self) -> &T {
        &self.value
    }
}

#[derive(Copy, Clone)]
// #[serde(as = "u8")]
pub enum Toggle {
    Off,
    On,
}

pub fn serialize_smart<T>(smart: &Smart<T>) {
    let _ = smart as &T;
}

pub fn serialize_toggle(toggle: &Toggle) {
    let _ = *toggle as u8;
}
```

I would like to support:

- #[serde(into = "I")]
- #[serde(from = "F")]
- #[serde(into = "I", from = "F")]
- #[serde(as = "A")]
- #[serde(as = "A", from = "F")]

I think the other two cases can be rejected as an error for now  until somebody comes up with a real use case for mixing `as` and `into`.

serde_codegen_internals/src/attr.rs
jbaublitz: I mainly just didn't want to pollute the `serde` namespace since all of the types are grouped together functionally. I will change that if that is not a concern.

dtolnay: I am not worried about polluting the namespace (not yet anyway!). I think it will be fine as long as we group these in a sensible way in our documentation and examples. For example:

```rust
#[serde(as = "&str", from = "String")]
#[serde(rename = "myStr")]

// rather than
#[serde(as = "&str")]
#[serde(rename = "myStr")]
#[serde(from = "String")]
```

dtolnay:
One catch is that Deref types require `self as ...` while C-like enum types require `*self as ...`. Any ideas for how to deal with that?

```rust
pub fn serialize_smart<T>(smart: &Smart<T>) {
    let _ = smart as &T;
}

pub fn serialize_toggle(toggle: &Toggle) {
    let _ = *toggle as u8;
}
```

One simple heuristic that might be good enough is to check whether the thing is a C-like enum, if it is then `*self`, otherwise `self`. Or we could punt on the Deref case by always using `*self` and wait for somebody to complain and possibly introduce a totally different attribute like #[serde(deref)].

I think at this point I am inclined to drop the `as` case entirely and just implement `into` and `from` as those seem a lot more useful and flexible. The C-like enum case will be better served by something like `#[serde(repr = "u8")]` that handles deserialization as well, and the Deref case is questionable whether anybody needs that.

jbaublitz:
I agree that the as case is not necessarily going to add anything at the macro level. If worst comes to worst, a user already has the option to implement this in an `Into` or `From` trait and take care of the details on their own. I will remove the `as`.

dtolnay:
Cool thanks. That means a lot of my review comments probably won't apply anymore.

serde_codegen_internals/src/attr.rs
jbaublitz: At this point with the code changes, I'm assuming this is one of the outdated comments. The only failure that can really happen is a type issue and at that point, it'll fail before it even runs the tests. Is this correct?

dtolnay: Compile-fail tests are supposed to test things that fail to compile. For example [this test](https://github.com/serde-rs/serde/blob/c488cec6412202280b99748c3a73c480d38aa038/test_suite/tests/compile-fail/unknown-attribute/container.rs#L5) is testing [this compile-time failure](https://github.com/serde-rs/serde/blob/c488cec6412202280b99748c3a73c480d38aa038/serde_codegen_internals/src/attr.rs#L274-L275). I would like a similar test for any `cx.error` failures added in this PR. That way we know we don't regress any of these error messages if serde_codegen_internals ever needs to be refactored. An example test would look like `#[serde(into = ")")] //~^ failed to parse type: ")"` (or whatever the message is).

dtolnay:
Oh if you're trying to get compile-fail tests to run locally, it is dumb right now. This should work:

```
cd $SERDE/test_suite/deps
cargo +nightly clean
cargo +nightly update
cargo +nightly build
cd $SERDE/test_suite
cargo +nightly test --features unstable
```

jbaublitz:
@dtolnay Do I need to make any additional changes?

PR #819, opened 2017-03-20
Zero copy deserialization

### LINKED ISSUE #492: What's serde zero-copy story so far?
Hi all, first of all huge kudos to all contributors! Serde is already a tried and proven framework for de/serializing! 

I was wondering what's the current state or plans to support zero-copy deserialization, aka deserializing into &str and &[u8].

dtolnay:
This would need a *lot* more work but fixes #492. What do you think @oli-obk, is all of this worth it?

The idea would be to support:

```rust
#[derive(Serialize, DeserializeBorrow)]
struct S<'a> {
    u: u8,
    s: &'a str,
}
```

### DIFF
--- a/serde/src/de/borrow/impls.rs
+++ b/serde/src/de/borrow/impls.rs
@@ -1,0 +1,77 @@
+#[cfg(feature = "std")]
+use std::borrow::Cow;
+#[cfg(all(feature = "collections", not(feature = "std")))]
+use collections::borrow::Cow;
+
+#[cfg(feature = "collections")]
+use collections::borrow::ToOwned;
+
+use core::fmt;
+use core::marker::PhantomData;
+
+use de::{Visitor, Error};
+use de::borrow::{DeserializeBorrow, DeserializerBorrow, VisitorBorrow};
+
+impl<'a> DeserializeBorrow<'a> for &'a str {
+    fn deserialize_borrow<D>(deserializer: D) -> Result<Self, D::Error>
+        where D: DeserializerBorrow<'a>
+    {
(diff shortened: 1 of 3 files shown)


dtolnay:
One negative aspect of this approach is that tuples can't work because of conflicting impls:

```rust
impl<'a, A, B> DeserializeBorrow<'a> for (A, B)
    where A: DeserializeBorrow<'a>,
          B: Deserialize;

impl<'b, A, B> DeserializeBorrow<'b> for (A, B)
    where A: Deserialize,
          B: DeserializeBorrow<'b>;

impl<'d, 'a: 'd, 'b: 'd, A, B> DeserializeBorrow<'d> for (A, B)
    where A: DeserializeBorrow<'a>,
          B: DeserializeBorrow<'b>;
```

One way around it would be to provide:

```rust
impl<'a, T> DeserializeBorrow<'a> for T
    where T: Deserialize
```

But that requires giving up this impl:

```rust
impl<'a, T> DeserializeBorrow<'a> for Cow<'a, T>
    where T: ToOwned,
          &'a T: DeserializeBorrow<'a> 
```

which would be bad because Cow\<str> is required for dealing with possibly-escaped strings in JSON for example, where you want `"s"` to deserialize as borrowed but `"\u0073"` to deserialize as owned.

oli-obk:
What about adding either a DeserializeBorrow bound to Deserializer or making it a marker trait and adding the method to Deserializer conditionally depending on whether there's an impl?

dtolnay:
Closing in favor of https://github.com/serde-rs/serde/issues/492#issuecomment-288873748.

PR #820, opened 2017-03-21
Re-export derives from serde

### LINKED ISSUE #797: Consider re-exporting derives from serde

dtolnay:
Fixes #797. @oli-obk I am still not settled on whether this should be on or off by default but I am leaning toward off.

### DIFF
--- a/serde/src/lib.rs
+++ b/serde/src/lib.rs
@@ -106,3 +106,50 @@ mod utils;
 // Generated code uses these to support no_std. Not public API.
 #[doc(hidden)]
 pub mod export;
+
+// Re-export #[derive(Serialize, Deserialize)].
+//
+// This is a workaround for https://github.com/rust-lang/cargo/issues/1286.
+// Without this re-export, crates that put Serde derives behind a cfg_attr would
+// need to use some silly feature name that depends on both serde and
+// serde_derive.
+//
+//     [features]
+//     serde-impls = ["serde", "serde_derive"]
+//
+//     [dependencies]
+//     serde = { version = "1.0", optional = true }
+//     serde_derive = { version = "1.0", optional = true }
+//
+//     # Used like this:
+//     # #[cfg(feature = "serde-impls")]
+//     # #[macro_use]
+//     # extern crate serde_derive;
+//     #
+//     # #[cfg_attr(feature = "serde-impls", derive(Serialize, Deserialize))]
+//     # struct S { /* ... */ }
+//
+// The re-exported derives allow crates to use "serde" as the name of their
+// Serde feature which is more intuitive.
+//
+//     [dependencies]
+//     serde = { version = "1.0", optional = true, features = ["derive"] }
+//
+//     # Used like this:
+//     # #[cfg(feature = "serde")]
+//     # #[macro_use]
+//     # extern crate serde;
+//     #
+//     # #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+//     # struct S { /* ... */ }
+//
+// The reason re-exporting is not enabled by default is that disabling it would
+// be annoying for crates that provide handwritten impls or data formats. They
+// would need to disable default features and then explicitly re-enable std.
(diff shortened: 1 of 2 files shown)


serde/src/lib.rs
```diff
@@ -106,3 +106,50 @@ mod utils;
...
+//     # Used like this:
+//     # #[cfg(feature = "serde")]
+//     # #[macro_use]
+//     # extern crate serde;
+//     #
+//     # #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+//     # struct S { /* ... */ }
+//
+// The reason re-exporting is not enabled by default is that disabling it would
+// be annoying for crates that provide handwritten impls or data formats. They
```
oli-obk: that provide **only** handwritten impls

I'd say these are really rare. I think the usability of serde increases a lot if the `serde` crate reexports the derive macros by default.

serde/src/lib.rs
```diff
@@ -106,3 +106,50 @@ mod utils;
...
+//
+// The re-exported derives allow crates to use "serde" as the name of their
+// Serde feature which is more intuitive.
+//
+//     [dependencies]
+//     serde = { version = "1.0", optional = true, features = ["derive"] }
+//
+//     # Used like this:
+//     # #[cfg(feature = "serde")]
+//     # #[macro_use]
```
oli-obk: I didn't know that worked (reexporting derives). That's awesome!

dtolnay:
On crates.io, of the 438 crates that depend on serde >=0.9, 17% of them do not depend on serde_derive.

> anterofit, app_units, arccstr, blob, bodyparser, bson, chrono, cleverbot_io, core-graphics, cryogen_prelude, cssparser, elastic, elastic_types_derive_internals, envy, euclid, eval, gelf, generic-array, geojson, gluon_language-server, hopper, hyper_serde, ipc-channel, iron-json-response, jfs, json-color, json_io, jsonrpc-lite, jsonrpc-macros, juniper, kubewatch, linked-hash-map, mage, mdbook, meval, microsalt, mysql, ndarray, objectid, ocean, offscreen_gl_context, plist, preferences, record-query, reqwest, rmp-serde, rmpv, rocket_contrib, ruma-identifiers, ruma-signatures, rust_decimal, serde-bench, serde-encrypted-value, serde_ignored, serde_json, serde-pickle, serde-protobuf, serde_test, serde-transcode, serde_urlencoded, serde_utils, serde-value, serde-xml-rs, serde_yaml, sharp_pencil, sidekiq, sidekiq-rs, spaceralk, string_cache, tera, toml, treeflection, treeflection_derive, uhttp_json_api, url_serde, uuid

Some broad categories are:

- Data formats (bson, serde_json, serde_yaml, toml)
- Only handwritten impls (chrono, linked-hash-map, uuid)
- Adapters (serde-encrypted-value, serde_ignored)
- General serde helpers (serde_utils, serde-value)
- Serde interop for specific libraries (bodyparser, hyper_serde, rocket_contrib, url_serde)
- Serde traits in the public API (hopper, jsonrpc-macros, kubewatch, mage, preferences)
- Standard library data structures only (cleverbot_io, geojson)

And lots more miscellaneous situations. I think this shows that depending on serde but not serde_derive is a fairly common and diverse use case.

For comparison, pretty much the same fraction of crates (19%) depend on serde >=0.9 but not serde_json so I don't think the number by itself justifies enabling the re-export by default. All three of serde and serde_derive and serde_json provide well-factored and distinguishable functionality so it is fair to expect users to depend on what they need.

I think this change is a last resort for the use case explained in the comment - preserving sanity around a serde cargo cfg. I do not intend to push this as the recommended way to import derives.

PR #824, opened 2017-03-27
Implement Serialize/Deserialize for OsStr/OsString

### LINKED ISSUE #823: Implementations for OsStr/ing

alexcrichton:
This commit implements the two serde traits for the libstd `OsStr` and
`OsString` types. This came up as a use case during implementing sccache where
we're basically just doing IPC to communicate paths around. Additionally the
`Path` and `PathBuf` implementations have been updated to delegate to the os
string ones.

These types are platform-specific, however, so the serialization/deserialization
isn't trivial. Currently this "fakes" a newtype variant for Unix/Windows to
prevent cross-platform serialization/deserialization. This means if you're doing
IPC within the same OS (e.g. Windows to Windows) then serialization should be
infallible. If you're doing IPC across platforms (e.g.  Unix to Windows) then
using `OsString` is guaranteed to fail as bytes from one OS won't deserialize on
the other (even if they're unicode).

Closes #823

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,17 @@
+environment:
+  matrix:
+    - APPVEYOR_RUST_CHANNEL: stable
+    - APPVEYOR_RUST_CHANNEL: nightly
+
+install:
+  # Install rust, x86_64-pc-windows-msvc host
+  - appveyor DownloadFile https://win.rustup.rs/ -FileName rustup-init.exe
+  - rustup-init.exe -y --default-host x86_64-pc-windows-msvc --default-toolchain %APPVEYOR_RUST_CHANNEL%
+  - set PATH=C:\msys64\usr\bin;%PATH%;C:\Users\appveyor\.cargo\bin
+  - rustc -vV
+  - cargo -vV
+
+build: false
+
+test_script:
+  - sh -c 'PATH=`rustc --print sysroot`/bin:$PATH ./travis.sh'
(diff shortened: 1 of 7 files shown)


dtolnay:
I have to run but the test is wrong - the windows test says "Widnows" but is skipped because of the early return in the previous test.

I would prefer to update the test macro to look like:

```rust
#[cfg(unix)]
test_osstring {
    {
        use std::os::unix::ffi::OsStringExt;
        OsString::from_vec(vec![1, 2, 3])
    } => &[
        /* ... */
    ],
}
#[cfg(windows)]
test_osstring {
    {
        use std::os::windows::ffi::OsStringExt;
        OsString::from_wide(&[1, 2, 3])
    } => &[
        /* ... */
    ],
}
```

If you don't get to it I will do this myself this afternoon and then publish a release for you.

alexcrichton:
Oops good catch, I couldn't quite get `#[cfg]` to work out in the macro unfortunately so I just extracted it to a separate test case

dtolnay:
Thanks, I released this in serde 0.9.12 and filed https://github.com/mozilla/sccache/issues/93 to follow up on the sccache side.

alexcrichton:
Thanks @dtolnay!

clarcharr:
@alexcrichton A bit late on this, but one comment: shouldn't the newtype have a separate variant for redox? Because stdlib has a separate implementation for redox and I assume that there's a reason for that.

alexcrichton:
I think that's a backwards compatible extension?

Issue #793, opened 2017-02-27
Stop ignoring so many doc tests


dtolnay:
Most of them are probably wrong. #792

dtolnay:
There are no more ignored tests.

PR #826, opened 2017-03-29
Add a 'de lifetime to the deserialize traits

dtolnay:
This is the boring part of #492 - just plumbing `'de` through all the relevant places.

In this PR all `#[derive(Deserialize)]` generated implementations will apply to any lifetime `'de`. The remaining work for #492 is to support a `#[serde(borrow)]` attribute that constrains the lifetimes that `'de` is allowed to be.

---

The weirdest part of this PR is in serde/src/macros.rs where it assumes that the lifetime of your deserializer is called `'de`. So in particular this would not work:

```rust
impl<'x> Deserializer<'x> for MyDeserializer {
    /* ... */

    forward_to_deserialize! {
        /* ... */
    }
}
```

I think settling on a particular lifetime name as a convention is fine and we can add a feature to `forward_to_deserialize!` to select a different lifetime name in the future if necessary.

---

There is an unfortunate interaction with type checking where any handwritten Deserialize bound on a type parameter conflicts with the automatically inferred Deserialize bound.

```rust
#[derive(Deserialize)]
struct Bounds<T: DeserializeOwned> {
    /* ... */
}
```

The same error is triggered by this simplified model:

```rust
trait Deserialize<'de> {}

trait DeserializeOwned: for<'de> Deserialize<'de> {}
impl<T> DeserializeOwned for T where T: for<'de> Deserialize<'de> {}

struct Bounds<T: DeserializeOwned>(T);
impl<'de, T: DeserializeOwned> Deserialize<'de> for Bounds<T> where T: Deserialize<'de> {}
```

```
error[E0283]: type annotations required: cannot resolve `T: Deserialize<'de>`
 --> src/main.rs:7:1
  |
7 | impl<'de, T: DeserializeOwned> Deserialize<'de> for Bounds<T> where T: Deserialize<'de> {}
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: required by `Deserialize`
```

### DIFF
--- a/serde/src/de/mod.rs
+++ b/serde/src/de/mod.rs
@@ -147,9 +147,9 @@ pub trait Error: Sized + error::Error {
     /// #     type Err = String;
     /// #     fn from_str(_: &str) -> Result<Self, String> { unimplemented!() }
     /// # }
-    /// impl Deserialize for IpAddr {
+    /// impl<'de> Deserialize<'de> for IpAddr {
     ///     fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
-    ///         where D: Deserializer
+    ///         where D: Deserializer<'de>
     ///     {
     ///         let s = try!(String::deserialize(deserializer));
     ///         s.parse().map_err(Error::custom)
@@ -256,7 +256,7 @@ pub trait Error: Sized + error::Error {
 /// # use std::fmt;
 /// # #[allow(dead_code)]
 /// # struct Example;
-/// # impl Visitor for Example {
+/// # impl<'de> Visitor<'de> for Example {
 /// # type Value = ();
 /// fn visit_bool<E>(self, v: bool) -> Result<Self::Value, E>
 ///     where E: Error
@@ -376,7 +376,7 @@ impl<'a> fmt::Display for Unexpected<'a> {
 /// # use std::fmt;
 /// # #[allow(dead_code)]
 /// # struct Example;
-/// # impl Visitor for Example {
+/// # impl<'de> Visitor<'de> for Example {
 /// # type Value = ();
 /// fn visit_bool<E>(self, v: bool) -> Result<Self::Value, E>
 ///     where E: Error
@@ -499,7 +499,7 @@ impl<T> DeserializeOwned for T where T: for<'de> Deserialize<'de> {}
 /// # #[allow(dead_code)]
 /// # enum Error {}
 /// # #[allow(dead_code)]
-/// fn func<T: Deserialize>() -> Result<T, Error>
+/// fn func<'de, T: Deserialize<'de>>() -> Result<T, Error>
 /// # { unimplemented!() }
 /// ```
 ///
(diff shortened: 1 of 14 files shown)


oli-obk:
> I think settling on a particular lifetime name as a convention is fine and we can add a feature to forward_to_deserialize! to select a different lifetime name in the future if necessary.

Yes, I like this too. Add a doc comment to the macro, so it's at least documented somewhere.

> There is an unfortunate interaction with type checking where any handwritten Deserialize bound on a type parameter conflicts with the automatically inferred Deserialize bound.

Can't `serde_derive` detect such bounds and produce different code?

serde/src/de/mod.rs
```diff
@@ -451,16 +451,36 @@ impl<'a> Display for Expected + 'a {
...
+///
+/// ```rust,ignore
+/// pub fn from_str<'a, T>(s: &'a str) -> Result<T>
+///     where T: Deserialize<'a>;
+///
+/// pub fn from_reader<R, T>(rdr: R) -> Result<T>
+///     where R: Read,
+///           T: DeserializeOwned;
+/// ```
+pub trait DeserializeOwned: for<'de> Deserialize<'de> {}
```
oli-obk: Aren't we reinventing `Deserialize<'static>`? I mean, `for<'de> Deserialize<'de>` is different from `Deserialize<'static>`, but `DeserializeOwned` should, to best of my knowledge be equivalent to `Deserialize<'static>`.

dtolnay:
> Can't `serde_derive` detect such bounds and produce different code?

Yes but I think that is a nice-to-have and should not block this feature. In reality you never want a Deserialize bound on the struct definition. You want it on the impl and serde_derive puts it there for you.

```rust
// never what you want
struct S<'de, T: Deserialize<'de>> { /* ... */ }

// what you want
struct S<T> { /* ... */ }
impl<'de, T> Deserialize<'de> for S<T> where T: Deserialize<'de> { /* ... */ }
```

Really the only situation where you want trait bounds in a struct definition is if the fields use associated types of the trait. In this case serde_derive having smarter logic for Deserialize bounds would not help.

```rust
trait Trait: DeserializeOwned {
    type Associated;
}

#[derive(Deserialize)]
struct S<T: Trait> {
    a: T::Associated,
}
```

The workaround is to handwrite the right bound using `#[serde(bound = "...")]`.

serde/src/de/mod.rs
dtolnay: Nope, `Deserialize<'static>` is actually the opposite of `for<'de> Deserialize<'de>`.

The first one means that data borrowed from the deserializer has an arbitrarily **long** lifetime i.e. your type can borrow 'static data from the deserializer. For example `&'static str` would implement `Deserialize<'static>`.

The second one means that data borrowed from the deserializer has an arbitrarily **short** lifetime i.e. you can't borrow any data from the deserializer because it may already be gone.

oli-obk: I understand that difference, but I mean in practical terms they aren't different, because in both cases, whatever type you deserialize to can't have any lifetimes other than `'static` (or must be something like `Option::<&'a Foo>::None`, which again, would turn `'a` into `'static`, since you can turn it into any other lifetime you need afterwards).

To remove all confusion: Can you think of any type that can be deserialized in the `for<'de> Deserialize<'de>` scheme, but not in the `Deserialize<'static>` scheme?

oli-obk: oooh, wait I think I got it... `struct Foo(i32)` might deserialize from `&str`, and we don't want to allocate to `String` first.

dtolnay:
I have not even attempted to fix doc tests that are ignored. I figure it will be easier to take care of that as part of #793.

Issue #825, opened 2017-03-28
Deserialize OsString on Windows


dtolnay:
#824 is missing Deserialize logic for OsString on Windows.

```
---- test_osstring stdout ----
    thread 'test_osstring' panicked at 'assertion failed: `(left == right)`
    (left: `Err(Message("invalid type: enum, expected os string"))`,
    right: `Ok("\u{1}\u{2}\u{3}")`)', C:\projects\serde\serde_test\src\assert.rs:43
```

clarcharr:
Also adding that ser/de isn't enabled on redox either.

dtolnay:
I filed #833 to track Redox support.

PR #834, opened 2017-03-31
Deserialize impls for &str and &[u8]

dtolnay:


oli-obk:
nightly windows fails with 

```
thread 'test_osstring' panicked at 'assertion failed: `(left == right)` (left: `Err(Message("invalid type: enum, expected os string"))`, right: `Ok("\u{1}\u{2}\u{3}")`)', C:\projects\serde\serde_test\src\assert.rs:43
```

dtolnay:
Yep the OsString impl is broken: #825.

PR #839, opened 2017-04-05
Remove questionable type conversions

### LINKED ISSUE #836: Audit type conversions
The standard I would like to set is that if Serde successfully deserializes some input into a particular struct, you can be confident that the input matches what your struct says it should look like.

Currently there are conversions that I think violate this, for example deserializing strings and collections from unit:

```rust
#[derive(Deserialize)]
struct Bad {
    s: String
}

serde_json::from_str::<Bad>("{\"s\":null}").unwrap();
```

To me, null in the input means you should have used Option\<String> or a deserialize_with function that treats null as default.

Let's go through and critically evaluate all of the type conversions that are being done and eliminate these confusing ones.

dtolnay:
Fixes #836.

Type | No longer deserialized from
--- | ---
() | empty seq
bool | "true" and "false" strings
number | FromStr
string | unit
various sequence types | unit
[T; 0] | unit
various map types | unit

These conversions are bad because they make untagged enums confusing.

```rust
#[macro_use]
extern crate serde_derive;

extern crate serde;
extern crate serde_json;

#[derive(Deserialize, Debug)]
#[serde(untagged)]
enum E {
    B(bool),
    S(String),
}

fn main() {
    // This should be E::S but is actually E::B because of implicit conversion.
    println!("{:?}", serde_json::from_str::<E>("\"false\"").unwrap());
}
```

It looks like bool from string and number from string were added only to support XML in https://github.com/serde-rs/serde/pull/70. @RReverser can you make serde-xml-rs work without those conversions? Basically deserialize_bool and deserialize_u64 etc need to not forward to deserialize_string anymore but have their own logic for handling the appropriate type.

The rest were added before we had deserialize_with. I think deserialize_with changes the game and we no longer want these implicit conversions.

### DIFF
--- a/serde/src/de/impls.rs
+++ b/serde/src/de/impls.rs
@@ -80,12 +80,6 @@ impl<'de> Visitor<'de> for UnitVisitor {
     {
         Ok(())
     }
-
-    fn visit_seq<V>(self, _: V) -> Result<(), V::Error>
-        where V: SeqVisitor<'de>
-    {
-        Ok(())
-    }
 }
 
 impl<'de> Deserialize<'de> for () {
@@ -112,16 +106,6 @@ impl<'de> Visitor<'de> for BoolVisitor {
         where E: Error
     {
         Ok(v)
-    }
-
-    fn visit_str<E>(self, s: &str) -> Result<bool, E>
-        where E: Error
-    {
-        match s.trim_matches(::utils::Pattern_White_Space) {
-            "true" => Ok(true),
-            "false" => Ok(false),
-            _ => Err(Error::invalid_type(Unexpected::Str(s), &self)),
-        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


oli-obk:
Yea, serde XML could have done without these hacks, but they were so common already, I didn't stop to think if it makes sense

RReverser:
> @RReverser can you make serde-xml-rs work without those conversions? Basically deserialize_bool and deserialize_u64 etc need to not forward to deserialize_string anymore but have their own logic for handling the appropriate type.

Sure. Are all deserialize_(integer) forwarded to deserialize_u64 and same for float?

dtolnay:
Yes all the number types are still interconverted so you only need to write logic for u64, i64, f64.

e-oz:
Quite idiotic change.

e-oz:
Especially "true/false" to boolean.

e-oz:
Lol, if you want to more polite version: looks like there is a universe, where JS is a strongly typed language, so your frontend will never send you "42" instead of 42. Hope we will find this universe.

PR #858, opened 2017-04-09
Derive for remote types

### LINKED ISSUE #496: Improve the newtype wrapper situation
Something like:

``` rust
serde_newtype! {
    pub struct MyDuration(time::Duration);

    struct time::Duration {
        secs: u64,
        nanos: u32,
    }
}
```

dtolnay:
Fixes #496.

For types with public fields, all you need to provide is a definition of the type.

```rust
#[serde(remote = "time::Duration")]
struct DurationDef {
    secs: u64,
    nanos: u32,
}
```

For structs with one or more private fields, also provide getters for the private fields and a conversion to the remote type.

```rust
#[serde(remote = "time::Duration")]
struct DurationDef {
    #[serde(getter = "time::Duration::as_secs")]
    secs: u64,
    #[serde(getter = "time::Duration::subsec_nanos")]
    nanos: u32,
}

impl From<DurationDef> for time::Duration {
    fn from(def: DurationDef) -> time::Duration {
        time::Duration::new(def.secs, def.nanos)
    }
}
```

Then these definitions can be used to serialize and deserialize the remote type.

```rust
struct Feature {
    #[serde(with = "DurationDef")]
    timestamp: time::Duration,
}
```

### DIFF
--- a/serde/src/private/ser.rs
+++ b/serde/src/private/ser.rs
@@ -8,6 +8,12 @@ use self::content::{SerializeTupleVariantAsMapValue, SerializeStructVariantAsMap
 #[cfg(feature = "std")]
 use std::error;
 
+/// Used to check that serde(getter) attributes return the expected type.
+/// Not public API.
+pub fn constrain<T: ?Sized>(t: &T) -> &T {
+    t
+}
+
 /// Not public API.
 pub fn serialize_tagged_newtype<S, T>(serializer: S,
                                       type_ident: &'static str,
--- a/serde_codegen_internals/src/ast.rs
+++ b/serde_codegen_internals/src/ast.rs
@@ -1,5 +1,6 @@
 use syn;
 use attr;
+use check;
 use Ctxt;
 
 pub struct Item<'a> {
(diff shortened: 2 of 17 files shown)


dtolnay:
And if you want a newtype wrapper rather than serialize_with / deserialize_with functions, it is easy to make one:

```rust
#[serde(remote = "time::Duration")]
struct DurationDef {
    secs: u64,
    nanos: u32,
}

struct Duration(#[serde(with = "DurationDef")] time::Duration);
```

Mark-Simulacrum:
How are the paths in `remote` interpreted? for `#[serde(remote = "time::Duration")]`, would I need to `use std::time;`?

oli-obk:
lgtm, but needs a compile-fail test for using #[serde(remote = "Foo")] with not enough fields, too many fields, wrongly named fields and wrongly typed fields

dtolnay:
@Mark-Simulacrum it works just like any other path anywhere. These refer to the same Duration:

```rust
#[serde(remote = "time::Duration")]
struct DurationDef { /* ... */ }

let d: time::Duration = /* ... */;
```

PR #860, opened 2017-04-09
Standardize on u32 for struct field index and enum variant index

### LINKED ISSUE #722: Struct fields and variant tags both go through deserialize_struct_field
In `#[derive(Deserialize)]` we are generating effectively the same Deserialize implementation for deciding which struct field is next vs which variant we are looking at. As a result, both implementations go through Deserializer::deserialize_struct_field which is unexpected for Deserializer authors trying to implement `EnumVisitor::visit_variant`.

dtolnay:
- Rename `deserialize_struct_field` to `deserialize_identifier` to indicate that it is used not just for struct fields but also for enum discriminants.
- Use u32 as the type of variant indices in the serializer API to match the way that our generated visitors expect visit_u32.

Fixes #722.

### DIFF
--- a/serde/src/de/impls.rs
+++ b/serde/src/de/impls.rs
@@ -1232,7 +1232,7 @@ impl<'de> Deserialize<'de> for Duration {
                     }
                 }
 
-                deserializer.deserialize_struct_field(FieldVisitor)
+                deserializer.deserialize_identifier(FieldVisitor)
             }
         }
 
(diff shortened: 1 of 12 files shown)


dtolnay:
@oli-obk I am merging this because it conflicts with all my other changes but lmk if you want to discuss this.

oli-obk:
I'm unhappy with the `identifier` name, but since the docs explain it well and I can't think of anything better, lgtm

PR #864, opened 2017-04-11
Put rc impls behind a cfg

### LINKED ISSUE #194: Rc<> gets deserialized incorrectly
The way serde currently treats ref-counted data structures means that it creates copies of the references objects. This is bad for several reasons:
- Potentially increased memory usage
- Equality comparison that relies on comparison of address breaks
- Interior mutability is not reflected in copies
  and maybe more that I can't think of right away. I think there should at least be big warnings in the documentation about this behaviour, but the support should be either fixed or deprecated, imo.

dtolnay:
Fixes #194. cc @shahn

### DIFF
--- a/serde/Cargo.toml
+++ b/serde/Cargo.toml
@@ -64,6 +64,11 @@ alloc = ["unstable"]
 #    https://doc.rust-lang.org/collections/
 collections = ["alloc"]
 
+# Opt into impls for Rc<T> and Arc<T>. Serializing and deserializing these types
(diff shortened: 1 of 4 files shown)


nox:
Wouldn't it be better to just document that people shouldn't expect that `Rc<T>` contents get reused after a roundtrip through serde?

nox:
This feature won't save anyone that uses serde and has a dependency that also uses serde with that feature enabled.

PR #876, opened 2017-04-14
Access traits

### LINKED ISSUE #873: Rename Seq/Map/Enum/VariantVisitor
Possible new names:

- `serde::de::SeqAccess`
- `serde::de::MapAccess`
- `serde::de::EnumAccess`
- `serde::de::VariantAccess`

I dislike that Visitor and SeqVisitor are both "visitors" but one of them is the responsibility of the data structure and the other is the responsibility of the data format. I have seen this trip people up where they aren't sure which one(s) they need to implement. I believe the current naming scheme is unhelpful.

Conceptually the only thing doing any visiting is the data structure's Visitor. The SeqVisitor etc have never been visitors but are providing access to the data contained inside a seq/map/enum/variant.

With the new scheme the data structure would be responsible for providing Deserialize(..) and Visitor, while the data format would be responsible for providing Deserializer and (..)Access.

Thoughts @oli-obk @nox?

dtolnay:
Fixes #873.

### DIFF
--- a/serde/src/de/ignored_any.rs
+++ b/serde/src/de/ignored_any.rs
@@ -8,7 +8,7 @@
 
 use lib::*;
 
-use de::{Deserialize, Deserializer, Visitor, SeqVisitor, MapVisitor, Error};
+use de::{Deserialize, Deserializer, Visitor, SeqAccess, MapAccess, Error};
 
 /// An efficient way of discarding data from a deserializer.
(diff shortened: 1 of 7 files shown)


dtolnay:
Moving forward because I have strong conviction that the previous names were poor. We can follow up with adjustments if necessary.

PR #882, opened 2017-04-17
Add Serialize/Deserialize for std types that provide interior mutability

### LINKED ISSUE #179: implement Serialize/Deserialize traits for RefCell
The struct I want to serialize contains a `Vec<Rc<RefCell<Foo>>>`. So, this is kind of half a feature request if a feature is needed in serde, or if not, then it's a question about what do I need to do? :smile:

jethrogb:
Fixes #179


dtolnay:
Thanks! I released this in serde 0.9.14.

nchouba:
can  we add the implementation for [Ref<'a,T>](https://doc.rust-lang.org/nightly/core/cell/struct.Ref.html) as well ?

PR #898, opened 2017-04-23
Remove usage of unstable core::num::Zero, which was removed upstream.

SimonSapin:
https://github.com/rust-lang/rust/pull/41437


SimonSapin:
For Servo we would also like this change to be back-ported to 0.9.x.

CC @nox

PR #899, opened 2017-04-23
Remove usage of unstable core::num::Zero, which was removed upstream.

SimonSapin:
https://github.com/rust-lang/rust/pull/41437

Backport of https://github.com/serde-rs/serde/pull/898 to 0.9.x

### DIFF
--- a/serde/Cargo.toml
+++ b/serde/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "serde"
-version = "0.9.14"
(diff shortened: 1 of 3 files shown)


SimonSapin:
I’ve realized that looking at each byte is incorrect for fat pointers (which can contain non-zero secondary component (length or vtable) but still have a null "data" component). But since `Deserialize` is not implemented for raw pointers anyway that ok.

PR #900, opened 2017-04-25
Replace deprecated MacroInput; completes c52e131

SuperFluffy:
The deprecated `syn::MacroInput` is still used in some places in `serde_derive`. It looks like commit c52e131 didn't fully update all instances to `syn::DeriveInput`.

This fixes it.


dtolnay:
Thank you!

PR #905, opened 2017-04-27
Fix #816 - adjacently-tagged enums honor deny_unknown_fields

### LINKED ISSUE #816: Tag/Content attributes should silently ignore unkown attributes
Currently if I use `#[serde(tag, content)]` on a tagged enum, it seems that serde would throw error when the JSON string contains attributes other than the tag and the content. I think silently ignore the unknown field could be better?

TedDriggs:
Implementation includes happy-path test; wasn't sure where to put the test for `deny_unknown_fields`, but was able to verify that locally in another crate.

#816

### DIFF
--- a/serde_derive/src/de.rs
+++ b/serde_derive/src/de.rs
@@ -795,9 +795,19 @@ fn deserialize_adjacently_tagged_enum(
 
     let expecting = format!("adjacently tagged enum {}", params.type_name());
     let type_name = cattrs.name().deserialize_name();
+    let deny_unknown_fields = cattrs.deny_unknown_fields();
+
+    /// If unknown fields are allowed, we pick the visitor that can
+    /// step over those. Otherwise we pick the visitor that fails on
+    /// unknown keys.
+    let field_visitor_ty = if deny_unknown_fields {
+        quote! { _serde::private::de::TagOrContentFieldVisitor }
+    } else {
+        quote! { _serde::private::de::TagContentOtherFieldVisitor }
+    };
 
     let tag_or_content = quote! {
-        _serde::private::de::TagOrContentFieldVisitor {
+        #field_visitor_ty {
             tag: #tag,
             content: #content,
         }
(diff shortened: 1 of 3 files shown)


test_suite/tests/test_macros.rs
```diff
@@ -751,6 +751,31 @@ fn test_adjacently_tagged_enum() {
         ],
     );
 
+    // unit with excess content (f, g, h)
+    assert_de_tokens(
```
dtolnay: Please add a test that checks the message when an unknown field is denied. See the tests in this file that use `assert_de_tokens_error`.

serde_derive/src/de.rs
```diff
@@ -795,9 +795,19 @@ fn deserialize_adjacently_tagged_enum(
 
     let expecting = format!("adjacently tagged enum {}", params.type_name());
     let type_name = cattrs.name().deserialize_name();
+    let deny_unknown_fields = cattrs.deny_unknown_fields();
+
+    /// If unknown fields are allowed, we pick the visitor that can
```
dtolnay: Please replace these doc comments with ordinary `//` comments. This makes it look like they somehow end up in rustdoc.

serde_derive/src/de.rs
```diff
@@ -844,9 +854,46 @@ fn deserialize_adjacently_tagged_enum(
...
+    };
+
+    /// When allowing unknown fields, we want to transparently step through keys we don't care
+    /// about until we find `tag`, `content`, or run out of keys.
+    let next_relevant_key = if deny_unknown_fields {
+        next_key
+    } else {
+        quote! {
+            {
+                let mut __rk : Option<_serde::private::de::TagOrContentField> = None;
```
dtolnay: Please use absolute paths for `Option` and `None` (below as well) to avoid bugs like https://github.com/serde-rs/serde/issues/737#issuecomment-276728419.

serde_derive/src/de.rs
TedDriggs: In my IDE (VSCode), `///` comments on local vars are supposed to be showing up, like with JSDoc comments on private functions or members.

dtolnay: Oh neat. Never mind then!

test_suite/tests/test_macros.rs
TedDriggs: Done. Also have success case test for `deny_unknown_fields`

PR #909, opened 2017-04-29
[WIP] Allow deriving seeded serialization/deserialization

### LINKED ISSUE #881: Derive stateful serialization/deserialization
I started looking into if it were possible to serialize/deserialize shared references in a way that preserves the reference graph and encountered this proof of concept https://github.com/TyOverby/bincode/issues/139#issuecomment-287259643. While working on generalizing that implementation I started thinking that while I could probably remove most of the duplication by using macros there would still be a non trivial amount of code left which I'd rather not have to write.

So I propose some extensions to serde so it is possible to easily create stateful deserializers and serializers. While this could maybe be implemented as a separate crate I think it would incur a fair amount of duplication if the crate were to properly handle attributes such as `#[serde(rename ...)]`.

# Deserialization

```rust
pub struct PassThroughSeed<S, T> {
    pub seed: S,
    _marker: PhantomData<T>,
}

#[derive(DeserializeSeed)]
#[serde(deserialize_seed="Seed")]
pub struct MyType {
    // Marking a field with a visitor which is constructed from `self.seed`
    #[serde(deserialize_seed_with="SomeTypeVisitor::new")]
    field: SomeType,
    // `NestedType` has `impl DeserializeSeed for PassThroughSeed<Seed, NestedType>`
    #[serde(deserialize_seed)]
    field2: NestedType,
    // If no deserialize_seed_with attribute is set just use the normal `Deserialize` implementation
    field3: i32
}

// Deriving `DeserializeSeed` generates a passthrough implementation
// (It does not use the seed itself but just serves as a provider of the seed to the fields)
impl DeserializeSeed for PassThrough<Seed, MyType> {
    // Normal struct derive except `*_seed` functions are called
}
```

# Serialization

```rust
trait SerializeSeed {
    type Seed;
    fn serialize_dag<S>(&self,
                        serializer: S,
                        seed: &Self::Seed)
                        -> Result<S::Ok, S::Error>
        where S: Serializer;
}


pub struct Wrapper<'a, S, T> {
    pub state: S,
    pub value: &'a T,
}

#[derive(SerializeSeed)]
#[serde(serialize_seed="Seed")]
pub struct MyType {
    #[serialize_seed_with="TypeVisitor::new"]
    field: Type,
    // Wraps `self.field2` using `Wrapper`
    #[serialize_seed]
    field2: Type,
    // Normal serialization
    field3: i32,
}
```

If the implementation seems sound I will try and implement it (wheter that be in serde or in separate cratte) however it might not be a top priority for me for a bit.

Marwes:
Current progress of #881. I added the `SerializeSeed` trait and have verified that it works for a few cases. Going to work on getting some form of `DeserializeSeed` working next.

Fixes #881

### DIFF
--- a/serde/src/ser/mod.rs
+++ b/serde/src/ser/mod.rs
@@ -119,6 +119,7 @@ mod impls;
 mod impossible;
 
 pub use self::impossible::Impossible;
+pub use self::impls::{Seeded, Unseeded};
 
 ////////////////////////////////////////////////////////////////////////////////
 
@@ -1824,102 +1825,4 @@ pub trait SerializeSeed {
     fn serialize_seed<S>(&self, serializer: S, seed: &Self::Seed) -> Result<S::Ok, S::Error>
     where
         S: Serializer;
-}
-
-/// Placeholder
-pub struct Seeded<'seed, S: ?Sized + 'seed, V> {
-    /// Placeholder
-    pub seed: &'seed S,
-    /// Placeholder
-    pub value: V,
-}
-
-impl<'seed, S: ?Sized, V> Seeded<'seed, S, V> {
-    /// Placeholder
-    #[inline]
-    pub fn new(seed: &'seed S, value: V) -> Self {
-        Seeded {
-            seed: seed,
-            value: value,
-        }
-    }
-}
-
-impl<'seed, T: ?Sized, V> Serialize for Seeded<'seed, T, V>
-where
-    V: SerializeSeed<Seed = T>,
-{
-    #[inline]
-    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
-    where
-        S: Serializer,
-    {
-        self.value.serialize_seed(serializer, self.seed)
-    }
-}
-
-/// Placeholder
-pub struct Unseeded<T>(pub T);
-
-impl<T> SerializeSeed for Unseeded<T>
-where
-    T: Serialize,
-{
-    type Seed = ();
-
-    #[inline]
-    fn serialize_seed<S>(&self, serializer: S, _: &Self::Seed) -> Result<S::Ok, S::Error>
-    where
-        S: Serializer,
-    {
-        self.0.serialize(serializer)
-    }
-}
-
-impl<T> SerializeSeed for Option<T>
-where
-    T: SerializeSeed,
-{
-    type Seed = T::Seed;
-
-    #[inline]
-    fn serialize_seed<S>(&self, serializer: S, seed: &Self::Seed) -> Result<S::Ok, S::Error>
-    where
-        S: Serializer,
-    {
-        match *self {
-            Some(ref value) => serializer.serialize_some(&Seeded::new(seed, value)),
-            None => serializer.serialize_none(),
-        }
-    }
-}
-
-impl<T> SerializeSeed for [T]
-where
-    T: SerializeSeed,
-{
-    type Seed = T::Seed;
-
-    #[inline]
-    fn serialize_seed<S>(&self, serializer: S, seed: &Self::Seed) -> Result<S::Ok, S::Error>
-    where
-        S: Serializer,
-    {
-        serializer.collect_seq(self.iter().map(|value| Seeded::new(seed, value)))
-    }
-}
-
-impl<T> SerializeSeed for Vec<T>
-where
-    T: SerializeSeed,
-{
-    type Seed = T::Seed;
-
-    #[inline]
-    fn serialize_seed<S>(&self, serializer: S, seed: &Self::Seed) -> Result<S::Ok, S::Error>
-    where
-        S: Serializer,
-    {
-        self[..].serialize_seed(serializer, seed)
-    }
 }
--- a/serde/src/ser/mod.rs
+++ b/serde/src/ser/mod.rs
@@ -1813,3 +1813,107 @@ where
         _ => None,
     }
 }
+
+
+/// Placeholder
+pub trait SerializeSeed {
+    /// TODO Placeholder (use a type parameter instead?)
+    type Seed: ?Sized;
+
+    /// Placeholder
+    fn serialize_seed<S>(&self, serializer: S, seed: &Self::Seed) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer;
+}
+
+/// Placeholder
+pub struct Seeded<'seed, 'a, S, V>
+where
+    S: ?Sized + 'seed,
+    V: ?Sized + 'a,
+{
+    /// Placeholder
+    pub seed: &'seed S,
+    /// Placeholder
+    pub value: &'a V,
+}
+
+impl<'seed, 'a, S, V> Seeded<'seed, 'a, S, V>
+where
+    S: ?Sized,
+    V: ?Sized,
+{
+    /// Placeholder
+    #[inline]
+    pub fn new(seed: &'seed S, value: &'a V) -> Self {
+        Seeded {
+            seed: seed,
+            value: value,
+        }
+    }
+}
+
+impl<'seed, 'a, T, V> Serialize for Seeded<'seed, 'a, T, V>
+where
+    T: ?Sized,
+    V: ?Sized + SerializeSeed<Seed = T>,
+{
+    #[inline]
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        self.value.serialize_seed(serializer, self.seed)
+    }
+}
+
+/// Placeholder
+pub struct Unseeded<'a, T: ?Sized + 'a>(pub &'a T);
+
+impl<'a, T> SerializeSeed for Unseeded<'a, T>
+where
+    T: ?Sized + Serialize + 'a,
+{
+    type Seed = ();
+
+    #[inline]
+    fn serialize_seed<S>(&self, serializer: S, _: &Self::Seed) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        self.0.serialize(serializer)
+    }
+}
+
+impl<T> SerializeSeed for Option<T>
+where
+    T: SerializeSeed,
+{
+    type Seed = T::Seed;
+
+    #[inline]
+    fn serialize_seed<S>(&self, serializer: S, seed: &Self::Seed) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        match *self {
+            Some(ref value) => serializer.serialize_some(&Seeded::new(seed, value)),
+            None => serializer.serialize_none(),
+        }
+    }
+}
+
+impl<T> SerializeSeed for [T]
+where
+    T: SerializeSeed,
+{
+    type Seed = T::Seed;
+
+    #[inline]
+    fn serialize_seed<S>(&self, serializer: S, seed: &Self::Seed) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        serializer.collect_seq(self.iter().map(|value| Seeded::new(seed, value)))
+    }
+}
--- a/serde/src/ser/mod.rs
+++ b/serde/src/ser/mod.rs
@@ -1827,19 +1827,17 @@ pub trait SerializeSeed {
 }
 
 /// Placeholder
-pub struct Seeded<'seed, 'a, S: ?Sized + 'seed, V: ?Sized + 'a>
-{
+pub struct Seeded<'seed, S: ?Sized + 'seed, V> {
     /// Placeholder
     pub seed: &'seed S,
     /// Placeholder
-    pub value: &'a V,
+    pub value: V,
 }
 
-impl<'seed, 'a, S: ?Sized, V: ?Sized> Seeded<'seed, 'a, S, V>
-{
+impl<'seed, S: ?Sized, V> Seeded<'seed, S, V> {
     /// Placeholder
     #[inline]
-    pub fn new(seed: &'seed S, value: &'a V) -> Self {
+    pub fn new(seed: &'seed S, value: V) -> Self {
         Seeded {
             seed: seed,
             value: value,
@@ -1847,7 +1845,7 @@ impl<'seed, 'a, S: ?Sized, V: ?Sized> Seeded<'seed, 'a, S, V>
     }
 }
 
-impl<'seed, 'a, T: ?Sized, V: ?Sized> Serialize for Seeded<'seed, 'a, T, V>
+impl<'seed, T: ?Sized, V> Serialize for Seeded<'seed, T, V>
 where
     V: SerializeSeed<Seed = T>,
 {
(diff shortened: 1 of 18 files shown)


serde/src/ser/mod.rs
```diff
@@ -1813,3 +1813,107 @@ where
...
+    #[inline]
+    fn serialize_seed<S>(&self, serializer: S, _: &Self::Seed) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        self.0.serialize(serializer)
+    }
+}
+
+impl<T> SerializeSeed for Option<T>
```
Marwes: This kind of duplication seems unfortunate but the only way of avoiding it is to implement all `Serialize` instances by forwarding them to `SerializeSeed`

serde_derive/src/ser.rs
```diff
@@ -16,7 +16,7 @@ use internals::{attr, Ctxt};
 
 use std::u32;
 
-pub fn expand_derive_serialize(input: &syn::DeriveInput) -> Result<Tokens, String> {
+pub fn expand_derive_serialize(input: &syn::DeriveInput, seed: bool) -> Result<Tokens, String> {
```
Marwes: Need to replace this boolean parameter with something more descriptive.

serde_derive/src/ser.rs
```diff
@@ -898,6 +930,51 @@ fn wrap_serialize_with(
...
+fn wrap_serialize_seed(
+    value: Tokens,
+) -> Tokens {
+    quote!({
+        &_serde::ser::Seeded::new(__seed, #value)
+    })
+}
+
+
+fn wrap_serialize_seed_with(
```
Marwes: Largely copy pasted from `wrap_serialize_with` so this should be merged with that function.

Marwes:
0acd32f and 919cc67 does some changes to the serde_test crate to display better errors when it fails at the cost of a lifetime parameter, I can remove those commits if so desired.

dtolnay:
This is amazing work. It will make these seed use cases a lot more accessible. I am not ready to accept this in Serde though. Would you be able to provide derive(SerializeSeed, DeserializeSeed) in an external serde_derive_seed crate? That way we can get some experience with how people use this and what the limitations are before bringing an ultimate version into Serde. You may need to duplicate some of the serde_derive code but that's fine.

I won't be able to accept the Token changes because the API is stable. Is there a different way to test this?

Marwes:
> Would you be able to provide derive(SerializeSeed, DeserializeSeed) in an external serde_derive_seed crate?

It would probably require duplicating the entire `serde_derive` crate as the `Seed` variants still needs to understand most or all attributes that `serde` normally uses. So it's a bit of a hassle but maybe its the best way? In general semver does make experimental additions a bit awkward. An alternative may be to add a feature like `unstable_derive_seed` and explain that using the feature opts in to a feature which may break outside for any version change?

> I won't be able to accept the Token changes because the API is stable. Is there a different way to test this?

Ah, though `serde_test` was just an internal crate. The changes were just to make the error messages contain more information so I may just remove them.

oli-obk:
You can keep rebasing your changes over any changes happening here. So basically take the crate as it is, change version and number and keep publishing it with your changes

Marwes:
Update: derive(DeserializeSeed) has been hacked together, see test_de_seed.rs for it being used to implement rc deserialization. Will start an implemention of https://github.com/gluon-lang/gluon/issues/207 tomorrow which should probably shake out a few more implementation issues. If that works out then I will start cleaning up the code to get this workable for other people as well (as a git dependency at least).

Marwes:
Status: 

Got stuck a long time getting deserialization of generic types working in some fashion. While it works for some simple types it has made me more convinced that the current `DeserializeSeed` interface isn't the right abstraction.

With the current trait one has to define a `Seed` type for every type that needs deserializing which leads to a lot of redundant code. In my mind it seeded deserialization would be significantly easier with a trait such as the one below which more closely mirrors the normal deserialization trait as well as the actual uses I can think of which all involve having a single seed which many different types uses for their deserialization. It is possible I am wrong however, would appreciate insight into why the current trait was chosen instead of any alternatives (cc @dtolnay  @sfackler) https://github.com/serde-rs/serde/issues/650.

```rust
trait DeserializeSeedEx<Seed> {
    fn deserialize<D>(seed: Seed, deserializer: D) -> Result<Self, D::Error>;
}
```

https://github.com/gluon-lang/gluon/pull/279 (Will update this PR later today with the mess of deserialization that generic types create)

dtolnay:
@Marwes that trait is 100% equivalent i.e. you can implement either one in terms of the other (see below). I haven't looked at your code but can you clarify what difficulty you were having with the existing trait?

The real DeserializeSeed is more convenient than DeserializeSeedExt because it lets you skirt the orphan rules. I can define a seed in my crate that deserializes to Vec\<T\> for example. With DeserializeSeedExt it would force you to do an inconvenient wrapper type dance.

```rust
impl<'de, S> DeserializeSeedEx<'de, S> for S::Value
    where S: DeserializeSeed<'de>
{
    fn deserialize<D>(seed: S, deserializer: D) -> Result<Self, D::Error>
        where D: Deserializer<'de>
    {
        seed.deserialize(deserializer)
    }
}

struct Ex<S, T> {
    seed: S,
    value: PhantomData<T>,
}

impl<'de, S, T> DeserializeSeed<'de> for Ex<S, T>
    where T: DeserializeSeedEx<'de, S>
{
    type Value = T;
        where D: Deserializer<'de>
    {
        T::deserialize(self.seed, deserializer)
    }
}
```

Marwes:
@dtolnay 
> I haven't looked at your code but can you clarify what difficulty you were having with the existing trait?

It basically boils down to me needing to do the wrapper type dance you mentioned before since each `derive(DeserializeSeed)` type needs an extra type declared.

https://github.com/gluon-lang/gluon/pull/279/files#diff-96350ec6a8ad57cc6b60f780a54fdf28R151

On its own this is not that bad, but the problems come in that when we deserialize fields we only know about the type of the field, not its `Seed`. So we need to annotate it using `deserialize_seed_with` which we would need to do anyway so that is not a problem. What is a problem is that the current `impl` has a seed type unique it itself so it becomes hard to do this in a generic way.

```rust
struct SeedA(State);
struct A {
   #[serde(deserialize_seed_with = "deserialize_b_from_a")]
   b: B
}

struct SeedC(State);
struct C {
   #[serde(deserialize_seed_with = "deserialize_b_from_c")]
   b: B
}
```

My first thought here is that we just define `deserialize_b<S>(seed: S, ...  where S: Into<State>` (or `AsMut<State>` ...) and we would then define `Into<State> for SeedA` etc. This mostly works but its still some unnecessary boilerplate here (impl `Into` as well as a `deserialize` function for each seed) which I'd rather not have.

My next iteration was something like `struct Ex` [link](https://github.com/gluon-lang/gluon/pull/279/files#diff-e3ba02432b2ff52d409d8cb4adb45865R39). This should eliminate both the `Into` impl and the `deserialize` function since we can now define a single [deserialize function](https://github.com/gluon-lang/gluon/pull/279/files#diff-e3ba02432b2ff52d409d8cb4adb45865R262). However this does not quite work.

First we need an [extra trait](https://github.com/gluon-lang/gluon/pull/279/files#diff-e3ba02432b2ff52d409d8cb4adb45865R62) which we need to implement for each type deriving `DeserializeSeed`. The reason for this is subtle but it is needed for this [line](https://github.com/gluon-lang/gluon/pull/279/files#diff-e3ba02432b2ff52d409d8cb4adb45865R92). I thought at first that it could just be `Seed<T>: DeserializeSeed<'de>`, skipping the `trait` but that causes rustc to fail typechecking since that bound depends on the trait itself, causing an infinite recursion. So the trait becomes an extra bit of boilerplate but its probably the best we can do.

Another issue is that the orphan rules prevent us from using a `Seed<State, T>` type defined in say `serde`. Each crate needs its own `Seed<T>` type which causes some friction between crates.

### DeserializeExt

In theory the `DeserializeEx` trait should make it possible to skip all of these problems since we can just pass the `Seed` directly to its `DeserializeEx` implementation (`deserialize_seed_with` can still be used if the field needs a different seed)

Marwes:
This now brings this PR to an actual working state (as in I have been able to implement everything I need in https://github.com/gluon-lang/gluon/pull/279 and it works for everything I throw at it).

It still lacks documentation which I will add soon but I will elaborate on the implementation a bit here first.

# Current implementation

## Container Attributes

### de_parameters

Container attribute which can be used to add more type parameters to a `DeserializeSeed` impl.

```
#[derive(DeserializeSeed)]
#[serde(deserialize_seed = "ExtraParameterNewtypeSeed<T>")]
#[serde(de_parameters = "T")]
struct ExtraParameterNewtype(
    Inner
);
```

This attribute is necessary either to allow derived implementations to be generic over some seed type, or to allow a seed type to be reused between types which do not have the same generic parameters.

### deserialize_seed

Used to indicate what the seed is for a `DeserializeSeed` derive. If this attribute is not specified then the derive errors.

```
#[derive(DeserializeSeed)]
#[serde(deserialize_seed = "Seed")]
struct Newtype(
    Inner
);
// impl DeserializeSeed for Seed {
//    type Value = Newtype;
//    ...
// }
```

### serialize_seed

Used to indicate what the seed is for a `SerializeSeed` derive. If this attribute is not specified then the derive errors.

```
#[derive(SerializeSeed)]
#[serde(serialize_seed = "Seed")]
struct Newtype(
    Inner
);
// impl SerializeSeed for NewType {
//    type Seed = Seed;
//    ...
// }
```

##  Field attributes

### deserialize_seed_with

Similiarily to `deserialize_with` this calls a custom function when deserializing the marked field.

By default, any field that is not marked with any attributes will just call the normal `Deserialize` implementation so this is the only way to forward to a DeserializeSeed implementation of a field.

```rust

#[derive(DeserializeSeed)]
#[serde(deserialize_seed = "Seed")]
struct Newtype(
    #[serde(deserialize_with = "deserialize_inner")]
    Inner
);
fn deserialize_inner<'de, D>(seed: &mut Seed, deserializer: D) -> Result<Inner, D::Error>
where
    D: Deserializer<'de>,
{
    ...
}
```

### serialize_seed

If a field uses the same seed as the type containing it then it an be marked with `serialize_seed` to forward to the `SerializeSeed` implementation instead of `Serialize`.

```rust
#[derive(SerializeSeed)]
#[serde(serialize_seed = "Cell<i32>")]
struct SeedStruct {
    #[serde(serialize_seed)]
    value: Inner,
}
```

### serialize_seed_with

Similiarily to `serialize_with` this calls a custom function when serializing the marked field.

By default, any field that is not marked with any attributes will just call the normal `Serialize` implementation so to call a `SerializeSeed` implementation this attribute or `serialize_seed` is required.

```rust
#[derive(SerializeSeed)]
#[serde(serialize_seed = "RefCell<NodeToId>")]
struct Node {
    data: char,
    #[serde(serialize_seed_with = "serialize_option_rc_seed")]
    left: Option<Rc<Node>>,
    #[serde(serialize_seed_with = "serialize_option_rc_seed")]
    right: Option<Rc<Node>>,
}

fn serialize_option_rc_seed<S>(
    self_: &Option<Rc<Node>>,
    serializer: S,
    map: &RefCell<NodeToId>,
) -> Result<S::Ok, S::Error>
where
    S: Serializer,
{
    ...
}
```

### seed_with

Does the same thing as the `with` attribute except it calls functions which expect a seed.


## Problems

As it is now, the `SerializeSeed` derive is much easier to use than `DeserializeSeed` since it lets us assume that all fields have the same seed as the type containing them letting us avoid the `serialize_seed_with` attribute most of the time. As mentioned in https://github.com/serde-rs/serde/pull/909#issuecomment-303726504 and https://github.com/serde-rs/serde/pull/909#issuecomment-303802344 there is another way of representing `DeserializeSeed` which brings it closer to how `SerializeSeed` works which should make it a `deserialize_seed` attribute viable. While I have managed to implement everything I need, I will probably try to add the `DeserializeSeedEx` trait (under a better name) and instead provide derivation for that trait instead. Since any explicit implementations of `DeserializeSeed` can still be used via `deserialize_seed_with` there shouldn't be any ergonomic disadvantage to those implementations compared to the current approach.

`DeserializeSeed` takes the seed by value which forces the seed to be cloned for each field. I'd really like to make the derive work well with `&mut` references so this could be avoided.

At a lower priority, `SerializeSeed` does not need to clone the seed since it is passed by `&` but it does require a `RefCell` instead. Potentially this could be fixed by adding `_seed` functions to `Serialize` just as `Deserialize` has but I don't know if this can be done in a backwards compatible manner.

## Where to now

The way I see it there are two ways of getting this released. Either I make a new crate (`serde_derive_seed`  or something similar) containing the deriving code as well as a `serde_seed` crate which provides the `*Seed`  traits and implementations (this extra crate is required to make deserializers and serializers work with derived `*Seed` implementations as otherwise these implementations will depend on this version on `serde` and not version 1.x of `serde`).

The second way is to merge this PR but hide all added functionality behind feature flags which are explicitly documented as unstable functionality and if one uses them one must lock the version of serde one uses.

# Conclusion

I'd say that barring the lack of documentation this is ready for review now so I will hold off any rebasing. The implementation is a bit more verbose than I had hoped for but I find it hard to reduce it further without ballooning the complexity.

dtolnay:
Great work @Marwes. I think providing these traits and derives in a separate crate is going to be more convenient than providing them as an unstable Serde feature that forces you to pin the version of Serde. That way users will continue to receive non-breaking Serde improvements, and the new functionality can be semantically versioned.

Once we get more experience with the use cases that this solves, and we are comfortable committing to the design, we can look at shipping this as part of Serde.

Marwes:
@dtolnay Agree.

My current plan is to move all additions to the `serde` crate into a `serde_seed` crate. There are a few refactorings/cleanups I think I may backport to try and the code in `serde` and `serde_seed` from diverging to much. This crate will in as much as possible mirror the structure of `serde` so that a potential merge later down the line should go smoothly. It will also re-export the entire `serde` crate for the benefit of the derive crate which ought to make it possible to only change `extern crate serde` to `extern crate serde_seed` (https://github.com/serde-rs/serde/blob/4fdba725fe30dc30658b8e71bead7d9b83ab9b0f/serde_derive/src/de.rs#L58).

For the derive crate I will create a `serde_derive_seed` crate which only exports derives for `DeserializeSeed` and `SerializeSeed` and any references to the `serde` crate will go through `serde_seed`. It will otherwise be identical to `serde_derive` and I will attempt to keep this rebased (or merged) with serde master to prevent it from diverging.

I hope to get this done and published before next week, possibly without proper documentation depending on how much time I need to finish "DeserializeSeedEx" (see earlier comment) and `&mut Seed` (I haven't completed porting the entire gluon library to rely on these yet, but considering how much boilerplate gets removed I really think it is the way to go).

dtolnay:
Sounds good!

ebkalderon:
@Marwes Any progress on this going on behind the scenes?

Marwes:
@ebkalderon I released a pair of crates for this last week https://github.com/serde-rs/serde/issues/881#issuecomment-315738922 but i forgot to comment here as well, sorry!

ebkalderon:
No worries, thank you for pointing me in the right direction.

Issue #913, opened 2017-05-02
hashset: duplicate entries are silently discarded when deserializing


lucab:
When working with an hashset, it looks like serde is currently deserializing non-unique entries simply by silently discarding all but the last occurence. I was a bit surpised as I discovered this when writing tests around supposedly malformed inputs (with duplicate elements), which were instead accepted by serde.

This probably also applies to other types of sets and maps. It may even be working as intended, but I think it may lead to unexpected surprises.

As an example, I was going through:
```
#[test]
fn test_hashset() {
    #[derive(Serialize, Deserialize, Debug, PartialEq, Eq, Hash)]
    struct Entry {
        id: u32,
        val: String,
    }
    let jsonin = r#"[{"id":1, "val":"a"}, {"id":1, "val":"a"}]"#;
    std::collections::HashSet<Entry> = serde_json::from_str(jsonin).unwrap();
}
```

This naive case will just result in the set having a single entry instead of the original two, which I find already a bit strange as I would have expected some "duplicate key" error. However, a similar example with a custom implementation of `Eq` and `Hash` (e.g. only keyed on `id`) will result in actual values being discarded without the consumer being aware of it.

Is this semantics on purpose? If so, is there is a way of opting-out in order to get a stricter deserializer? In general, would you consider changing this behavior to early-fail instead of keep-going?

dtolnay:
See the discussion in #916.

PR #916, opened 2017-05-06
de: ensure unique keys when deserializing to maps and sets

lucab:
BREAKING: deserializing inputs with duplicate keys to maps and sets now fails.

Sets and maps do not allow insertion of duplicate entries, which are
instead skipped or replaced.
This updates serde to check for duplicate entries when deserializing,
erroring when such cases are encourented.

Closes https://github.com/serde-rs/serde/issues/913

### DIFF
--- a/serde/src/de/impls.rs
+++ b/serde/src/de/impls.rs
@@ -553,6 +553,63 @@ macro_rules! seq_impl {
 }
 
 #[cfg(any(feature = "std", feature = "collections"))]
+macro_rules! set_impl {
+    (
+        $ty:ident < T $(: $tbound1:ident $(+ $tbound2:ident)*)* $(, $typaram:ident : $bound1:ident $(+ $bound2:ident)*)* >,
+        $access:ident,
+        $ctor:expr,
+        $with_capacity:expr,
+        $insert:expr
+    ) => {
+        impl<'de, T $(, $typaram)*> Deserialize<'de> for $ty<T $(, $typaram)*>
+        where
+            T: Deserialize<'de> $(+ $tbound1 $(+ $tbound2)*)*,
+            $($typaram: $bound1 $(+ $bound2)*,)*
+        {
+            fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+            where
+                D: Deserializer<'de>,
+            {
+                struct SeqVisitor<T $(, $typaram)*> {
+                    marker: PhantomData<$ty<T $(, $typaram)*>>,
+                }
+
+                impl<'de, T $(, $typaram)*> Visitor<'de> for SeqVisitor<T $(, $typaram)*>
+                where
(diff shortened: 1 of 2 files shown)


clarcharr:
Perhaps this could be made not-breaking by adding an opt-in attribute?

dtolnay:
The current behavior matches both `JSON.parse` in JavaScript and `encoding/json` in Go. Do you know of any JSON implementations that behave the way your PR does? I worry that this would break expectations.

```js
JSON.parse('{"a":0,"a":1}')
```

```go
package main

import (
	"encoding/json"
	"fmt"
)

func main() {
	var m map[string]int
	err := json.Unmarshal([]byte(`{"a":0,"a":1}`), &m)
	fmt.Println(m, err)
}
```

lucab:
@clarcharr As I said in #913, I would be fine with some way to opt-in this behavior (even if I personally believe this should be opt-out instead). But I think this should be per-deserializer instead of per-field, as library authors and consumers may have different usecases at hand.

@dtolnay Your comment is on-spot. I'm actually here because this behavior in Go is resulting in higher-level pain there: a seemingly random logic bug was traced down to a producer emitting a manifest with duplicate fields, with occasional re-ordering. Even if important/widespread, I won't take those as my ideal references.

Some related datapoints:
 * Haskell aeson is having a similar discussion right now https://github.com/bos/aeson/issues/531
 * using the same JSON dict with duplicate keys as input, currently serde deserializes it as a map/set (discarding previous keys) but refuses to deserialize it as a struct with named fields (erroring on duplicate keys). This uniforms the behavior toward the early-error side.

dtolnay:
Thanks for the link but it looks like they are pretty far from a consensus that treating this as an error is the right behavior.

For your use case would you be able to use your own map type for which you define Deserialize, or a `deserialize_with` attribute on the field to specify this logic for the deserialize behavior?

lucab:
Sorry, I didn't mean to imply that aeson is moving in this direction, just that a better comparison would be that haskell discussion (instead of golang/js stdlib legacy).

And yes, I can encode this in my own deserializers. I'm am however concerned about the default behavior of serde here, as I was surprised when I first encountered #913. I'm inclined to close this PR as I understand this requires a bit more thoughts, but I'd still would like to see something better than "redefine all maps/sets deserializers" for this directly in serde.

dtolnay:
I added this to the collection of deserialize_with helpers I intend to provide in #553. Thanks!

lucab:
@dtolnay thank you for the feedback. With context, I now understand your previous question better and I think it is a good middleground solution.

petreeftime:
As far as I can tell this issue is still not resolved, or the solution is somewhere in a crate I can't really find. Is there a nicer way to Error on duplicate keys except by re-implementing the Deserialize function for the specific container that might have duplicates?

PR #925, opened 2017-05-10
Take dependency on ident_case

TedDriggs:
I moved the rename rule logic from `serde_derive_internals::case` out into their own crate so I could use it in a separate project. This seems useful for anyone working on proc macros, so this PR would update serde to use that crate rather than keeping its own copy of the code.


dtolnay:
Thanks! I am okay with having this duplicated for now.

Side note: I would have tried to do this in a way that allows adding new rename rules without a breaking change, or given it a version less than 1.0.

TedDriggs:
> Side note: I would have tried to do this in a way that allows adding new rename rules without a breaking change, or given it a version less than 1.0.

In hindsight, this would've been a good idea; I was under the impression 1.x packages tried to avoid 0.x dependencies, so I set it for that. Do you think a 2.x with a `__Hidden` variant would be a good way to address this?

dtolnay:
There is no problem with 1.x packages having 0.x dependencies as long as they are not exposed in the public API.

I don't think a `__Hidden` variant would be a good way to address this. I would recommend making a list of all the ways you expect callers to interact with this crate and then designing an API for those interactions. In particular, pattern matching on an instance of `RenameRule` is probably not something callers will need to do.

PR #935, opened 2017-05-16
Add a combined skip attribute

### LINKED ISSUE #931: Warn when using only one of skip_serializing and skip_deserializing
I used the `skip_serializing` attribute for a field, but forgot to add `skip_deserializing` too. After serializing my data to a binary format (with `bincode`), I tried to deserialize it but got an error which took a long time to track down.

```
/* Example code */
pub struct Data {
    #[serde(skip_serializing)] // forgot skip_deserializing
    foo: u32,
    //...
}
```

If `Serialize` and `Deserialize` are both derived, then it never makes sense to use only one of `skip_serializing` and `skip_deserializing` -- I expect a derived serializer to always be compatible with the derived deserializer.

Here are some ways to prevent this:

  - Warn when both `Serialize` and `Deserialize` and derived but only one of `skip_serializing` and `skip_deserializing` is used.
  - Make `skip_serializing` imply `skip_deserializing` and vice versa.
  - Replace `skip_serializing` and `skip_deserializing` with a single attribute that means both.
  - (As a last resort: in the documentation for `skip_serializing`, add the warning "You may want to use `skip_deserializing` too.", and vice versa.)

This issue also applies to:

  - The `skip_serializing` and `skip_deserializing` variant attributes.
  - The `serialize_with` and `deserialize_with` field attributes.

What do you think?

spikefoo:
fixes #931

I added a combined skip field attribute.

If this looks fine, I'll add a combined skip variant attribute too.

### DIFF
--- a/serde_derive_internals/src/attr.rs
+++ b/serde_derive_internals/src/attr.rs
@@ -719,6 +719,12 @@ impl Field {
                         skip_deserializing.set_true();
                     }
 
+                    // Parse `#[serde(skip)]`
+                    MetaItem(Word(ref name)) if name == "skip" => {
+                        skip_serializing.set_true();
+                        skip_deserializing.set_true();
+                    },
+
                     // Parse `#[serde(skip_serializing_if = "...")]`
                     MetaItem(NameValue(ref name, ref lit)) if name == "skip_serializing_if" => {
                         if let Ok(path) = parse_lit_into_path(cx, name.as_ref(), lit) {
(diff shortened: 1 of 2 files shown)


spikefoo:
I'm not adding a combined skip variant attribute for now, because it appears `skip_serializing` and `skip_deserializing` have orthogonal behavior as variant attributes, at least with `bincode`: [bincode#184](https://github.com/TyOverby/bincode/issues/184).

cc @dtolnay

PR #949, opened 2017-06-04
Support std::time::SystemTime

### LINKED ISSUE #535: impl Serialize and Deserialize for std::time::SystemTime
If there is a different type that is better suited for timestamps in serialized data that's already supported by serde, I'd be happy to know about it. The serialized format should probably be [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601).

WiSaGaN:
Closes #535

### DIFF
--- a/serde/src/de/impls.rs
+++ b/serde/src/de/impls.rs
@@ -1364,6 +1364,132 @@ impl<'de> Deserialize<'de> for Duration {
 
 ////////////////////////////////////////////////////////////////////////////////
 
+#[cfg(feature = "std")]
+impl<'de> Deserialize<'de> for SystemTime {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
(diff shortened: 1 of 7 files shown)


WiSaGaN:
A naive implementation of `serde` support for `std::time::SystemTime`. Tested locally.

serde/src/de/impls.rs
```diff
@@ -1364,6 +1364,132 @@ impl<'de> Deserialize<'de> for Duration {
...
+            }
+        }
+
+        struct DurationVisitor;
+
+        impl<'de> Visitor<'de> for DurationVisitor {
+            type Value = Duration;
+
+            fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+                formatter.write_str("struct Duration")
```
dtolnay: Please change this to refer to SystemTime.

Issue #271, opened 2016-03-28
Macro serialization for struct methods


kardeiz:
Is there any plan to add annotations that would allow (public) struct methods to be serialized? Something like:

```
#[derive(Serialize, Deserialize, Debug)]
struct Foo { /* ... */ }
impl Foo {
    #[serde(serialize))]
    pub fn bar(&self) -> i32 {
        // ...
    }
}
```

For example, this would be similar to the `@JsonProperty` annotation for methods used by some Java JSON libraries.

I asked about this on [StackOverflow](http://stackoverflow.com/questions/36159031/add-struct-method-to-serde-serialization-output) and received a possible workaround, but no definitive answer.

It seems like this would be a pretty common scenario, where one would want to serialize something that might not be a field (unless I am missing something). I skimmed through the issues and didn't see any relevant discussion, but I apologize if this has been discussed elsewhere&mdash;

jwilm:
How is that "property" supposed to be deserialized?

kardeiz:
Right&mdash;I understand that this would be a one-way kind of thing. It's very possible that this would be out of the scope of `serde` since there would be no pure back-and-forth serialization-deserialization. I guess it might be more appropriate to structure this "view" of the struct through a separate struct, as suggested by Shepmaster on SO.

I just know that `@JsonProperty` and similar features are very convenient in the languages where I have used them...

dtolnay:
See #760 for a similar request - serializing constant values.

```rust
// must be serialized as {"type": "Program", "body": [...]}
pub struct Program {
    pub body: Vec<Stmt>
}

#[serde(tag = "type")]
pub enum Expr {
    // must be serialized as {"type": Literal", "value": null}
    Literal
}
```

Given that we never want to have people writing code inside of attributes, I think this is just a degenerate case of serializing method values.

TedDriggs:
@dtolnay, the enum tag case feels a bit special, in part because I don't know how one would access the tag name for a variant in a method on the enum.

dtolnay:
Oh in the enum case it is not the tag that would be added. That is already supported. See [playground](http://play.integer32.com/?gist=4d429c12b7f4c0e65d36cf486333567a). The user wants to add the constant field `"value": null`.

GyrosOfWar:
I've looked into this a little. I imagine it would be nice if you could do something like
```rust
#[derive(Deserialize)]
struct Test {
    value: i32
}

impl Test {
[...]

PR #950, opened 2017-06-06
[WIP] Serialization for struct methods

GyrosOfWar:
Addresses #271. 

This allows serialization of values that are computed by methods in an `impl` block of a struct or an enum. Only methods that take a &self parameter are supported. Example:
```rust
#[derive(Deserialize, Debug, PartialEq)]
#[serde(method_properties(data = "Test::as_string"))]
struct Test{
    pub value: i32,
}

impl Test {
    #[allow(unused)]
    pub fn as_string(&self) -> String {
        format!("The value is {}", self.value)
    }
}
```

TODO:
- [ ] Prevent duplicate keys in deserialized data
- [ ] How to specify names that are not valid words in an attribute? For example a fake field called da-ta or da.ta or da ta.
- [ ] How do these compose with field attributes, if at all? For example, say I want to serialize the value of Test::as_string but use a skip_serializing_if or serialize_with attribute on it. 
- [ ] How do these compose with container attributes? If there is a rename_all on the container, does that apply to this fake field?
- [ ] Weird issue in test case with enum: struct variant has len of 1 even though the method value and the field both get deserialized.
- [ ] More tests.


dtolnay:
I am closing because it looks like this is not being worked on anymore. Please open a different PR if you are still interested in this.

PR #961, opened 2017-06-18
Add methods to provide field indices

### LINKED ISSUE #960: Pass field index to SerializeStruct::serialize_field
Similar to how `Serializer::serialize_*_variant` receives both the index and name of the variant. This would enable CBOR to deserialize correctly from a packed representation (integer keys) with fields that are conditionally skipped - https://github.com/pyfisch/cbor/issues/37.

sfackler:
Closes #960

This change does mean that if you run a new serde_derive with an old serde the build will fail. That seems like an uncommon situation to be stuck in though.

### DIFF
--- a/serde/src/ser/mod.rs
+++ b/serde/src/ser/mod.rs
@@ -1703,9 +1703,9 @@ pub trait SerializeMap {
 ///         where S: Serializer
 ///     {
 ///         let mut rgb = serializer.serialize_struct("Rgb", 3)?;
-///         rgb.serialize_field("r", &self.r)?;
-///         rgb.serialize_field("g", &self.g)?;
-///         rgb.serialize_field("b", &self.b)?;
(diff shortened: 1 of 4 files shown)


sfackler:
I'd like to update serde_test's `Token` to be able to store the index, but I'm not sure what the back compat situation is there.

dtolnay:
Are there any restrictions that need to be placed on how this is used? In particular, is the following valid and how would Bincode be expected to behave in this case?

```rust
rgb.serialize_indexed_field("g", 1, &self.g)?;
rgb.serialize_indexed_field("r", 0, &self.r)?;
rgb.serialize_indexed_field("b", 2, &self.b)?;
```

dtolnay:
https://github.com/serde-rs/serde/issues/960#issuecomment-317225162 proposes a different approach.

sfackler:
Closing in favor of #1022.

Issue #745, opened 2017-02-03
Allow integer tags for internally tagged enums


dtolnay:
See [this use case](https://www.reddit.com/r/rust/comments/5rwe3w/new_in_serde_096_untagged_and_internally_tagged/ddaq5lf/).

> Would the internally tagged enum support allow me to handle schema versioning defined like this?
>
> ```
> {
>     "schema_version": 1,
>     ...
> }
> ```
>
> I've never gotten a good answer about how I'd do that with Serde.

cc @ssokolow

dtolnay:
Attributes support non-string literals now right? This could be as simple as allowing:

```rust
#[derive(Serialize, Deserialize)]
#[serde(tag = "schema_version")]
enum E {
    #[serde(rename = 1)]
    V1(...),
    #[serde(rename = 2)]
    V2(...),
}
```

dtolnay:
Also boolean tags?

```rust
#[derive(Serialize, Deserialize)]
#[serde(tag = "error")]
enum Response {
    #[serde(rename = false)]
    Ok(QueryResult),
    #[serde(rename = true)]
    Err(QueryError),
}
```

Phaiax:
Any plans to progress on this feature? I could give it a try, but I would need a bit of mentoring / pointing to the right places.

dtolnay:
I have not started working on this. I would love a PR! Happy to provide guidance if you run into any trouble.

Noxime:
Any updates on this?

NotBad4U:
Hi :smile: 
We need this feature for [sozu #240](https://github.com/sozu-proxy/sozu/issues/240)  to handle configuration versioning.

I'd like to implement it. I saw that someone (#973) started working on it but abandoned it.
Can I use it as a starting point or do you recommend another approach ?

Arnavion:
@NotBad4U You can use a string tag (the version enum's variant name) for the configuration version.

dtolnay:
@NotBad4U I think #973 is the right approach. Literals in attributes will be stable in rust 1.30 so we can support `#[serde(rename = 0)]`.

WiSaGaN:
I guess this is blocked on https://github.com/serde-rs/serde/pull/1392?

Ekleog:
For what it's worth, I think there's an additional use case for this (though it's technically not for internally tagged enums, it'd hopefully be fixed the same way): `#[repr(i32)]` enums and the like.

[...]

PR #973, opened 2017-07-01
WIP: Allow #[serde(rename=str|bool|int)] for enum variants.

Phaiax:
(This commit contains untested code and a dirty hacks so that the rename="str" still works)

#745 Issue for this PR

I got started and it was not too hard, serde derive is written very well. (I'm still confused by the deserializiation part, but I have not looked into it to much)

Some questions:
1. Non string literals are not stabilized yet. I think we should support a fallback with `#[serde(rename="1", rename_as="int")]`. This is necessary since serde supports old rustc versions anyway, so even with [stabilization](https://github.com/rust-lang/rust/issues/34981) this will not go away.
    ```
    error: non-string literals in attributes, or string literals in top-level positions, are experimental (see issue #34981)
       --> tests/test_macros.rs:713:9
        |
    713 |         #[serde(rename=true)]
        |         ^^^^^^^^^^^^^^^^^^^^^
        |
        = help: add #![feature(attr_literals)] to the crate attributes to enable
    ```
2. Currently I implemented a type `StrBoolInt`, but I had to define it twice, once in `serde_derive_internals` and once again in `serde_derive`, so that I can implement `ToTokens`.
3. The second `StrBoolInt` is stuffed into `fragment.rs`. Should I use a new file or is that ok?
3. I decided to only allow non-string renames for internally and adjacently tagged enums. Correct?
4. As far as I understood it, `Serializers` have the choice to serialize either with the name of the variant or with the id. What happens if `rename=2` is stated for attribute with field number 1? Should I just replace the whole the `derserialize_u32` part as soon as an int-rename is encountered?
5. What is that `[#serde(field_identifier/variant_identifier)]` setting? It is not mentioned in the docs.

Thx, Daniel

### DIFF
--- a/serde/src/export.rs
+++ b/serde/src/export.rs
@@ -14,9 +14,7 @@ pub use lib::marker::PhantomData;
 pub use lib::option::Option::{self, None, Some};
 pub use lib::result::Result::{self, Ok, Err};
 
-pub use self::string::from_utf8_lossy;
-pub use self::string::from_int;
-pub use self::string::from_bool;
+pub use self::string::{from_utf8_lossy, from_int, from_bool};
 
 mod string {
     use lib::*;
@@ -42,22 +40,30 @@ mod string {
     }
 
     pub fn from_bool(b : bool) -> &'static str {
-        match b {
-            true => "true",
-            false => "false",
+        if b {
+            "true"
+        } else {
+            "false"
         }
     }
 
     #[cfg(any(feature = "std", feature = "alloc"))]
     pub fn from_int(i: u64) -> Vec<u8> {
-        format!("{}", i).into_bytes()
+        use lib::fmt::Write;
+        let mut buf = String::with_capacity(20);
+        write!(&mut buf, "{}", i).ok();
+        buf.into_bytes()
     }
 
     #[cfg(not(any(feature = "std", feature = "alloc")))]
     pub fn from_int(i: u64) -> [u8; 20] {
-        let buf = [0; 20];
-        let wrap = Wrapper { buf: &buf };
-        write!(wrap, "{}", i);
+        use lib::fmt::Write;
+        // len(str(1<<64)) = 20
+        let mut buf = [0; 20];
+        {
+            let mut wrap = Wrapper { buf: &mut buf };
+            write!(wrap, "{}", i).ok();
+        }
         buf
     }
 
--- a/serde/src/export.rs
+++ b/serde/src/export.rs
@@ -15,6 +15,8 @@ pub use lib::option::Option::{self, None, Some};
 pub use lib::result::Result::{self, Ok, Err};
 
 pub use self::string::from_utf8_lossy;
+pub use self::string::from_int;
+pub use self::string::from_bool;
 
 mod string {
     use lib::*;
@@ -37,5 +39,42 @@ mod string {
         // white-on-black question mark. The user will recognize it as invalid
         // UTF-8.
         str::from_utf8(bytes).unwrap_or("\u{fffd}\u{fffd}\u{fffd}")
+    }
+
+    pub fn from_bool(b : bool) -> &'static str {
+        match b {
+            true => "true",
+            false => "false",
+        }
+    }
+
+    #[cfg(any(feature = "std", feature = "alloc"))]
+    pub fn from_int(i: u64) -> Vec<u8> {
+        format!("{}", i).into_bytes()
+    }
+
+    #[cfg(not(any(feature = "std", feature = "alloc")))]
+    pub fn from_int(i: u64) -> [u8; 20] {
+        let buf = [0; 20];
+        let wrap = Wrapper { buf: &buf };
+        write!(wrap, "{}", i);
+        buf
+    }
+
+    #[cfg(not(any(feature = "std", feature = "alloc")))]
+    struct Wrapper<'a> {
+        buf: &'a mut [u8],
+    }
+
+    #[cfg(not(any(feature = "std", feature = "alloc")))]
+    impl<'a> fmt::Write for Wrapper<'a> {
+        // Could panic if buf is too small.
+        fn write_str(&mut self, s: &str) -> fmt::Result {
+            let bytes = s.as_bytes();
+            self.buf[..bytes.len()].copy_from_slice(bytes);
+            let this : &mut[u8] = mem::replace(&mut self.buf, &mut []);
+            self.buf = &mut this[bytes.len()..];
+            Ok(())
+        }
     }
 }
--- a/serde_derive/src/de.rs
+++ b/serde_derive/src/de.rs
@@ -705,11 +705,6 @@ fn deserialize_internally_tagged_enum(
         }
     };
 
-    // TODO: just to keep the compiler quiet for now, this must not be strinfied here!
-    let variant_names_idents = variant_names_idents.iter().map(|&(ref variant, ref ident)|
-                                                            (variant.clone(), ident.clone()))
-                                                     .collect();
-
     let variant_visitor = Stmts(deserialize_generated_identifier(variant_names_idents, cattrs, true),);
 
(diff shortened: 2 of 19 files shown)


Phaiax:
Hey @dtolnay .

I think this is ready for a first review. 

As for my Questions.

1. The implementation treats `#[serde(rename=3)]` the same as `#[serde(rename="3", rename_as="int")]` and gives errors if the parsing fails or the `rename_as` is used with a non string literal for `rename`. 
2. still same question as above.
3. still same question as above.
4. still same question as above.
5. I implemented it so that _deserializing by index_ is deactivated as soon as one rename has an int type. 
6. still same question as above.
7. This PR will collide with #963 (same lines edited). I use u64 for the int type.
8. Naming of structs. See my own review.

serde_derive/src/de.rs
```diff
@@ -1366,14 +1366,29 @@ fn deserialize_custom_identifier(
...
+    fields: &[(StrBoolInt, Ident)],
     is_variant: bool,
     fallthrough: Option<Tokens>,
 ) -> Fragment {
-    let field_strs = fields.iter().map(|&(ref name, _)| name);
-    let field_bytes = fields.iter().map(|&(ref name, _)| quote::ByteStr(name));
-
-    let constructors: &Vec<_> = &fields
+    let field_strs = fields.iter().filter_map(|&(ref n, _)| n.as_str());
+    let field_bytes = fields.iter().filter_map(|&(ref n, _)| n.as_str().map(|n| quote::ByteStr(n)));
```
Phaiax: maybe rename into `matcharms_str`. (`field_strs` was the original name)

serde_derive/src/de.rs
```diff
@@ -1384,7 +1399,36 @@ fn deserialize_identifier(
...
+
+    let (bytes_to_str, int_to_str, bool_to_str) = if fallthrough.is_some() {
+        (None, None, None)
+    } else {
+        let conversion_bytes = quote! {
+            let __value = &_serde::export::from_utf8_lossy(__value);
+        };
+        let conversion_int = quote! {
+            let __value = &_serde::export::from_int(__value);
+            let __value = &_serde::export::from_utf8_lossy(__value);
```
Phaiax: First convert into [u8;20] or Vec<u8> then interpret as utf8

serde_derive/src/de.rs
```diff
@@ -1384,7 +1399,36 @@ fn deserialize_identifier(
...
+        quote! {
+            _serde::export::Err(_serde::de::Error::unknown_variant(__value, VARIANTS))
+        }
+    } else {
+        quote! {
+            _serde::export::Err(_serde::de::Error::unknown_field(__value, FIELDS))
+        }
+    };
+
+    let visit_int = if constructors_ints.is_empty() && is_variant {
```
Phaiax: Note when a deserialized int is interpreted as index and when as a `renamed=int`

serde_derive/src/fragment.rs
```diff
@@ -73,3 +73,88 @@ impl ToTokens for Match {
...
+    Bool(bool),
+    Int(u64),
+}
+
+impl StrBoolInt {
+    /// Get a string description for use in error messages.
+    ///
+    /// It will be used as second argument in
+    /// `_serde::Serializer::serialize_struct(__serializer, name, len)`
+    /// while serializing the inner struct in adjacently tagged enums.
```
Phaiax: This may be a problem, depending on how a serializer serializes a struct name. But I guess that formats that require a tagged enum do not serialize the variant/struct name. (This case is part of the tests and has a comment there also)

serde_derive/src/fragment.rs
```diff
@@ -73,3 +73,88 @@ impl ToTokens for Match {
         }
     }
 }
+
+use internals::attr::StrBoolInt as StrBoolIntInternal;
+
+/// An extended name, used for the type tag for enums.
+#[derive(Debug, Clone)]
+pub enum StrBoolInt {
```
Phaiax: Correct file for this struct?

serde_derive/src/ser.rs
```diff
@@ -392,7 +392,10 @@ fn serialize_externally_tagged_variant(
     cattrs: &attr::Container,
 ) -> Fragment {
     let type_name = cattrs.name().serialize_name();
-    let variant_name = variant.attrs.name().serialize_name();
+    let variant_name = match variant.attrs.name().serialize_name() {
+        attr::StrBoolInt::Str(s) => s,
+        _ => unreachable!(),
```
Phaiax: `check.rs` prevents this case

test_suite/tests/test_macros.rs
```diff
@@ -696,6 +696,262 @@ fn test_internally_tagged_enum() {
...
+    );
+
+    assert_de_tokens_error::<InternallyTagged>(
+        &[
+            Token::Map { len: Some(1) },
+            Token::Str("type"),
+            Token::Bool(false)
+        ],
+        "unknown variant `false`, expected one of `abc`, `true`, `3`, `4`, `5`, `6`",
+    );
```
Phaiax: the error message doesn't differentiate between string literals and non string literals ('3' vs 3). Should be ok anyway I hope.

test_suite/tests/test_macros.rs
```diff
@@ -696,6 +696,262 @@ fn test_internally_tagged_enum() {
...
+    assert_tokens(
+        &AdjacentlyTagged::B { b: 1 },
+        &[
+            Token::Struct { name: "AdjacentlyTagged", len: 2 },
+
+            Token::Str("t"),
+            Token::Bool(true),
+
+            Token::Str("c"),
+            // TODO: Is this the correct way to name the struct? stringify the rename?
```
Phaiax: that is the comment I referred to earlier in the review.

dtolnay:
> Currently I implemented a type `StrBoolInt`, but I had to define it twice, once in `serde_derive_internals` and once again in `serde_derive`, so that I can implement `ToTokens`.

`serde_derive` depends on `serde_derive_internals` so it should have access to the `StrBoolInt` in `serde_derive_internals`. We can add a dependency of `serde_derive_internals` on `quote` if necessary and provide the impl there.

> The second `StrBoolInt` is stuffed into `fragment.rs`. Should I use a new file or is that ok?

It should be in `serde_derive_internals`.

> I decided to only allow non-string renames for internally and adjacently tagged enums. Correct?

Yes I agree with this choice.

> What is that `[#serde(field_identifier/variant_identifier)]` setting? It is not mentioned in the docs.

They generate a Deserialize impl for reading field names in a struct or variant names in an enum in a way that is compatible with how derive(Deserialize) behaves on structs and externally tagged enums. See the example of field_identifier in [Manually implementing Deserialize for a struct](https://serde.rs/deserialize-struct.html). This is helpful when you want custom Deserialize logic but the field names / variant names do not need to be custom.

serde/src/export.rs
```diff
@@ -38,4 +38,49 @@ mod string {
...
+    #[cfg(any(feature = "std", feature = "alloc"))]
+    pub fn from_int(i: u64) -> Vec<u8> {
+        use lib::fmt::Write;
+        let mut buf = String::with_capacity(20);
+        write!(&mut buf, "{}", i).ok();
+        buf.into_bytes()
+    }
+
+    #[cfg(not(any(feature = "std", feature = "alloc")))]
+    pub fn from_int(i: u64) -> [u8; 20] {
```
dtolnay: This seems fine. Nicely done.

serde_derive/src/fragment.rs
dtolnay: I am okay with the behavior as you implemented it.

test_suite/tests/test_macros.rs
dtolnay: That's fine.

Ten0:
What happened to this ?

PR #978, opened 2017-07-06
Update to proc-macro2

dtolnay:
:tired_face: @alexcrichton @mystor

PR #980, opened 2017-07-06
Update to the proc_macro2 crate and rewritten `syn`

alexcrichton:


alexcrichton:
Note that this is a PR against the `proc-macro2` branch, not `master`. It's intended that this updates https://github.com/serde-rs/serde/pull/978.

alexcrichton:
cc @mystor as well

Some things I've learned as part of this transition:

* Overall the transition was pretty smooth, API-wise nothing was too bad.
* `Delimited` kinda sucks. It doesn't have all the `Vec` APIs so using it can be sort of verbose.
* The biggest footgun is having a `syn` AST that's in an inconsistent state. For example you could have a `WhereClause` with a bunch of where predicates *without* a `where` token. This happened a ton here and I had to fix up a bunch of places.

Curious what y'all think of those points!

mystor:
I haven't had the time to read the actual patch yet, so I'm just commenting based on your comment. 

For point 2, I think in many cases we could probably get away with the `into_vec` and `from(vec)` methods, as usually `rustc` won't be producing diagnostics pointing to delimiters. 

That being said, we should definitely persue better ergonomics for `Delimited`. The obvious solution is to implement more of `Vec`'s API, which is an unfortunate maintenance burden. My other idea would be to store the delimiters and the AST nodes in two separate Vec objects internally, so that we can provide a `&mut` to the AST node `Vec`. When tokenising the AST nodes, we could pull tokens from the delimiter vec until we run out of tokens, and then we'd start generating them from thin air with dummy spans. That way we provide the ergonomics of ignoring the tokens and using a `Vec<AST Node>` directly unless you need to worry about delimiter spans.

I filed https://github.com/dtolnay/syn/issues/177 for point 3.

serde_derive/src/de.rs
```diff
@@ -163,7 +173,10 @@ fn needs_deserialize_bound(attrs: &attr::Field) -> bool {
 // Fields with a `default` attribute (not `default=...`), and fields with a
 // `skip_deserializing` attribute that do not also have `default=...`.
 fn requires_default(attrs: &attr::Field) -> bool {
-    attrs.default() == &attr::Default::Default
+    match *attrs.default() {
+        attr::Default::Default => true,
```
mystor: The extra-traits change feels unrelated to me, but I suppose you're doing it to try to save on compile time? It feels a bit like it should be done in a separate patch to me.

serde_derive_internals/Cargo.toml
```diff
@@ -12,8 +12,16 @@ readme = "README.md"
...
+[dependencies.syn]
+git = 'https://github.com/dtolnay/syn'
+default-features = false
+features = ["parsing", "extra-traits"]
 
 [badges]
 travis-ci = { repository = "serde-rs/serde" }
+
+[features]
+extra-traits = ["syn/extra-traits"]
```
mystor: extra-traits is listed twice, both as a required and optional feature. I think it should only be listed in one of those two places.

dtolnay:
Great, this is enough to convince me that things are still usable. I don't think we need to polish this further right now.

I responded in https://github.com/dtolnay/syn/issues/123#issuecomment-313664408 but I think we can go ahead and publish the current syn/synom/quote under different names and address some of the points here over the next couple weeks.

alexcrichton:
@dtolnay do you have a preference on coordination of merging this? Would you prefer to only merge the update into `serde_derive` when syn/synom/etc are published? (that seems like a reasonable constraint to me)

Would you otherwise want to merge this into the `proc-macro2` branch here for continued development of the proc-macro2 integration?

alexcrichton:
Apart from https://github.com/dtolnay/syn/issues/177 I believe I've addressed the issues mentioned here

PR #986, opened 2017-07-10
refactor: Implement Deserialize of wrapper types with a macro

Marwes:
Backported commit from the derive_seed branch which removes a bit of boiler plate (which helps when porting this module to the seeded variants)


Marwes:
Didn't realize this broke 1.13. I changed the macro to not have an empty where clause and reverted the impl for `Cell` since I accidentally removed the `Copy` bound. Travis is still broken though but that does not seem to be due to this change.

PR #995, opened 2017-07-21
Workaround for "no bin target named serde_derive_tests_no_std"

### LINKED ISSUE #988: No bin target named `serde_derive_tests_no_std`
In the test_suite directory, `cargo clippy` fails with this message.

```
error: no bin target named `serde_derive_tests_no_std`
```

As though it sees that bin target in test_suite/no_std/Cargo.toml but tries to build that bin target against test_suite/Cargo.toml.

@oli-obk I don't think we changed anything significant on our end so this seems like a Clippy or Cargo regression.

dtolnay:
Fixes #988.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,5 +5,4 @@ members = [
     "serde_derive_internals",
     "serde_test",
     "test_suite",
-    "test_suite/no_std",
 ]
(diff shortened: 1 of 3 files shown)


dtolnay:
AppVeyor is borked.
http://help.appveyor.com/discussions/problems/7115-curl-cant-download-files-through-ssl

PR #1001, opened 2017-07-26
Inherit the visibility of remote struct definition

### LINKED ISSUE #1000: Remote type visibility
The `#[serde(remote)]` attribute generates `deserialize` and `serialize` methods in an impl for the local type but they're private - this prevents a remote "Def" type in a module from being exported outside of it. There are workarounds (newtype or redefine proxy functions) but would there be any disadvantage to just marking these `pub`?

Or maybe traits could be used?

```rust
trait DeserializeInto<'de> {
    type Output;

    fn deserialize_into<D: Deserializer<'de>>(d: D) -> Result<Self::Output, D::Error>;
}

trait SerializeFrom {
    type Input;

    fn serialize_from<S: Serializer>(value: &Self::Input, s: S) -> Result<S::Ok, S::Error>;
}
```

(something like that, could be named `RemoteDeserialize` instead and I didn't really consider type parameters vs associates types)

dtolnay:
For remote derive this will give the derived methods the same visibility as the definition.

```rust
#[serde(remote = "...")]
struct Def; // Def::serialize and deserialize are private to this module

#[serde(remote = "...")]
pub struct Def; // Def::serialize and deserialize are public

#[serde(remote = "...")]
pub(crate) struct Def; // Def::serialize and deserialize are pub(crate)
```

Fixes #1000. @arcnmx

### DIFF
--- a/serde_derive/src/de.rs
+++ b/serde_derive/src/de.rs
@@ -28,9 +28,10 @@ pub fn expand_derive_deserialize(input: &syn::DeriveInput) -> Result<Tokens, Str
     let body = Stmts(deserialize_body(&cont, &params));
 
     let impl_block = if let Some(remote) = cont.attrs.remote() {
+        let vis = &input.vis;
         quote! {
             impl #de_impl_generics #ident #ty_generics #where_clause {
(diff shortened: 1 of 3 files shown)


arcnmx:
Such a prompt response! Looks like a good solution to me.

PR #1008, opened 2017-08-03
Add #[serde(empty_struct)]

spinda:
This allows forcing unit variants and structs to be (de)serialized as an empty struct, so for example you get `{}` instead of `null` in JSON.

Sample usage:

```rust
#[serde(empty_struct)]
struct Empty;
```


spinda:
It would, but if the type is part of a public API then it's an implementation detail unnecessarily exposed to the rest of the world, making things slightly more inconvenient and confusing for end users ("why do I need to put braces here, when there are no fields?").

Without this attribute in `serde_derive`, it takes quite a lot of code to implement equivalent functionality in the case of unit variants. Unit structs are easy to support on top of the code required for variants.

spinda:
That sounds like a nice, general solution. I'll investigate implementing something like it.

dtolnay:
Sounds good. I filed #1013 to follow up.

Issue #1013, opened 2017-08-06
Allow de/serialize_with on enum variants


dtolnay:
This still needs to be fleshed out some more, but the idea would be to support things like empty_struct from https://github.com/serde-rs/serde/pull/1008 implemented in user code rather than in Serde.

```rust
#[derive(Serialize)]
#[serde(tag = "t", content = "c")]
enum E {
    Hello { foo: String },

    // serialize as `{ "t": "World", "c": {} }`
    #[serde(serialize_with = "empty_struct")]
    World,
}

fn empty_struct<signature TBD> { /* ... */ }
```

dtolnay:
@spinda may be interested in working on a design.

spinda:
I would like to propose the following ("design A"):

```rust
#[derive(Serialize)]
#[serde(tag = "t", content = "c")]
enum E {
    // normal variant (de)serialization
    Hello { world: String },

    // (de)serialize_with can be used to add special behavior to variants;
    // for tuple structs, each inner value is passed in order to the serialize
    // function, and the deserialize function produces a tuple (see below)
    #[serde(serialize_with = "my_special_serializer")]
    #[serde(deserialize_with = "my_special_deserializer")]
    Quux(String, u8),

    // struct variants are also supported, and are treated in the same way as
    // tuple variants, with inner values ordered by how they appear in the
    // definition
    #[serde(serialize_with = "my_special_serializer")]
    #[serde(deserialize_with = "my_special_deserializer")]
    Xyzzy {
        foo: String,
        bar: u8,
    },

    // for unit variants, the serialize function is passed only the serializer,
    // and the deserialize function is expected to produce ()
    #[serde(serialize_with = "serialize_empty_struct")]
    #[serde(deserialize_with = "deserialize_empty_struct")]
    Baz,
}

fn my_special_serializer<S>(foo: &String, bar: &u8, serializer: S) -> Result<S::Ok, S::Error>
    where S: Serializer,
{
    unimplemented!()
}

fn my_special_deserializer<'de, D>(deserializer: D) -> Result<(String, u8), D::Error>
    where D: Deserializer<'de>,
{
    unimplemented!()
}

fn serialize_empty_struct<S>(serializer: S) -> Result<S::Ok, S::Error>
    where S: Serializer,
{
    unimplemented!()
}

fn deserialize_empty_struct<'de, D>(deserializer: D) -> Result<(), D::Error>
    where D: Deserializer<'de>,
{
    unimplemented!()
}
```

[...]

Issue #631, opened 2016-11-28
Support big arrays


dtolnay:
Servo does this to support big arrays in one of their proc macros:

https://github.com/servo/heapsize/blob/44e86d6d48a09c9cbc30a122bc8725b188d017b2/derive/lib.rs#L36-L41

Let's do the same but only if the size of the array exceeds our biggest builtin impl.

Thanks @nox.

dtolnay:
One relatively easy workaround for serialization is coercing to a slice:

```rust
struct S {
    #[serde(serialize_with = "<[_]>::serialize")]
    arr: [u8; 256],
}
```

Deserialization is still annoying I think.

sdleffler:
Hey folks, this feature is important to me because I'd like to be able to serialize a 512-bit hash (so, 64 bytes) and because the serde impls necessarily only go up to `[u8; 32]` I cannot serialize a `[u8; 64]`.

As workarounds I'm considering using `[[u8; 32]; 2]`, `GenericArray` or just lazily using a `Box<[u8]>`. I'm piqued by the idea of the workaround shown above - @dtolnay did you ever find a deserialization workaround?

Would it be okay to add impls up to 64? Or is there some reason that hasn't been done?

clarcharr:
Perhaps we should add impls for the sizes that `arrayvec` provides? See:

```
impl<T> Array for [T; 40]
impl<T> Array for [T; 48]
impl<T> Array for [T; 50]
impl<T> Array for [T; 56]
impl<T> Array for [T; 64]
impl<T> Array for [T; 72]
impl<T> Array for [T; 96]
impl<T> Array for [T; 100]
impl<T> Array for [T; 128]
impl<T> Array for [T; 160]
impl<T> Array for [T; 192]
impl<T> Array for [T; 200]
impl<T> Array for [T; 224]
impl<T> Array for [T; 256]
impl<T> Array for [T; 384]
impl<T> Array for [T; 512]
impl<T> Array for [T; 768]
impl<T> Array for [T; 1024]
impl<T> Array for [T; 2048]
impl<T> Array for [T; 4096]
impl<T> Array for [T; 8192]
impl<T> Array for [T; 16384]
impl<T> Array for [T; 32768]
impl<T> Array for [T; 65536]
```

dtolnay:
@clarcharr I would prefer to stick with what the standard library does, which is [0 to 32 (inclusive)](https://doc.rust-lang.org/std/primitive.array.html).

[...]

PR #1015, opened 2017-08-08
implement (de)serialize_with for variants

spinda:
Work in progress for #1013. So far I've implemented the `serialize_with` half. Any early feedback would be appreciated.

### DIFF
--- a/test_suite/tests/test_annotations.rs
+++ b/test_suite/tests/test_annotations.rs
@@ -811,6 +811,74 @@ fn test_serialize_with_enum() {
     );
 }
 
+#[derive(Debug, PartialEq, Serialize)]
+enum SerializeWithVariant {
+    #[serde(serialize_with="serialize_unit_variant_as_i8")]
+    Unit,
+
+    #[serde(serialize_with="SerializeWith::serialize_with")]
+    Newtype(i32),
+
+    #[serde(serialize_with="serialize_some_other_variant")]
+    Tuple(String, u8),
+
+    #[serde(serialize_with="serialize_some_other_variant")]
+    Struct {
+        f1: String,
+        f2: u8,
+    },
+}
+
+fn serialize_unit_variant_as_i8<S>(serializer: S) -> Result<S::Ok, S::Error>
+    where S: Serializer,
+{
+    serializer.serialize_i8(0)
+}
+
+fn serialize_some_other_variant<S>(f1: &String,
+                                   f2: &u8,
+                                   serializer: S)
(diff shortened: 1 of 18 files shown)


test_suite/tests/test_annotations.rs
```diff
@@ -811,6 +811,77 @@ fn test_serialize_with_enum() {
...
+    #[serde(serialize_with="SerializeWith::serialize_with")]
+    Newtype(i32),
+
+    #[serde(serialize_with="serialize_some_other_variant")]
+    Tuple(String, #[serde(skip_serializing_if="is_zero")] u8),
+
+    #[serde(serialize_with="serialize_some_other_variant")]
+    Struct {
+        f1: String,
+        #[serde(skip_serializing_if="is_zero")]
```
dtolnay: I would prefer not to support skip attributes inside of a variant with serialize_with until somebody has a compelling use case for it. For now the derive should panic so it fails to compile. I feel like it is just as convenient and more obvious for this logic to be encompassed inside the serialize_with function:

```rust
fn serialize_some_other_variant<S>(f1: &String,
                                   f2: &u8,
                                   serializer: S)
                                   -> Result<S::Ok, S::Error>
    where S: Serializer
{
    if is_zero(*f2) {
        /* ... */
    } else {
        /* ... */
    }
}
```

spinda: Sure, I'll rip that part out. I was split on this and ended up erring on the side of compatibility with existing features.

spinda:
@dtolnay That should wrap up the implementation (no longer WIP).

oli-obk:
The clippy thing is unfortunate. @dtolnay do you think it would make sense to allow derefs in `serialize_with` and `deserialize_with`? We'd have to generate an intermediate function/closure at the expanded call site, but that should be doable.

dtolnay:
We already allow derefs in serialize_with and deserialize_with. That is why https://github.com/serde-rs/serde/issues/631#issuecomment-279271870 works, for example. The tests added in this PR can be changed as clippy says: use &str instead of &String.

The generated code is doing the equivalent of this, which works:

```rust
struct __SerializeWith<'__a> {
    values: (&'__a String, &'__a u8),
}

fn main() {
    let sw = __SerializeWith { values: (&"".to_owned(), &0) };
    serialize_some_other_variant(sw.values.0, sw.values.1);
}

fn serialize_some_other_variant(_: &str, _: &u8) {}
```

spinda:
Oh, so you can. I didn't realize that. I've pushed a fix.

PR #1018, opened 2017-08-16
Fix warning and link error in compiletest

spinda:
This first fixes a deprecation warning in compiletest: compiletest::default_config() is being replaced by compiletest::Config::default().

Secondly, all the compiletests were failing due to rustc not being able to locate the serde_derive crate. The [compiletest README][0] recommends using config.link_deps() to ensure dependency paths are properly set. Adding this to serde's compiletest.rs causes the compiletests to run properly.

[0]: https://github.com/laumann/compiletest-rs#to-use-in-your-project

### DIFF
--- a/test_suite/tests/compiletest.rs
+++ b/test_suite/tests/compiletest.rs
@@ -13,7 +13,7 @@ extern crate compiletest_rs as compiletest;
 use std::env;
 
 fn run_mode(mode: &'static str) {
-    let mut config = compiletest::default_config();
+    let mut config = compiletest::Config::default();
 
     config.mode = mode.parse().expect("invalid mode");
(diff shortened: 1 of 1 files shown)


dtolnay:
The Travis build passed before this change and fails with this change. Could you take a look?

```
can't find crate for `serde_derive` [E0463]
```

spinda:
After staring at this harder, I've realized that the reason the doctests were failing locally was that I hadn't run `cargo build` inside `test_suite/deps`, as `travis.sh` does. Using the `link_deps` method fails on Travis as `travis.sh` compiles multiple copies of the `serde` crait (to test different feature configurations).

With https://github.com/serde-rs/serde/commit/0681cd50032a548f1ee060110c957997769a14b7 having committed the deprecation fix, there isn't anything left to contribute here, so I'll just close this.

PR #1022, opened 2017-08-19
Inform serializers about skipped fields.

### LINKED ISSUE #960: Pass field index to SerializeStruct::serialize_field

sfackler:
Closes #960.

### DIFF
--- a/serde/src/ser/mod.rs
+++ b/serde/src/ser/mod.rs
@@ -1728,6 +1728,7 @@ pub trait SerializeStruct {
         T: Serialize;
 
     /// Indicate that a struct field has been skipped.
+    #[inline]
     fn skip_field(&mut self, key: &'static str) -> Result<(), Self::Error> {
         let _ = key;
         Ok(())
@@ -1779,6 +1780,7 @@ pub trait SerializeStructVariant {
         T: Serialize;
 
     /// Indicate that a struct variant field has been skipped.
+    #[inline]
     fn skip_field(&mut self, key: &'static str) -> Result<(), Self::Error> {
         let _ = key;
         Ok(())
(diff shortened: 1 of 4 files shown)


dtolnay:
I would like to see a cbor PR working with this before I would merge it. I am on board with this idea but it would be tough to undo later if it turns out we are missing something...

Have you looked into how this interacts with unconditionally skipped fields?

```rust
struct Sfackler {
    a: u8,
    #[serde(skip)]
    b: u8,
    c: u8,
}
```

It looks like Serialize will call serialize_field("a"), skip_field("b"), serialize_field("c") so presumably "c" would be serialized with index 2, but then Deserialize invokes deserialize_struct("Sfackler", &["a", "c"]) in which there is no field with index 2.

sfackler:
Yeah, the behavior for unconditionally skipped fields is wrong.

I'm not sure this is sufficient to handle compact struct field names - we'd also need to support numeric deserialization of field identifiers.

dtolnay:
We used to have numeric deserialization of field identifiers but got rid of it because nobody ever used it. I would be okay putting it back.

sfackler:
I've fixed the handling of skipped fields and enabled identifier deserialization from u32.

Here's an example of use for CBOR packed serialization: https://github.com/sfackler/cbor/commit/d2409da8166866f514eec26b0dbcf1ce7631e777

dtolnay:
Thanks for persisting with this :heart:. I will try to squeeze in some of the other reviews tonight and get this released no later than tomorrow night.

sfackler:
Thanks!

PR #1043, opened 2017-09-05
SCREAMING-KEBAB-CASE support

greyblake:
Hey! Thanks for the nice library =)

Here is a small patch for support of `SCREAMING-KEBAB-CASE`, that is used in some APIs. Name is inspired by already existing `SCREAMING_SNAKE_CASE` and `kebab-case`.

### DIFF
--- a/serde_derive_internals/src/case.rs
+++ b/serde_derive_internals/src/case.rs
@@ -27,6 +27,8 @@ pub enum RenameRule {
     ScreamingSnakeCase,
     /// Rename direct children to "kebab-case" style.
     KebabCase,
+    /// Rename direct children to "SCREAMING-KEBAB-CASE" style.
+    ScreamingKebabCase
 }
 
 impl RenameRule {
@@ -47,6 +49,7 @@ impl RenameRule {
             }
             ScreamingSnakeCase => SnakeCase.apply_to_variant(variant).to_ascii_uppercase(),
             KebabCase => SnakeCase.apply_to_variant(variant).replace('_', "-"),
+            ScreamingKebabCase => ScreamingSnakeCase.apply_to_variant(variant).replace('_', "-")
         }
     }
 
@@ -74,6 +77,7 @@ impl RenameRule {
             }
             ScreamingSnakeCase => field.to_ascii_uppercase(),
             KebabCase => field.replace('_', "-"),
+            ScreamingKebabCase => ScreamingSnakeCase.apply_to_field(field).replace('_', "-")
         }
     }
 }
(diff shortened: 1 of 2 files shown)


greyblake:
@dtolnay Hey! The one that I am interesting in is [Bittrex API](https://bittrex.com/home/api).
Example: https://bittrex.com/api/v1.1/public/getmarketsummaries    
I agree, it may look weird, but they use market names like `BTC-LTC`, `BTC-ETH`, etc. So I thought it would be good to have something like `SCREAMING-KEBAB-CASE` to parse market names into enum automatically.

serde_derive_internals/src/case.rs
```diff
@@ -27,6 +27,8 @@ pub enum RenameRule {
     ScreamingSnakeCase,
     /// Rename direct children to "kebab-case" style.
     KebabCase,
+    /// Rename direct children to "SCREAMING-KEBAB-CASE" style.
+    ScreamingKebabCase
```
clarcharr: rustfmt: comma here

serde_derive_internals/src/case.rs
```diff
@@ -47,6 +49,7 @@ impl RenameRule {
             }
             ScreamingSnakeCase => SnakeCase.apply_to_variant(variant).to_ascii_uppercase(),
             KebabCase => SnakeCase.apply_to_variant(variant).replace('_', "-"),
+            ScreamingKebabCase => ScreamingSnakeCase.apply_to_variant(variant).replace('_', "-")
```
clarcharr: also comma here

serde_derive_internals/src/case.rs
```diff
@@ -74,6 +77,7 @@ impl RenameRule {
             }
             ScreamingSnakeCase => field.to_ascii_uppercase(),
             KebabCase => field.replace('_', "-"),
+            ScreamingKebabCase => ScreamingSnakeCase.apply_to_field(field).replace('_', "-")
```
clarcharr: also comma here

dtolnay:
Thanks for the review @clarcharr. I removed the comments because I don't care about formatting in PRs -- see https://github.com/serde-rs/serde/issues/438#issuecomment-279338348 for my philosophy.

@greyblake those are values in the JSON, not keys, so to clarify are you deserializing into something like this?

```rust
#[derive(Deserialize)]
#[serde(rename_all = "PascalCase")]
struct Summary {
    market_name: MarketName,
    /* other fields */
}

#[derive(Deserialize)]
#[serde(rename_all = "SCREAMING-KEBAB-CASE")]
enum MarketName {
    BtcLtc,
    BtcEth,
}
```

greyblake:
@dtolnay Yes, those are values. The example you provided is exactly what I do in my code base.

Issue #1020, opened 2017-08-17
`Wrapping` support


dhardy:
Please consider adding support for [`std::num::Wrapping`](https://doc.rust-lang.org/std/num/struct.Wrapping.html).

Use-case: using `#[derive(Serialize, Deserialize)]` on RNGs like [`ChaChaRng`](https://github.com/rust-lang-nursery/rand/blob/master/src/chacha.rs#L30).

Of course, a work-around is possible, but `derive` won't work without support for `Wrapping`.

dtolnay:
Added in #1072, released in 1.0.17.

PR #1072, opened 2017-10-31
Add std::num::Wrapping impl

LinearZoetrope:
Addresses #1020 

This would really help with cutting down the code bloat in:

https://github.com/rust-lang-nursery/rand/pull/189

By making more of the structs `derive`able. We'd still be blocked waiting for blanket impls that are generic over array/tuple size constants for two of the RNGs, but this would help with a lot of code.

Unfortunately, since `Wrapping` is in `std`, there's no way to do this cleanly without directly modifying `serde` or manually implementing for/working around every type that uses a `Wrapping` forever. This is an extremely small change.


serde/src/de/impls.rs
```diff
@@ -1825,3 +1825,20 @@ where
...
+impl<'de, T> Deserialize<'de> for Wrapping<T>
+where 
+    T: Deserialize<'de> 
+{
+    fn deserialize<D>(deserializer: D) -> Result<Wrapping<T>, D::Error> 
+        where D: Deserializer<'de>
+    {
+        Deserialize::deserialize(deserializer).map(Wrapping)
+    }
+}
```
oli-obk: missing newline at end of file

serde/src/de/impls.rs
```diff
@@ -1825,3 +1825,20 @@ where
...
+
+#[cfg(feature = "std")]
+impl<'de, T> Deserialize<'de> for Wrapping<T>
+where 
+    T: Deserialize<'de> 
+{
+    fn deserialize<D>(deserializer: D) -> Result<Wrapping<T>, D::Error> 
+        where D: Deserializer<'de>
+    {
+        Deserialize::deserialize(deserializer).map(Wrapping)
```
oli-obk: Deserializing a `Wapping<u8>` from `500` will fail. Does that match your use case?

serde/src/ser/impls.rs
```diff
@@ -664,3 +664,19 @@ impl Serialize for OsString {
...
+#[cfg(feature = "std")]
+impl<T: Serialize> Serialize for Wrapping<T> {
+    #[inline]
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        self.0.serialize(serializer)
+    }
+}
```
oli-obk: also missing newline

serde/src/de/impls.rs
LinearZoetrope: Can you clarify? Do you mean that if you deserialize a number that exceeds the max for that type it won't wrap? That's fine, I didn't expect it to use the wrapping functionality at deserialization time.

LinearZoetrope:
Should I fix the newlines in a new commit, or do an amend?

LinearZoetrope:
I pushed a commit fixing the newlines, if you'd prefer it to be squashed I can do that too.

serde/src/de/impls.rs
hcpl: Then isn't this something `Wrapping(u8::deserialize(deserializer)?)` can already accomplish without this impl?

oli-obk: Yea that's what I meant. Just wanted to make sure. It's backwards compatible to change later if we want to.

oli-obk:
> I pushed a commit fixing the newlines, if you'd prefer it to be squashed I can do that too.

fine with me

serde/src/de/impls.rs
LinearZoetrope: @hcpl If you just want to (de)serialize a `Wrapping` that's fine, but it becomes somewhat more problematic if you want to `Derive` a struct containing one. I guess you could decorate every field with a wrapping with a `with` attribute, but it's overall cleaner to do this.

hcpl: So there are 2 desired behaviours for `Wrapping`: accept and produce values in the certain range, or only do so for producing values while accepting anything.

Logically speaking, if we were to provide wrapping behaviour in `impl Deserialize`, we would have to do custom number parsing to wrap _any_ possible number for consistency's sake.

Also, `Wrapping<T>` doesn't have any bounds on `T`, so we end up having to rely on the unstable specialization feature for numeric types.

So yeah, easier to leave things this way.

dtolnay:
I added some unit tests in 9b08915a18086f6dc7df850c421f202cf6a00d8c and released this in Serde 1.0.17.

PR #1075, opened 2017-10-31
Clarify the README example for local builds

### LINKED ISSUE #1074: README instructions result in compile error
I'm trying to compile an empty rust project (just a Cargo.toml and a main.rs that contains nothing more than the code in the README) and I'm getting a compile error.

llamaforge:mmdb michaeldehaan$ cargo build
    Updating registry `https://github.com/rust-lang/crates.io-index`
 Downloading serde v1.0.17
   Compiling serde v1.0.17
   Compiling mmdb v0.1.0 (file:///Users/michaeldehaan/Devel/mmdb)
error[E0463]: can't find crate for `serde_derive`
 --> src/main.rs:2:1
  |
2 | extern crate serde_derive;
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^ can't find crate

error: aborting due to previous error(s)

hcpl:
Fixes #1074.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,9 +20,22 @@ You may be looking for:
 
 ## Serde in action
 
+**Cargo.toml**:
+
+```toml
+[dependencies]
+serde = "1.0"
+serde_derive = "1.0"
+serde_json = "1.0"
+```
+
+Click on the "Run" button to have a quick taste:
+
 <a href="http://play.integer32.com/?gist=9003c5b88c1f4989941925d7190c6eec" target="_blank">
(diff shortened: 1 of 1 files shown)


dtolnay:
I combined this with 7f9ba155cb7c57a2ec279d385e15342e1f2edfdf to explain what each dependency is for. That way people are less likely to blindly copy the serde_json dependency into their project that does not use JSON. And I put that stuff behind a \<details> so that the real code doesn't end up too far down the page. Let me know what you think.

hcpl:
Good idea, looks even better to me!

PR #1080, opened 2017-11-05
serde_test requires serde 1.0.16 to work

xfix:
This is due to implementing is_human_readable which was added in serde 1.0.16.


xfix:
I'm developing a crate that depends on `serde`, and on `serde_test` in tests. `serde` version requirement is set to `1.0.0` as I'm fairly convinced that a crate should work with such a version. However, the `serde_test` version was upgraded to take advantage of possible bugfixes that could cause test to incorrectly pass (it can happen) and it was a development dependency anyway.

xfix:
Anyway, 1.0.16 version requirement is in place. Should have checked docs.rs instead of git history (which is quite misleading).

dtolnay:
Makes sense. Thank you.

PR #1086, opened 2017-11-07
Print the contents of the expected token when a serialize assert fails

Marwes:
Before
```
expected Token::Str but serialized as Str(“F9168C5E-CEB2-4FAA-B6BF-329BF39FA1E4")
```
After
```
expected Token::Str(“f9168c5e-ceb2-4faa-b6bf-329bf39fa1e4”) but serialized as Str(“F9168C5E-CEB2-4FAA-B6BF-329BF39FA1E4")
```


Marwes:
` serde_test/src/de.rs` handles this is a much simpler way but unfortunately I get lifetime errors from some uses when I tried it so instead I needed to do this.

Issue #855, opened 2017-04-08
Deserialize into a pre-existing struct


dtolnay:
Example from @burntsushi - when iterating over rows of a CSV file, we would like to be able to reuse the same five String buffers on every row.

```rust
#[derive(Debug, Deserialize, PartialEq)]
struct MBTARow {
    trip_id: String,
    arrival_time: String,
    departure_time: String,
    stop_id: String,
    stop_sequence: i32,
    stop_headsign: String,
    pickup_type: i32,
    drop_off_type: i32,
    timepoint: i32,
}
```

The use case is similar to [`Clone::clone_from`](https://doc.rust-lang.org/nightly/core/clone/trait.Clone.html#method.clone_from).

dtolnay:
The analogous approach would be something like:

```rust
pub trait Deserialize<'de>: Sized {
    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
        where D: Deserializer<'de>;

    fn deserialize_from<D>(&mut self, deserializer: D) -> Result<(), D::Error>
        where D: Deserializer<'de>
    {
        *self = Deserialize::deserialize(deserializer)?;
        Ok(())
    }
}
```

dtolnay:
@nox mentioned being interested in working on this.

Gankro:
Note the semantics we'd like in webrender is that it blindly clobbers the source struct. On failure it's in a partially overwritten (but memory-safe) state. We want this because serde errors are hard aborts for us (all sources are trusted), and we want to minimize work.

Not sure if that matches what y'all want.

dtolnay:
Yes, partially overwritten but memory-safe state is acceptable.

The important thing is that this is about reusing allocations, not merging one deserialized struct into another. So if you have:

```rust
struct S {
    a: Option<T>,
    b: Option<T>,
}
```

Deserializing `{"a": ..., "b": ...}` followed by deserializing `{"a": ...}` over top of the same data needs to result in b=None, not retaining its previous value.

Gankro:
Blugh, this is getting to a bit of a clumsy mess. Just to clarify, I basically need to duplicate the entire of the deserialization subsystem with _from variants?

Like every single method on Visitor will need a _from variant? Or I'll need to make a FromVisitor trait...?

dtolnay:
@Gankro I hope not... [Here is a quick prototype](https://play.integer32.com/?gist=8bd069878ba92a508ce8d7e72233a59e).

[...]

PR #1094, opened 2017-11-16
Add and derive deserialize_from

Gankro:
The actual deserialize work is WIP, because I probably need to refactor it so it can be feature-flagged off. I think the first two commits are "good to go", though.

One thing missing from this design is deriving deserialize_from for exactly C-like enums, but that can obviously be done later.

Pushing this up now because my brain is melting trying to do the non-C-like enum work I need for webrender, and so I'm gonna take a bit of a "break" by dealing with review complaints.

### DIFF
--- a/serde/src/de/impls.rs
+++ b/serde/src/de/impls.rs
@@ -15,6 +15,7 @@ use de::{Deserialize, Deserializer, EnumAccess, Error, SeqAccess, Unexpected, Va
 use de::MapAccess;
 
 use de::from_primitive::FromPrimitive;
+use private::de::DeserializeFromSeed;
 
 #[cfg(any(feature = "std", feature = "alloc"))]
 use private::de::size_hint;
@@ -50,6 +51,7 @@ impl<'de> Deserialize<'de> for () {
 ////////////////////////////////////////////////////////////////////////////////
 
 struct BoolVisitor;
+
 
 impl<'de> Visitor<'de> for BoolVisitor {
     type Value = bool;
@@ -210,6 +212,8 @@ impl<'de> Deserialize<'de> for char {
 
 #[cfg(any(feature = "std", feature = "alloc"))]
 struct StringVisitor;
+#[cfg(any(feature = "std", feature = "alloc"))]
+struct StringFromVisitor<'a>(&'a mut String);
 
 #[cfg(any(feature = "std", feature = "alloc"))]
 impl<'de> Visitor<'de> for StringVisitor {
@@ -255,12 +259,72 @@ impl<'de> Visitor<'de> for StringVisitor {
 }
 
 #[cfg(any(feature = "std", feature = "alloc"))]
+impl<'a, 'de> Visitor<'de> for StringFromVisitor<'a> {
+    type Value = ();
+
+    fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+        formatter.write_str("a string")
+    }
+
+    fn visit_str<E>(self, v: &str) -> Result<(), E>
+    where
+        E: Error,
+    {
+        self.0.clear();
+        self.0.push_str(v);
+        Ok(())
+    }
+
+    fn visit_string<E>(self, v: String) -> Result<(), E>
+    where
+        E: Error,
+    {
+        *self.0 = v;
+        Ok(())
+    }
+
+    fn visit_bytes<E>(self, v: &[u8]) -> Result<(), E>
+    where
+        E: Error,
+    {
+        match str::from_utf8(v) {
+            Ok(s) => {
+                self.0.clear();
+                self.0.push_str(s);
+                Ok(())
+            }
+            Err(_) => Err(Error::invalid_value(Unexpected::Bytes(v), &self)),
+        }
+    }
+
+    fn visit_byte_buf<E>(self, v: Vec<u8>) -> Result<(), E>
+    where
+        E: Error,
+    {
+        match String::from_utf8(v) {
+            Ok(s) => {
+                *self.0 = s;
+                Ok(())
+            }
+            Err(e) => Err(Error::invalid_value(Unexpected::Bytes(&e.into_bytes()), &self),),
+        }
+    }
+}
+
+#[cfg(any(feature = "std", feature = "alloc"))]
 impl<'de> Deserialize<'de> for String {
     fn deserialize<D>(deserializer: D) -> Result<String, D::Error>
     where
         D: Deserializer<'de>,
     {
         deserializer.deserialize_string(StringVisitor)
+    }
+
+    fn deserialize_from<D>(&mut self, deserializer: D) -> Result<(), D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        deserializer.deserialize_string(StringFromVisitor(self))
     }
 }
 
@@ -421,6 +485,7 @@ forwarded_impl!((), Box<CStr>, CString::into_boxed_c_str);
 struct OptionVisitor<T> {
     marker: PhantomData<T>,
 }
+struct OptionFromVisitor<'a, T: 'a>(&'a mut Option<T>);
 
 impl<'de, T> Visitor<'de> for OptionVisitor<T>
 where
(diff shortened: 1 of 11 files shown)


Gankro:
Oh also obviously BLATANTLY MISSING: TESTS. ANY TESTS.

Gankro:
Any tips on testing this kind of thing would be appreciated. My local experiments have mostly consisted of asserting that serializing and deserializing round-trips correctly.

Gankro:
Currently grinding through the serde test suite, lots of little derive bugs.

Starting to regret changing a bunch of things from `-> Body` to `-> (Option<Common>, Body, Option<FromBody>)` in order to prevent duplicating a static.

Gankro:
Pushed fixes:

* Fixed bad cfg's in builtin impl commit
* Got existing test-suite passing with derive changes

One thing I didn't realize until now is that deserialize_from kinda doesn't make any sense when doing "remote" impls?

serde/src/de/impls.rs
```diff
@@ -66,13 +68,37 @@ impl<'de> Visitor<'de> for BoolVisitor {
     }
 }
 
+struct BoolFromVisitor<'a>(&'a mut bool);
```
oli-obk: I don't think primitive types need this specialization. The pointer indirection will probably be less performant than the default implementation. I'm easily convinced otherwise by benchmarks ;)

serde/src/de/impls.rs
```diff
@@ -254,6 +391,59 @@ impl<'de> Visitor<'de> for StringVisitor {
...
+        *self.0 = v;
+        Ok(())
+    }
+
+    fn visit_bytes<E>(self, v: &[u8]) -> Result<(), E>
+    where
+        E: Error,
+    {
+        match str::from_utf8(v) {
+            Ok(s) => {
```
oli-obk: forward to `visit_str` instead of repeating the code

serde/src/de/impls.rs
```diff
@@ -457,6 +655,45 @@ where
...
+        *self.0 = None;
+        Ok(())
+    }
+
+    #[inline]
+    fn visit_some<D>(self, deserializer: D) -> Result<(), D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        // Some enum is opaque, can't do anything good here :(
```
oli-obk: can't you do

```rust
match *self.0 {
    Some(ref mut data) => data.deserialize_from(deserializer),
    None => Some(T::deserialize(deserializer)?),
}
```
```

File: serde/src/de/impls.rs [hunk 4]
```diff
@@ -554,63 +800,117 @@ macro_rules! seq_impl {
...
+                    fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+                        formatter.write_str("a sequence")
+                    }
+
+                    #[inline]
+                    fn visit_seq<A>(mut self, mut $access: A) -> Result<(), A::Error>
+                    where
+                        A: SeqAccess<'de>,
+                    {
+                        $clear(&mut self.0);
```
oli-obk: Isn't the real optimization here to not clear, but overwrite all already existing elements, if the new sequence has more elements, append, if it has less, drop the leftover ones?

dtolnay:
I have fallen behind on PRs so I haven't had a chance to review the implementation yet, but -- I would be interested in exploring some ways to modularize the serde and serde_derive crates better so that large but uncommonly required features like this one can be delivered in their own crate. The [`serde_state`](https://github.com/Marwes/serde_state) / [`serde_derive_state`](https://crates.io/crates/serde_derive_state) feature is in a similar situation where it seems somewhat overwhelming to maintain as part of Serde but the community that wants the feature is willing to maintain it on the side.

Maybe this means factoring a lot of the serde_derive code into a separate pre-1.0 helper crate that is depended on by serde_derive, serde_derive_state, and a deserialize_from derive crate.

Would you be interested in figuring out a structure that would work for deserialize_from as part of getting this integrated? Otherwise would it be possible to provide deserialize_from in its own crate somehow, even if it means some code duplication with serde_derive?

Gankro:
I don't really see how deserialize_from could be seperated out, realisitically? It's part of the trait definition.

serde/src/de/impls.rs
Gankro: It's been a while since I implemented this, but I seem to recall getting wins from doing this. Gonna start profiling tomorrow, will find number then.

serde/src/de/impls.rs
Gankro: The deserialize() impl doesn't, and this code seems clearer without it. (happy to change if you disagree)

serde/src/de/impls.rs
Gankro: Good point, but hard to say if this is justified. Very workload-specific.

serde/src/de/impls.rs
Gankro: You can't do this for BTreeSet and BinaryHeap, so I just did the generically correct thing. Seems like it could be left as future work to split this up and further optimize it?

oli-obk:
> I don't really see how deserialize_from could be seperated out, realisitically? It's part of the trait definition.

You can make a `DeserializeFrom` trait which is implemented for all `Deserialize` types

Gankro:
How would that work? It can't be blanketed, because actual type-specific code needs to be emitted.

dtolnay:
Okay I can see how shipping this outside of the Deserialize trait definition would be problematic.

I am on board with accepting this as long as it does not affect compile times for anyone not using deserialize_from (basically the same concern as for [clone_from](https://github.com/rust-lang/rust/pull/27939)). As is, when I tried this it raises compile time for the serde crate by 13% (5.9 seconds to 6.7 seconds) which is a bit much for something that few people will use. Worse, Deserialize impls already compile very slow (https://github.com/serde-rs/json/issues/313) so something like 2x more generated code in every impl would not be nice.

Could you propose what you think is the best way to feature gate this?

Up to you how you want to get this merged. If it is easier, we can merge the serde pieces up front and follow up in one or more other PRs with the serde_derive pieces as those are written and tested.

One thing I don't like here is all the deserialize_from impls that are conceptually no different from the default implementation. For example the bool one:

```rust
fn visit_bool<E>(self, v: bool) -> Result<(), E> {
    *self.0 = v;
    Ok(())
}
```

Could you share some benchmarks that show these outperform the default deserialize_from? Is there an easy way to tell which types benefit from a custom deserialize_from? Or is a custom deserialize_from always preferable?

Gankro:
Ok finally got a chance to double check primitive stuff. Looks like, at least on webrender, removing the primitive impls is neutral, or possibly even profitable (the deserializer went from 7.1% to 6.6% of backend time; probably noise).

I'll take them out, then.

Gankro:
Ok I refactored out all the deserialize_from derivation code out so it could be feature'd out. Ultimately we weren't sharing *a ton* of code, so it's not a huge loss. Writing it that way helped me incrementalize the development process for myself better, but that doesn't matter now.

I also did a kinda naive addition of deserialize_from to the test suite.

Gankro:
Webrender, and therefore gecko, will likely be moving forward with a soft fork of serde in the next few days with these patches, plus the additional patch found on https://github.com/Gankro/serde/tree/deserialize_from_enums3

Even if you aren't ready to make a call on merging this, we'd certainly appreciate more thorough review. :)

oli-obk:
I'd be on board with merging this PR as it is if we make the `deserialize_from` feature require nightly. Then we can play around with it a little without having to immediately stabilize it

Gankro:
That makes sense to me.

Gankro:
Wait, would that also put the trait method behind the feature?

Gankro:
Also CC https://github.com/serde-rs/serde/issues/855

oli-obk:
Hmm right, that could be messy with derives and such :/

dtolnay:
I started making progress on the review -- pushed a few fixup commits as I go. Some remaining items:

- Before merging, I would like to get a sense of what this means for the future of [`csv::Reader`](https://docs.rs/csv/1.0.0-beta.5/csv/#example-with-serde) and [`serde_json::StreamDeserializer`](https://docs.serde.rs/serde_json/struct.StreamDeserializer.html). Is there a nice way to adapt these APIs to take advantage of deserialize_from? (I have not started thinking about this; it may turn out to be an easy answer.)

- Benchmarks! It seems obvious that this would be faster in some usage patterns, but I would want to see at least one basic not-too-contrived benchmark that I can run using a real existing data format to see an improvement from this.

- Currently for every struct with named fields we emit two totally identical copies of the `__Field` enum and the `__FieldVisitor` and its `Visitor` and `Deserialize` impls, which is a ton of redundant code. These should be emitted in a higher scope and shared by `deserialize` and `deserialize_from`. It is not critical to fix this before merging; will file an issue to follow up.

Gankro:
1. I don't know what's interesting about those, I'll take a look in a bit.
2. I'll try to cut out some basic workloads. Where do you put benches?
3. Yes, I originally put in a fair bit of work trying to avoid this, but I backtracked on it when I deintegrated stuff to support cfg(). This ends up being pretty nasty to thread through the code for things like deserializing struct-like enum variants. Not clear it's worth the effort.

dtolnay:
The CSV reader and JSON stream deserializer are interesting because they are APIs designed for deserializing lots of instances of the same type that are processed one at a time -- seems like an obvious candidate for deserialize_from. CSV is why we filed #855 in the first place so we may as well be certain that we are addressing their use case in the best way.

There aren't really benchmarks of Serde by itself since nothing interesting happens without a data format, but JSON benchmarks live [here](https://github.com/serde-rs/json-benchmark) and Bincode benchmarks live [here](https://github.com/TyOverby/bincode-bench). I wouldn't bother with either of those, not looking for anything fancy just a 30 line thing I can run and see an improvement. I would do this myself but I am about to call it a night. 😴

PR #1097, opened 2017-11-21
Allow uint Unit enum variants

alex-shapiro:
Serde's `ContentDeserializer` and `ContentRefDeserializer` currently expect unit variants to be encoded as strings. This PR updates both deserializers to recognize `u8`, `u16`, `u32`, and `u64`  values as unit variants. 

This change allows `rmp_serde` to optimize unit variant encoding as the variant index (min size 1 byte), instead of its current representation as a map `{variant_idx => nil}` (min size 3 bytes).

### DIFF
--- a/serde/src/private/de.rs
+++ b/serde/src/private/de.rs
@@ -1246,6 +1246,9 @@ mod content {
                 Some(Content::Map(v)) => {
                     de::Deserializer::deserialize_any(MapDeserializer::new(v), visitor)
                 }
+                Some(Content::Seq(v)) => {
+                    de::Deserializer::deserialize_any(SeqDeserializer::new(v), visitor)
+                }
                 Some(other) => Err(de::Error::invalid_type(other.unexpected(), &"struct variant"),),
                 _ => Err(de::Error::invalid_type(de::Unexpected::UnitVariant, &"struct variant"),),
             }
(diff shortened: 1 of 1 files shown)


serde/src/private/de.rs
```diff
@@ -1100,6 +1100,10 @@ mod content {
                     (variant, Some(value))
                 }
                 s @ Content::String(_) | s @ Content::Str(_) => (s, None),
+                u @ Content::U8(_) |
+                u @ Content::U16(_) |
+                u @ Content::U32(_) |
+                u @ Content::U64(_) => (u, None),
                 other => {
                     return Err(de::Error::invalid_type(other.unexpected(), &"string or map"),);
```
hcpl: Error message now should be "string, map or unsigned integer".

serde/src/private/de.rs
```diff
@@ -1525,6 +1529,10 @@ mod content {
                     (variant, Some(value))
                 }
                 ref s @ Content::String(_) | ref s @ Content::Str(_) => (s, None),
+                ref u @ Content::U8(_) |
+                ref u @ Content::U16(_) |
+                ref u @ Content::U32(_) |
+                ref u @ Content::U64(_) => (u, None),
                 ref other => {
                     return Err(de::Error::invalid_type(other.unexpected(), &"string or map"),);
```
hcpl: Same as here: https://github.com/serde-rs/serde/pull/1097/files#r152168430.

alex-shapiro:
@hcpl fixed

dtolnay:
Closing as inactive. Thanks anyway!

PR #1106, opened 2017-12-01
Inline various deserialization helper methods.

khuey:
Inlining `cautious` helps LLVM figure out what is going on with array deserialization.  I figured we might as well do the others while we're here.

### DIFF
--- a/serde/src/private/de.rs
+++ b/serde/src/private/de.rs
@@ -198,6 +198,7 @@ pub mod size_hint {
         helper(iter.size_hint())
     }
 
+    #[inline]
     pub fn cautious(hint: Option<usize>) -> usize {
         cmp::min(hint.unwrap_or(0), 4096)
     }


serde/src/private/de.rs
```diff
@@ -191,17 +191,20 @@ where
 pub mod size_hint {
     use lib::*;
 
+    #[inline]
     pub fn from_bounds<I>(iter: &I) -> Option<usize>
```
oli-obk: I don't think placing `#[inline]` on generic methods does anything.

khuey: Yeah that would make sense, since code generation has to be pushed to the caller anyways.

serde/src/private/de.rs
```diff
@@ -191,17 +191,20 @@ where
...
         helper(iter.size_hint())
     }
 
+    #[inline]
     pub fn cautious(hint: Option<usize>) -> usize {
         cmp::min(hint.unwrap_or(0), 4096)
     }
 
+    #[inline]
     fn helper(bounds: (usize, Option<usize>)) -> Option<usize> {
```
dtolnay: I believe the point of this helper (as opposed to writing the same code in from_bounds) is to keep the logic from being instantiated for all the different `I` types used with from_bounds. Does inline affect that and is it offset by the benefit of inline here?

serde/src/private/de.rs
dtolnay: I wouldn't worry about an extra inline as we certainly do the same on lots of other generic methods. If rustc or clippy ever starts to warn, we will clean them up.

serde/src/private/de.rs
khuey: I haven't measured anything for this, like I said I was just in the neighborhood.  I'd be happy to drop it if you're concerned about it.

dtolnay: I don't have an intuition for whether inline would be better or worse here. Let's drop this one and stick to the one you measured a benefit from. Thanks!

khuey:
Updated.

PR #1111, opened 2017-12-04
allow calling a function on result when deserializing

cheme:
I have this kind of need (quick implementation in pull, might require some 'check' rules if incompatible with others derive attribute).
This is simply a container attribute to hook a function call at the end of a macro derived deserialize: 

```rust
#[serde(finish_deserialize = "finalize_point2")]
struct Point2 {
    #[serde(rename = "X")]
    x: i32,
    #[serde(rename = "Y")]
    y: i32,
}
fn finalize_point2<E>(p : &mut Point2) -> Result<(),E>
  where E : de::Error {
  if p.x == 5 {
    p.y = 8
  } else {
    p.y = 0
  }
  Ok(())
}
```

Basically running a post action on the struct after deserializing.

It may seems useless but is helpful in some cases. Basically when a struct requires a modification after deserializing and you start composing over it. My use case was a lazy initialized field (byte64 encoding of another id) that requires  to be initialized immediately when used as a Rc in another struct (deriving this other struct by macro requires that the initialization is done before the Rc creation).

This hook can also be use to insert some verification over multiple fields of a struct (without hand writing the deserialization).

On paper it does not make totally sense (checking or initialization should be done during deserialization upon an inner state), but I think it could simplify some use case (in mine I would have to define my Rc as another struct with a specific derive implementation).

### DIFF
--- a/serde_derive/src/de.rs
+++ b/serde_derive/src/de.rs
@@ -25,7 +25,7 @@ pub fn expand_derive_deserialize(input: &syn::DeriveInput) -> Result<Tokens, Str
     let params = Parameters::new(&cont);
     let (de_impl_generics, _, ty_generics, where_clause) = split_with_de_lifetime(&params);
     let dummy_const = Ident::new(format!("_IMPL_DESERIALIZE_FOR_{}", ident));
-    let body = Stmts(deserialize_body(&cont, &params));
+    let body = Stmts(finish_deserialize(&cont, &params));
     let delife = params.borrowed.de_lifetime();
 
     let impl_block = if let Some(remote) = cont.attrs.remote() {
@@ -242,6 +242,20 @@ fn deserialize_body(cont: &Container, params: &Parameters) -> Fragment {
             }
             Body::Struct(_, _) => unreachable!("checked in serde_derive_internals"),
         }
+    }
+}
+
+fn finish_deserialize(cont: &Container, params: &Parameters) -> Fragment {
+    let body = deserialize_body(&cont, &params);
+    if let Some(path) = cont.attrs.finish_deserialize() {
+        let body = Expr(body);
+        quote_block!{
+          let mut __res = try!(#body);
+          try!(#path(&mut __res));
+          Ok(__res)
+        }
+    } else {
+        body
     }
 }
 
--- a/serde_derive_internals/src/attr.rs
+++ b/serde_derive_internals/src/attr.rs
@@ -112,6 +112,7 @@ pub struct Container {
     from_type: Option<syn::Ty>,
     into_type: Option<syn::Ty>,
     remote: Option<syn::Path>,
+    finish_deserialize: Option<syn::Path>,
     identifier: Identifier,
 }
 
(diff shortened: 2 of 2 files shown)


dtolnay:
Thanks! Could you talk through some of the tradeoffs of the signature being `&mut T -> Result<(), Error>` vs `T -> Result<T, Error>`? Do you believe either one works for any use case? Do we need to support both in order to cover all the use cases? Is there a way to support both without having two different attributes? How does this interact with https://github.com/Keats/validator/issues/22 and what changes would you expect to see in the validator crate to take advantage of this new attribute?

cheme:
While testing, I realize that the signature is wrong as there is no way to return generically a Deserializer Error. 
I was expecting serde to have the equivalent of the trait `Rustc_serialize::Decoder` function `fn error(&mut self, err: &str) -> Self::Error;` for its Deserializer trait.
Therefore returning result is currently impracticable (I do not know of a way to generically instantiate deserialize error).

I think `T -> Result<T, Error>` would be fine, but considering my previous point it should simply be `T -> T`.

If the Deserializer trait would have an additional method `fn error(&mut self, err: &str) -> Self::Error`, the signature could be `T, &mut D -> Result<T, D::Error>` with D being Deserializer, but I guess the impact of changing Deserializer is big.

I did not know the validator crate, at first glance calling validate function in the call back seems easy but it involves converting the Validation error into the Deserializing crates error type and we got the same problem as describe before : we need a way to spawn a Deserializer error (plus Validator error are a bit more complex than a error spawn from a simple &str as in rustc_serialize).

dtolnay:
Errors are created using the [`serde::de::Error`](https://docs.serde.rs/serde/de/trait.Error.html) trait, for example [here](https://github.com/serde-rs/serde/blob/v1.0.23/serde/src/de/impls.rs#L1709-L1713).

Marwes:
Had an idea for a more alternative approach. It is more verbose for this use case but otherwise more flexible.

```rust
#[derive(Deserialize)]
// Generates a deserialize function instead of an implementation of `Deserilize`
// fn test_deserialize<D>(deserializer: D) -> Result<Test, D::Error> { ... }
#[serde(freestanding = "test_deserialize)]
struct Test { ... }


// Now `Deserialize` can be implemented manually on the type

impl<'de> Deserialize<'de> for Test {
    fn deserialize<D>(deserializer: D) -> Result<Test, D::Error> {
        let test = test_deserialize(deserializer)?;
        // validate
        Ok(test)
    }
}
```

I suppose it is also is a bit strange that it does not actually derive `Deserialize` but it should still work (I am pretty sure) and might be an acceptable trade off for the extra flexibility.

cheme:
@dtolnay indeed, that is the function that I needed, all seems fine with it ([eg](https://github.com/cheme/votingmachine/blob/master/src/vote/striples.rs#L220), I guess for the validator crates replacing the call to 'check' by 'validate' could be fine to). 

@Marwes the alternative has the advantage to avoid confusion regarding the function proto and also to expose the deserializer (yet this could also be done with the other approach by putting the deserializer in the proto). I do not know what is the simpler (the call back is similar to 'skip_serializing_with' and the alternative is similar to 'remote').

dtolnay:
The approach proposed by @Marwes seems really promising to me. Let's put this on hold for a bit as we sort out #1118. @cheme feel free to chime in with your thoughts over there. Thanks!

Issue #1105, opened 2017-11-30
Container `#[serde(default)]` attribute ignored for `skip`ed fields


epage:
```rust
#[derive(Debug, PartialEq, Serialize, Deserialize)]
#[serde(deny_unknown_fields, default)]
struct Skipped {
    required: Option<String>,
    #[serde(skip)] optional: String,
}

impl Default for Skipped {
    fn default() -> Self {
        Self {
            required: Some("Hello".to_owned()),
            optional: "World".to_owned(),
        }
    }
}
```
Note that `optional` is skipped and has a default value.  When deserialized to `Skipped`, `String::default()` is used instead of being filled in by `Skipped::default`, like is done when `required` is missing.

See https://play.rust-lang.org/?gist=d32b5f2e2f37bd6c8c87e403f2a26f6e&version=stable

Binero:
Traced back the issue to this block: https://github.com/serde-rs/serde/blob/0c34e06e518c81315ec738584db0cffba6f5677a/serde_derive_internals/src/attr.rs#L884-L888

Two possible behaviours, one of which is breaking, and one of which adds some complexity to the `Field` structure. 

1. Simply require any field which has `#[serde(skip)]` set to also have either `#[serde(default)]` itself, or in its parent.
2. When the container of the `Field` has `#[serde(default)]` set, do not automatically set `#[serde(default)]` on the children.

The second option would require exposing some context to the field.

dtolnay:
@Binero fixed this in #1115 and I released the fix in serde_derive 1.0.24. Thanks!

PR #1115, opened 2017-12-06
Solved #1105.

Binero:
When a field should be skipped during deserialization, it will not use its own Default implementation
when the container structure has `#[serde(default)]` set.

### DIFF
--- a/serde_derive/src/de.rs
+++ b/serde_derive/src/de.rs
@@ -451,15 +451,15 @@ fn deserialize_seq(
         attr::Default::Default => {
             Some(
                 quote!(
-                let __default: Self::Value = _serde::export::Default::default();
-            ),
+                    let __default: Self::Value = _serde::export::Default::default();
+                ),
             )
         }
         attr::Default::Path(ref path) => {
             Some(
                 quote!(
-                let __default: Self::Value = #path();
-            ),
+                    let __default: Self::Value = #path();
+                ),
             )
         }
         attr::Default::None => {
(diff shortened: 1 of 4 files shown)


serde_derive/src/de.rs
```diff
@@ -447,7 +447,30 @@ fn deserialize_seq(
         };
     }
 
+    let let_default = match *cattrs.default() {
+        attr::Default::Default => {
+            Some(
+                quote!(
+                let __default: Self::Value = _serde::export::Default::default();
+            ),
```
oli-obk: Can you indent this line and the one above + the same thing in the `Path` match arm

serde_derive_internals/src/ast.rs
```diff
@@ -98,44 +98,61 @@ impl<'a> Body<'a> {
     }
 }
 
-fn enum_from_ast<'a>(cx: &Ctxt, variants: &'a [syn::Variant]) -> Vec<Variant<'a>> {
+fn enum_from_ast<'a>(
+    cx: &Ctxt,
+    variants: &'a [syn::Variant],
+    container: &attr::Container,
```
oli-obk: I wonder if we should just pass a reference to the `default` field. Information frugality and stuff ^^

Binero: It's a trade-off between the `Container` knowing that `Field` does inside, and `Field` knowing what `Container` does inside. I decided on the latter, but switching it around seems fine. I'm on it.

oli-obk: hm? I meant just passing `container_default: &Default`

PR #1159, opened 2018-02-16
[no_std] Add instructions in README.MD

niklasad1:
Hi,
This PR adds some information how to use Serde in ```no_std``` because it was not obvious to me

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,16 @@ about Serde. If IRC is not your thing or you don't get a good response, we are
 happy to respond to [GitHub issues](https://github.com/serde-rs/serde/issues/new)
 as well.
 
+## no-std
+Serde has a feature named std enabled by default. 
+In order to to use Serde in a no_std context the std feature has to be disabled.
+To do that add the following to your Cargo.toml:
+
+```
+[dependencies]
+serde = { version = "1.0.27", default-features = false }
+```
+
 ## License
 
 Serde is licensed under either of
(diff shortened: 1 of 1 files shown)


oli-obk:
Are there any additional things to take care of? like something special about `serde_derive`?

niklasad1:
@oli-obk 

I used the following when I tested it:
```
serde = { version = "1.0.27", default-features = false }
serde_derive = "1.0.27"
```
But ```serde_json``` seem not work to as it depends on the ```ǹum_crate```.

oli-obk:
So... I looked into it a bit more.

I think often you'll want to activate the [alloc](https://github.com/serde-rs/serde/blob/master/serde/Cargo.toml#L58) feature, in order to be able to use `Vec` and `String` if you have a heap (which isn't uncommon anymore, even on embedded). Can you just add a short comment about that?

niklasad1:
@oli-obk 
Yeah, that is a good point but also to use ```alloc``` you have run the nightly channel of the compiler but 
I don't add any info about that.  Because, it is implied by running in ```no_std``` context.

Hopefully, I managed to write it like you wanted otherwise feel free to modify it or ping me.

Also, it might good with an example for this. What do you think?

dtolnay:
I merged this but moved it to its own page on https:​//serde.rs to keep all documentation in one place.

**https://serde.rs/no-std.html**

Issue #1161, opened 2018-02-21
Check if internally-tagged enum's type tag name collides with one of the fields of a struct variant


H2CO3:
A (somewhat) minimal example:

```
#[macro_use]
extern crate serde_derive;
extern crate serde;
extern crate serde_json;

#[derive(Serialize, Deserialize)]
#[serde(tag = "qux")]
enum Foo {
    Lol {
        qux: &'static str,
    },
}

fn main() {
    let foo = Foo::Lol { qux: "bar" };
    let json = serde_json::to_string_pretty(&foo).unwrap();
    println!("{}", json);
}
```
Stdout:
```
{
  "qux": "Lol",
  "qux": "bar"
}
```

I would expect this to be a compile-time error since it's known at proc macro expansion time that the serialized map will contain two identical keys.

It would also be nice if serde issued a warning when keys are not identical yet very similar, e.g. `foobar` and `foo_bar`, or only differ in capitalization, since that might very well indicate a typo.

PR #1170, opened 2018-03-07
Disallow variant field names to conflict with tag of internally-tagged enum

H2CO3:
Fixes https://github.com/serde-rs/serde/issues/1161.

### DIFF
--- a/serde_derive_internals/src/check.rs
+++ b/serde_derive_internals/src/check.rs
@@ -16,7 +16,7 @@ pub fn check(cx: &Ctxt, cont: &Container) {
     check_getter(cx, cont);
     check_identifier(cx, cont);
     check_variant_skip_attrs(cx, cont);
-    check_internally_tagged_variant_name_conflict(cx, cont);
+    check_internal_tag_field_name_conflict(cx, cont);
 }
 
 /// Getters are only allowed inside structs (not enums) with the `remote`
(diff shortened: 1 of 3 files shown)


oli-obk:
Could you add a test for it?

H2CO3:
Sure thing! Where do I add one? I'm afraid I don't completely understand the structure of the `serde_tests` crate, is that the one I need? Thanks.

H2CO3:
@oli-obk Hey, could you help me with adding tests? I can't even run the test suite. I have latest nightly, but when I run 

    cargo +nightly test --features="unstable"

then basically every compile-time test fails, with errors such as

    error: tests/compile-fail/remote/wrong_de.rs:10: unexpected error: '10:1: 10:27: can't find crate for `serde_derive` [E0463]'

and

    found crate `serde_derive` compiled by an incompatible version of rustc [E0514]

What am I missing?

oli-obk:
That's an unfortunate side effect of `compiletest-rs`. You need to run `cargo clean` or even `rustup override set nightly` instead of `cargo +nighty`

H2CO3:
I've already tried cleaning; in fact, I've deleted and re-`git clone`'d the entire repository, without success :( `rustup override set stable` doesn't seem to have helped either, I still get the very same errors.

Just to be clear: I've set up my cargo config so that all `serde`-related dependencies are searched for locally. Cargo build reports:

    Compiling serde v1.0.27 (file:///Users/H2CO3/Projects/serde/serde)
    Compiling serde_derive_internals v0.19.0 (file:///Users/H2CO3/Projects/serde/serde_derive_internals)
    Compiling serde_derive v1.0.27 (file:///Users/H2CO3/Projects/serde/serde_derive)
    Compiling serde_test v1.0.27 (file:///Users/H2CO3/Projects/serde/serde_test)
    Compiling serde_test_suite v0.0.0 (file:///Users/H2CO3/Projects/serde/test_suite)

Am I correct that this is what should be done?

H2CO3:
@dtolnay hey there! Thanks for the tests and the guidance. I'm glad you found this useful. 👍

PR #1191, opened 2018-03-24
impl Serialize and Deserialize for std::num::NonZero*

SimonSapin:
… gated on the `unstable` Cargo feature.

These are new standard library types. Tracking issue: https://github.com/rust-lang/rust/issues/49137

### DIFF
--- a/serde/src/de/impls.rs
+++ b/serde/src/de/impls.rs
@@ -1918,6 +1918,7 @@ where
 ////////////////////////////////////////////////////////////////////////////////
 
 #[cfg(feature = "unstable")]
+#[allow(deprecated)]
 impl<'de, T> Deserialize<'de> for NonZero<T>
 where
     T: Deserialize<'de> + Zeroable,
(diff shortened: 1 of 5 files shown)


serde/src/ser/impls.rs
```diff
@@ -363,6 +364,32 @@ where
...
+macro_rules! nonzero_integers {
+    ( $( $T: ident, )+ ) => {
+        $(
+            #[cfg(feature = "unstable")]
+            impl Serialize for ::std::num::$T {
+                fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+                where
+                    S: Serializer,
+                {
+                    self.clone().get().serialize(serializer)
```
oli-obk: Why is this cloned?

serde/src/de/impls.rs
```diff
@@ -1934,6 +1935,39 @@ where
     }
 }
 
+macro_rules! nonzero_integers {
+    ( $( $T: ident, )+ ) => {
+        nonzero_integers!(@ $(::std::num::$T,)+);
```
oli-obk: The @ trick doesn't work on 1.13

serde/src/ser/impls.rs
SimonSapin: I copied this from the generic `NonZero<T>` impl which does not assume `T: Copy` (`get` takes `self` by value), but these concrete types are indeed `Copy` so `Clone` is not necessary. I’ve removed it.

serde/src/de/impls.rs
SimonSapin: I’ve verified that it does if I swap the order of the macro’s two variants/patterns.

SimonSapin:
On Travis:
```
lib: serde
    Updating registry `https://github.com/rust-lang/crates.io-index`
   Compiling serde v1.0.35 (file:///home/travis/build/serde-rs/serde/serde)
error: Unrecognized option: 'features'
error: Could not compile `serde`.
The command "./travis.sh" exited with 101.
```

I *think* this is when running `cargo clippy --features 'rc unstable' -- -Dclippy`, but it seems unrelated to this PR…

dtolnay:
The clippy trouble is https://github.com/rust-lang-nursery/rust-clippy/issues/2566.

PR #1213, opened 2018-04-10
Support #[derive(Serialize, Deserialize)] when using associated types

### LINKED ISSUE #1208: Infer correct trait bounds for use of a type parameter associated type
The following expands to `impl<T: Trait> Serialize for Struct<T>` which fails to compile.

```rust
trait Trait {
    type Assoc;
}

#[derive(Serialize)]
struct Struct<T: Trait> {
    assoc: T::Assoc,
}
```

The builtin derives already get this partially right. For example `#[derive(Debug)]` on the above expands to `impl<T: Debug + Trait> Debug for Struct<T> where T::Assoc: Debug`. This is partially right because it has `T::Assoc: Debug`, but wrong because it also has `T: Debug`.

Osspial:
Prior to this, structs and enums that had fields using associated types wouldn't have the appropriate where bounds set. This fixes that, meaning that this now builds:

```rust
trait Trait {
    type Assoc;
}

#[derive(Serialize)]
struct Struct<T: Trait> {
    assoc: T::Assoc,
}
```

Fixes #1208.

### DIFF
--- a/serde_derive/src/bound.rs
+++ b/serde_derive/src/bound.rs
@@ -55,21 +55,22 @@ pub fn with_where_predicates_from_fields<F, W>(
     generics: &syn::Generics,
     trait_bound: &syn::Path,
     from_field: F,
-    where_ty: W,
+    gen_bound_where: W,
 ) -> syn::Generics
 where
     F: Fn(&attr::Field) -> Option<&[syn::WherePredicate]>,
-    W: for<'a> Fn(&attr::Field) -> Option<&syn::ExprPath>,
+    W: for<'a> Fn(&attr::Field) -> bool,
 {
     let predicates = cont.data
         .all_fields()
         .flat_map(|field| {
-            let field_bound: Option<syn::WherePredicate> = match where_ty(&field.attrs) {
-                Some(_) => None,
-                None => {
+            let field_bound: Option<syn::WherePredicate> = match gen_bound_where(&field.attrs) {
+                true => {
                     let field_ty = field.ty;
-                    Some(parse_quote!(#field_ty: #trait_bound))
-                }
+                    let predicate: syn::WherePredicate = parse_quote!(#field_ty: #trait_bound);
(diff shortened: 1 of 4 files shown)


serde_derive/src/bound.rs
```diff
@@ -50,18 +50,35 @@ pub fn with_where_predicates(
...
+    W: for<'a> Fn(&attr::Field) -> bool,
 {
     let predicates = cont.data
         .all_fields()
-        .flat_map(|field| from_field(&field.attrs))
-        .flat_map(|predicates| predicates.to_vec());
+        .flat_map(|field| {
+            let field_ty = field.ty;
+            let field_bound: Option<syn::WherePredicate> = match (field_ty, gen_bound_where(&field.attrs)) {
+                (&syn::Type::Path(ref ty_path), true) => match (ty_path.path.segments.first(), generics.params.first()) {
```
dtolnay: Please make this work even if the associated type is not from the first generic parameter.

```rust
struct AssocDerive<S, T: AssocSerde> {
    s: S,
    assoc: T::Assoc,
}
```

serde_derive/src/bound.rs
```diff
@@ -50,18 +50,35 @@ pub fn with_where_predicates(
...
+pub fn with_where_predicates_from_fields<F, W>(
     cont: &Container,
     generics: &syn::Generics,
+    trait_bound: &syn::Path,
     from_field: F,
+    gen_bound_where: W,
 ) -> syn::Generics
 where
     F: Fn(&attr::Field) -> Option<&[syn::WherePredicate]>,
+    W: for<'a> Fn(&attr::Field) -> bool,
```
dtolnay: The `'a` lifetime parameter is never used.

dtolnay:
Thanks btw! This is going to fix a quite common stumbling block.

Osspial:
Those errors should be fixed now.

PR #1258, opened 2018-05-14
Don't rely on symlinks as they do not work (well) on windows

Marwes:
While symlink support on windows is improving it is not solid enough to
just work yet. This just changes the path to point to the real file
which makes `cargo test --all` work on windows again.

See for instance.

https://github.com/git-for-windows/git/issues/1646 (merged but not
released)

https://github.com/git-for-windows/git/issues/1197


Marwes:
Damn, I didn't attempt to do `cargo publish`... . No clue how to get this working then (`cargo build` etc works fine though, arguably a bug with rustc/cargo but probably a low prio/hard to fix one...)

Marwes:
Could perhaps remove `serde_derive_internals` from the work space, that would let `cargo test --all` work at least.

dtolnay:
Would it work if we symlink individual .rs files rather than symlinking the directory?

Marwes:
Possibly, it is rather unclear what works and what does not so I will need to test it.

Marwes:
`ln ` is apparently broken in git bash as well as the linux subsystem, may need to add the symlink from linux and checkout the repo in windows to test.

Marwes:
Found out how to get symlinks working finally (directory symlinks as well).

Since a while back symlinks do work on Windows but only if [Developer Mode](https://docs.microsoft.com/en-us/windows/uwp/get-started/enable-your-device-for-development) is enabled. After I enabled that + having a git released after february this year (which enables symlinks in git) then git is able to create the symlinks as it should.

So there is no action necessary here, unless we want to document that Developer Mode is necessary on Windows (this issue might be enough perhaps)?

dtolnay:
Does the version with symlinked files instead of a symlinked directory work without Developer Mode and with an old git? I would be okay landing it this way if it does.

dtolnay:
That's too bad. Thanks for checking. I think you're right, let's leave it as is and leave this thread for people to land on from searches. Thanks anyway!

Issue #1136, opened 2018-01-07
i128 and u128 integers missing Deserialize impls


PlasmaPower:
They should behind the `unstable` flag.

PlasmaPower:
u128 and i128 are going to be stabilized very soon. However, from what I can tell, adding support to serde will be a breaking change. Is serde 2.0 planned?

SeunLanLege:
They've been stabilized :confetti_ball: when will be this be included in the next release?

oli-obk:
We can't add them directly, because that would break all users of older rustc stable versions

We could add it under a (stable) feature flag

cgm616:
So I'm trying to add this behind a feature flag in cgm616/serde, but I come across a super weird error when attempting to compile. Can someone check this out for me?

I'm running

```shell
> cargo build --all-features
```

and getting this output:

```
﻿   Compiling serde_test v1.0.41 (file:///Users/cgm616/Documents/Developing/rust/serde/serde_test)
error[E0046]: not all trait items implemented, missing: `deserialize_i128`, `deserialize_u128`
   --> serde_test/src/de.rs:126:1
    |
126 | impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ missing `deserialize_i128`, `deserialize_u128` in implementation
    |
    = note: `deserialize_i128` from trait: `fn(Self, V) -> std::result::Result<<V as serde::de::Visitor<'de>>::Value, <Self as serde::Deserializer<'de>>::Error>`
    = note: `deserialize_u128` from trait: `fn(Self, V) -> std::result::Result<<V as serde::de::Visitor<'de>>::Value, <Self as serde::Deserializer<'de>>::Error>`

error[E0046]: not all trait items implemented, missing: `deserialize_i128`, `deserialize_u128`
   --> serde_test/src/de.rs:655:1
    |
655 | impl<'de> de::Deserializer<'de> for BytesDeserializer {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ missing `deserialize_i128`, `deserialize_u128` in implementation
    |
[...]

PR #1259, opened 2018-05-15
Build script that does nothing

dtolnay:
Eventually we will want a build script that enables Serde impls for i128 and u128 -- https://github.com/serde-rs/serde/issues/1136. As a first step here is a build script that does nothing to see whether we can roll this out without breaking anyone's workflow, without having a supported feature at stake in the event that it needs to be rolled back.

### DIFF
--- a/serde/Cargo.toml
+++ b/serde/Cargo.toml
@@ -10,7 +10,8 @@ documentation = "https://docs.serde.rs/serde/"
 keywords = ["serde", "serialization", "no_std"]
 categories = ["encoding"]
 readme = "README.md"
-include = ["Cargo.toml", "src/**/*.rs", "README.md", "LICENSE-APACHE", "LICENSE-MIT"]
+include = ["Cargo.toml", "build.rs", "src/**/*.rs", "README.md", "LICENSE-APACHE", "LICENSE-MIT"]
+build = "build.rs"
 
 [badges]
 travis-ci = { repository = "serde-rs/serde" }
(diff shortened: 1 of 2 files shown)


dtolnay:
After releasing this I would like to give people a week to complain before we start using this to enable any observable behavior.

oli-obk:
Would it be overkill to use the rustc_version crate?

Good idea on the slow rollout. We could probably publish an rc version first when we try out an actual feature.

dtolnay:
The rustc_version crate adds 2.5 CPU-seconds of compile time on my computer above what this build script adds, which seemed excessive. We don't need a production quality semver parser or complete rustc metadata handling.

hcpl:
FYI there is also the `version_check` crate which compared to `rustc_version` is much more lightweight feature-wise and doesn't have dependencies.

dtolnay:
Nice, I hadn't seen `version_check` before. That looks a lot more reasonable for our use case.

dtolnay:
Thanks for demonstrating the implementation with version_check! In my own applications I would definitely use it. But for Serde I am going to stick with some simple parsing in the build script. I believe one of the defining challenges for Serde over the next few years will be avoiding bloat (or the perception of bloat) and there is a vast difference in perception between having dependencies and having no dependencies. I took a look at `cargo clean && time cargo build --no-default-features` before this PR, with just a build script, and with build script + version_check and the build script by itself adds 9% to compile time which is already close to the limit of what it makes sense to accept for the sake of i128 impls; build script + version_check adds 18%. But let's revisit if this approach turns out unreliable or unmaintainable.

PR #1260, opened 2018-05-16
Add Avro to the list of supported data formats

flavray:
Hello there,

We've been working on a serde-compatible [Apache Avro](http://avro.apache.org/docs/current/) rust client library. It supports all the main Avro features, with a serde-friendly interface.

The crate is documented, tested and (slightly 🙂) benchmarked.

Crate link: https://crates.io/crates/avro-rs
Crate repository: https://github.com/flavray/avro-rs

### DIFF
--- a/serde/src/lib.rs
+++ b/serde/src/lib.rs
@@ -59,6 +59,8 @@
 //!   *(deserialization only)*
 //! - [Redis], deserialize values from Redis when using [redis-rs].
 //!   *(deserialization only)*
+//! - [Avro], a binary data serialization format with support for
+//!   schematized data.
 //!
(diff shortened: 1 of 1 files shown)


flavray:
Thanks!

Good idea, I agree the `-rs` suffix is indeed not great (it's actually been added only because `avro` was already taken) and it seems that there is no on-going work on the current `avro` crate.

I will try to contact the owner of the crate. However, wouldn't it break any (hypothetical) application depending on the current `avro` crate? I am also unsure about how crates.io deals with changes to inner repositories (things like library versions).
Do you know of any documentation we could follow in order to do this?

PR #1263, opened 2018-05-18
Add Serde impls for i128 and u128

dtolnay:
Fixes https://github.com/serde-rs/serde/issues/1136.

### DIFF
--- a/serde/src/de/from_primitive.rs
+++ b/serde/src/de/from_primitive.rs
@@ -178,3 +178,91 @@ impl_from_primitive_for_uint!(u32);
 impl_from_primitive_for_uint!(u64);
 impl_from_primitive_for_float!(f32);
 impl_from_primitive_for_float!(f64);
+
+serde_if_integer128! {
+    impl FromPrimitive for i128 {
+        #[inline]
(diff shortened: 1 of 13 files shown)


cgm616:
What's the process like to get this merged? Support for i128 and u128 would be awesome.

dtolnay:
I am planning to land this on 25 May to give people some time to complain about our new build script. It becomes complicated to reverse course on the build script after starting to stick observable behavior behind it.

dtolnay:
@oli-obk can you notice any problems with this approach?

dtolnay:
Thanks, I really appreciate the second opinion. In terms of CI -- our existing 1.13.0 builder confirms that things continue to compile on a compiler without these types. Our nightly builder that runs the test suite is running test_de::test_small_int_to_128 and test_ser::test_integer128. The usefulness of those test cases is limited by not having 128-bit tokens in serde_test (#1281) but at least they confirm that deserialize_?128 and serialize_?128 are invoked as intended. Upon implementing #1280 at least there will be a Deserializer directly in the Serde codebase that deals with 128-bit integers that can be tested. And serde_json will provide end-to-end test coverage when we add 128-bit support there.

PR #1270, opened 2018-05-20
Transparent attribute to specify that representation is the same as its only field

### LINKED ISSUE #1054: Attribute to specify that type representation is the same as its only field
By analogy with [`#[repr(transparent)]`](https://github.com/rust-lang/rfcs/blob/master/text/1758-repr-transparent.md).

```rust
// This should serialize and deserialize directly as String, rather than as newtype.
#[serde(transparent)]
struct Transparent(String);

// Generated code.
impl Serialize for Transparent {
        where S: Serializer
    {
        self.0.serialize(serializer)
    }
}

impl<'de> Deserialize<'de> for Transparent {
        where D: Deserializer<'de>
    {
        Deserialize::deserialize(deserializer).map(Transparent)
    }
}
```

Should also work for structs with one field, and structs and tuple structs with all but one skipped field.

@bluss

dtolnay:
Fixes #1054.

### DIFF
--- a/serde_derive/src/internals/attr.rs
+++ b/serde_derive/src/internals/attr.rs
@@ -105,6 +105,7 @@ impl Name {
 /// Represents container (e.g. struct) attribute information
 pub struct Container {
     name: Name,
+    transparent: bool,
     deny_unknown_fields: bool,
     default: Default,
     rename_all: RenameRule,
@@ -181,6 +182,7 @@ impl Container {
     pub fn from_ast(cx: &Ctxt, item: &syn::DeriveInput) -> Self {
         let mut ser_name = Attr::none(cx, "rename");
         let mut de_name = Attr::none(cx, "rename");
+        let mut transparent = BoolAttr::none(cx, "transparent");
         let mut deny_unknown_fields = BoolAttr::none(cx, "deny_unknown_fields");
         let mut default = Attr::none(cx, "default");
         let mut rename_all = Attr::none(cx, "rename_all");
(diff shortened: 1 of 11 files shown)


serde_derive/src/internals/check.rs
```diff
@@ -278,6 +279,31 @@ fn check_adjacent_tag_conflict(cx: &Ctxt, cont: &Container) {
     }
 }
 
+/// Enums and unit structs cannot be transparent.
```
oli-obk: Technically single variant enums are just structs

serde_derive/src/internals/attr.rs
```diff
@@ -777,7 +789,6 @@ pub enum Default {
 }
 
 impl Default {
-    #[cfg(feature = "deserialize_in_place")]
```
oli-obk: Is this change intended?

serde_derive/src/ser.rs
```diff
@@ -186,6 +188,52 @@ fn serialize_body(cont: &Container, params: &Parameters) -> Fragment {
...
+        }
+        if transparent_field.is_some() {
+            panic!("Ambiguous transparent field");
+        }
+        transparent_field = Some(field);
+    }
+
+    match transparent_field {
+        Some(transparent_field) => transparent_field,
+        None => panic!("No field can be transparent"),
```
oli-obk: That message feels suboptimal. I know why it's worded this way because I'm reading the PR, but still found it confusing

serde_derive/src/internals/check.rs
dtolnay: True but let's leave that until someone asks for it.

serde_derive/src/internals/attr.rs
dtolnay: I believe so, deciding whether a field during `derive(Deserialize)` is transparent now checks `field.attrs.default().is_none()`.

serde_derive/src/ser.rs
dtolnay: Thanks, I improved the message and added some compile-fail tests.

PR #1275, opened 2018-05-21
Re-enable testing of nightly proc macro

dtolnay:


dtolnay:
Fixed in https://github.com/rust-lang/rust/pull/50946.

PR #1278, opened 2018-05-24
Implement for std::num::NonZero* on Rust 1.28+

### LINKED ISSUE #1274: Stabilize impls for NonZero*
These types were stabilized in https://github.com/rust-lang/rust/pull/50808 and will be released in Rust 1.28.

We should use version detection through the Serde build.rs script to automatically provide the impl on sufficiently new compilers.

SimonSapin:
… regardless of the `unstable` feature. Fix #1274.


SimonSapin:
Done.

PR #1297, opened 2018-06-01
Use compile_error! instead of panicking

adamcrume:
As a side effect, this changes error messages from:

```
error: proc-macro derive panicked
   --> test_suite/tests/test_ser.rs:597:10
    |
597 | #[derive(Serialize)]
    |          ^^^^^^^^^
    |
    = help: message: 2 errors:
                # unknown serde field attribute `typo`
                # expected serde with attribute to be a string: `with = "..."`
```

to

```
error: 2 errors:
    # unknown serde field attribute `typo`
    # expected serde with attribute to be a string: `with = "..."`
   --> test_suite/tests/test_ser.rs:597:10
    |
597 | #[derive(Serialize)]
    |          ^^^^^^^^^
```

### DIFF
--- a/serde_derive/src/lib.rs
+++ b/serde_derive/src/lib.rs
@@ -53,6 +53,7 @@ extern crate proc_macro2;
 
 mod internals;
 
+use std::str::FromStr;
 use proc_macro::TokenStream;
 use syn::DeriveInput;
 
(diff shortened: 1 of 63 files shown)


serde_derive/src/lib.rs
```diff
@@ -71,7 +72,7 @@ pub fn derive_serialize(input: TokenStream) -> TokenStream {
     let input: DeriveInput = syn::parse(input).unwrap();
     match ser::expand_derive_serialize(&input) {
         Ok(expanded) => expanded.into(),
-        Err(msg) => panic!(msg),
+        Err(msg) => TokenStream::from_str(&format!("compile_error!({:?});", msg)).unwrap(),
```
dtolnay: Typically macros try to avoid working with Rust source code as strings -- strings are unfriendly to our text editor or IDE when it comes to brace matching, syntax highlighting, indentation, etc. Serde's derive currently uses the [`quote`](https://github.com/dtolnay/quote) crate instead of strings of source code.

Here it's just one line so there may not be an appreciable advantage, but I think it would be better to be consistent with the code generation done in the rest of serde_derive and use `quote!` here as well.

PR #1298, opened 2018-06-01
Use try_borrow for serializing RefCell

xfix:


xfix:
Nevermind, `custom` requires `std`.

PR #1303, opened 2018-06-03
Support Duration in no-std mode on new compilers

### LINKED ISSUE #1123: Allow Duration on libcore once it's stable
There's currently an accepted issue (rust-lang/rust#46520) to move the `std::time::Duration` type to `core::time::Duration`. Once this is stable, there should be a change to `serde/src/lib.rs` to change the import of `Duration` so that it's allowed on `no_std` builds.

dtolnay:
Fixes #1123.

### DIFF
--- a/serde/build.rs
+++ b/serde/build.rs
@@ -24,6 +24,12 @@ fn main() {
         println!("cargo:rustc-cfg=de_rc_dst");
     }
 
+    // Duration available in core since Rust 1.25:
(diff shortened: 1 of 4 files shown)


oli-obk:
should we be adding builders for all versions mentioned in the build script to prevent accidentally inserting things under the relevant cfg, if that is available only on a newer rustc?

oli-obk:
e.g. if something is changed in the Duration serialization, but that change works only on a recent rustc, our test suite will not catch it atm

Issue #1310, opened 2018-06-07
Conditionally enabling tagged enums for non "self-describing" (de)serializers.


AGausmann:
Consider the following `enum` declaration which I am using for dynamic typing:

```rust
#[derive(Serialize, Deserialize)]
pub enum Property {
    Boolean(bool),
    Integer(i64),
    Number(f64),
    Text(String),
}
```

This is used in many places in my program, usually within a `Properties` struct which encapsulates a `String` -> `Property` map. I have implemented `Serialize` and `Deserialize` for `Properties` which simply delegates to its inner map. The rest of `Properties` implementation is irrelevant at this time, but I can add it later if needed.

Here are two scenarios where `Properties` is commonly used:

1. **Small configuration or metadata files.** Human-readable and human-editable formats like JSON or TOML would be nice in this scenario. In such formats, using tagged enums would make this less human-friendly and a bit redundant, as each enum variant can get unambiguously (de)serialized as its inner value if I'm using the formats I've previously noted, since they both can reasonably support `Deserializer::deserialize_any`. I can provide my own implementation to convert most types to one of those variants, or even potentially use `#[serde(untagged)]`, though I don't think it will automatically cast other integer/float layouts and would therefore not work in my situation.

2. **Files containing massive arrays where each element contains a `Properties` map.** Since the most reasonable way for users to interact with the elements in the array is through a specialized program, I would like to be space-efficient by serializing these arrays using `Bincode` or another compact binary format. Many of the most efficient binary formats do not support `deserialize_any` because they aren't "self-describing" (the term used in `deserialize_any` documentation); they assume you know the structure of the data, and I do... for the most part. In this scenario, tags are necessary; untagged enums would fail because they don't sufficiently describe their contents. 

Currently, I can not provide a single implementation of `Serialize` and `Deserialize` to cover both cases in the way that I have described, as it is impossible for the traits' functions to determine whether tags are necessary for the given format.

This has implications for situations other than mine. Right now, if you serialize an untagged enum using Bincode, it will actually write the enum values without a tag. Once you try to deserialize it, it won't know what to do because it can't support `deserialize_any`. While this could be written off as a logic error by the user of the library who should know better, this is a reasonable thing that could happen to an unsuspecting user who may not actually know what went wrong.

[...]

PR #1311, opened 2018-06-08
Special trait for deserialize_any

AGausmann:
Related issue: #1310 

Prototype implementation of the features discussed in related issue #1310. The goal is to create an effective way to distinguish "self-describing" data formats, the ones that can determine the type of the object during deserialization and can therefore implement  `Deserializer::deserialize_any`.  Currently this is being done by moving `deserialize_any` to a new trait so this check can be performed at compile-time.

The names that I have given to the new traits and functions are placeholders until we decide on proper names; I don't know what I want to call them yet.

Comments, suggestions, commits from maintainers, and any other contributions are more than welcome. This is my first time messing around in Serde source code, so I'm not familiar with a lot of the conventions and available utilities.

**TODO**

- [ ] Proper names
- [ ] Replace deprecated calls (should fix errors, wait until names are finalized)
- [ ] Documentation/doctests (wait until names are finalized)
- [ ] New tests?

### DIFF
--- a/serde/src/de/mod.rs
+++ b/serde/src/de/mod.rs
@@ -559,6 +559,14 @@ pub trait Deserialize<'de>: Sized {
         *place = Deserialize::deserialize(deserializer)?;
         Ok(())
     }
+
+    #[warn(missing_docs)] //TODO
+    fn deserialize_with_any<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: DeserializerAny<'de>,
+    {
+        Self::deserialize(deserializer)
+    }
 }
 
 /// A data structure that can be deserialized without borrowing any data from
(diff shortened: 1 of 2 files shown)


AGausmann:
I've run into a situation that I've never had before. Since `Deserializer::deserialize_any` is deprecated, it would make sense for me to make its implementation optional, having a sane default implementation. 

Since implementors of `Deserialize` who have not upgraded may be calling `Deserializer`s who have, I was thinking about delegating the old function to the new `DeserializerAny::deserialize_any` by default. However, I can not assume that that trait is also implemented. I don't know of any way (if there is a way) to provide a default implementation only for implementers of `DeserializerAny`.

dtolnay:
Could you write out an implementation of deserialize_with_any for something like serde_yaml::Value? The existing Deserialize impl is [here](https://github.com/dtolnay/serde-yaml/blob/0.7.5/src/value/de.rs#L13-L119).

Naively it seems like it would require -Any versions of all the data format traits involved: SeqAccessAny, MapAccessAny, etc.

AGausmann:
Regarding my previous comment, I don't think it's worth trying to solve that issue. Leaving the old deprecated function without a default is no change which will be technically a non-breaking change although users will still have to provide an implementation for it.

re @dtolnay:  That example shows an important point about situations where `deserialize_any` is the only option that makes sense. In this case you would obviously have to move the current `Deserialize::deserialize` implementation to `Deserialize::deserialize_with_any`, but you'd also have to reimplement the former in some way. I would guess using the default enum implementation would be correct but that makes no sense in the contexts that `serde_yaml::Value` is normally used in.

AGausmann:
Closed per #1310

PR #1324, opened 2018-06-28
Suppress 'extern crate' warning for rust 2018

jechase:
With the 2018 preview, and presumably when the edition is officially released, `extern crate` will be unidiomatic, and will generate warnings.

This suppresses the warning for the `extern crate` generated in the `Serialize` and `Deserialize` derives.

### DIFF
--- a/serde_derive/src/de.rs
+++ b/serde_derive/src/de.rs
@@ -74,6 +74,7 @@ pub fn expand_derive_deserialize(input: &syn::DeriveInput) -> Result<TokenStream
         #[allow(non_upper_case_globals, unused_attributes, unused_qualifications)]
         const #dummy_const: () = {
             #[allow(unknown_lints)]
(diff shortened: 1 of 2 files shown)


jechase:
So this might not actually be needed in the long run. From IRC:

```
 @steveklabnik | jechase: generally you don't do that; you pick one edition and write your code in it. editions are compatible, so if you write your  
               | code in rust 2018, programs using rust 2015 can use it, and vice versa                                                               
       jechase | Right, but this isn't me writing code, this is serde_derive writing code. I guess the alternative is "don't generate warnings inside 
               | of macros?"                                                                                                                          
 @steveklabnik | ah so                                                                                                                                
 @steveklabnik | i *think* the intention is that macros carry the edition they were deifned in                                                        
 @steveklabnik | so that this doesn't just happen                                                                                                     
 @steveklabnik | i *think* 
```

PR #1330, opened 2018-07-01
Add SemVer stability badge to README

greysteil:
Adds a SemVer stability badge to the readme that displays the percentage of CI runs that pass when updating `serde` between SemVer compatible versions. Data comes from Dependabot (disclosure: I built it).

Here's what the badge looks like:

[SemVer](https://dependabot.com/compatibility-score.html?dependency-name=serde&package-manager=cargo&version-scheme=semver)

The score isn't quite 100%, maybe because of users' flaky tests. I do my best to filter those out, but hopefully a 98% score still makes it clear that users can be confident that serde follows SemVer pretty much perfectly 🙂


greysteil:
👍 - I'll let you know once that work is done!

PR #1347, opened 2018-08-05
Implement Serialize and Deserialize for RangeInclusive

c410-f3r:
Copied implementation from Range, is it OK?

Edit: I will fix the failing tests but `RangeInclusive` structure was modified in [#51622](https://github.com/rust-lang/rust/pull/51622), should we wait for the next Rust release?

### DIFF
--- a/serde/build.rs
+++ b/serde/build.rs
@@ -36,6 +36,12 @@ fn main() {
         println!("cargo:rustc-cfg=integer128");
     }
 
+    // Inclusive ranges methods stabilized in Rust 1.27:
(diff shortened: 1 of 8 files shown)


c410-f3r:
@dtolnay Tests are fixed. Not sure about the additional field since every new instance of `RangeInclusive` [starts with the same value](https://doc.rust-lang.org/beta/src/core/ops/range.rs.html#394-396) `is_empty: None`.

dtolnay:
I published these impls in serde 1.0.71.

Issue #1335, opened 2018-07-07
`--features alloc` no longer compiles on nightly-2018-07-07


dtolnay:
```console
$ cargo build --no-default-features --features alloc

error[E0432]: unresolved import `alloc::BTreeMap`
   --> serde/src/lib.rs:196:21
    |
196 |     pub use alloc::{BTreeMap, BTreeSet, BinaryHeap, LinkedList, VecDeque};
    |                     ^^^^^^^^ no `BTreeMap` in the root

error[E0432]: unresolved import `alloc::BTreeSet`
   --> serde/src/lib.rs:196:31
    |
196 |     pub use alloc::{BTreeMap, BTreeSet, BinaryHeap, LinkedList, VecDeque};
    |                               ^^^^^^^^ no `BTreeSet` in the root

error[E0432]: unresolved import `alloc::BinaryHeap`
   --> serde/src/lib.rs:196:41
    |
196 |     pub use alloc::{BTreeMap, BTreeSet, BinaryHeap, LinkedList, VecDeque};
    |                                         ^^^^^^^^^^ no `BinaryHeap` in the root

error[E0432]: unresolved import `alloc::LinkedList`
   --> serde/src/lib.rs:196:53
    |
196 |     pub use alloc::{BTreeMap, BTreeSet, BinaryHeap, LinkedList, VecDeque};
    |                                                     ^^^^^^^^^^ no `LinkedList` in the root

error[E0432]: unresolved import `alloc::VecDeque`
   --> serde/src/lib.rs:196:65
    |
196 |     pub use alloc::{BTreeMap, BTreeSet, BinaryHeap, LinkedList, VecDeque};
    |                                                                 ^^^^^^^^ no `VecDeque` in the root
```

PR #1359, opened 2018-08-20
Fix compilation under `rc` feature

Pratyush:
Currently compilation fails when enabling both the `alloc` and `rc` features because of an update to `liballoc`. A previous PR (https://github.com/serde-rs/serde/issues/1335) to fix this didn't catch that `Arc` now lives under `alloc::sync`. This is a quick fix =)

### DIFF
--- a/serde/src/lib.rs
+++ b/serde/src/lib.rs
@@ -196,7 +196,7 @@ mod lib {
     pub use std::rc::{Rc, Weak as RcWeak};
 
     #[cfg(all(feature = "rc", feature = "alloc", not(feature = "std")))]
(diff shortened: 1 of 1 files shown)


dtolnay:
I published 1.0.72 with this fix.

Pratyush:
Great, thanks! I was wondering why this slipped past CI, and it seems like travis.sh doesn't check "alloc + rc" anywhere. To fix this, would I just add a line to that effect here: https://github.com/serde-rs/serde/blob/befc7edc17aa50aa1a666d0b2fdd61538ae98635/travis.sh#L52 ?

dtolnay:
Yes that would be a good idea. :)

PR #1365, opened 2018-08-23
Disable i128 integers on Emscripten targets

koute:
Emscripten doesn't currently support i128 and will fail with the following error if they're used:

```
LLVM ERROR: Instruction not yet supported for integer types larger than 64 bits
```


dtolnay:
Thanks! I published this in 1.0.74.

PR #1376, opened 2018-09-06
Fix internally tagged enum deserialization with unknown fields

dreid:
Currently an internally tagged enum with a unit variant will fail to deserialize if the map contains unknown fields.

[Example on Playground](https://play.rust-lang.org/?gist=58c1609b0ed43cb8275ea3f1e4fa0c31&version=nightly&mode=debug&edition=2015)

This change attempts to address that by having the `InternallyTaggedUnitVisitor.visit_map` method consume the map before returning.

I've added a test case for the simple enum case and for the case where the unit enum variant is flattened into a struct.

### DIFF
--- a/serde/src/private/de.rs
+++ b/serde/src/private/de.rs
@@ -231,8 +231,8 @@ mod content {
 
     use super::size_hint;
     use de::{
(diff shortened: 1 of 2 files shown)


dtolnay:
I released this fix in 1.0.77.

PR #1399, opened 2018-09-29
Some docs

derekdreery:
Just some extra documentation I wrote whilst trying to understand the codebase.

### DIFF
--- a/serde_derive/src/internals/ast.rs
+++ b/serde_derive/src/internals/ast.rs
@@ -6,24 +6,36 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
+//! A serde ast, parsed from the syn ast and ready for codegen.
+
 use internals::attr;
 use internals::check;
 use internals::{Ctxt, Derive};
 use syn;
 use syn::punctuated::Punctuated;
 
+/// A source data structure annotated with `#[derive(Derialize)]` and/or `#[derive(Deserialize)]`,
+/// parsed into an internal representation.
 pub struct Container<'a> {
+    /// The struct or enum name (without generics).
     pub ident: syn::Ident,
+    /// Attributes on the structure, parsed for serde.
     pub attrs: attr::Container,
+    /// The contents of the struct or enum.
     pub data: Data<'a>,
+    /// Any generics on the struct or enum.
     pub generics: &'a syn::Generics,
 }
 
+/// The fields of a struct or enum.
+///
+/// Analagous to `syn::Data`.
 pub enum Data<'a> {
     Enum(Vec<Variant<'a>>),
     Struct(Style, Vec<Field<'a>>),
 }
 
+/// A variant of an enum.
 pub struct Variant<'a> {
     pub ident: syn::Ident,
     pub attrs: attr::Variant,
@@ -31,6 +43,7 @@ pub struct Variant<'a> {
     pub fields: Vec<Field<'a>>,
 }
 
+/// A variant of a struct.
 pub struct Field<'a> {
     pub member: syn::Member,
     pub attrs: attr::Field,
@@ -40,13 +53,18 @@ pub struct Field<'a> {
 
 #[derive(Copy, Clone)]
 pub enum Style {
+    /// Named fields.
     Struct,
+    /// Many unnamed fields.
     Tuple,
+    /// One unnamed field.
     Newtype,
+    /// No fields.
     Unit,
 }
 
 impl<'a> Container<'a> {
+    /// Convert the raw syn ast into a parsed container object, collecting errors in `cx`.
     pub fn from_ast(cx: &Ctxt, item: &'a syn::DeriveInput, derive: Derive) -> Container<'a> {
         let mut attrs = attr::Container::from_ast(cx, item);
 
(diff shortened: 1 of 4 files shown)


serde_derive/src/internals/ast.rs
```diff
@@ -6,31 +6,44 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
+//! A serde ast, parsed from the syn ast and ready for codegen.
```
dtolnay: We have moved away from using the word "codegen" to describe procedural macros because it is unnecessarily intimidating. Please find a different way to phrase this without "codegen".

serde_derive/src/internals/ast.rs
```diff
@@ -6,31 +6,44 @@
...
 
+/// A variant of an enum.
 pub struct Variant<'a> {
     pub ident: syn::Ident,
     pub attrs: attr::Variant,
     pub style: Style,
     pub fields: Vec<Field<'a>>,
 }
 
+/// A variant of a struct.
```
dtolnay: A struct always has a single variant but it may have multiple fields, so fields are not the same as variants.

serde_derive/src/internals/case.rs
```diff
@@ -6,6 +6,9 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
+//! Stuff to handle where the case of the source (e.g. `my-field`) is different to the
```
dtolnay: Please write something like "case conversions" instead of "stuff".

dtolnay: Please change "different to" to "different from", which also works in American English.
https://en.oxforddictionaries.com/usage/different-from-than-or-to

serde_derive/src/internals/ctxt.rs
```diff
@@ -10,18 +10,27 @@ use std::cell::RefCell;
...
+/// References can be shared since this type uses run-time exclusive mut checking.
 #[derive(Default)]
 pub struct Ctxt {
+    // The contents will be set to `None` during checking. This is so that checking can be
+    // enforced.
     errors: RefCell<Option<Vec<String>>>,
 }
 
 impl Ctxt {
+    /// Create a new context object
```
dtolnay: I don't think that explaining `fn new` creates a new object adds any value. Please remove this comment.

serde_derive/src/internals/ctxt.rs
```diff
@@ -10,18 +10,27 @@ use std::cell::RefCell;
...
     pub fn new() -> Self {
         Ctxt {
             errors: RefCell::new(Some(Vec::new())),
         }
     }
 
+    /// Add an error to the context object
```
dtolnay: Since you've added periods in other files, please end this comment with a period as well. Ditto `fn error`.

serde_derive/src/internals/ast.rs
derekdreery: I changed "for codegen" to "to generate rust code".

serde_derive/src/internals/ast.rs
derekdreery: Sorry this is a typo, it should say `field`.

serde_derive/src/internals/case.rs
derekdreery: TIL

derekdreery: I've actually reworded this completely.

serde_derive/src/internals/ctxt.rs
derekdreery: I think I included this doc because the `Drop` behavior is non-standard. I've expanded it a little. If you still don't think it's necessary I'm happy to remove :).

serde_derive/src/internals/case.rs
```diff
@@ -6,6 +6,9 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
+//! Utility code to assist converting the case of the source (e.g. `my-field`, `MY_FIELD`) to the
```
dtolnay: What distinction is this making between utility code and ordinary code? I would recommend avoiding the word utility in naming or documenting anything. [https://blog.golang.org/package-names#TOC_5.](https://blog.golang.org/package-names#TOC_5.) is relevant.

serde_derive/src/internals/ast.rs
```diff
@@ -6,31 +6,44 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
+//! A serde ast, parsed from the syn ast and ready to generate rust code.
```
dtolnay: Please capitalize Serde, Syn, Rust (here and elsewhere).

serde_derive/src/internals/case.rs
```diff
@@ -6,6 +6,9 @@
...
+//! rust-styled field/variant (e.g. `my_field`, `MyType`).
```
dtolnay: This code actually only converts in the opposite direction, from name in Rust to name in the input, so I find this explanation misleading even if technically correct. The conversion from name in the input to name in Rust happens elsewhere by emitting a match-expression.

derekdreery: Can you check you're happy with this now.

PR #1424, opened 2018-11-17
Use more spans for error messages

hcpl:
Basically, this PR makes errors that look like this:

```
error[E0277]: the trait bound `Foo: std::default::Default` is not satisfied
  --> src/lib.rs:11:21
   |
11 | #[derive(Serialize, Deserialize)]
   |                     ^^^^^^^^^^^ the trait `std::default::Default` is not implemented for `Foo`
   |
   = note: required by `std::default::Default::default`
```

to look like this instead:

```
error[E0277]: the trait bound `Foo: std::default::Default` is not satisfied
  --> src/lib.rs:17:5
   |
17 | /     #[serde(rename = "f")]
18 | |     #[serde(default)]
19 | |     c: Foo,
   | |__________^ the trait `std::default::Default` is not implemented for `Foo`
   |
   = note: required by `std::default::Default::default`
```

I _believe_ I covered every possible invalid case, though I possibly missed some due to the sheer amount of supported derive configurations.

I am not sure if spans for attributes should be included in general --- on one hand, they consume vertical terminal estate; on the other hand, they may give a clue why the error happened at all (in the example above the `std::default::Default` requirement came from `#[serde(default)]`).

Also this needs compile-fail tests and I haven't figured out how to write those yet.


hcpl:
I added missing test cases as UI tests.

Remote type-spanned errors still not finished yet, some internal restructuring needs to be made.

PR #1426, opened 2018-11-22
Add prefix_all attribute

Phlosioneer:
The prefix_all attribute will add a given prefix to all the fields in
a struct, or all the variants in an enum. It is applied
after `rename_all` and respects `rename`.

This feature is incredibly useful when dealing with enum encodings that prefix each variant with the name of the enum. For example, in some work I'm doing:
```json
{
  "result": "NetworkResult_Success"
}
```

I may not have chosen the best way to add the prefix to a string. It required an extra allocation per rename. I did my best to keep it efficient.

I ran clippy, and added tests. I'm also opening a corresponding PR in the book's repo to add documentation for it. If I've forgotten anything, let me know, and I'll do my best to reply promptly.

### DIFF
--- a/serde_derive/src/internals/ast.rs
+++ b/serde_derive/src/internals/ast.rs
@@ -86,11 +86,17 @@ impl<'a> Container<'a> {
             Data::Enum(ref mut variants) => {
                 for variant in variants {
                     variant.attrs.rename_by_rule(attrs.rename_all());
+                    if let Some(prefix) = attrs.prefix_all() {
+                        variant.attrs.apply_prefix(prefix);
+                    }
                     for field in &mut variant.fields {
                         if field.attrs.flatten() {
                             has_flatten = true;
                         }
                         field.attrs.rename_by_rule(variant.attrs.rename_all());
+                        if let Some(prefix) = variant.attrs.prefix_all() {
+                            field.attrs.apply_prefix(prefix);
+                        }
                     }
                 }
             }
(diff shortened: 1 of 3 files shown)


Phlosioneer:
I had no idea serde_with was a thing at all. I didn't think it was possible to extend serde_derive as a 3rd party crate.

As to your point: I think it's fairly common. But I understand. It was far, far easier for me to extend serde_derive than to make a custom proc macro; and I couldn't figure out how to wrangle the deserialize trait to be anything resembling "compose-able", so that wasn't an option. If you don't want it, I can just keep my patched version.

I will say this: _without_ this feature, prefix systems are awful. I had ~140 `#[serde(rename = "...")]` that were just prefixes.

dtolnay:
> As to your point: I think it's fairly common.

It may be common but I am still a bit skeptical because the only previous request for a prefix_all sort of thing would not have been addressed by this change. The use case is described [here](https://docs.rs/serde_with/0.2.3/serde_with/macro.with_prefix.html#example). They needed to use different prefixes across different uses of the same type.

I am closing the PR because I would like to move in the direction of making serde_derive integrate better with logic found in external libraries like [serde_with](https://docs.rs/serde_with/0.2.4/serde_with/) and [serde_aux](https://docs.rs/serde-aux/0.6.1/serde_aux/) rather than adding more logic into serde_derive. I would love your help on that sort of design.

Thanks anyway!

Phlosioneer:
I'd be willing to help set that up. I've gotten much more familiar with the design. If something like this could easily be written as a plugin sort of thing, it would be just as good. I think it's also important to suggest ways to extend serde's functionality in the documentation, so that more people will think to search for crates like `serde_with` and `serde_aux`.

Is there an issue for discussion about options?

dtolnay:
Filed as #1427.

Issue #1427, opened 2018-11-24
Plugins for serde_derive


dtolnay:
I would like to have serde_derive integrate better with use-case-specific logic in external crates, rather than adding further use-case-specific logic into serde_derive.

As an example, #1426 wants to prefix every variant of an enum with some constant prefix. One way would be to build this logic into serde_derive.

```rust
#[derive(Deserialize)]
#[serde(prefix_all = "NetworkResult_")]
enum NetworkResult {
    Success, // deserializes from "NetworkResult_Success"
}
```

As there can be a question of how common a use case is and whether that justifies the tradeoff in compile time and complexity in serde_derive, it would be nicer to have such logic provided outside of serde_derive. The precise mechanism would need to be designed.

Observe that with the stabilization of proc_macro_attribute in Rust 1.30, one could write an attribute macro that inserts serde attributes into a derive invocation.

```rust
// expands to #[serde(rename = "NetworkResult_Success")] on the variant
#[serde_prefix_all = "NetworkResult_"]
#[derive(Deserialize)]
enum NetworkResult {
    Success,
}
```
 
If necessary, serde_derive could provide some basic indirection to make the invocations feel less jarring.

```rust
extern crate serde_prefix_all;
use serde_prefix_all::prefix_all;

#[derive(Deserialize)]
#[serde(apply = "prefix_all", prefix = "NetworkResult_")] // expands to #[prefix_all(prefix = "NetworkResult_")]
enum NetworkResult {
    Success,
}
```

Part of this work would involve polishing up [`serde_derive_internals`](https://docs.rs/serde_derive_internals/0.24.1/serde_derive_internals/) because many plugins would presumably want to parse serde attributes. For example in the prefix_all use case the plugin would need to be aware of `serde(rename_all = "...")` and `serde(rename = "...")` attributes and contain logic that respects them. So far `serde_derive_internals` has been an internal dependency with no attention to its public API.

We will also need to catalogue some known use cases that would be met by plugin support. Digging through support tickets or feature requests or Stack Overflow questions could help put together a compelling list.

And we should consult with the authors of crates like [serde_with](https://docs.rs/serde_with/0.2.4/serde_with/) and [serde_aux](https://docs.rs/serde-aux/0.6.1/serde_aux/) for buy-in once a design has been fleshed out.

Mentioning: @Phlosioneer

Phlosioneer:
[...]

PR #1429, opened 2018-11-26
Refactor the renaming code

Phlosioneer:
It's now concentrated in the Name struct. Some clones were avoided,
too.

Partly in preparation for #1427

### DIFF
--- a/serde_derive/src/de.rs
+++ b/serde_derive/src/de.rs
@@ -1855,7 +1855,7 @@ fn deserialize_untagged_newtype_variant(
 }
 
 fn deserialize_generated_identifier(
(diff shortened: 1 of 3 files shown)


Phlosioneer:
Oops, forgot to check back for test failures.

Phlosioneer:
Well then... how do you propose to make _any_ progress on #1427? If the API can't be changed, then trying to make room for plugins is pointless; they would have to re-do much of the work in `serde_derive_internals`, and it raises the barrier of entry extremely high (you need to have a thorough understanding of how proc macros work).

If a tiny refactoring like this is too much, how do you propose to add any way for plugins?

PR #1432, opened 2018-11-29
Add serde_postgres to list of data formats

Aaronepower:


Aaronepower:
@dtolnay Okay, I've updated it so now all public items are documented. Let me know if there are any more needed changes.

Issue #1437, opened 2018-12-05
Enums inside untagged enums cannot decode in formats which use variant index rather than name


daboross:
In any format which encodes an enum as its index rather than it's name, if an enum is used within data for an untagged enum, the deserialization fails.

Example of what I mean:

```rust
#[derive(Serialize, Deserialize, Debug, PartialEq)]
#[serde(untagged)]
enum UntaggedEnum {
    A(InnerEnum),
}

#[derive(Serialize, Deserialize, Debug, PartialEq)]
enum InnerEnum {
    B,
    C,
}
```

`UntaggedEnum::A(InnerEnum::B)` will succeed a round-trip in formats which encode enum variants as string names, but it will fail in formats which use the variant index.

This was introduced in https://github.com/serde-rs/serde/commit/e1db820c9fa614eceb23baa4e7f2cbd4fc2dfa21#diff-751ea35094a96d89c2d04ff6a61e703aR2139 and serde v1.39.0.

----

The actual problem is two-fold.

First, when `InnerEnum` deserializes it's enum variant, it uses `deserializer.deserialize_identifier`:

```rust
impl<'de> _serde::Deserialize<'de> for __Field {                                      
    #[inline]                                                                         
    fn deserialize<__D>(__deserializer: __D) -> _serde::export::Result<Self, __D::Error>
    where                                                                               
        __D: _serde::Deserializer<'de>,                                                 
    {                                                                                   
        _serde::Deserializer::deserialize_identifier(__deserializer, __FieldVisitor)    
    }                                                                                   
} 
```

The generated `__FieldVisitor` accepts `u64` for the variant index, `str` for the variant name, or `bytes` for the variant name. But when deserializing, it requests an identifier.

The second part is that `ContentRefDeserializer` and `ContentDeserializer` only pass through string/byte values when `deserialize_identifier` is requested. [This method](https://github.com/serde-rs/serde/blob/v1.0.80/serde/src/private/de.rs#L1421) rejects integers for `deserialize_identifier`.

When these two facts are combined, it makes it impossible for `InnerEnum` inside `UntaggedEnum` to be deserialized in any format which uses variant index rather than name.

Background for `Context`:

When a `#[serde(untagged)]` enum deserializes data, it first deserializes it into a `Content`, then passes `ContentRefDeserializer` and/or `ContentDeserialize` to each variant in turn, exiting as soon as one succeeds.

---

[...]

PR #1438, opened 2018-12-05
Accept integers variant encoding in Content

daboross:
This allows `ContentDeserializer` and `ContentRefDeserializer` to deserialize unsigned integers in `deserialize_identifier`, which then allows enums inside untagged enums to be correctly decoded in formats which encode the enum variant as an integer.

Fixes https://github.com/serde-rs/serde/issues/1437. CC https://github.com/3Hren/msgpack-rust/issues/178

### DIFF
--- a/serde/src/private/de.rs
+++ b/serde/src/private/de.rs
@@ -1419,6 +1419,10 @@ mod content {
                 Content::Str(v) => visitor.visit_borrowed_str(v),
                 Content::ByteBuf(v) => visitor.visit_byte_buf(v),
                 Content::Bytes(v) => visitor.visit_borrowed_bytes(v),
+                Content::U8(v) => visitor.visit_u8(v),
+                Content::U16(v) => visitor.visit_u16(v),
+                Content::U32(v) => visitor.visit_u32(v),
+                Content::U64(v) => visitor.visit_u64(v),
                 _ => Err(self.invalid_type(&visitor)),
             }
         }
@@ -2121,6 +2125,10 @@ mod content {
                 Content::Str(v) => visitor.visit_borrowed_str(v),
                 Content::ByteBuf(ref v) => visitor.visit_bytes(v),
                 Content::Bytes(v) => visitor.visit_borrowed_bytes(v),
+                Content::U8(v) => visitor.visit_u8(v),
+                Content::U16(v) => visitor.visit_u16(v),
+                Content::U32(v) => visitor.visit_u32(v),
+                Content::U64(v) => visitor.visit_u64(v),
                 _ => Err(self.invalid_type(&visitor)),
             }
         }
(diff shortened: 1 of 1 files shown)


daboross:
For msgpack support, no - `rmp-serde` does an optimization of storing any unsigned integer value as the smallest type that supports the value. So in practice most enum variants, being less than `256`, get encoded as `u8`. The decoding doesn't do any translation so it will just decode to `u8`.

This could be changed in `rmp-serde`, but I _think_ that would be a breaking change for that crate, and this would still be broken in the meantime.

dtolnay:
If the smallest type is used, can we support only u8 for now, until such time as someone specifically requests for u16 to work as well?

hcpl:
> can we support only u8 for now, until such time as someone specifically requests for u16 to work as well?

I'm afraid the potential of u8 -> u16 change in representation can cause subtle breakages not apparent from reading the user's code alone (e.g. if the format relies on the exact type to calculate offsets).

How about making it configurable, like `#[repr(u8)]` in Rust? This does increase complexity and API surface support, but also eliminates these kinds of problems in the future.

The default repr is `u32`, for backwards compatibility.

daboross:
> I'm afraid the potential of u8 -> u16 change in representation can cause subtle breakages not apparent from reading the user's code alone (e.g. if the format relies on the exact type to calculate offsets).

As far as I know all `#[derive(Serialize)]` serializers will serialize variants as u32, and it's only a quirk of msgpack that it ever gets converted to anything smaller.

As long as we're OK with only supporting enums up to 256 variants in size, dropping u16 sounds alright.

Part of me wants to say "in optimized code rust should be able to figure out that `visit_u8`,`visit_u16`,`visit_u32` is all forwarded to `visit_u64` for enum field visitors and collapse the three branches into one anyways", but I know relying on advanced optimizations like that is flaky.

> How about making it configurable, like #[repr(u8)] in Rust? This does increase complexity and API surface support, but also eliminates these kinds of confusions in the future.

I'm not entirely sure how this would work. This is internal to `ContentDeserializer`, something that as far as I know doesn't receive any configuration at the moment. Even if we did accept configuration, who would give it, and what exactly would it configure? 

I guess we could have something like `const ACCEPTS_U16_IDENTIFIER: bool` / `const ACCEPTS_U8_IDENTIFIER: bool` in `Visitor` but I'm inclined to view that as way too much complexity for the gain of a few bytes of program binary per enum deserialized inside a `#[serde(untagged)]` enum.

Do you have a concrete plan for how this configuration could work and who would use it, when?

dtolnay:
Let's start with supporting only u8 for now.

In serde_derive 2.0 we may try to gate this sort of code behind Cargo cfgs so that for example Content only knows how to deserialize numeric enum variants if msgpack (or a different format that enables that cfg) is present in your dependency tree.

daboross:
I've pushed a commit to drop `u16` and `u64` in favor of supporting only `u8` and `u32`. I left `u32` in since bincode does serialize variants as `u32`, unlike `msgpack-rust`.

I can squash the commits if necessary.

dtolnay:
Bincode does not support untagged enums so I would remove u32 as well.

daboross:
Oh - sounds good.

daboross:
Sorry I've been slow on the uptake for this. Would it be best if I squashed my commits on this branch or is the format alright as is?

PR #1447, opened 2018-12-20
Support Container attributes rename_all only for Serialize or Deserialize

vincascm:
Support Container attributes rename_all only for Serialize or Deserialize


dtolnay:
Could you make sure that your vinoca.org email is associated with your GitHub account so that these commits are correctly attributed to you? https://github.com/settings/emails

Issue #1456, opened 2019-01-13
Allow a field to map to multiple different external field names when deserializing.


Lymia:
I am parsing a file format where a field can appear with multiple different names when deserializing. Currently, it seems the only similar option for this is `#[serde(rename = "foo")]`, which no longer allows the field to be deserialized under its original name.

It would be nice to add a `#[serde(alias = "foo")]` attribute that would allow a field to be deserialized under multiple different names. For example:

```rust
#[derive(Deserialize)]
struct Struct {
    #[serde(alias = "foo")] bar: u32,
}
```

This feature would allow, for example, both `{"foo": 0}` and `{"bar": 0}` to deserialize to `Struct { bar: 0 }` when using `serde_json`. This seems like it would be relatively simple to implement, and I would be willing to PR an implementation of this if there's no problems with the feature idea.

dtolnay:
How would you specify a field that is deserializable under two different names neither of which is its Rust field name?

Check out #595 which is maybe more intuitive in that case. I would accept a PR to support #595.

Lymia:
My thought process was that it would be something like `#[serde(alias = "alias1", alias = "alias2", rename = "alias3")]`, but I the multiple rename syntax is certainly more intuitive when you don't want your Rust field name as a deserialize name.

I suppose my only objection to #595 is that it's somewhat clunky to use for my use case, because I'd have to write `#[serde(rename(deserialize = "foo", deserialize = "bar"))] foo: u32` for all the fields with aliases.

This isn't a huge deal in my case, because I only have a handful of them, but I can see it getting out of hand in formats with more. It could be an option to implement both, I guess, since many of the same code changes would support both.

dtolnay:
Sounds good. I would accept an implementation where `serde(alias = "...")` is equivalent to `serde(rename(deserialize = "..."))` except that we keep accepting the original field name as long as all of the renames are specified by `alias` and not by `rename`.

dtolnay:
Implemented in #1458. Thanks again!

Issue #595, opened 2016-10-19
Support multiple deserialize renames for the same field


dtolnay:
``` rust
#[derive(Serialize, Deserialize)]
struct S {
    #[serde(rename(deserialize = "old_name", deserialize = "new_name"))]
    new_name: u8
}
```

Asked by @SuperFluffy in #rust.

dtolnay:
@SuperFluffy for now the workaround would be to implement Deserialize something like this:

```rust
impl Deserialize for S {
    fn deserialize<D>(deserializer: &mut D) -> Result<Self, D::Error>
        where D: Deserializer
    {
        #[derive(Deserialize)]
        struct Helper {
            old_name: Option<u8>,
            new_name: Option<u8>,
        }

        let helper: Helper = try!(Deserialize::deserialize(deserializer));

        let s = /* do the conversion, maybe fail if both are set */;
        Ok(s)
    }
}
```

scull7:
I believe this would be useful with type of structure also: https://github.com/serde-rs/serde/issues/970#issuecomment-438153519

dtolnay:
Implemented in #1458.

PR #1458, opened 2019-01-15
Implements alias annotation and allow multiple deserialization renames.

Lymia:
This implements #1456 and #595.

Some things in the implementation I'm not sure of:
 * This is only implement for fields and enum variants. While it'd make sense on containers for some formats like RON, it doesn't seem like the API actually allows for this.
 * The current implementation does not reflect the fact that it accepts aliases in the `VARIANTS`/`FIELDS` arrays passed to `deserialize_struct`, etc. I believe this is the right behavior because encoders like `bincode` care about the length of this list, and the `flatten` attribute causes field names that don't appear in this list anyway, but I can't be 100% sure.


dtolnay:
Released in serde_derive 1.0.85.

PR #1466, opened 2019-01-29
Impl Serialize/Deserialize for std::ops::{Bound, RangeFrom, RangeTo, RangeToInclusive}

0nkery:
Implemented `Serialize` and `Deserialize` traits for various interval-related types from `std`.

Could help me to setup `cfg`s etc properly?

### DIFF
--- a/serde/build.rs
+++ b/serde/build.rs
@@ -14,6 +14,15 @@ fn main() {
     let target = env::var("TARGET").unwrap();
     let emscripten = target == "asmjs-unknown-emscripten" || target == "wasm32-unknown-emscripten";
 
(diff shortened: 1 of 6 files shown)


dtolnay:
Could you explain more about your use case for these impls? How is all of this going to be used in what you are building?

0nkery:
Yep, that's right.

dtolnay:
Thanks, I merged the PR but pared it down to just the Bound impls in a later commit. I am not opposed to the other impls but I would prefer not to accept code that I don't know of anyone wanting to use.

Published in 1.0.86.

0nkery:
Thanks! We've decided to use `Bound` only anyway.

PR #1472, opened 2019-01-31
Add a bunch of tuple impls

maplant:
I need a lot more auto tuple impls for some macros I'm righting.
Unfortunately, rust doesn't have variadic generics :-(


maplant:
That is unfortunate, serde compiles so quickly on my machine I didn't notice.
Perhaps there's a middle ground on increasing the number that might be reachable? 
I guess it's probably reasonable to just punt on this issue until the core language problem has been fixed.

Regardless, for anyone else who experiences this issue while auto generating code, there is a solution: fold a tuple into pairs of pairs

PR #1483, opened 2019-02-17
Implement representing enums as integers

RalfJung:
I have not worked with syn/quote before, so I hope this makes some sense. I have no idea what the point of all the phantom stuff in `__Visitor` is, but I copied that from elsewhere in the same file. Also, I used `u64` because that's what is used in the [example in the docs](https://serde.rs/enum-representations.html), but I've been wondering if `u128` wouldn't be more correct. Maybe it should be `#[serde(u64)]` to leave room for expanding that to more integer types in the future?

Fixes https://github.com/serde-rs/json/issues/349

### DIFF
--- a/serde_derive/src/de.rs
+++ b/serde_derive/src/de.rs
@@ -1675,6 +1675,7 @@ fn deserialize_integer_enum(
         .filter(|variant| !variant.attrs.skip_deserializing())
         .enumerate()
     {
+        assert!(variant.attrs.deserialize_with().is_none());
         let variant_ident = &variant.ident;
         let const_name = Ident::new(&format!("__variant{}", idx), Span::call_site());
         consts.push(quote!{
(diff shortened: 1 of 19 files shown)


dtolnay:
> Also, I used u64 because that's what is used in the example in the docs, but I've been wondering if u128 wouldn't be more correct.

I don't think either is what we want. We don't want all such enums taking 16 bytes in bincode, for example. The serde-repr derives require a `#[repr(uN)]` attribute that they use as the serialized representation.

RalfJung:
> This one is implemented in [dtolnay/serde-repr](https://github.com/dtolnay/serde-repr). Sorry, looks like I didn't find that serde_json issue when publishing it. Thanks anyway.

Well, diving into serde for one afternoon was still interesting and I learned some things, so it wasn't entirely in vain. ;)

PR #1484, opened 2019-02-17
Array impls using MaybeUninit

dtolnay:
This commit improves compile time of Serde by 13% in exchange for three lines of unsafe code deserializing fixed size arrays and 1.9x slower performance on the following array deserialization microbenchmark:

```rust
#![feature(test)]

extern crate test;
use test::Bencher;

#[bench]
fn bench(b: &mut Bencher) {
    let array: [u32; 32] = [
        01, 02, 03, 04, 05, 06, 07, 08, 09, 10,
        11, 12, 13, 14, 15, 16, 17, 18, 19, 20,
        21, 22, 23, 24, 25, 26, 27, 28, 29, 30,
        31, 32,
    ];
    let bytes = bincode::serialize(&array).unwrap();
    assert_eq!(bytes.len(), 32 * 4);
    b.iter(|| {
        for _ in 0..100 {
            let _ = bincode::deserialize::<[u32; 32]>(&bytes).unwrap();
        }
    });
}
```

Will need to do some profiling and figure out whether the performance can be recouped.

### DIFF
--- a/serde/src/de/impls.rs
+++ b/serde/src/de/impls.rs
@@ -4,6 +4,9 @@ use de::{
     Deserialize, Deserializer, EnumAccess, Error, SeqAccess, Unexpected, VariantAccess, Visitor,
 };
 
+#[cfg(manually_drop)]
+use backport::MaybeUninit;
+
 #[cfg(any(core_duration, feature = "std", feature = "alloc"))]
 use de::MapAccess;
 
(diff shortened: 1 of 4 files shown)


serde/src/de/impls.rs
```diff
@@ -961,9 +964,67 @@ impl<'de, T> Deserialize<'de> for [T; 0] {
...
+    {
+        let mut guard = PartialArrayDrop {
+            mem: self.mem,
+            initialized_count: 0,
+        };
+        while guard.initialized_count < self.len {
+            match try!(seq.next_element()) {
+                Some(element) => {
+                    unsafe {
+                        ptr::write(self.mem.offset(guard.initialized_count as isize), element);
```
oli-obk:
```suggestion
                        ptr::write(self.mem.add(guard.initialized_count), element);
```

(stable since 1.26)

serde/src/de/impls.rs
```diff
@@ -961,9 +964,67 @@ impl<'de, T> Deserialize<'de> for [T; 0] {
...
+    mem: *mut T,
+    initialized_count: usize,
+}
+
+#[cfg(manually_drop)]
+impl<T> Drop for PartialArrayDrop<T> {
+    fn drop(&mut self) {
+        for i in (0..self.initialized_count).rev() {
+            unsafe {
+                ptr::drop_in_place(self.mem.offset(i as isize));
```
oli-obk:
```suggestion
                ptr::drop_in_place(self.mem.add(i));
```

PR #1486, opened 2019-02-20
impl Serialize and Deserialize for core::cmp::Reverse

vorot93:


dtolnay:
Is it possible to share your code that would need to use these impls? I have not come across Reverse\<T\> needing to be serialized.

vorot93:
I see it as a very convenient way to serialize a sorted collection in the reverse order. Like this:

```
let mut my_coll = BTreeSet::new();
my_coll.insert(Reverse(1));
my_coll.insert(Reverse(2));
my_coll.insert(Reverse(3));

assert_eq!(serde_json::to_string(&my_coll).unwrap(), "[3, 2, 1]");
```

dtolnay:
Thanks, that's fair. You will need to add a build flag in build.rs because core::cmp::Reverse is new in 1.19.0 and we support back to 1.13.0.

vorot93:
@dtolnay Done.

dtolnay:
Released in 1.0.89.

PR #1519, opened 2019-05-04
de: Make flattened structs and aliases interact correctly.

### LINKED ISSUE #1504: Field aliases do not work in combination with `flatten`

### Expected behavior

### Observed behavior

### How to reproduce

```
use serde_json;

pub struct A {
    x: i64,
}

pub struct B {
    #[serde(flatten)]
    a: A,
}

     assert_eq!(b.a.x,12);
     Ok(())
}
```

emilio:
The issue is that FlatStructAccess has no access to the aliases of the struct
it's deserializing.

Ideally we'd try to serialize the key rather than checking whether we're going
to use it before-hand, then actually take the value out, but that happens to be
tricky with the current seed API.

So we need to somehow get the aliased names back to FlatStructAccess. Introduce
a serialize_struct-like API that takes them in a backwards-compatible way. For
parallelism, and since we also support aliases on enum variants, also extend the
struct_variant API in a similar way. I'm open to better ways to fix it, but I
can't think of any other that isn't a breaking change...

Fixes #1504.

### DIFF
--- a/serde/src/de/mod.rs
+++ b/serde/src/de/mod.rs
@@ -1112,6 +1112,28 @@ pub trait Deserializer<'de>: Sized {
     where
         V: Visitor<'de>;
 
+    /// Hint that the `Deserialize` type is expecting a struct with a particular
+    /// name and fields.
+    ///
+    /// `_fields_with_aliases` includes all valid field names, including
+    /// aliases.  `fields` only includes the canonical struct fields.
+    ///
+    /// Use this if you care about aliased fields. For backwards compatibility,
+    /// by default this calls `deserialize_struct`. If you implement this, you
+    /// probably want `deserialize_struct` to call this instead.
+    fn deserialize_struct_with_aliases<V>(
+        self,
+        name: &'static str,
+        fields: &'static [&'static str],
+        _fields_with_aliases: &'static [&'static str],
+        visitor: V,
+    ) -> Result<V::Value, Self::Error>
+    where
+        V: Visitor<'de>
+    {
+        self.deserialize_struct(name, fields, visitor)
+    }
+
     /// Hint that the `Deserialize` type is expecting an enum value with a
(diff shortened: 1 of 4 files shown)


emilio:
That's sort-of a breaking change, plus other `serialize_struct` implementations that use the `fields` use it as a hint to pre-allocate storage, which in the case of aliases may vastly over-allocate.

dtolnay:
Ah, I haven't heard of Deserializer impls pre-allocating storage based on deserialize_struct. Do you have an example of a format like that? Ordinarily it would be the other way around; the Deserializer impl is aware how much data is in the input, and the Deserialize impl needs to pre-allocate based on MapAccess::size_hint.

emilio:
Blerg, I was misreading code (was reading Serializer impls indeed).

But searched for a bit and found bits that would look at best fishy with that change. For example bincode serializes structs as tuples, and takes `fields.len()`:

 * https://searchfox.org/mozilla-central/rev/06578bfadb5bdc5faee81f7e9b3c3fed21e616e0/third_party/rust/bincode/src/de/mod.rs#361

I haven't dug about whether this would break bincode for sure. I'm happy to dig out a bit. If it's just lying about a too large size hint it may be ok I guess....

dmarcuse:
Is there any change to the status of this PR? I just ran into this issue in one of my projects, and it would be nice to have it fixed assuming this implementation covers all bases

jnicholls:
Same here. It's August and no movement. @dtolnay @emilio can this conversation be resuscitated or should we take up the mantle?

jplatte:
Ping @dtolnay

cosmicchipsocket:
Ran into this while trying to make field name changes in a backward-compatible way.

dtolnay (commented):
Fixed a different way by #2387.

Issue #1524, opened 2019-05-12
Feature Request: Support for `[serde(try_into = "type", try_from = "type")]`


fanzeyi:
A little background, I'm trying to make toml work with one of the enum I have (`toml` does not work with enums). I discovered the `serde(into = "type", from = "type")` attributes, and it works perfectly with my case. However, I do want to have the ability to handle unexpected input when serializing/de-serializing the enum. 

I think it would be nice if serde could support a similar attribute `try_into` and `try_from`. An example could be:

```rust
use std::convert::TryFrom;

#[derive(Debug, Clone, Deserialize)]
#[serde(try_from = "String")]
enum Color {
    Rgb(u8, u8, u8)
}

impl TryFrom<String> for Color {
    type Error = String;

    fn try_from(value: String) -> Result<Self, Self::Error> {
        // ... manually parsing string
        Err("Some error message".to_string())
    }
}
```

One complication of supporting `try_into` and `try_from` is error handling. I'm not familiar with implementing serializer and deserializer with serde but I saw there is `Error::Message` in the [example](https://serde.rs/error-handling.html) of implementing a data format. I wonder if it possible to have the user to return a String error message and passed that to the actual seralizer/deserializer etc.

Please let me know what do you think about this idea. Thanks!

dtolnay:
I would accept a PR for a `try_from` attribute. I would prefer not to do `try_into` yet until someone justifies a use case for it. The existing `into` attribute is already sort of confusing with the way it needs to clone self.

davidbarsky:
Just wanted to check—I think this issue could be closed because #1526 is merged?

PR #1526, opened 2019-05-12
Add serde attribute `try_from`

fanzeyi:
This pull request adds a new serde attribute `try_from` which will make serde try to convert the type specified in the attribute to the type serde is trying to deserialize into. (#1524)

For example,

```rust
use serde::Deserialize;
use serde_json;
use std::convert::TryFrom;

#[derive(Deserialize, Debug)]
#[serde(try_from = "u32")]
enum Color {
    Red,
}

impl TryFrom<u32> for Color {
    type Error = String;

    fn try_from(value: u32) -> Result<Self, Self::Error> {
        if value > 10 {
            Err("out of range".into())
        } else {
            Ok(Color::Red)
        }
    }
}

fn main() {
    let result: Color = serde_json::from_str("2").unwrap();
    println!("result = {:?}", result);
    let _: Color = serde_json::from_str("100").unwrap();
}
```

will produce the following output:

```
result = Red
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error("out of range", line: 0, column: 0)', src/libcore/result.rs:997:5
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

## Details

* `serde_derive` will generate the `try_from` call when generating deserializing code
* The generated code will map the result type to a call to `serde::Error::custom`
* Similar to `from` attribute, `try_from` is incompatible with `transparent`
* `from` and `try_from` attributes cannot be specified at the same time

### DIFF
--- a/serde/src/export.rs
+++ b/serde/src/export.rs
@@ -1,5 +1,5 @@
 pub use lib::clone::Clone;
-pub use lib::convert::{From, Into};
+pub use lib::convert::{From, Into, TryFrom};
 pub use lib::default::Default;
 pub use lib::fmt::{self, Formatter};
 pub use lib::marker::PhantomData;
--- a/serde_derive/src/de.rs
+++ b/serde_derive/src/de.rs
@@ -269,6 +269,8 @@ fn deserialize_body(cont: &Container, params: &Parameters) -> Fragment {
         deserialize_transparent(cont, params)
     } else if let Some(type_from) = cont.attrs.type_from() {
         deserialize_from(type_from)
+    } else if let Some(type_try_from) = cont.attrs.type_try_from() {
+        deserialize_try_from(type_try_from)
     } else if let attr::Identifier::No = cont.attrs.identifier() {
         match cont.data {
             Data::Enum(ref variants) => deserialize_enum(params, variants, &cont.attrs),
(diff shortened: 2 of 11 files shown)


fanzeyi:
Hmm... This will break serde's backward compatibility (1.13+) as try_from / try_into just stabilized 😢 

Would it be acceptable to make it a feature?

bluebear94:
Would it be possible to make this feature conditionally available to versions that support `TryFrom`?

davidbarsky:
@fanzeyi Hi! I'm not a maintainer, but looking at how the `alloc` feature is supported in Serde on Rust versions later than 1.36, I can _somewhat_ confidently infer that placing the `try_from` behind a feature flag is the correct approach. Depending on how much time you have, I'd be happy to do this myself in a separate PR that builds atop of yours.

fanzeyi:
@davidbarsky Let me ask @dtolnay today.

davidbarsky:
Ah, my bad for misunderstanding. Thanks for the clarification!

jhpratt:
@dtolnay Slightly odd request, but would it be possible to allow a custom `TryFrom` implementation for back-compatibility? I was looking into dropping the MSRV for the time crate from 1.34 to 1.32, and am using `TryFrom` via standback. Standback just copies code from std/core as appropriate, so it's generally compatible.

If it's not possible, I suppose I could write out what is currently derived. I presume it isn't too common to do this, though.

Issue #1533, opened 2019-05-23
Self referring in [serde(from)]


a-tomasi:
serde 1.0.91
serde_json 1.0.39 (it is not his fault)
rustc 1.36-nightly (with 1.34 is the same)

I would expect this code not to compile but it does.

```rust
use serde::{Deserialize};

#[derive(Deserialize)]
#[serde(from="Point")]
struct Point {}

fn main() {
   let _ = serde_json::from_str::<Point>("{}");
}
```
because at runtime this is the result:
  
```
thread 'main' has overflowed its stack
fatal runtime error: stack overflow
Aborted (core dumped)
```
I know that I have written something 'bad' but this is dangerous.

dtolnay:
The generated code is:

```rust
impl<'de> serde::Deserialize<'de> for Point {
    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
    where
        D: serde::Deserializer<'de>,
    {
        Result::map(
            <Point as serde::Deserialize>::deserialize(deserializer),
            From::from,
        )
    }
}
```

which correctly reports a warning if handwritten.

```console
warning: function cannot return without recursing
  --> src/main.rs:12:5
   |
12 | /     fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
13 | |     where
14 | |         D: serde::Deserializer<'de>,
   | |____________________________________^ cannot return without recursing
...
17 |               <Point as serde::Deserialize>::deserialize(deserializer),
   |               -------------------------------------------------------- recursive call site
   |
   = help: a `loop` may express intention better if this is on purpose
   = note: `#[warn(unconditional_recursion)]` on by default
```

We tried putting a `deny(unconditional_recursion)` in #1537 to prevent this bug, but it didn't work due to https://github.com/rust-lang/rust/issues/61504.

Let's try it again after the rustc bug has been fixed.

PR #1537, opened 2019-05-26
Avoid circular #[serde(from="")] using a AvoidFromMyself trait.

treere:
Fix to #1533.

I create a trait (AvoifFromMyself) and I try to impl it for the structure and for the "from".
If the structure and from are the same it cannot compile.

This is the best solution I have found but I hope there is a better one. 
This is a bit too complex.

 ps: the error message is terrible

### DIFF
--- a/serde_derive/src/de.rs
+++ b/serde_derive/src/de.rs
@@ -29,10 +29,17 @@ pub fn expand_derive_deserialize(input: &syn::DeriveInput) -> Result<TokenStream
     let delife = params.borrowed.de_lifetime();
     let serde = cont.attrs.serde_path();
 
+    let avoid_circular_from = quote! {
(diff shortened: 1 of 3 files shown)


treere:
I agree. Use `unconditional_recursion` is better.

I do not understand why you piece of code works but if I add `#[deny(unconditional_recursion)]` in the procedural macro the compiler still compiles my test.

Where can I read something to figure out why?

dtolnay:
I don't know, it could be a compiler bug. If you can put together a minimal repro (from scratch not using serde_derive or any other libraries) then please report it in rust-lang/rust.

treere:
Reported to
https://github.com/rust-lang/rust/issues/61504

Issue #773, opened 2017-02-18
Better message when failing to match any variant of an untagged enum


dtolnay:
This message could be a lot better.

https://github.com/serde-rs/serde/blob/v0.9.7/serde_derive/src/de.rs#L750-L756

```rust
// TODO this message could be better by saving the errors from the failed
// attempts. The heuristic used by TOML was to count the number of fields
// processed before an error, and use the error that happened after the
// largest number of fields. I'm not sure I like that. Maybe it would be
// better to save all the errors and combine them into one message that
// explains why none of the variants matched.
let fallthrough_msg = format!("data did not match any variant of untagged enum {}", type_ident);
```

dtolnay:
Closing as more complicated than it's worth. I would be willing to consider a PR that implements better error messages but I don't plan to pursue this as a priority.

jschaf:
I don't have a PR but I'd like to add my experience with this error in case someone want to pick this up.

I've hit this error as a consumer (not developer) of the [vector] logging binary. The error is confusing because I'm not sure what went wrong.  Specifically, here was the rough mental debugging I went through:

```
 ERROR vector: Configuration error: data did not match any variant of untagged enum 
    LuaConfig for key `transforms.journald_lua_transform`
```

- Did I forget a field? Which one?
- Is there an extra field? Which one?
- Is the error referring to a key called `data`?
- What is an untagged enum? Is it relevant to this error?

[vector]: https://github.com/timberio/vector

abeluck:
> I've hit this error as a consumer (not developer) of the [vector](https://github.com/timberio/vector) logging binary. The error is confusing because I'm not sure what went wrong. Specifically, here was the rough mental debugging I went through:
> 
> ```
>  ERROR vector: Configuration error: data did not match any variant of untagged enum 
>     LuaConfig for key `transforms.journald_lua_transform`
> ```

@jschaf sounds like i'm hitting the exact error you are (even down to `transforms.journald_lua_transform`). What was your eventual fix?

jschaf:
> @jschaf sounds like i'm hitting the exact error you are (even down to transforms.journald_lua_transform). What was your eventual fix?

Sorry, I've forgotten most of the context. It might have been mismatch between the docs which show version 2 rules, and the vector implementation at the time which defaulted to version 1. I remember diving through the vector source code to figure out what version 1 expected.

If it helps, here's my current working invocation of that transform:

```toml
# Transform to extract whitelisted fields from the JournalD src.
[...]

PR #1544, opened 2019-06-01
Collect errors when deserializing untagged enums

killercup:
This also adds a verbose-debug feature that enables the new behavior.

Previously:

> data did not match any variant of untagged enum ReferenceOr at line 6 column 4

Now, with `verbose-debug` feature enabled:

> data did not match any variant of untagged enum `ReferenceOr`
> 	- attempted to deserialize `Reference` but failed with: missing field `$ref`
> 	- attempted to deserialize `Item` but failed with: invalid value: string "lol", expected expected format `\dXX` at line 6 column 4

cf. #773

### DIFF
--- a/serde/Cargo.toml
+++ b/serde/Cargo.toml
@@ -58,3 +58,6 @@ alloc = ["unstable"]
 # does not preserve identity and may result in multiple copies of the same data.
 # Be sure that this is what you want before enabling this feature.
 rc = []
+
+# Opt into verbose errors that are more expensive to compute.
+verbose-debug = ["serde_derive/verbose-debug"]
--- a/serde_derive/Cargo.toml
+++ b/serde_derive/Cargo.toml
@@ -18,6 +18,7 @@ appveyor = { repository = "serde-rs/serde" }
 [features]
 default = []
 deserialize_in_place = []
+verbose-debug = []
 
 [lib]
 name = "serde_derive"
--- a/serde_derive/src/de.rs
+++ b/serde_derive/src/de.rs
@@ -1621,35 +1621,69 @@ fn deserialize_untagged_enum(
         .iter()
         .filter(|variant| !variant.attrs.skip_deserializing())
         .map(|variant| {
-            Expr(deserialize_untagged_variant(
+            let de_any = Expr(deserialize_untagged_variant(
                 params,
                 variant,
                 cattrs,
                 quote!(_serde::private::de::ContentRefDeserializer::<__D::Error>::new(&__content)),
-            ))
+            ));
+
+            if cfg!(feature = "verbose-debug") {
+                let variant_name = variant.ident.to_string();
+                quote! {
+                    _serde::export::Result::map_err(
+                        #de_any,
+                        |e| -> __D::Error {
+                            _serde::de::Error::custom(
+                                format!("attempted to deserialize `{}` but failed with: {}", #variant_name, e.to_string())
+                            )
+                        },
+                    )
+                }
+            } else {
+                quote! { #de_any }
+            }
         });
 
-    // TODO this message could be better by saving the errors from the failed
-    // attempts. The heuristic used by TOML was to count the number of fields
-    // processed before an error, and use the error that happened after the
-    // largest number of fields. I'm not sure I like that. Maybe it would be
-    // better to save all the errors and combine them into one message that
-    // explains why none of the variants matched.
-    let fallthrough_msg = format!(
-        "data did not match any variant of untagged enum {}",
-        params.type_name()
-    );
+    if cfg!(feature = "verbose-debug") {
+        let type_name = params.type_name();
 
-    quote_block! {
-        let __content = try!(<_serde::private::de::Content as _serde::Deserialize>::deserialize(__deserializer));
+        quote_block! {
+            let __content = try!(<_serde::private::de::Content as _serde::Deserialize>::deserialize(__deserializer));
+            let mut fallthrough_msg = format!(
+                "data did not match any variant of untagged enum `{}`",
+                #type_name
+            );
 
-        #(
-            if let _serde::export::Ok(__ok) = #attempts {
-                return _serde::export::Ok(__ok);
-            }
-        )*
+            #(
+                match #attempts {
+                    _serde::export::Ok(__ok) => return _serde::export::Ok(__ok),
+                    _serde::export::Err(__err) => {
+                        fallthrough_msg.push_str("\n\t- ");
+                        fallthrough_msg.push_str(&__err.to_string());
+                    }
+                }
+            )*
 
-        _serde::export::Err(_serde::de::Error::custom(#fallthrough_msg))
+            _serde::export::Err(_serde::de::Error::custom(fallthrough_msg))
+        }
+    } else {
+        let fallthrough_msg = format!(
+            "data did not match any variant of untagged enum {}",
+            params.type_name()
+        );
+
+        quote_block! {
+            let __content = try!(<_serde::private::de::Content as _serde::Deserialize>::deserialize(__deserializer));
+
+            #(
+                if let _serde::export::Ok(__ok) = #attempts {
+                    return _serde::export::Ok(__ok);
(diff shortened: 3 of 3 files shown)


killercup:
See this more as a draft that just happens to help me debug stuff locally :)

Bikeshedding of feature name and error format very welcome.

WyseNynja:
Just used this to debug some nested enums that had a typo. Very helpful! Thanks!

killercup:
@dtolnay are you interested in merging this? I'm not sure what Serde's approach to verbose debugging is/whether there is any precedence for this.

DerrikMilligan:
What would be holding us back from merging this in? Seems fantastic.

piegamesde:
I cannot get this branch to work. Compiling throws a lot of errors like:

```
error[E0277]: the trait bound `serde_yaml::Mapping: serde::Deserialize<'_>` is not satisfied
  --> src/config.rs:48:2
   |
48 |     #[serde(default)]
   |     ^ the trait `serde::Deserialize<'_>` is not implemented for `serde_yaml::Mapping`
   |
   = note: required by `serde::de::SeqAccess::next_element`
```

Stargateur:
how to use it along with serde_json cause I can override serde version but not serde_json version of serde

WyseNynja:
You can use `[patch]` in your Cargo.toml. Check out this section of the rust book: https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html

Stargateur:
> 
> You can use `[patch]` in your Cargo.toml. Check out this section of the rust book: https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html
> […](#)
> On Aug 9, 2020, at 4:09 AM, Antoine PLASKOWSKI ***@***.***> wrote: ﻿ how to use it along with serde_json cause I can override serde version but not serde_json version of serde — You are receiving this because you commented. Reply to this email directly, view it on GitHub, or unsubscribe.

This would work cause there would be two version of serde, and so as @piegamesde reported this cause Deserialize to not be implement for serde_json version of serde.

yaymukund:
Thanks for doing this work. Out of curiosity, is this behind a flag because the message is too long or because collecting the debugging information comes at a performance cost (or something else)?

killercup:
Just had a use for this again because the error messages serde gives me when reading in a config file are not helpful to users.

@dtolnay I'd really appreciate if you could have a look at this :)

andrewdavidmackenzie:
+1 for adding this or more extensive error messages - some failures are *very* hard to debug.

calavera:
Can any of the maintainers take a look at this? Debugging deserialization problems into enums is basically impossible without this.

panther99:
I know it has been asked already but I think maintainers should seriously consider checking this. It looks like a small change and it will be of huge help to all of us using untagged enums in our projects.

torkleyy:
Is there any reason why this is not being considered? This would be so helpful for actually finding errors in YAML files.

sztomi:
I really need something like this. Unfortunately I was not able to use the git version in the PR repo because it's a bit outdated. My use-case: I'm querying an API that either returns an error JSON or the value requested. The API is apparently unstable because after some time, my code no longer works. I'm trying to work out what's not matching in the JSON, but it's really difficult because it's a huge chunk and it almost fits. Is it a type not matching? Is it a field that became optional? I have no way of knowing without looking at each field in my struct and trying to manually match that against the JSON I got. 

My workaround is likely going to be: identify an error response without deserializing (i.e. dumb string matching) and then deserializing only the concrete types and use serde_path_to_error to gain some insight.

Stargateur:
@dtolnay ?

nirvana-msu:
@dtolnay there seems to be a fair amount of interest for this issue. We'd all appreciate if you let us know what your concerns are, and if you see a path for it to get merged.

kurtbuilds:
For anyone who encounters this issue today, you need a branch where the serde version matches your other dependencies.

I took a similar commit (that uses no compiler feature flag) from this branch: https://github.com/jonasbb/serde/tree/dev and cherry picked it on top of master.

Result:
https://github.com/kurtbuilds/serde is version 1.0.144 of serde, and it will give you better error messages. Use it with these lines in your Cargo.toml

```toml
[patch.crates-io]
serde_derive = { git = "https://github.com/kurtbuilds/serde" }
```

Hopefully this is mainlined soon.

AlexTMjugador:
Serde has such a fast release cadence that the repository that @kurtbuilds kindly provided is already out of date. If anyone wants this patch, they practically need to keep their own automatically-updated fork with the patch (using tools such as the `wei/pull` GitHub Action), or else bend over backwards to pin to a Serde version and make sure that nothing ends up pulling a slightly different one, including dependencies of dependencies. Both approaches are inconvenient and contribute to technical debt in downstream applications.

What has been blocking this improvement for more than three years now? It's a really, really welcome change for those that use Serde to parse user-provided configuration.

coltfred:
@dtolnay Did you have any thoughts on this approach? When trying to build user facing tools/services this would be a big help.

If you have thoughts on a different approach that's cool too. It'd be great to get some collaboration going on it.

AlexTMjugador:
https://github.com/serde-rs/serde/pull/2403 introduced new merge conflicts for this PR, which I've just fixed in my fork (if anyone's using my fork and notices any problem due to that, please file an issue on the fork). If conflicts are keeping this PR from being merged, I'd be happy to contribute and submit my updated patches upstream.

yaotthaha:
When will this PR be merged? This is really useful! There is so little misinformation now!

Emilgardis:
One way to substitute this is with [serde_path_to_error](https://docs.rs/serde_path_to_error/latest/serde_path_to_error/), for me, that has completely satisfied what this pr did.

AlexTMjugador:
Thank you @dtolnay for your clarification on the state of this PR - I appreciate being decisive and explicit with PRs as important for certain users as this one has been for years.

GeeWee:
When trying to use serde-untagged I also ran into issues with [not being able to roll the parser back](https://github.com/dtolnay/serde-untagged/issues/3) thus not being able to easily deserialize an enum containing multiple structs. However I guess an alternate implementation could potentially either fork/copy over the `Content` API or buffer to some other intermediate struct such as `serde_json::Value` - it does however require re-implementing a fair bit of code for each struct just to get error messages that are decent.

AlexTMjugador:
Adding to some of the points exposed by @33KK:

> Forking serde and serde_derive is obviously complicated since that would also require patching the entire dependency chain to use that fixed version. That is on top of maintaining the fork, keeping it up to date and resolving any conflicts.

[I've been doing that maintenance work for a while now](https://github.com/ComunidadAylas/serde), so other people don't have to. Maintaining that fork has been straightforward so far, barring the time when precompiled binaries where introduced, but they ended up being reverted for a plethora of reasons. That does not solve the problem that patching `serde` and `serde_derive` is only viable for end applications, but fortunately those are the most affected by these cryptic error messages.

> [...] I think this feature is pretty much basic and a must-have, since currently it's basically impossible to tell what went wrong in an untagged enum, especially when it's the top level type or has a lot of children, then the entire error is completely worthless.

While I agree with this, I recently stumbled upon and documented the little-known [`expecting` container attribute](https://serde.rs/container-attrs.html#expecting), which allows setting a custom fixed error message when deserializing untagged enums. By design, such a message lacks any situational information, unlike the changes proposed by this PR, but applications can write it so to be user-friendly enough for most circumstances, and supersede this PR for the applicable use cases.

milesj:
It's a shame this is still an issue. Most of the issues with our configuration that is reported by users is caused by enums/unions swallowing the actual inner error, and we have no way to present that to the user.

We also use `expecting` but that also doesn't really work well. 

For example, say we have a field that accepts a string ID, or a vec of string IDs. Very simple enum, as we can do something like `expecting = "expected a string or list of strings"`. But say we add custom parsing rules for an `ID` type that whitelists the allowed characters. When the user enters an invalid character, that parse error is not bubbled up, and they have no idea why their ID/config isn't working.

PR #1572, opened 2019-07-12
Implement serialization and deserialization for std::sync::atomic types

### LINKED ISSUE #1496: Support for atomics data types?
Hi,

it seems like there is no Serialize support for atomic data type: https://doc.rust-lang.org/core/sync/atomic/index.html

Is there any reason for this?

A first implementation would be to simply load them using `Ordering::Relaxed` and that should be quite enough.

Roguelazer:
Fixes #1496.

The tests for deserialization are kind of cheesy because Atomic* types don't implement PartialEq. Let me know if you want me to try to make them better.

### DIFF
--- a/serde/build.rs
+++ b/serde/build.rs
@@ -68,6 +68,10 @@ fn main() {
     if minor >= 28 {
         println!("cargo:rustc-cfg=num_nonzero");
     }
+
+    if minor >= 34 {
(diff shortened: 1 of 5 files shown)


fdubois1:
I guess it is related : https://github.com/serde-rs/serde/issues/1579
Thanks for your help

PR #1617, opened 2019-08-30
Fix unused variable warning when field of adjacently tagged enum is skipped

arilotter:
Replace skip_seralizing tuple members with `_` in generated serializer.
I'm not totally convinced this is the right solution, since it requires creating 2x the `Vec`s, but it's the most "correct" solution I've found.

### DIFF
--- a/serde_derive/src/ser.rs
+++ b/serde_derive/src/ser.rs
@@ -723,6 +723,8 @@ fn serialize_adjacently_tagged_variant(
             where
                 __S: _serde::Serializer,
             {
+                // Some members of this tuple will be unused if they're `skip_serializing`
+                #[allow(unused_variables)]
                 let (#(#fields_ident,)*) = self.data;
                 #inner
             }


arilotter:
I agree! I wrote a message in the #serde irc channel asking if `#[allow(unused_variables)]` was the way to go, but I thought I'd try without it in case y'all wanted to avoid that :)

I'm very happy to make the fix a simple warning suppression, now my project will finally compile with no warnings :D :pray:

serde_derive/src/ser.rs
```diff
@@ -723,6 +723,7 @@ fn serialize_adjacently_tagged_variant(
             where
                 __S: _serde::Serializer,
             {
+                #[allow(unused_variables)]
```
dtolnay: Please add a comment as a reminder of why this needs to be here, i.e. under what circumstances the variables would be unused.

PR #1618, opened 2019-09-01
Versioning support for deserialization

fredpointzero:
# Summary

This PR provides a versioning feature to support deserialization of previous format structure. 

When deserializing a from a source, we provide a version map which associate the version number to use for each data type. In that case, any type that has an explicit version number will be deserialize from the appropriate type.

Note: some tests are still not green, but this is an early PR to preview the feature. It is expected to be improved based on feedback in this PR before merging.

Note 2: 2 features were added for serde: `versioning` and `versioning-derive`. Use the latter to have versioning supported in generated derive impls.

# Motivation

When software are developped and used at the same time the data formats may change from one version to another and persisting data may be produced by a specific version and loaded by another version.

## Goal of this PR

Versioning is a complex topic and here we try to solve common use cases.

We aim at solving the case were a type or a set of types in a deserializer's data needs to be upgraded to their latest format. This is the case when a mandatory property was added or removed, or an existing property changed.

_Note: There already is support for added optional properties_

Example:
Let's have a file containing these structure with those version number: `A: 1, B: 1, C: 2` and the current version numbers are: `A: 3, B: 2, C: 4`.

Then in latest code version, we have the former data structures versions, let's call them: `Av1, Av2, Bv1, Cv1, Cv2, Cv3`.

Deserializing, whenever a structure `A`, `B` or `C` is ran into, then it is loaded with the appropriate format (in our case it will be `Av1`, `Bv1` and `Cv2`) and then converted to `A`, `B` or `C` using the `From` trait.

## Non goal of this PR

This is based on types that can be upgraded individually. Types that needs to be upgraded together is way more complex to handle and usually relies on domain specific deserializer.

So, these data format should be handle with specific `Deserialize` traits implementations.

## Unsupported Serde feature with versioning

Versioning is not supported for all serde features

### Not supported in `deserialize_in_place` calls

Deserializing in place with versioning support is way more complicated, so we don't deak with this in the current PR.

### Not supported with `deserialize_with` callback

In case of explicit deserialization, versioning is not handled.

### Versioning is only supported for structs and enums

There is no use case where versioning tuples and the unit type is useful.

# Guide-level explanation

## Authoring versions of a type

To describe the previous versions of a type, we use the `#[serde(versions(...))]` attribute.

Example:

```rust
// Version 1 of struct A
// It must implement Deserialize, so it can be loaded by serde
#[derive(Deserialize)]
// It must be identified by A during deserialization
#[serde(rename = "A")]
struct Av1 {
  a: u8
}

// Current version of struct A
// It must implement Deserialize
#[derive(Deserialize)]
// We use the versions attribute to define the previous versions
#[serde(versions("Av1"))]
// So, Version n°1 of A is Av1, Versions n°2 (current) of A is A
struct A {
  // We moved a property
  b: u8
}

// A must implement From for all previous type, so we implement From<Av1>
impl From<Av1> for A {
  fn from(v: Av1) -> Self {
    Self {
      b: v.a
    }
  }
}
```

## Runtime execution

The versioning feature happen in two steps: get the `VersionMap` which provides the version number to use for each specific type and the actual deserialization which uses the `VersionMap` to choose the proper type to deserialize.

The deserializer provides the version_map to use, if the deserializer does not implement the version_map getter, we can use a wrapper for this: `VersionedDeserializer<D>`.

Note: `VersionedDeserializer<D>` is not yet implemented

Example:

```rust
#[derive(Deserialize)]
struct AInMap {
  a: A,
}

fn main() {
  // Use ron as data format for this example
  use ron;
  // First get a header
  // Here, we use the version 1 of `A`
  let versions: serde::VersionMap = ron::de::from_str(r#"{ "A": 1 }"#).unwrap();
  
  // Let's deserialize some values
  
  // Deserialize directly A
  let mut deserializer = serde::de::VersionedDeserializer::new(
    ron::de::Deserializer::from_str(r#"A(a: 1)"#),
    &versions
    );
  let value = A::deserialize(&mut deserializer);
  assert_eq!(value, A { b: 1 });
  
  // Deserialize A contained in a struct property
	let mut deserializer = serde::de::VersionedDeserializer::new(
    ron::de::Deserializer::from_str(r#"AInMap(a: A(a: 2))"#),
    &versions
    );
  let value = AInMap::deserialize(&mut deserializer);
  assert_eq!(value.a, A { b: 2});
}
```

# Reference-level explanation

## Versioning serde attributes

For the versioning to work, 3 constraints must be ensured.

An additional constraint is required for type with `#[serde(default)]`

### Declare previous versions with the `versions` serde attribute

To declare that type `Av1`, `Av2` are the previous versions of type `A`, they must be declared with the `versions` serde attribute on the type `A`.

```rust
#[derive(Deserialize)]
#[serde(versions("Av1", "Av2"))]
struct A { ... }
```

Note: `#[serde(versions("Av1", "Av2"))]` is a shorthand for `#[serde(versions(version(type = "Av1"), version(type = "Av2")))]`

### Previous type must implements `serde::de::Deserialize` and the same deserialize name as the current version type

```rust
#[derive(Deserialize)]
#[serde(rename = "A")]
struct Av1 { ... }
#[serde(rename(deserialize = "A"))]
struct Av2 { ... }
```

### The current version type must implement `From` for previous types

```rust
impl From<Av1> for A { ... }
impl From<Av2> for A { ... }
```

### For types with `#[serde(default)]`

If a previous type is deserialized with `#[serde(default)]`, the it must be declared with the `default` attribute in the `versions` attribute. (Otherwise, there will be a compile error)

```rust
#[derive(Deserialize)]
#[serde(default, rename(deserialize = "A"))]
struct Av1 { ... }
impl Default for Av1 { ... }

#[derive(Deserialize)]
#[serde(versions(version(type = "Av1", default)))]
// Declare that `Av1` is the type for version 1 and is deserialize with `default`
struct A { ... }
```

# Implementation strategy

## Overview

We can use a `match` to choose the appropriate type to deserialize when appropriate.

```rust
// pseudo code
match version_map.get("A") {
  Some(1) => deserialize::<Av1>(deserializer, version_map).map(|v| v.into),
  Some(2) => deserialize::<Av2>(deserializer, version_map).map(|v| v.into),
  _ => deserialize::<A>(deserializer, version_map)
}
```

We only need to insert this match at the begining of the `Deserialize::deserialize` implementation.

## Providing the `VersionMap` and `Deserializer` trait update

The version map needs to be provided so the specialization implementation can choose the right version to deserialize.

To do so, the `Deserializer` trait has one more function:

```rust
fn version_map(&self) -> Option<&VersionMap> { None }
```

This is a method to get the `VersionMap` to use.

So whenever the deserialize is available, we can use it to access the `VersionMap`. However, quite a lot of context are visitors that are generated by the derive, so we need to provide in the member of the visitors the `VersionMap`.

# Rationale and alternatives

## Modifying `Deserializer` to provide the `VersionMap`

The `Deserializer` was chosen to provide the `VersionMap` the deserializer is available in most of the deserialization context. Also, it is wasy to create a wrapper on an existing serialize to make it provide the functionality.

The alternative is to provide the `VersionMap` as an argument, but this would change a lot fn signature, or require to duplicate code to support a version with the `VersionMap` as argument.

# Unresolved Questions

* `VersionMap` instead of being a type alias to `HashMap<String, usize>`, it can be
  * A trait, but visitor will need to be generic over it
  * An opaque type we implement, to provide only minimal API for it

* How to return properly an error for version number that are not handled currently?

# Future possibilities

## Supporting `deserialize_with` for versioning

We could provide to the callback enough information so the user can handle the versioning himself. For instance, we could provide the `VersionMap` and a list of previous type when the `#[serde(versions(...))]` attribute is used.


fredpointzero:
> 
> 
> Thanks! I haven't gotten a chance to look at the implementation yet but my first impression is that this sounds like a lot of complexity in Serde for something that not that many people will end up benefiting from. Can you think of any ways to provide similar functionality through a separate crate that doesn't require these code changes in serde or serde_derive?

There are several approach we can use:

# Reduce the amount of impacted code for this feature

The main issue is to provide the `VersionMap` in the context of visitors, currently I chose to add a member to all generated visitors, which is very intrusive.

Maybe I can add a method to `SeqAccess` and `MapAccess` to provide the `VersionMap`.

# Use an extern crate to add the feature

This was the initial approach actually, but ended up to integrate it with serde.

Here the approach I tried:

I created a new trait

```rust
trait DeserializeVersioned<'de>: serde::de::Deserialize<'de> {
	fn deserialize_versioned<D: Deserializer<'de>>(deserializer: D, version_map: VersionMap);

	/// Get the next element in a sequence with versioning support
	fn next_element_versioned<S: SeqAccess<'de>>(
	  seq: &mut S,
	  _versions: Option<&VersionMap>,
	) -> Result<Option<Self>, S::Error> {
	  seq.next_element::<Self>()
	}

	/// Get the next map value with versioning support
	fn next_value_versioned<M: MapAccess<'de>>(
	  map: &mut M,
	  _versions: Option<&VersionMap>,
	) -> Result<Self, M::Error> {
	  map.next_value::<Self>()
	}
}
```
and tried to generate the implementations with a derive macro.

But I need to reproduce what serde is generating for the deserialize method, and this is where it gets complicated.

I tried to use the `serde_derive` crate, but nothing is exposed, so I can't use its API to reproduce the generated deserialize.

I am not fan of copy pasting the `serde_derive` code, it will be hard to maintain whenever `serde_derive` updates.

But I am not even sure it is possible like this, the issue is to provide the version_map vor visitors which is tightly linked to the implementation of `serde_derive`.

If I were to go this way, I think I will need to expose a lot of `serde_derive` api to avoid the copy paste, and maybe update some of the `serde_derive` api to makes easier reuse of `serde_derive` code.

Lymia:
Maybe one approach would be to replace the `VersionMap` with a generic `TypeMap` that can store arbitrary data (which is similarly invasive, but more generic), then implementing the macros that specifically generate use this mechanism for versioning in its own crate?

fredpointzero:
> 
> 
> Maybe one approach would be to replace the `VersionMap` with a generic `TypeMap` that can store arbitrary data (which is similarly invasive, but more generic), then implementing the macros that specifically generate code that uses this mechanism for versioning in its own crate?

I think it just move the issue and don't solve the intrusive aspect of this PR.

However, I am wondering if this can't be way more simpler.

There is actually a mechanism that is quite close: `#[serde(from = "FromType")]`. And with a quick test this support recursive deserialization [(playground link)](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=5f98d0c40ca27dbde0028a1f5a2b0d86).

The main difference is that the `from` feature chose statically from which type to deserialize, and to have the versioning feature we need to only choose the specialized version. So it makes me thinking that updating the call sites of `SeqAccess::next_element` and `MapAccess::next_value` are unnecessary.

In that case the PR can actually be quite small:
1. Add the `fn version_map(&self) -> Option<&VersionMap>` on the `Deserializer`
2. Update the generated deserialize from the derive to include a match case that choose the appropriate version right before the dispatch.

Although, when using the `default` feature, we still need add a bound (so a third update here).

But in the end visitor won't need to be updated.

@dtolnay  Do you think this kind of update can be accepted if the changes are way less intrusive? Also do you think this could work?

Another way, would be to provide a mechanism for a dynamic dispatch in the generated deserialize that can be driven by a crate built on top of this. (So making the `from` feature more generic). But I guess it will make barely possible to add constraints like the one required for the `default` bounds.

fredpointzero:
@dtolnay It seems that only using the match case at the beginning of the deserialize is enough to provide the versioning feature. So this implementation is way less intrusive.

Do you think you can approve such a PR? It needs still some cleanup and returning the invalid version number error properly. But, the intent is here ^^

dtolnay:
I would prefer if we could explore what it would look like to address versioning without any code changes in serde or serde_derive. Instead of asking every Deserializer to hold the version map, consider doing something like [serde-encrypted-value](https://github.com/palantir/serde-encrypted-value) where a single Deserializer impl (in a separate crate) holds the version map and wraps an arbitrary underlying Deserializer. The wrapper Deserializer would recognize when a type with a version is being deserialized and handle the right conversion.

fredpointzero:
I did not think about this kind of solution.
I tried it ad it works (FYI, the repo is here: https://github.com/fredpointzero/serde-version).

The implementation replaces all calls to deserialize, next_element, next_value, next_key, and variant for all *Access traits. This will work whenever a seed is not used, but it will be 90% of the use case, so this is fine.

Thanks for the feedback, I'll close the PR and continue the work on the separate crate then.

PR #1620, opened 2019-09-05
Export std error type so no_std data formats don't need a "std" feature

dtolnay:
This PR adds `serde::ser::StdError` and `serde::de::StdError` which are either a re-export of `std::error::Error` (if "std" is enabled) or a new identical trait (otherwise).

```rust
#[cfg(feature = "std")]
pub use std::error::Error as StdError;

#[cfg(not(feature = "std"))]
pub trait StdError: Debug + Display {
    fn source(&self) -> Option<&(Error + 'static)> { None }
}
```

Serde's error traits `serde::ser::Error` and `serde::de::Error` require `std::error::Error` as a supertrait, but only when Serde is built with "std" enabled. Data formats that don't care about no\_std support should generally provide their error types with a `std::error::Error` impl directly:

```rust
#[derive(Debug)]
struct MySerError {...}

impl serde::ser::Error for MySerError {...}

impl std::fmt::Display for MySerError {...}

// We don't support no_std!
impl std::error::Error for MySerError {}
```

Data formats that *do* support no\_std may either have a "std" feature of their own:

```toml
[features]
std = ["serde/std"]
```

```rust
#[cfg(feature = "std")]
impl std::error::Error for MySerError {}
```

... or else provide the std Error impl unconditionally via Serde's re-export:

```rust
impl serde::ser::StdError for MySerError {}
```

---

<sup>FYI @oli-obk, @cbeck88</sup>

### DIFF
--- a/serde/src/de/mod.rs
+++ b/serde/src/de/mod.rs
@@ -125,6 +125,13 @@ mod utf8;
 
 pub use self::ignored_any::IgnoredAny;
 
+#[cfg(feature = "std")]
+#[doc(no_inline)]
+pub use std::error::Error as StdError;
+#[cfg(not(feature = "std"))]
+#[doc(no_inline)]
+pub use std_error::Error as StdError;
+
 ////////////////////////////////////////////////////////////////////////////////
 
 macro_rules! declare_error_trait {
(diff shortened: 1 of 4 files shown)


dtolnay:
It looks like a re-export like this used to exist but was lost to time. This is from 0.7.15:

https://github.com/serde-rs/serde/blob/78e74886be410b9ef0709f5ef11a13f875e6ef0a/serde/src/error.rs#L28-L44

https://github.com/serde-rs/serde/blob/78e74886be410b9ef0709f5ef11a13f875e6ef0a/serde/src/ser/mod.rs#L3-L6

https://github.com/serde-rs/serde/blob/78e74886be410b9ef0709f5ef11a13f875e6ef0a/serde/src/ser/mod.rs#L17

PR #1622, opened 2019-09-07
Fix (de)serialization of new types where the internal type is skipped

### LINKED ISSUE #1396: #[serde(skip)] in tuple enum variants silently ignored
```rust
#[macro_use]
extern crate serde_derive; // 1.0.78
extern crate serde; // 1.0.78

struct S;

#[derive(Serialize)]
enum E {
    V(#[serde(skip)] S),
}

```

```
error[E0277]: the trait bound `S: serde::Serialize` is not satisfied
 --> src/lib.rs:7:10
  |
7 | #[derive(Serialize)]
  |          ^^^^^^^^^ the trait `serde::Serialize` is not implemented for `S`
  |
  = note: required by `serde::Serializer::serialize_newtype_variant`

```

[Playground](https://play.rust-lang.org/?gist=52468306cc60f55fd62d11eba4a1ca45&version=stable&mode=debug&edition=2015)

The attribute should either work (thus serializing like `enum E { V() }` or give an error/warning that it has no meaning in this position.

Xaeroxe:
Fixes #1396 

The derive logic for newtype variants was not considering the possibility that the one value in the newtype might be skipped. This pull request fixes that. Additionally it adds some test cases that failed before, and now succeed.

### DIFF
--- a/serde_derive/src/de.rs
+++ b/serde_derive/src/de.rs
@@ -1698,7 +1698,7 @@ fn deserialize_externally_tagged_variant(
                     attr::Default::Path(ref path) => quote!(
                         #path()
                     ),
-                    _ => unimplemented!(),
+                    attr::Default::None => unimplemented!(),
                 };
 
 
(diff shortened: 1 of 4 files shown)


Xaeroxe:
This isn't quite ready for merge yet. The bug fix still needs to be applied to the functions for other forms of tagging in enums. I'll work on that in the morning.

Xaeroxe:
Ok, this has been applied to other enum representations and is ready for review.

Issue #1550, opened 2019-06-06
Feature Request: allow "from" and "into" as field attributes


Dgame:
I often have to convert attributes from or to strings and in some cases convert an object to a vector. Is it possible that such cases are implemented as `#[serde(from = string)]` / `#[serde(into = string)]` / `#[serde(into = vec)]`? Currently I'm solving such problems with `deserialize_with` (as you can see [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=2c56ecb5cf8d7ae5aac26cda6c2764c3), but I'm sure that my deserialization functions could be improved and that these problems also happen to others, so it might be worth considering.

Since I don't want to open another issue: There is another annoyance I had to solve, namely extracting fields from objects. An example can be found [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=5878e20cc74e3fdeac30eee00423e688) where I want to extract `$value`. Another example would be extracting `#text`.

PR #1623, opened 2019-09-07
Add new field attributes: from/try_from/from_str/into

nolanderc:
Addresses #1550.

The new field attributes `from`, `try_from`, `from_str` and `into` act as a shorthand for the `deserialize_with` and `serialize_with`. They mirror the functionality found in the container attributes with the same names. This PR adds a new kind of field attribute: `from_str`. If set, it first deserializes the field as a string and then attempts to convert it into the correct type by calling `FromStr::from_str`.

## Example

```rust
#[derive(Debug, Deserialize)]
struct Foo {
    #[serde(from = "i32")]
    sign: Sign,
    #[serde(from_str)]
    number: u32,
}

#[derive(Debug, PartialEq)]
enum Sign {
    Negative,
    Zero,
    Positive,
}

impl From<i32> for Sign {
    fn from(v: i32) -> Sign {
        match v {
            _ if v < 0 => Sign::Negative,
            _ if v == 0 => Sign::Zero,
            _ => Sign::Positive
        }
    }
}

fn main() {
    let json = r#"{ "sign": -1, "number": "14" }"#;

    let foo: Foo = serde_json::from_str(json).unwrap();

    assert_eq!(foo.sign, Sign::Negative);
    assert_eq!(foo.number, 14);
}
```

### DIFF
--- a/serde/src/private/de.rs
+++ b/serde/src/private/de.rs
@@ -184,6 +184,28 @@ where
     deserializer.deserialize_bytes(CowBytesVisitor)
 }
 
+#[cfg(any(feature = "std", feature = "alloc"))]
+pub fn deserialize_type_from<'de: 'a, 'a, D, U, T>(deserializer: D) -> Result<T, D::Error>
+where
+    D: Deserializer<'de>,
+    T: From<U>,
+    U: Deserialize<'de>
+{
+    Ok(T::from(U::deserialize(deserializer)?))
+}
+
+#[cfg(any(feature = "std", feature = "alloc"))]
+pub fn deserialize_type_try_from<'de: 'a, 'a, D, U, T>(deserializer: D) -> Result<T, D::Error>
+where
+    D: Deserializer<'de>,
+    T: std::convert::TryFrom<U>,
+    T::Error: std::fmt::Display,
+    U: Deserialize<'de>
+{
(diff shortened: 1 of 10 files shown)


Mingun (commented):
Also, why you implement `from` and `try_from`, but only `into`?

rimutaka:
@dtolnay , David, this is an interesting PR. This "from" / "into" approach can make for less boilerplate than with using variant type enums. I'm happy to spend some to finish what @nolanderc started, if you think it has legs.

davidv1992:
@dtolnay what is the current status of this PR. The feature offered by it is interesting for me in a current project, so I would also be willing to put some time into this if that means it has a good chance of getting it merged.

abonander:
If you're adding `from_str`, then `to_string` would be pretty neat. It'd be nice not to have to write a wrapper for that.

gavrilikhin-d:
Agree that this feature would be very useful

jbirnick:
I also need exactly this!

Many crates implement e.g. `From<String>` on their type, but have no `serde` support.

Please add this feature. We need it!

JHenneberg:
@jbirnick I do not know what the best is, but I am using `deserialize_with = try_from_string_to_u32` of course this not working then for mixed types.

stevenroose:
Any updates on this or similar work?

Issue #1556, opened 2019-06-25
Deserializing richer types than in Serde's data model is extremely painful


jyc:
First of all, thanks for this very useful crate! It is very nice to have a framework like this centralized in one place.

Unfortunately deserializing PDFs is proving to be very annoying in Serde, due to what seem like artificial limitations in its API.

PDF supports not just strings, but also "names". Both are byte strings with some bytes not allowed.

    (some string in PDF)
    /SomeName

PDF allow allows indirect object references, which are 2-tuples of an object ID and generation ID. These are distinct from PDF arrays.

    0 3 R % This references object number 0 in generation 3.
    [/Im /An /Array]

Deserializing a string containing a PDF value to a generic PDF value object requires a lot of gymnastics. This is because strings and names are both just strings of bytes in serde's data model, and indirect object references are just arrays. So I had to create a hundred lines of boilerplate so that the Visitor for PDF values could distinguish between them (see the end of this issue).

This is a consequence of Serde's decision to have visit_newtype_struct not take any &'static str as a tag, presumably because most Serde deserializers aren't deserializing a rich syntax like PDF and are fine with ignoring newtypes. It would be really nice if a future version of Serde could avoid this. I really only need a way to indicate to the visitor what newtype struct this is. If visit_newtype_struct instead looked like:

    fn visit_newtype_struct<D>(
        self,
        tag: &'static str,
        deserializer: D
    ) ...

... I could delete the fake deserializer and fake Enum/VariantAccess impls.

I see that there are similar gymnastics in the TOML deserializer for datetimes: https://github.com/alexcrichton/toml-rs/blob/master/src/de.rs#L758

```
struct AtomDeserializer {
    atom: &'static str,
}

impl<'de> de::Deserializer<'de> for AtomDeserializer {
    type Error = DeserErr;

    fn deserialize_any<V>(self, visitor: V) -> DeserResult<V::Value>
    where
        V: de::Visitor<'de>,
    {
        visitor.visit_borrowed_str(self.atom)
    }

    forward_to_deserialize_any! {
        bool i8 i16 i32 i64 i128 u8 u16 u32 u64 u128 f32 f64 char str string
        bytes byte_buf option unit unit_struct newtype_struct seq tuple
        tuple_struct map struct enum identifier ignored_any
    }
}

impl<'de> de::Deserializer<'de> for ObjectID {
    type Error = DeserErr;

    fn deserialize_any<V>(self, visitor: V) -> DeserResult<V::Value>
    where
[...]

PR #1643, opened 2019-10-07
Implement the serialization of tagged values

pyfisch:
Add a new method called serialize_tagged_value to the Serializer trait.
The method has a default implementation serializing just the value for all
formats without tags.

[Tagged values are the most requested feature](https://github.com/pyfisch/cbor/issues/157) of the `serde_cbor` crate and they are an important part of working with CBOR documents. But tagged values do not fit neatly into serde's data model therefore this small change to serde is required to enable serialization of tagged values in CBOR and possibly other data formats.

Simple CBOR serialization with a string tagged as an URL (tag 32):

```rust
let mut ser = serde_cbor::Serializer::new(&mut buf);
assert!(ser.serialize_tagged_value("cbor", &32u64, "http://example.org/").is_ok());
```

To support tagged values in a serializer for a data format it is necessary to implement `serialize_tagged_value` and write a custom serializer as I have done for CBOR: https://github.com/pyfisch/cbor/blob/6ad5ea3ade62ceadc20c5bcc832ac85f6c1d3901/src/ser.rs#L511-L695

This PR deliberately only includes serialization and not deserialization as both parts can be discussed and used independently from each other.

What do you think about this extension to serde?

### DIFF
--- a/serde/src/ser/mod.rs
+++ b/serde/src/ser/mod.rs
@@ -1247,6 +1247,27 @@ pub trait Serializer: Sized {
         len: usize,
     ) -> Result<Self::SerializeStructVariant, Self::Error>;
 
+    /// Serialize a value with a semantic tag for the given format.
+    ///
+    /// Tagged values are a mechanism supported by some data formats
+    /// to add additional information to serialized values and to
+    /// introduce user-defined subtypes.
+    ///
+    /// The default behavior is to ignore the tag and to serialize
+    /// just the value.
+    fn serialize_tagged_value<T: ?Sized, U: ?Sized>(
+        self,
+        _format: &'static str,
+        _tag: &U,
+        value: &T,
+    ) -> Result<Self::Ok, Self::Error>
+    where
+        T: Serialize,
+        U: Serialize,
+    {
+        value.serialize(self)
+    }
+
     /// Collect an iterator as a sequence.
     ///
     /// The default implementation serializes each item yielded by the iterator


pyfisch:
@dtolnay Did you have a look at this? Do you see any problems with this approach?

vmx:
I've implemented the deserialization part based on #408: https://github.com/vmx/serde/tree/tagged-value-deserialization

I dare to cross-post my comment from https://github.com/pyfisch/cbor/issues/157#issuecomment-545049252 in hope that this issue might have more people looking at it:

I'm not happy with the result. One missing thing is to deserialize the value (without the tag) by default. I just couldn't get it working. I always got trait bounds errors. I hope someone can help with that.

I also don't really like the `Tagger`. I tried several different ways. Traits with generics, traits with associated types, using a deserializer for the tag, implementing a `TagAccess` trait similar to the `SeqAccess` trait. I could none of those working. If anyone has a good idea, please let me know. Meanwhile I changed the `Tagger` to return an Any-type instead of having methods for each type to access it. I like it better, but I'm also open to change that to some other approach.

vmx:
@dtolnay Is there any chance that tagged values become part of the Serde Data Model given the design constraints outlined at https://github.com/pyfisch/cbor/issues/157#issue-503136891 (unrelated to the implementation proposed in later comments)?

I'm asking as only if there is a chance, it [makes sense for @pyfisch](https://github.com/pyfisch/cbor/issues/157#issuecomment-546622939) and me to spend more time on this. I really need tagged values and else I would need to look into other ways solving it and I fear that I would eventually end up with a less good re-implementation of Serde.

pyfisch:
Thanks for your comment. I am disappointed by your response.

> Beyond that, I would recommend building dedicated libraries to get full coverage of individual formats

`serde-cbor` already covers 95% of CBOR it just needs this *tiny bit of help* from serde to push it across the finish line. Building a separate library to implement all of serde would be a lot of work and you'd lose all the benefits of integration into the serde ecosystem.

dtolnay:
From my perspective the trouble is that many other data formats also need a different *tiny bit of help*, and even if each one is only a 5% extension, we really need to draw a line to keep the overall complexity of the core under control.

There should be ways for a library with its own traits to integrate with the Serde ecosystem, such as providing blanket impls for its traits for any type that implements Serde's traits.

vmx:
I agree that projects shouldn't just blindly implement whatever is requested from users. It's important to keep scope so you don't end up in some hard to maintain state where lots of additions are only used by a small subset of people.

Though I think things are different with Serde and this request to add tags:

 - Serde is the de facto standard library in the Rust ecosystem to serialize/deserialize things. It looks like it started as "I want to serialize my Rust types" library, but in my opinion it long outgrew that use case and is used in a bigger scope now. I think if you are implementing some serialization the go-to crate is Serde. Just using something else is not as easy as with other libraries for e.g. error handling or logging, where there are several implementations for different tasts and you could pick the one that fits best. Of course you could create your own library, but so much work has been put into Serde over the years, it would be a waste re-implementing a similar library.
 - This is not the request from a single library author that needs one special feature that is only for his implementation. Several formats work around this limitation in the same hacky way (CBOR, MessagePack, TOML). If such a common pattern emerges, I think it makes sense to move it to the upstream library and solve it once for good there. There's also more formats that have tag support (BSON, YAML) that could leverage that feature. Correct me if I'm wrong, but it would also apply to formats that don't have direct tag support, but implement types that are not part of the Serde Data Model (e.g. TOML's date type).
 - It has been discussed [in](https://github.com/serde-rs/serde/pull/301) [the](https://github.com/serde-rs/serde/pull/408) [past](https://github.com/serde-rs/serde/pull/455) between @pyfisch, @oli-obk, @dtolnay, @erickt and it sounded like there was consensus that this feature is a good idea if a good implementation is found.

rklaehn:
Is there possibly a way to keep the serde data model as it is, but just add a mechanism for different data formats that have a bit that is outside the serde data model to work? Just a lightweight optional communication channel between values and serializers/deserializers?

This could be beneficial not just for tagged values, but also for other things that are currently done in a slightly hackish way, like e.g. RawValue support in serde_json, where a "magic value" is used as a communication channel between a special value and a certain serializer.

The tag hack that @vmx came up with in serde_cbor ( https://github.com/pyfisch/cbor/pull/151 ) and the RawValue hack ( https://github.com/serde-rs/json/blob/master/src/raw.rs#L220) kinda have the same basic approach.

I have not yet spent time thinking about this, but would you be willing to add something to make the life of formats that go beyond the serde data model easier if it does *not* include extending the serde data model?

pyfisch:
>  Correct me if I'm wrong, but it would also apply to formats that don't have direct tag support, but implement types that are not part of the Serde Data Model (e.g. TOML's date type).

Yes this could be used for the TOML date type or to differentiate between undefined and null in data formats which usually map to unit in the serde data model.

> Just a lightweight optional communication channel between values and serializers/deserializers?

Tags are kind of a lightweight communication channel. I made sure that they are entirely optional and specific to a given data format so messages are not confused between different formats. I have not checked if the proposed tags are sufficient for deserializing a `RawValue` but serialization would work and I am confident that a design for deserialization would be found too that enables both tags and raw values.

rklaehn:
> Tags are kind of a lightweight communication channel. I made sure that they are entirely optional and specific to a given data format so messages are not confused between different formats. I have not checked if the proposed tags are sufficient for deserializing a RawValue but serialization would work and I am confident that a design for deserialization would be found too that enables both tags and raw values.

If many of these ugly hacks like RawValue etc. can be made nicer with an additional communication channel, that probably increases the chances of getting the additional communication channel merged.

Would having an additional &str parameter for newtype structs be sufficient (as described in https://github.com/serde-rs/serde/issues/1556 )

pyfisch:
I don't think an additional &'static str parameter would be sufficient as for tags I need to pass two pieces of information, namely the format CBOR and the tag e.g. 42.
But I have not thought about this much, so it may also be possible.

Am 15. November 2019 10:04:09 MEZ schrieb "Rüdiger Klaehn" <notifications@github.com>:

rklaehn:
> I don't think an additional &'static str parameter would be sufficient as for tags I need to pass two pieces of information, namely the format CBOR and the tag e.g. 42. But I have not thought about this much, so it may also be possible.

I am just trying to think of something really minimal that has a chance to get accepted. You could of course encode CBOR:42 in a string, but that would be super ugly. Maybe make the additional newtype parameter a `&'static [u8]` so all the different tagged formats can make up whatever they need to transfer to the visitor?

PR #1650, opened 2019-10-15
Reduce size of output .wasm file

est31:
This applies suggestions mentioned in these two documents (the subset that actually helps to reduce the size):

* https://rustwasm.github.io/book/reference/code-size.html
* https://github.com/johnthagen/min-sized-rust

Size before: 2.8M
Size after: 480K

Size of wa-serde-derive-0.1.101.crate before: 684K
Size of wa-serde-derive-0.1.101.crate after: 144K


est31:
CI fails because it seems to be broken for this branch. But locally, after I ran the bash script, running `cargo test` inside the tes_suite directory reports no errors.

est31:
@pauan has pointed me to https://github.com/rustwasm/wasm-pack-template/issues/27 . Some of the changes of this PR might have a performance impact. Maybe we should wait until we have benchmarks to measure the speedup/slowdown of this change.

est31:
TODO:

* check out souper: https://github.com/WebAssembly/binaryen/pull/1638
* check @alexcrichton 's benchmark to measue impact: https://github.com/alexcrichton/watt/tree/watt-baseline

est31:
PR got stale and without a single positive sign from the maintainer since I'll just close this..

Issue #1668, opened 2019-11-06
Untagged unit variant fails to deserialize from `visit_none()`


H2CO3:
I'm writing a serialization format that uses explicit optionals. Similarly to `serde_json::Value`, my `Value` enum has a `Null` variant representing the null value. When asked to `deserialize_any()`, this null value calls `visit_none()` on the supplied visitor. However, `UntaggedUnitVisitor` doesn't implement `visit_none()`, only `visit_unit()`, so this fails.

It seems to me that `UntaggedUnitVisitor` should ideally implement `visit_none()` because `OptionVisitor` treats unit and `None` as the same, so this would be the consistent behavior. 

Another reason is that although I could fix the error by making `Value::deserialize_any()` call `visitor.visit_unit()` instead of `visitor.visit_none()`, I'd rather not change this behavior, because untagged enums are what I would consider more of a less-used special case, whereas `Option` has very regular structure and semantics, and it is used a lot more frequently too.

Mingun:
@H2CO3 , I found, that the problem here was in calling `deserialize_any`. If derived code would call `deserialize_unit` to request the unit variant for untagged enum, would that solve problem with your format?

PR #1669, opened 2019-11-06
Allow untagged unit variants to deserialize from `Visitor::visit_none()`

H2CO3:
Fixes https://github.com/serde-rs/serde/issues/1668.

Note: the test suite fails to compile on my machine for some reason. I still added a test verifying the expected behavior nevertheless – could you please check if it's correct?

### DIFF
--- a/serde/src/private/de.rs
+++ b/serde/src/private/de.rs
@@ -2510,6 +2510,13 @@ mod content {
         {
             Ok(())
         }
+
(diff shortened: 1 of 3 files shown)


H2CO3:
God, I can't figure out how to add the appropriate test. This is clearly not a round-trip case. I give up. 😄

H2CO3:
Thank you!

PR #1677, opened 2019-11-14
Flip heterogeneous to homogeneous for seq types

djc:


djc:
But in that case, all values are actually `serde_json::Value`s? If you use `Vec<T>` and `HashSet<T>` as examples, as I understand it those are also examples of homogeneous data types, where every element is of type `T`. At the very least, I find this a very surprising use of the word homogeneous.

dtolnay:
The Rust representation only loosely informs what those types look like in the data model, so the fact that the bool and string are stored in memory as a Value on the Rust side isn't relevant to the data model or observable to the serializer.

Here is a more basic example of a heterogeneous serde seq that also contains a serde bool and serde string:

```rust
struct S;

impl Serialize for S {
    fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
        let mut seq = serializer.serialize_seq(Some(2))?;
        seq.serialize_element(&true)?;
        seq.serialize_element("...")?;
        seq.end()
    }
}
```

I would accept a change to make this clearer in the documentation.

PR #1695, opened 2019-12-15
Add #[serde(rename_all_fields = "foo")] attribute

### LINKED ISSUE #1061: Apply rename_all to enum member "fields"
When using enums with struct-like members it would be useful if `rename_all` on the enclosing enum could apply to the members, too. For example:

```rust
#[serde(untagged, rename_all = "camelCase" )]
pub enum Value {
  Null { null_value: () },
  String { string_value: String },
  Boolean { boolean_value: bool },
}
```
could do

```json
Value::String("foo".to_string())
=> 
{ "stringValue": "foo" }
  ------^-----
        |
This here is what I want
```

Currently the annotation has to be placed on the fields in addition:

```rust
#[serde(untagged)]
pub enum Value {
    Null {
        #[serde(rename = "nullValue")]
        null_value: ()
    },
    String {
        #[serde(rename = "stringValue")]
        string_value: String
    },
    /* etc. */
}
```

(If you're wondering what is going on with this type, it's [Google Datastore](https://cloud.google.com/datastore/docs/reference/rest/v1/projects/runQuery#Value)).

jplatte:
fixes #1061 

If you don't like the style change in the second commit, I can remove it from this PR (or you can just cherry-pick the first commit).

### DIFF
--- a/serde_derive/src/internals/attr.rs
+++ b/serde_derive/src/internals/attr.rs
@@ -193,9 +193,21 @@ impl Name {
     }
 }
 
+#[derive(Copy, Clone)]
 pub struct RenameAllRules {
     serialize: RenameRule,
     deserialize: RenameRule,
+}
+
+impl RenameAllRules {
+    /// Returns a new `RenameAllRules` with the individual rules of `self` and
+    /// `other_rules` joined by `RenameRules::or`.
+    pub fn or(&self, other_rules: &Self) -> Self {
+        Self {
+            serialize: self.serialize.or(&other_rules.serialize),
+            deserialize: self.deserialize.or(&other_rules.deserialize),
+        }
+    }
 }
 
 /// Represents struct or enum attribute information.
@@ -205,6 +217,7 @@ pub struct Container {
     deny_unknown_fields: bool,
     default: Default,
     rename_all_rules: RenameAllRules,
+    rename_all_fields_rules: RenameAllRules,
     ser_bound: Option<Vec<syn::WherePredicate>>,
     de_bound: Option<Vec<syn::WherePredicate>>,
     tag: TagType,
@@ -288,6 +301,8 @@ impl Container {
         let mut default = Attr::none(cx, DEFAULT);
         let mut rename_all_ser_rule = Attr::none(cx, RENAME_ALL);
         let mut rename_all_de_rule = Attr::none(cx, RENAME_ALL);
+        let mut rename_all_fields_ser_rule = Attr::none(cx, RENAME_ALL_FIELDS);
+        let mut rename_all_fields_de_rule = Attr::none(cx, RENAME_ALL_FIELDS);
         let mut ser_bound = Attr::none(cx, BOUND);
         let mut de_bound = Attr::none(cx, BOUND);
         let mut untagged = BoolAttr::none(cx, UNTAGGED);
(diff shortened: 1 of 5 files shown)


jplatte:
I'll try again since Triagebot Notifications are now a thing:

r? @dtolnay 

I've opened this PR because you wrote over at https://github.com/serde-rs/serde/issues/1061#issuecomment-451716365 that you'd accept a PR for this. If your opinion has changed, I won't be upset if this is closed without further reasoning.

kjeremy:
@dtolnay  it looks like this was approved a few months back but not merged in.

jplatte:
@kjeremy The person who left the review doesn't seem to be associated with serde in any way.

jplatte:
Rebased to trigger new CI runs, the old ones had been stuck for months.

jplatte:
CI fixed, no merge conflics. Ping @dtolnay

roberts-ivanovs:
Are there no plans on merging this in? This really would be a feature I'd love to have.

kaiba42:
Plus one on this merge...

llgerard:
Would love this!

pmhpereira:
Any update on this? Would love to have this merged as well.

jplatte:
Rebased to fix conflicts with the syn 2.0 upgrade.

serde_derive/src/internals/attr.rs
```diff
@@ -341,6 +356,43 @@ impl Container {
...
+                                            cx.error_spanned_by(de, err);
+                                        }
+                                    }
+                                }
+                            }
+                        }
+                        syn::Data::Struct(syn::DataStruct { struct_token, .. }) => {
+                            let msg = "#[serde(rename_all_fields)] can only be used on enums";
+                            cx.error_spanned_by(struct_token, msg);
+                        }
```
dtolnay, line 390: Returning a success from `parse_nested_meta` closure in this case, without having parsed the rest of the `rename_all_fields` attribute, is going to cause bogus "expected ','" errors:

```console
error: #[serde(rename_all_fields)] can only be used on enums
 --> src/main.rs:5:1
  |
5 | struct Struct {}
  | ^^^^^^

error: expected `,`
 --> src/main.rs:4:27
  |
4 | #[serde(rename_all_fields = "asdf")]
  |                           ^
```

The whole attribute needs to be parsed before you validate it.

serde_derive/src/internals/attr.rs
```diff
@@ -341,6 +356,43 @@ impl Container {
...
+                        syn::Data::Union(syn::DataUnion { union_token, .. }) => {
+                            let msg = "#[serde(rename_all_fields)] can only be used on enums";
+                            cx.error_spanned_by(union_token, msg);
+                        }
```
dtolnay, line 394: I think using the `struct`/`union` token for the span of this error is not a good choice. The problem is not that the user wrote a struct. The problem is that the user wrote a `rename_all_fields` attribute on their struct. The fix is not for the user to turn their struct into an enum. The fix is to change the attribute. So the error should point to the attribute.

serde_derive/src/internals/attr.rs
```diff
@@ -341,6 +356,43 @@ impl Container {
                             }
                         }
                     }
+                } else if meta.path == RENAME_ALL_FIELDS {
+                    // #[serde(rename_all_fields = "foo")]
+                    // #[serde(rename_all_fields(serialize = "foo", deserialize = "bar"))]
+                    match item.data {
+                        syn::Data::Enum(_) => {
+                            let one_name = meta.input.peek(Token![=]);
+                            let (ser, de) = get_renames(cx, RENAME_ALL, &meta)?;
```
dtolnay, line 365:
```suggestion
                            let (ser, de) = get_renames(cx, RENAME_ALL_FIELDS, &meta)?;
```

Otherwise this refers to the wrong attribute name in error messages.

```console
error: malformed rename_all attribute, expected `rename_all(serialize = ..., deserialize = ...)`
 --> src/main.rs:4:27
  |
4 | #[serde(rename_all_fields(asdf = "lowercase"))]
  |                           ^^^^
```

dtolnay (changes requested):
Thanks!

serde_derive/src/internals/attr.rs
jplatte: Makes sense. When I just updated this, I noticed the (container-level) `#[serde(default)]` does the same thing of pointing to the enum / struct token. Should I send another PR to change that?

dtolnay: Yep, let's fix that too. Good catch.

dtolnay (approved):
Looks good! Thank you.

PR #1490, opened 2019-02-25
Add `default_expr` and `default_fn`

bossmc:
A proposal to resolve #368, this adds `default_fn = ` as an alias for `default = ` and also adds `default_expr = ` to allow providing any valid expression (e.g, a single literal, or a custom function call, or even a block) which will be used as the generator for the default value when deserialising.

If `proc_macro_diagnostic` were stabilised, serde could emit a `deprecated` warning for use of `default = `, but, as it's not, I've just left both available.

Issue #368, opened 2016-06-10
Support default literals


dtolnay:
```rust
#[serde(default="0.1")]
pub n: f32
```

There is an ambiguity when the field is of function type so we need to figure that out.

dtolnay:
This needs to be fleshed out a lot more before it can be implemented so please chime in if you have ideas for how this should work.

I am glad we implemented `#[serde(default="...")]` and it has been widely useful but backward compatibility is going to be tricky. To start brainstorming, here are a few possible alternatives:

- Another level of nesting
  - `#[serde(default(expr="0.1"))]` - we insert the expression verbatim into the generated code
  - This is flexible but clunky.
- Look for parentheses to denote an expression
  - `#[serde(default="(0.1)")]`
  - This is hacky and doesn't provide a path forward for deprecating the current syntax.
- Use a different attribute name
  - `#[serde(missing="0.1")]`, `#[serde(default_expr="0.1")]`
  - But `default` is such a perfect fit.
- Treat as an expression if it fails to parse as a path
  - `#[serde(default="0.1")]`
  - This is ambiguous in many cases, for example enums: `#[serde(default="Color::Yellow")]`

sfackler:
Seems like an easy route would be `#[serde(default_value = "0.1")]`. It's a bit more wordy but would allow us to get the functionality out there, and we can always deprecate `default_value` if a better approach comes around.

rekka:
Would it be possible to support closures? Something like:

```rust
#[serde(default="||{ 0.1 }")]
pub n: f32
```

skorokithakis:
The closure route would be the most logical for me (and the first thing I tried when exploring how to do what I wanted). However, I think `default_expr` would be by far the clearest.

I would have preferred `default=3` to be the literal mode and `default_func` the function-based approach, but eh, can't do much about that now.

dtolnay:
As reported in #1031, unit variants would be nice to support as well. Rust no longer enforces that the value in an attribute is a literal.

Fiedzia:
Any progress on that? Right now I am forced to write functions like
'''
fn one() -> u32 { return 1}
'''
and it really looks silly

alexbool:
I agree with @sfackler and think `#[serde(default_value = "0.1")]` is a way to go

richardwhiuk:
Surely `#[serde(default_value = 0.1)]` is better - and then any expression can be provided?

In fact, that can replace any of the current usage right?

[...]

PR #1704, opened 2020-01-03
Create an extended version of syn::Meta that accepts  paths and expressions in value position.

AGausmann:
As discussed in #368 and #1490, there is a desire for supporting a larger set of syntaxes in attributes, including the ability to pass paths and expressions as values.

Right now, attribute parsing uses `syn::Meta`, but this only allows for literals in value position. This means that any other desired syntax must be wrapped in string literals (e.g. `#[serde(default = "defaults::one")]`). The long-term goal is to eventually replace such string literals with paths, and to be able to support other syntax elements in value position, such as expressions, for new kinds of options like `default_value`.

The approach that I'm taking is to duplicate and extend `syn::Meta`. I haven't bothered with changing the names or the basic structure as it seems to be otherwise sufficient, and this allows us to use it as a drop-in replacement with minimal changes. The main difference is that `lit: Lit` in `MetaNameValue` is replaced with `lit: MetaValue`, where `MetaValue` is an enum that allows types other than `Lit` to be parsed in its place.


AGausmann:
Where I'm at now: I've replaced all uses of syn::Meta with the extended version, and it now supports path parsing. The only test failures are UI mismatches due to the different high-level parsing logic and also due to a new message "expected path or literal" in value position.

Adding support for expressions isn't as straightforward. First, [`syn::Expr`](https://docs.rs/syn/1.0.13/syn/enum.Expr.html) will require syn's `"full"` feature. I don't know what impact this will have on compile times, but it's worth considering whether that's acceptable. Second, there is ambiguity between paths, literals and expressions, as paths and literals are also valid expressions (e.g. `std::mem::replace` being an expression returning a function pointer). What I could do is completely replace `MetaValue` with `Expr`; though it might not be semantically correct to parse paths as expressions in all cases (e.g. `#[serde(with = some::module)]`, it is still possible to match against the parsed expression and extract the path.

AGausmann:
I'm having second thoughts about exactly specifying the type of values in `MetaNameValue`. It might make more sense to simply store it as a `TokenStream` and have a method like `fn parse_value::<T: Parse>(&self) -> Result<T>`. This is simultaneously the most flexible and most specific option, as it allows the value to be any kind of syntax element, and that type can be determined by each option in its own parsing code. We would be able to parse many other things like `where` clauses, types, and lifetimes. 

This will also result in better diagnostic error messages for the end-user. If the type were specified within the structure of `MetaNameValue`, then it would also have to be parsed that early, and the error message would have to be as generic as possible, as nothing is known about what specific type is expected until the `MetaNameValue` is passed to an option for further parsing.

This pattern already exists (e.g. [`syn::Attribute::parse_args`](https://docs.rs/syn/1.0.13/syn/struct.Attribute.html#method.parse_args)), so there is some precedent to try it out.

AGausmann:
I had some issues with ambiguity when parsing values as `TokenStream`s, as they will parse every remaining token, including the punctuating commas. I'm looking into ways to resolve that.

```
error: expected serde tag attribute to be a string: `tag = "..."`
   --> test_suite/tests/test_gen.rs:604:19
    |
604 |     #[serde(tag = "t", content = "c")]
    |                   ^^^^^^^^^^^^^^^^^^
```

AGausmann:
Ready for discussion / review

AGausmann:
I've been working on the next step, modifying the parsing frontend to accept inline syntax, you can see progress on my branch [`inline_values`](https://github.com/AGausmann/serde/compare/extended_meta...AGausmann:inline_values).

Something I ran into is an ambiguity with regard to where clauses. Since they use commas as punctuation, you can't write something like `#[serde(where = E: Clone, F: Clone)]`. They would have to be written as separate invocations of the `where` option.

agausmann:
@dtolnay any feedback? There are a few UI failures but to me they don't seem to be any harder to understand, though a bit more verbose.

dtolnay:
Hi @agausmann, I am still interested in this change but I haven't gotten a chance to take a look yet. It is still definitely on the list (https://triage.rust-lang.org/notifications?user=dtolnay).

pksunkara:
I did a brief review, looks good to me. But I don't know the internals of serde that well to forsee any future problems.

PR #1705, opened 2020-01-03
impl Serializer for &mut fmt::Formatter

jethrogb:
Allow using `&mut fmt::Formatter` as a serializer for primitive types.

One use-case I have in mind in particular is implementing `Display` for enums that also implement `Serialize`.

If you believe this impl should exist, due to coherence, this crate is the only place it could exist.

### DIFF
--- a/serde/src/ser/impls.rs
+++ b/serde/src/ser/impls.rs
@@ -1,6 +1,6 @@
 use lib::*;
 
-use ser::{Error, Serialize, SerializeTuple, Serializer};
+use ser::{Error, Impossible, Serialize, SerializeTuple, Serializer};
 
 ////////////////////////////////////////////////////////////////////////////////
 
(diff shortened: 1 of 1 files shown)


dtolnay:
Thanks! Could you give an example of how this could be used?

jethrogb:
```rust
#[derive(Serialize)]
#[serde(rename_all="kebab-case")]
enum MessageType {
    StartRequest,
    EndRequest,
}

impl fmt::Display for MessageType {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        self.serialize(f)
    }
}
```

peterjoel:
This is actually a nice idea. I frequently end up with a dev dependency on `serde_json`, just to do a string comparison with a serialised value. This could be very convenient.

It doesn't actually need to implement `Serializer` for `Formatter` though; a new struct could delegate for the purpose, and live in a separate crate. It could also be implemented as a macro that generates a `Display` or `Debug` impl, without the need for an extra type.

dtolnay:
Published in 1.0.108.

PR #1706, opened 2020-01-06
Deserialize adjacently tagged newtype variants with optional content as None instead of erroring when content field is missing

### LINKED ISSUE #1553: Optional content for enum variants?
Hi!
I need to de/serialize JSON messages such as the following:
```javascript
{ ...,  "command": "cmd1", "arguments": { <command-dependent> } }
```
So far so good, this fits the adjacently tagged enum representation.   
The wrinkle is that for some commands the `arguments` field may be optional.   I tried modelling this as 
```rust
#[serde(tag = "command", content = "arguments")]
enum Command {
    cmd1(Option<Cmd1Arguments>)
    ...
}
```
but Serde is not liking it and complains about "missing field \`arguments\`".

Any suggestions? (besides custom serialization)

zth0:
Fixes #1553

### DIFF
--- a/serde_derive/src/de.rs
+++ b/serde_derive/src/de.rs
@@ -1379,53 +1379,37 @@ fn deserialize_adjacently_tagged_enum(
         }
     };
 
-    fn maybe_optional(variant: &Variant) -> bool {
-        match variant.style {
-            Style::Unit => true,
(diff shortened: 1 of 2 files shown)


dtolnay:
Published in 1.0.109.

Issue #1717, opened 2020-01-14
untagged enums do not support 128 bit integers


Licenser:
The `Content` enum 

https://github.com/serde-rs/serde/blob/ff70409215b62fa8cd138ce47534c432352c85bb/serde/src/private/de.rs#L230

does not include `i128` or `u128` fields, this makes it impossible to use those values in untagged enums.

mikonieminen:
I've run into similar issues and raised a bug in `serde_json`: https://github.com/serde-rs/json/issues/740

I've created a fork with tests showing that tagged variant works while untagged variant fails: https://github.com/mikonieminen/json/blob/128bit-bug/tests/128bit.rs

Mingun:
This is duplicate of #1682

PR #1679, opened 2019-11-21
Improves `u128`/`i128` handling 

bkchr:
Besides that it makes it possible to have `u128`/`i128` in buffered `Content`.

PR #1719, opened 2020-01-14
Update support for 128 numbers in untagged enums

Licenser:
This adds some support for untagged enums as mentioned in #1717  . But it causes some problems w/ serde-json and I'm not sure how to best handle this. Any suggestions?

### DIFF
--- a/serde/src/de/impls.rs
+++ b/serde/src/de/impls.rs
@@ -192,7 +192,7 @@ macro_rules! impl_deserialize_num {
         visit_integer_method!(i32, visit_i32, from_i32, Signed, i64);
         visit_integer_method!(i64, visit_i64, from_i64, Signed, i64);
 
-        visit_integer_method!(u8, visit_u8, from_u8, Unsigned, u64) ;
+        visit_integer_method!(u8, visit_u8, from_u8, Unsigned, u64);
         visit_integer_method!(u16, visit_u16, from_u16, Unsigned, u64);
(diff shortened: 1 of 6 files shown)


serde/src/de/impls.rs
```diff
@@ -141,12 +192,14 @@ macro_rules! impl_deserialize_num {
         visit_integer_method!(i32, visit_i32, from_i32, Signed, i64);
         visit_integer_method!(i64, visit_i64, from_i64, Signed, i64);
 
-        visit_integer_method!(u8, visit_u8, from_u8, Unsigned, u64);
+        visit_integer_method!(u8, visit_u8, from_u8, Unsigned, u64) ;
```
teor2345, line 195: Nitpick: looks like there's an accidental extra space here

```suggestion
        visit_integer_method!(u8, visit_u8, from_u8, Unsigned, u64);
```

teor2345: (But I guess rustfmt will get rid of all the extra spaces, so it might not really matter.)

Licenser: thank you!

dtolnay (commented):
I'll close this PR since it does not resolve the serde_json use case, as you noticed.

There have been some other PRs along the same lines which I've also closed -- please see #1679 and #1979. In particular https://github.com/serde-rs/serde/pull/1979#pullrequestreview-735514847 has a brief proposal of what direction to go with supporting this use case.

PR #1733, opened 2020-02-01
allow user control on has_atomic32 and has_atomic64

semarie:
on OpenBSD, for i386 architecture, we use a slightly different specification that the one in `src/librustc_target/spec/i686_unknown_openbsd.rs`. we are targeting olders CPUs, possibly without atomic64.

this PR adds a user configurable way to specify if the target has atomic32 and atomic64.

it uses environment variables `SERDE_HAS_ATOMIC32_targetname` and `SERDE_HAS_ATOMIC64_targetname` to force a specific value, and fallback to hardcoded definition.


semarie:
@dtolnay I understand your point of vue. my problem is "patching the build script" means "patching the build script" for all rust programs using serde in OpenBSD. which is lot of work. I will look to upstream `src/librustc_target/spec` change so.

Thanks.

PR #1749, opened 2020-03-06
Allow floats to be deserialized from ints in tagged unions

### LINKED ISSUE #1748: Small integers don't get converted to floats when deserializing
Ok I'm pretty sure when you deserialize a value that comes out as something like a `U8`, and feed it into a struct that is expecting an `f32`, it gives an error, even though it should just convert the value.

This code demonstrates the problem:

```
use serde::Deserialize;

#[derive(Deserialize)]
pub struct Foo {
  pub step: Step,
}

#[derive(Deserialize)]
#[serde(tag = "type")]
pub enum Step {
  One(One),
  Two(Two),
}

#[derive(Deserialize)]
pub struct One {
}

#[derive(Deserialize)]
pub struct Two {
  pub value: f32,
}

fn main() {

    // Encode as int.
    let json = r#"
{
    "step": {
        "type": "Two",
        "value": 1
    }
}
"#;

    // The corresponding CBOR, where the value 1 is encoded as an integer (the
    // final byte).
    let cbor: Vec<u8> = vec![0xA1, 0x64, 0x73, 0x74, 0x65, 0x70, 0xA2, 0x64,
                             0x74, 0x79, 0x70, 0x65, 0x63, 0x54, 0x77, 0x6F,
                             0x65, 0x76, 0x61, 0x6C, 0x75, 0x65, 0x01];

    eprintln!("Decoding JSON");
    let _json_dec: Foo = serde_json::from_str(&json).unwrap();
    eprintln!("Decoding CBOR");
    let _cbor_dec: Foo = serde_cbor::from_slice(&cbor).unwrap();
}
```

It outputs:

```
Decoding JSON
Decoding CBOR
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ErrorImpl { code: Message("invalid type: integer `1`, expected f32"), offset: 0 }', src/libcore/result.rs:1188:5
```

What is going on here? I investigated by running the above code though `cargo expand` and then adding `println!()`s in it. The differences start here:

```
            let __tagged = match _serde::Deserializer::deserialize_any(
                __deserializer,
                _serde::private::de::TaggedContentVisitor::<__Field>::new("type"),
            ) {
                _serde::export::Ok(__val) => __val,
                _serde::export::Err(__err) => {
                    return _serde::export::Err(__err);
                }
            };

            match __tagged.tag {
                __Field::__field0 => { // "One"
                  _serde::export::Result::map(
                    <One as _serde::Deserialize>::deserialize(
                        _serde::private::de::ContentDeserializer::<__D::Error>::new(
                            __tagged.content,
                        ),
                    ),
                    Step::One,
                  )
                },
                __Field::__field1 => { // "Two"
                  dbg!(&__tagged.content);
                  _serde::export::Result::map(
                    <Two as _serde::Deserialize>::deserialize(
                        _serde::private::de::ContentDeserializer::<__D::Error>::new(
                            __tagged.content,
                        ),
                    ),
                    Step::Two,
                  )
                },
            }
```

The `dbg!()` output for the JSON and CBOR cases are:

```
// JSON
&__tagged.content = Map(
    [
        (
            Str(
                "value",
            ),
            U64(
                1,
            ),
        ),
    ],
)
```

```
// CBOR
&__tagged.content = Map(
    [
        (
            Str(
                "value",
            ),
            U8(
                1,
            ),
        ),
    ],
)
```

Ok that seems fine surely? But then the code eventually gets to [this bit of Serde](https://github.com/serde-rs/serde/blob/master/serde/src/private/de.rs#L1174):

```
        fn deserialize_f32<V>(self, visitor: V) -> Result<V::Value, Self::Error>
        where
            V: Visitor<'de>,
        {
            match self.content {
                Content::F32(v) => visitor.visit_f32(v),
                Content::F64(v) => visitor.visit_f64(v),
                Content::U64(v) => visitor.visit_u64(v),
                Content::I64(v) => visitor.visit_i64(v),
                _ => Err(self.invalid_type(&visitor)),
            }
        }
```

At this point, `self.content` is `Content::U64` for JSON, but `Content::U8` for CBOR, so it fails with the `invalid_type` error.

But why does it work for JSON? Well, JSON basically always decodes numbers as either `U64`, `I64`, or `F64`. It never generates any other Serde types so that switch always works.

But it gets a bit weirder, if you simplify the code, removing the tagged union like this:

```
use serde::Deserialize;

#[derive(Deserialize)]
pub struct Foo {
  pub test: f32,
}

fn main() {

    // Encode as int.
    let json = r#"
{
    "test": 1
}
"#;

    // The corresponding CBOR, where the value 1 is encoded as an integer (the
    // final byte).
    let cbor: Vec<u8> = vec![0xA1, 0x64, 0x74, 0x65, 0x73, 0x74, 0x01];

    eprintln!("Decoding JSON");
    let _json_dec: Foo = serde_json::from_str(&json).unwrap();
    eprintln!("Decoding CBOR");
    let _cbor_dec: Foo = serde_cbor::from_slice(&cbor).unwrap();
}
```

Then it works fine! What? I `cargo expand`ed again, and I *think* it's because without the tagged union it doesn't have to deserialize into a `serde::Value` first, so it must take a slightly different code path that never gets to `deserialize_f32`.

Anyway I guess the fix is add all the other integer types to `deserialize_f32` and `deserialize_f64`.

Timmmm:
This code is hit when deserializing internally tagged unions. Only some type conversions to `f32` and `f64` were allowed previously, but if you aren't in a tagged union, any number type can be converted to `f32` or `f64` (see `from_primitive.rs`).

There is a test in #1748 that now works, and I have also tested that this works in my real app.

Fixes #1748

### DIFF
--- a/serde/src/private/de.rs
+++ b/serde/src/private/de.rs
@@ -1036,6 +1036,25 @@ mod content {
                 _ => Err(self.invalid_type(&visitor)),
             }
         }
+
+        fn deserialize_float<V>(self, visitor: V) -> Result<V::Value, E>
+        where
+            V: Visitor<'de>,
+        {
+            match self.content {
+                Content::F32(v) => visitor.visit_f32(v),
+                Content::F64(v) => visitor.visit_f64(v),
+                Content::U8(v) => visitor.visit_u8(v),
+                Content::U16(v) => visitor.visit_u16(v),
+                Content::U32(v) => visitor.visit_u32(v),
+                Content::U64(v) => visitor.visit_u64(v),
+                Content::I8(v) => visitor.visit_i8(v),
+                Content::I16(v) => visitor.visit_i16(v),
+                Content::I32(v) => visitor.visit_i32(v),
+                Content::I64(v) => visitor.visit_i64(v),
+                _ => Err(self.invalid_type(&visitor)),
+            }
+        }
     }
 
     fn visit_content_seq<'de, V, E>(content: Vec<Content<'de>>, visitor: V) -> Result<V::Value, E>
@@ -1175,25 +1194,14 @@ mod content {
         where
             V: Visitor<'de>,
         {
-            match self.content {
-                Content::F32(v) => visitor.visit_f32(v),
-                Content::F64(v) => visitor.visit_f64(v),
-                Content::U64(v) => visitor.visit_u64(v),
-                Content::I64(v) => visitor.visit_i64(v),
-                _ => Err(self.invalid_type(&visitor)),
-            }
+            self.deserialize_float(visitor)
         }
 
         fn deserialize_f64<V>(self, visitor: V) -> Result<V::Value, Self::Error>
         where
             V: Visitor<'de>,
         {
-            match self.content {
-                Content::F64(v) => visitor.visit_f64(v),
-                Content::U64(v) => visitor.visit_u64(v),
-                Content::I64(v) => visitor.visit_i64(v),
-                _ => Err(self.invalid_type(&visitor)),
-            }
+            self.deserialize_float(visitor)
         }
 
         fn deserialize_char<V>(self, visitor: V) -> Result<V::Value, Self::Error>


Timmmm:
Have to abandon and recreate this for reasons.

PR #1768, opened 2020-04-03
Hide generated dummy const in rustdoc

robo9k:
When creating a binary crate with a `src/main.rs` and `#[derive(Deserialize, Serialize)]`, running `cargo doc` on it will result in new doc items for generated constants:
```rust

struct Dummy {
}

fn main() {
    println!("Hello, world!");
}
```
rustdoc

This is because Cargo runs with `--document-private-items` internally for `src/main.rs`. If you put the same code in a `src/lib.rs` the private items will not be documented by default, so you won't see them there for default `cargo doc`.
The items will be visible both for `src/lib.rs` and `src/main.rs` if the `struct` itself is `pub`.

I believe the generated `_IMPL_{}_FOR_{}` constants should not be visible in rustdoc and this merge request seems to fix the issue for me.

### DIFF
--- a/serde_derive/src/dummy.rs
+++ b/serde_derive/src/dummy.rs
@@ -29,6 +29,7 @@ pub fn wrap_in_const(
     };
 
     quote! {
+        #[doc(hidden)]
         #[allow(non_upper_case_globals, unused_attributes, unused_qualifications)]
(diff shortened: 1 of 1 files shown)


dtolnay:
Published in 1.0.106.

Issue #1747, opened 2020-03-06
Derive a Serialize implementation that works with packed structs


TannerRogalsky:
As per this [this comment](https://github.com/rust-lang/rust/issues/46043#issuecomment-425668016), `serde_derive` does not notice if a struct is `#[repr(packed)]` and "`#[derive(Copy)]` in order to derive a correct implementation. Instead it issues (at time of writing) a warning of "`#[derive]` can't be used on a `#[repr(packed)]` struct that does not derive Copy (error E0133)".

Would it be possible for `serde_derive` to take these annotations into account and generate a correct `Serialize` implementation?

PR #1791, opened 2020-05-01
Add support for packed structs.

alvardes:
Fixes: #1747

### DIFF
--- a/serde_derive/src/internals/attr.rs
+++ b/serde_derive/src/internals/attr.rs
@@ -222,6 +222,7 @@ pub struct Container {
     identifier: Identifier,
     has_flatten: bool,
     serde_path: Option<syn::Path>,
+    is_packed: bool,
 }
 
 /// Styles of representing an enum.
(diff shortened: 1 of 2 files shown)


alvardes:
@dtolnay I force-pushed to trigger the new checks. Seems to pass.

serde_derive/Cargo.toml
```diff
@@ -26,7 +26,7 @@ proc-macro = true
 [dependencies]
 proc-macro2 = "1.0"
 quote = "1.0"
-syn = { version = "1.0", features = ["visit"] }
+syn = { version = "1.0", features = ["visit", "extra-traits"] }
```
dtolnay, line 29: "extra-traits" is pretty expensive in compile time. Please avoid introducing a dependency on it.

alvardes: I replaced the use of `==` on `syn::AttrStyle` with a structural match.

dtolnay:
Published in 1.0.110.

PR #1794, opened 2020-05-02
Add prefix_all container attribute

peterjoel:
Usage:

```
#[derive(Serialize, Deserialize, Debug, PartialEq)]
#[serde(prefix_all = "PREFIX_", rename_all = "camelCase")]
struct PrefixedCamelCase {
    serialize: bool,      // => "PREFIX_serialize"
    serialize_seq: bool,  // => "PREFIX_serializeSeq"
}
```

If used in combination with `rename_all`, as above, the prefix is applied after the case conversion. The rationale is that you could get the other behaviour by just writing the prefix in the desired case to begin with, and this way also allows for the additional flexibility of choosing the prefix to be a different case.

There is one weird edge case, however. If you want to add a prefix, and also rename all so that the final result is in camelCase, you actually need to use `rename_all = "PascalCase"`, so that the first letter of the field name is capitalised. I think the choice I made is the right one, but it seemed worth mentioning in case I didn't consider something.

### DIFF
--- a/serde_derive/src/internals/attr.rs
+++ b/serde_derive/src/internals/attr.rs
@@ -20,7 +20,7 @@ use syn::NestedMeta::{Lit, Meta};
 // user will see errors simultaneously for all bad attributes in the crate
 // rather than just the first.
 
-pub use internals::case::RenameRule;
+pub use internals::rename::{ConvertCase, RenameRule};
 
 struct Attr<'c, T> {
     cx: &'c Ctxt,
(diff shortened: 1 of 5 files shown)


peterjoel:
Thanks, I didn't know about that crate! Unfortunately, the performance isn't good enough for my needs, because it allocates multiple times for each field name, every time it serializes. 

That said, I can use a proc macro to generate `#[serde(rename)]` attributes for each field, which gives the same effect at no extra cost. So closing this seems reasonable.

dtolnay:
I believe serde_with includes some proc macros already, so consider asking whether they would be willing to incorporate yours.

peterjoel:
Yep, thanks. I'm taking a proper look at it now.

PR #1813, opened 2020-05-17
Improve `packed` repr matching and add tests.

TannerRogalsky:
Further improves the patches related to #1747 which worked well as long as `packed` was the first attribute in the repr annotation. Now as long as the attribute exists it will behave appropriately.


alvardes:
Does this work for `repr(packed(2))`? Can you add a test for that style too?

TannerRogalsky:
@alvardes It does but in order to validate I added the tests anyway. :)

alvardes:
Thanks!

dtolnay:
Published in 1.0.114.

PR #1831, opened 2020-06-07
Collect lifetimes inside macro invocations

### LINKED ISSUE #1562: Error: "field has no lifetimes to borrow" if type is behind a macro
```Rust
use serde::Deserialize;
use std::borrow::Cow;

macro_rules! t {
    ($t:ty) => { $t }
}

#[derive(Deserialize)]
struct B<'a> {
    #[serde(borrow="'a")]
    f: t!(Cow<'a , str>)
}
```

This generates error:
```
error: field `f` has no lifetimes to borrow
  --> src\main.rs:10:5
   |
10 | /     #[serde(borrow="'a")]
11 | |     f: t!(Cow<'a , str>)
   | |________________________^
```

but If I manually substitute type that macro expands to then everything works fine.

taiki-e:
Fixes #1562

### DIFF
--- a/serde_derive/src/internals/attr.rs
+++ b/serde_derive/src/internals/attr.rs
@@ -1,6 +1,6 @@
 use internals::symbol::*;
 use internals::{ungroup, Ctxt};
-use proc_macro2::{Group, Span, TokenStream, TokenTree};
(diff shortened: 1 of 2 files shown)


dtolnay (approved):
Thanks!

PR #1840, opened 2020-06-16
Add FlexBuffers to serde ecosystem list

rw:
Add FlexBuffers to serde ecosystem list


CasperN:
Hi @dtolnay, I updated the Flexbuffers documentation to point at our serde example. Did you have something more in mind for Serde documentation? If so, could you provide more details or an example?

Issue #939, opened 2017-05-19
Support a #[serde(validate = "some_function")] attribute on fields


sfackler:
I occasionally run into fields in a serde-deserializable type that require some extra validation after deserialization, for example "a u32 >= 2". The way to do this now is to make a custom deserialize function that deserializes and then does the check:

```rust
#[derive(Deserialize)]
struct Thing {
    #[serde(deserialize_with = "validate_foo")]
    foo: u32,
}

fn validate_foo<'de, D>(d: D) -> Result<u32, D::Error>
    where D: de::Deserializer<'de>
{

    let value = u32::deserialize(d)?;

    if value < 2 {
        return Err(de::Error::invalid_value(de::Unexpected::Unsigned(value as u64),
                                            &"a value at least 2"));
    }

    Ok(value)
}
```

However, it'd be nice if it were possible to specifically tell serde to use the standard deserializer but apply a validation function after the field is deserialized:

```rust
#[derive(Deserialize)]
struct Thing {
    #[serde(validate = "validate_foo")]
    foo: u32,
}

fn validate_foo<E>(v: &u32) -> Result<(), E>
    where E: de::Error
{
    if value < 2 {
        Err(de::Error::invalid_value(de::Unexpected::Unsigned(value as u64),
                                     &"a value at least 2"));
    } else {
        Ok(())
    }
}
```

Keats:
That's very similar to one of my crate (https://github.com/Keats/validator) with a wip new version at https://github.com/Keats/validator/pull/27

lucab:
I guess this is a duplicate of https://github.com/serde-rs/serde/issues/642.

dtolnay:
I don't think this is a duplicated of #642. This is for a field attribute and that one is for a container attribute. They are useful in different situations.

TedDriggs:
@sfackler, how would you feel about this:

```rust
#[derive(Deserialize)]
struct Thing {
    #[serde(and_then = "validate_foo")]
    foo: u32,
}

fn validate_foo<E>(v: u32) -> Result<u32, E>
    where E: de::Error
{
    if value < 2 {
        Err(de::Error::invalid_value(de::Unexpected::Unsigned(value as u64),
                                     &"a value at least 2"));
    } else {
        Ok(value)
    }
}
```

That would enable both validation and post-deserialize transformation, which is something I've found myself reaching for in the past.

Mingun:
I think this one is superseded by #1550 and can be closed in preference to it

[...]

PR #1858, opened 2020-07-16
Implemented 'and_then' struct field attribute, fixing #939

jwillbold:
Implements the 'and_then' struct field attribute. It can be used for validation but also for post-processing.
Example:

```Rust
fn check_above_2<E>(v: u32) -> Result<u32, E> where E: serde::de::Error {
    if v > 2 {
        Ok(v)
    } else {
        Err(serde::de::Error::invalid_value(
            serde::de::Unexpected::Unsigned(v as u64), &"a value above 2")
        )
    }
}

fn check_vec_not_empty<E>(v: Vec<u32>) -> Result<Vec<u32>, E> where E: serde::de::Error {
    match v.is_empty() {
        false => Ok(v),
        true => Err(serde::de::Error::invalid_value(
            serde::de::Unexpected::Seq, &"a non-empty vector'")
        ),
    }
}

struct Struct {
    #[serde(and_then="check_above_2")]
    field_above_2: u32,
    #[serde(and_then="check_vec_not_empty")]
    vec_non_empty: Vec<u32>,
    field_unconditional: u32
}
```

This addresses Issue #939.

### DIFF
--- a/serde_derive/src/de.rs
+++ b/serde_derive/src/de.rs
@@ -2482,10 +2482,14 @@ fn deserialize_map(
         .filter(|&&(field, _)| !field.attrs.skip_deserializing() && !field.attrs.flatten())
         .map(|(field, name)| {
             let missing_expr = Match(expr_is_missing(field, cattrs));
-
+            let and_then = match field.attrs.and_then() {
+                Some(and_then) => quote! { try!(#and_then(#name)) },
+                None => quote! { #name }
+            };
+            
             quote! {
                 let #name = match #name {
-                    _serde::export::Some(#name) => #name,
+                    _serde::export::Some(#name) => #and_then,
                     _serde::export::None => #missing_expr
                 };
             }
--- a/serde_derive/src/internals/attr.rs
+++ b/serde_derive/src/internals/attr.rs
@@ -1137,6 +1137,7 @@ pub struct Field {
     getter: Option<syn::ExprPath>,
     flatten: bool,
     transparent: bool,
+    and_then: Option<syn::ExprPath>
 }
 
 /// Represents the default to use for a field when deserializing.
@@ -1181,6 +1182,7 @@ impl Field {
         let mut borrowed_lifetimes = Attr::none(cx, BORROW);
         let mut getter = Attr::none(cx, GETTER);
         let mut flatten = BoolAttr::none(cx, FLATTEN);
+        let mut and_then = Attr::none(cx, AND_THEN);
 
         let ident = match &field.ident {
             Some(ident) => unraw(ident),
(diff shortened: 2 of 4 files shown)


dtolnay (commented):
Thanks, the implementation of this looks good. Some questions on the design:

- Does it ever make sense for the input and output of the `and_then` function to have different types? Notice that [`Result::and_then`](https://doc.rust-lang.org/std/result/enum.Result.html#method.and_then) takes `FnOnce(T) -> Result<U, E>`. In serde presumably that would be `fn and_then<E: Error>(t: T) -> Result<U, E>`.

- In the `check_above_2` and `check_vec_not_empty` example you gave, that exact use case is already supported by `deserialize_with` with the 1 line difference of inserting a `deserialize` call at the top:

    ```rust
    fn de_above_2<'de, D>(deserializer: D) -> Result<u32, D::Error> where D: Deserializer<'de> {
        let v = u32::deserialize(deserializer)?;
        /* identical to check_above_2 body */
    }

    fn de_nonempty_vec<'de, D>(deserializer: D) -> Result<Vec<u32>, D::Error> where D: Deserializer<'de> {
        let v = Vec::deserialize(deserializer)?;
        /* identical to check_vec_not_empty body */
    }

    #[derive(Deserialize)]
    struct Struct {
        #[serde(deserialize_with = "de_above_2")]
        field_above_2: u32,
        #[serde(deserialize_with = "de_nonempty_vec")]
        vec_non_empty: Vec<u32>,
    }
    ```

    Is the point of the new attribute only to save the 1 line `deserialize` call in these cases, or is there another example of how you envision this being used that illustrates a bigger improvement?

jwillbold:
Thanks for reviewing the code! 

- I don't see how different types would be useful since the value must be stored in a pre-defined field-type.
- My understanding is that all use-cases of ``and_then`` can be implemented through custom deserialization. However, when I see custom deserialization for a vector, I expect significantly customized deserialization behavior. On the other hand, when I see ``and_then``, it is much clearer that default deserialization takes place plus a custom operation. This custom operation is specified by the name of the ``and_then`` function, such as ``check_vec_not_empty``. Thus, ``and_then`` is a much more readable and understandable solution.

jwillbold:
Thanks for reviewing the code! 

- I don't see how different types would be useful since the value must be stored in a pre-defined field-type.
- My understanding is that all use-cases of ``and_then`` can be implemented through custom deserialization since your code is pretty much exactly what happens internally. However, I think ``and_then`` implements validation checking in a more Rusty-style than a custom deserialization. Also, judging by the amount of issues cross-linked to #939, I feel like many people first look into something like ``and_then`` before thinking about custom deserialization.

dtolnay (commented):
I think I would lean toward continuing to recommend `deserialize_with` for this use case, instead of adding another separate attribute. We can do a better job in documentation with examples of using `deserialize_with` for validation. Regarding issues cross-linked to #939, that happens if an issue features high in search results, so I definitely wouldn't conclude it's a good or necessary solution from that.

Thanks anyway for the PR!

raphaelcohn:
@dtolnay Pity this was closed, as it is a far more ergonomic way to do this than `deserialize_with`... which is tedious to work with and requires lots of boilerplate or glue code to combine with existing data validation logic for custom types with a more restrict type (eg an u31), etc.

PR #1875, opened 2020-08-10
Use alloc and core crates directly, even in std

kylefleming:
I've created a proof-of-concept PR that simplifies the `no_std` situation.

The basic idea is to use the lowest-requirement stdlib crate (`core`, `alloc`, or `std`) where a type/function exists, regardless of whether it's reexported from a higher level stdlib crate, even if that higher level stdlib crate is also available.

This PR:
* Turns on `no_std` unconditionally for all feature configurations
* Selectively imports `alloc` or `std`, depending on feature configuration, but does so in a purely additive way (pulls in `alloc` if either the `std` or `alloc` features are enabled, and pulls in `std` if the `std` feature is enabled)
* Uses `core` for anything that exists in `core`, even if `std` is available
* Uses `alloc` for anything that exists in `alloc`, even if `std` is available

Note: This means that, for example, for a build with the `std` feature enabled, the `core`, `alloc`, and `std` crates are all pulled in.

Do you think this is a reasonable approach to the `no_std` situation?

### DIFF
--- a/serde/Cargo.toml
+++ b/serde/Cargo.toml
@@ -36,7 +36,7 @@ derive = ["serde_derive"]
 
 # Provide impls for common standard library types like Vec<T> and HashMap<K, V>.
 # Requires a dependency on the Rust standard library.
-std = []
+std = ["alloc"]
 
 # Provide impls for types that require unstable functionality. For tracking and
 # discussion of unstable functionality please refer to this issue:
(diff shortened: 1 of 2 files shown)


kylefleming:
> Could you clarify whether this is intended to be a maintainability improvement only or if there are any downstream-facing advantages to the approach in the PR?

Purely for maintainability.

kylefleming:
Looking at the failing CI checks, it appears there are further complications with older Rust versions that might make this approach untenable.

PR #1877, opened 2020-08-13
Implement Deserialize for &[u8; N], as with &[u8]

### LINKED ISSUE #1857: `#[derive(serde::Deserialize)]` fails to implement for `&[u8; 20]`, but works for `[u8;20]`
Here is a link [to the playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=825c161aefc77d63857d372e958faabb) illustrating this.

Screenshot 2020-07-16 at 15 01 23

The manual implementation looks [like this](https://github.com/Byron/git-oxide/blob/9f46efd625d45a9ad947e9f7bc6f31f4426f3cfc/git-object/src/borrowed/id.rs#L36-L61) but currently [panics](https://github.com/Byron/git-oxide/blob/9f46efd625d45a9ad947e9f7bc6f31f4426f3cfc/git-object/src/borrowed/id.rs#L61) if the underlying slice is not of the desired length.

#### Motivation

I just recently learned that borrowed arrays are similar to a slice, but encodes the desired length of the slice at the same time. This makes such a type self-documenting in my use case, and is thus preferable over using bare slices.

As it is possible to convert slices into arrays of respective sizes using `TryInto`, I believe this is a default feature naturally supported in Rust.

This is why I would have hoped borrowed arrays to be supported in serde as well.

#### Hopes

Unless there is a technical limitation for this, I hope there is a chance to have deserialization implemented for borrowed arrays up to a size of 32 elements.

In any case, thanks for all your efforts.

#### Other Observations

* `serde::Serialize` can be auto-implemented without problems

#### Meta

* tested *serde 1.0.114* on MacOS using rustc v1.44.1

mcy:
This change adds implementations of `Deserialize` for `&[u8; 0..=32]`, with the same byte-array semantics as `&[u8]`, by threading in existing `TryFrom` implementations in `core`.

This also doesn't work on older Rust versions, unfortunately, so it needs to be conditioned by that, or we can just use the same pointer casts that `core` does.

Unfortunately, I worry that this may break round-tripping. I'm not familiar enough with Serde's test suite to be able to add the appropriate tests, so guidance would be appreciated there.

Fixes #1857, at any rate.

### DIFF
--- a/serde/src/de/impls.rs
+++ b/serde/src/de/impls.rs
@@ -495,6 +495,56 @@ impl<'de: 'a, 'a> Deserialize<'de> for &'a [u8] {
     }
 }
 
+struct FixedBytesVisitor<A> {
+    marker: PhantomData<A>,
+}
+
+macro_rules! borrowed_byte_array_impls {
+    ($($len:expr),*) => {$(
+        impl<'a> Visitor<'a> for FixedBytesVisitor<[u8; $len]> {
+            type Value = &'a [u8; $len];
+
+            fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+                formatter.write_str(concat!("a borrowed byte array of length ", $len))
+            }
+
+            fn visit_borrowed_bytes<E>(self, v: &'a [u8]) -> Result<Self::Value, E>
+            where
+                E: Error,
+            {
+                let len = v.len();
+                v.try_into().map_err(|_| E::invalid_length(len, &stringify!($len)))
+            }
+
+            fn visit_borrowed_str<E>(self, v: &'a str) -> Result<Self::Value, E>
+            where
+                E: Error,
+            {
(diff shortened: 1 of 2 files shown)


dtolnay:
In particular, it looks like this commit increases compile time of serde (`cargo clean && time cargo build`) by 7.5% from 5.2 seconds to 5.6 seconds on my machine, which I don't think is justified by the benefit here.

mcy:
I currently use the `deserialize_with` solution. It feels a bit gross, but it works I guess.

I wonder if const generics would make this less awful. I get the impression that the current stabilization MVP floating around IRLO is probably "good enough" to make this (and probably the implementations for non-borrowed arrays) generic.

In any case, do you think this is the kind of thing that's worth sticking a specialized error message into the derive macros for? It's trivial to add yourself but non-obvious to people starting out with serde.

PR #1888, opened 2020-09-11
Fix hand-written enum variant deserializations to allow u64 discriminant

joshtriplett:
Automatically generated enum variant deserializers allowed any integer
type as the discriminant, but the hand-written ones for specific enum
types such as Result or IpAddr only allowed types up to u32. This broke
some non-human-readable deserializers for these enums, with
deserializers that emit any integer type as a u64. Switch the visit_u32
methods to visit_u64 methods to allow discriminants to have any size up
to a u64.

### DIFF
--- a/serde/src/de/impls.rs
+++ b/serde/src/de/impls.rs
@@ -1313,7 +1313,7 @@ macro_rules! variant_identifier {
                         formatter.write_str($expecting_message)
                     }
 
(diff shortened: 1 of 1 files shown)


dtolnay (approved):
Thanks! This was missed from #963.

joshtriplett:
Thanks for the quick review and merge, and the 1.0.116 release!

Issue #586, opened 2016-10-14
Case-insensitive attribute for serde_derive


alexreg:
Could we have a case-insitive option for the `serde` attribute on types and fields?

e.g. Something like:

```
#[derive(Serialize, Deserialize, Debug)]
#[serde(case_insensitive)] // would allow "Foo", "foo", "FOO", etc.
pub struct Foo {
    #[serde(case_insensitive)] // would allow "a" or "A"
    pub a: u32,
    pub b: i16,
    #[serde(case_insensitive, rename = "foo")] // would allow "foo", "FOO", "fOo", etc.
    pub c: String,
}
```

dtolnay:
A serde(case_insensitive) attribute would be great!

Currently the generated code for a struct like Foo contains:

```rust
fn visit_str<E>(self, field: &str) -> Result<Field, E>
    where E: de::Error
{
    match field {
        "a" => Ok(Field::field0),
        "b" => Ok(Field::field1),
        "foo" => Ok(Field::field2),
        _ => Ok(Field::ignore),
    }
}
```

The attribute would need to change this to do case insensitive comparison. The only case insensitive comparison in Rust I can find right now is [`std::ascii::AsciiExt::eq_ignore_ascii_case`](https://doc.rust-lang.org/std/ascii/trait.AsciiExt.html#tymethod.eq_ignore_ascii_case). Support for unicode-aware comparison is tracked in https://github.com/rust-lang/rfcs/issues/791.

dtolnay:
Here is one workaround for now - a deserialize_with function that converts to lowercase before deserializing.

```rust
#[macro_use]
extern crate serde_derive;

extern crate serde;
extern crate serde_json;

use serde::de::{self, Deserialize, Deserializer};
use serde_json::Value;

use std::collections::BTreeMap as Map;

#[derive(Deserialize, Debug)]
struct Data {
    aaa: u8,
    bbb: u8,
}

#[derive(Deserialize, Debug)]
struct Wrl {
    #[serde(deserialize_with = "case_insensitive")]
    data: Data,
}

fn case_insensitive<T, D>(deserializer: D) -> Result<T, D::Error>
    where T: Deserialize,
          D: Deserializer
{
    let map = Map::<String, Value>::deserialize(deserializer)?;
    let lower = map.into_iter().map(|(k, v)| (k.to_lowercase(), v)).collect();
    T::deserialize(Value::Object(lower)).map_err(de::Error::custom)
}

fn main() {
[...]

PR #1902, opened 2020-10-06
Added #[serde(case_insensitive)] container attribute for case-insensitive identifier deserialization

lasa01:
Added an attritube that enables case-insensitive container field/variant identifier deserialization.

A similar attribute been requested in #586, which was closed because case-insensitive parsing was provided by the serde-aux crate.
The case insensitive implementation there however seems to first parse the data into a map, convert all keys into lowercase and then deserialize the map into the target type.
This is definitely slower than having the code generated by derive compare the field names using case-insensitive comparison.
The serde_aux implementation also depends on serde_json::Value and might not work for other data formats.

This PR uses eq_ignore_ascii_case() for case-insensitive comparison, which limits it to ASCII characters. This should be sufficient for most use cases, and the standard library currently doesn't have a case insensitive comparsion function that supports any characters.

Also added tests test_case_insensitive_struct, test_case_insensitive_enum and test_case_insensitive_bytes.

### DIFF
--- a/serde_derive/src/de.rs
+++ b/serde_derive/src/de.rs
@@ -1905,6 +1905,7 @@ fn deserialize_generated_identifier(
         is_variant,
         fallthrough,
         !is_variant && cattrs.has_flatten(),
+        cattrs.case_insensitive(),
     ));
 
     let lifetime = if !is_variant && cattrs.has_flatten() {
@@ -2012,6 +2013,7 @@ fn deserialize_custom_identifier(
         is_variant,
         fallthrough,
         false,
+        cattrs.case_insensitive(),
     ));
 
     quote_block! {
@@ -2042,6 +2044,7 @@ fn deserialize_identifier(
     is_variant: bool,
     fallthrough: Option<TokenStream>,
     collect_other_fields: bool,
+    case_insensitive: bool,
 ) -> Fragment {
     let mut flat_fields = Vec::new();
     for (_, ident, aliases) in fields {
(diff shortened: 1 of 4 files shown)


lasa01:
Found a small bug, deriving Deserialize on an empty struct with case_insensitive attribute causes the following error:
```
expected expression, found keyword `else`
expected expression

main.rs(5, 10): Error originated from macro here
proc-macro derive produced unparseable tokens
```

lasa01:
This doesn't work on flattened structs, need to investigate it further.

ctaggart:
Does this allow enum variants to be case insensitive? Here, I want to be able to deserialize `NetworkSourceDeleted` or `NETWORKSOURCEDELETED` values as well.

``` rust
    #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
    pub enum State {
        #[serde(rename = "provisioning")]
        Provisioning,
        #[serde(rename = "deprovisioning")]
        Deprovisioning,
        #[serde(rename = "succeeded")]
        Succeeded,
        #[serde(rename = "failed")]
        Failed,
        #[serde(rename = "networkSourceDeleted")]
        NetworkSourceDeleted,
    }
```

lasa01:
You can mark the whole enum as case insensitive, then all variant names
would be case insensitive. The attribute won't work on a single variant
though.

lasa01:
It works now on flattened structs. However, you need to add the #[serde(case_insensitive)] attribute on both the parent struct and the flattened struct. If the attribute is only on the parent struct, the parent struct's fields are case insensitive and the flattened struct's fields case sensitive. If the attribute is not on the parent struct, everything is case sensitive regardless of the flattened struct's attributes.

Also, there is quite a lot of duplicate code between `CaseInsensitiveFlatStructAccess` and `FlatStructAccess` and between `CaseInsensitiveFlatMapDeserializer` and `FlatMapDeserializer`, not sure if this can be improved.

Looks like the build also now fails on older Rust versions without the eq_ignore_ascii_case function. Previously the function was used only in the code the derive generated, and the derive already required Rust 1.31 or newer, so it wasn't a problem. Now it's also used in the case insensitive flat map deserializer which is part of the main crate even without derives enabled. Not sure how this should be solved.

dtolnay (commented):
> Looks like the build also now fails on older Rust versions without the eq_ignore_ascii_case function. Previously the function was used only in the code the derive generated, and the derive already required Rust 1.31 or newer, so it wasn't a problem. Now it's also used in the case insensitive flat map deserializer which is part of the main crate even without derives enabled.

I landed #1966 to skip compiling the derive helpers on older builds. They should only be needed on 1.31+.

sagudev:
Any progress?

ctaggart (approved):
This pull request is looking pretty polished to me. I just tested the latest commit 11ee221 in https://github.com/ctaggart/azure-sdk-for-rust/pull/2 and it fixes the deserialization problems with the Azure Storage service. I'd love to see this merged and released.

avnerbarr:
is this being released?

avitex:
@dtolnay If docs are holding this back, I can make a PR for documenting this in https://github.com/serde-rs/serde-rs.github.io/blob/master/_src/container-attrs.md

avitex:
Looking for comments on #2161 which is an alternative solution that is more generic over the problem

albx79:
I came here looking for case-insensitive enum deserialization. What's the state of this PR? I don't understand if it's been superseded (and should be closed), or if it's still to be merged (what's keeping it?).

Thanks

erikjara:
Almost 2024 💀

deminearchiver:
It's 2024, will this ever be merged?

erikjara:
2025 🎉

DuckyBlender:
please guys

DorianCoding:
2025... still waiting the snow...

apiraino:
Can we please stop **once and for all** posting unhelpful comments? Let's please be mindful that notifications go to all that subscribed this PR (me included).

Thanks for your cooperation :pray:

PR #1916, opened 2020-10-22
Add new container attribute #[serde(expecting = "...")] for specifing custom expectation message

### LINKED ISSUE #1883: Provide an easy way to customize only the error message
**EDIT** After thinking about it some more, I feel it would be better to improve this error message rather than add API surface area. #1544 gets close, but I believe those errors are still too opaque for Cargo's purposes, and it's hidden behind a flag. Anyway, my original post:

-----------------

I have an untagged union:

```rust
#[serde(untagged)]
enum MyEnum {
  Foo(Vec<String>),
  Bar(usize)
}
```

If all variants fail, you get this error message:

https://github.com/serde-rs/serde/blob/9f331cc25753edd71ad7ab0ea08a430fefaa90e1/serde_derive/src/de.rs#L1648

Is there a good way to customize the error message but keep the default `#[derive(Deserialize)]` implementation? Perhaps a new attribute?:

```rust
#[serde(untagged, expecting = "a number or a list of strings")]
enum MyEnum {
  ...
}
```

## Advantages:

- You don't have to reimplement `Deserialize`
- Concise

## Disadvantages

- How common is this use-case? We have to do this a couple of times [within Cargo](https://github.com/rust-lang/cargo/commit/cb53e1b2056a02bd1eb83d926e1cc37e56180368).
- Is there a better solution I am overlooking?

Mingun:
This can be especially useful together with
- `#[serde(field_identifier)]`
- `#[serde(variant_identifier)]`
- `#[serde(untagged)]` -- for that case attribute completely replaces error message

because by default they all have very generic messages.

New attribute usage:
```rust
// default message would be 
// `data did not match any variant of untagged enum OneOrMany`
#[derive(Deserialize)]
#[serde(untagged, expecting = "expected single element or array of elements")]
struct OneOrMany<T> {
  One(T),
  Many(Vec<T>),
}
```

As two side-effects, first commit fixes panic in `assert_de_tokens_error` for enums when token stream doesn't represent an enum, so we can actually test implementation of `Deserialize`d type instead of implementation of test deserializer.

The second small improvement is adding enum name to default message for internally tagged enums:
`expected internally tagged enum`
replaced by
`expected internally tagged enum {}`

Closes #1883

### DIFF
--- a/serde/src/private/de.rs
+++ b/serde/src/private/de.rs
@@ -824,15 +824,17 @@ mod content {
     /// Not public API.
     pub struct TaggedContentVisitor<'de, T> {
         tag_name: &'static str,
+        expecting: &'static str,
         value: PhantomData<TaggedContent<'de, T>>,
     }
 
     impl<'de, T> TaggedContentVisitor<'de, T> {
         /// Visitor for the content of an internally tagged enum with the given
         /// tag name.
-        pub fn new(name: &'static str) -> Self {
+        pub fn new(name: &'static str, expecting: &'static str) -> Self {
             TaggedContentVisitor {
                 tag_name: name,
+                expecting: expecting,
                 value: PhantomData,
             }
         }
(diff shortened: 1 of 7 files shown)


dtolnay (approved):

messense:
This should really be documented in https://serde.rs/container-attrs.html

PR #1917, opened 2020-10-23
Allow to borrow data when deserializing field identifiers

Mingun:
Now that will not be ultimatively failed, if input can be borrowed:
```rust
#[derive(Deserialize)]
#[serde(field_identifier)]
enum Field<'a> {
  Field1,
  ...
  #[serde(borrow)]
  Other(&'a str),
}
```


dtolnay (approved):

dtolnay (commented):
Thanks -- I've merged this with some fixes in #1959.

PR #1918, opened 2020-10-27
Fix incorrect message in serializer tokens and correctly implement next_entry_seed

Mingun:
Two fixes in `serde_test`:
1. `assert_ser_tokens` had mixed up the expected and actual values
2. `Compact` and `Readable` wrappers didn't override `next_entry_seed` and used default implementation. Difference is noticeable for deserializers that rely on call of `next_entry_seed` (for example, quick-xml currently support only `next_entry_seed`)


dtolnay (approved):
Thanks!

PR #1919, opened 2020-10-27
Preserve is_human_readable flag on delegating (de)serializers

Mingun:
Added tests asserting that deserializer correctly reports readable/compact format sign and fix them.

### DIFF
--- a/serde/src/de/value.rs
+++ b/serde/src/de/value.rs
@@ -1000,16 +1000,40 @@ where
 
 ////////////////////////////////////////////////////////////////////////////////
 
-/// A deserializer holding a `SeqAccess`.
+/// Represents [seq accessor] as a [deserializer], that forwards all calls
+/// to [`Visitor::visit_seq`].
+///
+/// Contains seq accessor and human readable capability of the deserializer.
+///
+/// [seq accessor]: ../trait.SeqAccess.html
+/// [deserializer]: ../trait.Deserializer.html
+/// [`Visitor::visit_seq`]: ../trait.Visitor.html#method.visit_seq
 #[derive(Clone, Debug)]
 pub struct SeqAccessDeserializer<A> {
+    /// Sequence assessor represented as a deserializer. Deserializer will forward
+    /// all calls to [`visit_seq()`] with this sequence as an argument.
+    ///
+    /// [`visit_seq()`]: ../trait.Visitor.html#method.visit_seq
     seq: A,
+
+    /// [`is_human_readable()`] attribute of the original deserializer, that
+    /// creates `seq` accessor.
+    ///
+    /// [`is_human_readable()`]: ../trait.Deserializer.html#method.is_human_readable
+    is_human_readable: bool,
 }
 
 impl<A> SeqAccessDeserializer<A> {
-    /// Construct a new `SeqAccessDeserializer<A>`.
+    /// Construct a new `SeqAccessDeserializer<A>` with readable representation
+    /// (`self.is_human_readable() == true`).
     pub fn new(seq: A) -> Self {
-        SeqAccessDeserializer { seq: seq }
+        SeqAccessDeserializer { seq: seq, is_human_readable: true }
+    }
+
+    /// Construct a new `SeqAccessDeserializer<A>` with specified representation
+    /// (`self.is_human_readable()`).
+    pub fn with_representation(seq: A, is_human_readable: bool) -> Self {
+        SeqAccessDeserializer { seq: seq, is_human_readable: is_human_readable }
     }
 }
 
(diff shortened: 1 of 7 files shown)


Mingun:
@dtolnay , FYI, I've rebased and fix conflict

dtolnay (changes requested):
Could you provide some background on why this is warranted or what use case this is intended for?

Mingun:
Emmm... It's just a bug, for me it's enough to fix it

dtolnay (commented):
I'll go ahead and close this in the absence of a use case. It's challenging in general to pin down what is meant by "delegating (de)serializers". Is `I32Deserializer` a "delegating deserializer" if the i32 it's responsible for got extracted from a particular data format? I wouldn't want so many deserializers to be forced with propagating is_human_readable, which was intended to be best effort only. Overall my sense is that this change is introducing bloat that is disproportionate to the significance of the use case.

Mingun:
> Is `I32Deserializer` a "delegating deserializer" if the i32 it's responsible for got extracted from a particular data format?

If it used transparently to implement another deserializer then yes. Anyway, ability to specify `is_human_readable` property should be possible. Another possible solution will use [`Compact` and `Readable` adapters](https://github.com/serde-rs/serde/blob/fa6712d2bf1a6f1f72dfabbc6475d3086d2bbf02/serde_test/src/configure.rs#L9-L12), but that approach will not allow to select format based on the another deserializer format. `FlatMapDeserializer` is a good example. It is used internally and completely transparently for the user and loosing of the initial `is_human_readable` is just a bug. [New tests](https://github.com/serde-rs/serde/pull/1919/files#diff-7fc9f5f32f23d2349b3de346c0ca373a31801d3a7fd46ed9ed79b3c79c4b509d) clearly shows this.

> I wouldn't want so many deserializers to be forced with propagating is_human_readable, which was intended to be best effort only

I don't understand what is "to be best effort only". It is violation of API contract and it is a bug. All changed deserializers is used internally when you convert initial deserializer under the hood to another deserializer in order to implement special features, like flattering or internally tagged enums.

So I wonder why it is not enough for you, what another justification is required for fixing bug other than tests that shows that bug?

nakedible:
@dtolnay I have a concrete use case. I want to use the same structure to serialize and deserialize to JSON and CBOR. If I'm serializing to CBOR, I want to serialize as bytes, as that is natively supported. If I'm serializing to JSON, I want to serialize as `{"$binary":"base64"}`.

Normally, this works fine, I can use `is_human_readable()` in both serializer and deserializer. But, if I place the same value in a `#[serde(untagged)]` enum, it stops working and always returns human readable as true. The reason for this is that when using untagged enums, the deserializer gets wrapped as `ContentRefDeserializer`, which does not carry over the `is_human_readable` flag, which this pull is trying to fix.

So I'd like to echo the sentiment from @Mingun – either *fix* `is_human_readable` so it will always show the correct value even in nested deserializers – or remove the feature entirely, because right now it is unreliable.

hannometer:
@dtolnay It is very, very unintuitive that the `is_human_readable` is not maintained when using `serde(flatten)` (or other attributes). 
Many people stumble over this, as evident by the different issues (this one, #2704, #2172)

In my case, I refactored a struct and moved some members into a separate struct. In order to maintain compatibilty with previously serialized data, I added `serde(flatten)` to the new member. The serialization for the affected objects differs between human readable and non-human readable output. This broke (de-)serialization functionality. 

I could live with the "is_human_readable" being ignored by both the serialization and deserialization when using `serde(flatten)`. It might result in less efficient (de)serialization for binary formats. But it's worse: the serializer **does** make use of the `is_human_readable`, but the deserializer ignores it.

Example:
```rust
use std::net::Ipv4Addr;

struct Addresses {
    ip: Ipv4Addr,
}

struct A {
    #[serde(flatten)]
    addresses: Addresses,
}

fn main() {
    let container = A {
        addresses: Addresses {
            ip: "127.0.0.1".parse().unwrap(),
        },
    };

    let serialized = minicbor_serde::to_vec(&container).unwrap();
    let deserialized: A = minicbor_serde::from_slice(&serialized).unwrap();

    assert_eq!(container.addresses.ip, deserialized.addresses.ip);
}
```
with `Cargo.toml`:
```toml
[package]
name = "serde-flatten-bug"
version = "0.1.0"
edition = "2021"

[dependencies]
serde = { version = "1.0.219", features = ["derive"] }
minicbor-serde = { version = "0.4.1", features = ["full"] }
```

Output:
```
thread 'main' panicked at src/main.rs:23:67:
called `Result::unwrap()` on an `Err` value: DecodeError(Error { err: Message, pos: None, msg: "invalid type: sequence, expected IPv4 address" })
stack backtrace:
    [...]
```

PR #1920, opened 2020-10-28
Show correct location in error messages by tracking caller of utility `assert_tokens` functions

Mingun:
Currently when error originates in the `assert_tokens` and similar functions error position pointed to location inside that helper function, which is not much helpful. This PR improves the situation is some cases. There still have cases where error location points to inside of, for example, `serde_test::ser::Serializer`, because of too many levels of calls, generated by derive macro. I don't know how to improve situation here -- maybe not panic inside of (de)serializer but to return an error?

### DIFF
--- a/serde_test/build.rs
+++ b/serde_test/build.rs
@@ -1,0 +1,33 @@
+use std::env;
+use std::process::Command;
+use std::str;
+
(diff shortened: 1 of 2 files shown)


dtolnay (approved):
Neat!

Issue #1621, opened 2019-09-06
Deserializing internally tagged enum: error line information inaccurate


karlri:
When deserializing a internally tagged enum, everything works as expected when the input is correct, however:

If an error occurs while deserializing "inside" the internally tagged enum, the error line information points to the wrong place.

Note, simpler way to trigger bug in comments section.

Example code:
```
use serde_derive::*;
use serde_json::*;

#[derive(Deserialize)]
pub struct Event {
    event_id: u32,
    #[serde(flatten)]
    variant: EventVariant,
}

#[derive(Deserialize)]
#[serde(tag = "kind")]
pub enum EventVariant {
    Trigger { at: String },
    Interval { start: String, end: String },
}

fn main() {
    // at: 14 is wrong, it should be string. Expect error on line 3.
    let json_str = r#"{
        "kind": "Trigger",
        "at": 14,
        "event_id": 42
    }"#;
    let res: Result<Event> = from_str(json_str);
    match res {
        Err(e) => {
            assert_eq!(e.line(), 3)
        },
        Ok(_) => unreachable!()
    }
}
```

The result of running this code:
```
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `5`,
 right: `3`', src/main.rs:29:13
```

Problem:
Line information is incorrect.

Actual behavior:
error line refers to end of outermost struct: line 5

Expected behavior:
error line refers to the exact line where the parse error occurred: line 3

karlri:
Determined it has nothing to do with flatten. It is internal tagging only that triggers the issue.

Event simpler example:
`use serde_derive::*;
use serde_json::*;

#[derive(Deserialize, Debug)]
#[serde(tag = "kind")]
pub enum EventVariant {
    Trigger { at: String },
    Interval { start: String, end: String },
}

fn main() {
    // at: 14 is wrong, it should be string. Expect error on line 3.
    let json_str = r#"{
        "kind": "Trigger",
        "at": 14
    }"#;
    let res: Result<EventVariant> = from_str(json_str);
    match res {
        Err(e) => {
            println!("{:?}", e);
            assert_eq!(e.line(), 3)
        },
        Ok(_) => unreachable!()
    }
}`
Output:
`Error("invalid type: integer `14`, expected a string", line: 0, column: 0)
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `0`,
 right: `3`', src/main.rs:21:13
[...]

PR #1922, opened 2020-11-03
Optimize internal tagged enums -- do not use internal buffer if tag is the first field

### LINKED ISSUE #1495: Non-externally-tagged enum representations could be more efficient
I'm still working on producing a decent reproducible case for a better description of the issue, but decided to file an issue now for a braindump of what I came up with so far and to initiate a discussion.

Few days ago I tried to rewrite a parser for a pretty deep AST structure serialised as JSON. As is very common for AST structures, it's represented as a tree of internally-tagged objects like `{"type": "...", ...}` that should be parsed into a tree of Rust tagged enums.

Initially the code used `json` crate, parsing into a `json::JsonValue` and then it traversing that intermediate representation and manually created corresponding Rust types.

I moved it to use `serde_json::Value` which made performance slightly worse, but that was fine as it was meant only as an intermediate step. The endgame I was planning to end up with is using `#[derive(Deserialize)] #[serde(tag = "...")]`, as it would, in theory, avoid an intermediate JSON tree altogether and could statically parse everything into static structures, and so make everything much faster.

Unfortunately, after doing all this migration and ending up with just static structures, I found that this made entire process >2x slower when parsing directly from string. This felt extremely surprising and, at first, I thought that the problem is somewhere in my code, but, after digging into `serde`'s implementation I don't think so anymore.

Currently, it seems that using any non-externally-tagged representation causes Serde to collect entire input (JSON or not) into an intermediate `Content` tree, which explains the difference I was seeing - now, when using something like `serde_json::from_value`, there are two intermediate representations in place, and, when using `serde_json::from_str`, there is one but probably with less efficient matching than original manual `match obj["type"] { ... }`.

Now that I see it, I understand that this buffering into an intermediate is necessary to support cases where serde parses from a streaming reader and tag happens to be somewhere in the end of the object (like `{..., "type": "..."}`), but I think we could do much better for common cases and decided to raise this issue to discuss some ideas.

0. Before doing anything else, it's probably worth adding at least the AST case (which is quite common for JSON) as an internally-tagged benchmark to https://github.com/serde-rs/json-benchmark, as currently this workload is not represented there at all.
1. While JSON doesn't have any ordering, we could utilise the fact that more often than not tag does come up first in internally-tagged objects (and adjacently-tagged too). If we see that we already found it as the first element of the `MapAccess`, then we could check the tag value and pass the unconsumed part of `MapAccess` directly to proper static deserializer (this is what I've started to attempt, but seems to require quite a bit of refactoring which makes me uneasy).
2. Same idea as above, but could be extended to support tags somewhere in the middle too - buffer anything before the tag is found into a `Content`, then read the tag and then join `Content` + unconsumed part of `MapAccess` as a deserializer for the value. Although this one seems like overly complicated for unlikely benefits (it's not very common to have tags serialised somewhere in the middle).
3. Could allow deserialisers to mark themselves as "seekable"? This way any deserialiser over a string or a byte slice etc. could mark itself as such, and allow quickly skipping over any key-value pairs until it finds the tag, and then serde could "rewind" it back to the beginning of the map for value payload deserialisation, all without parsing into an intermediate dynamic buffer. Also seems more complicated than (1) and might be wasteful when skipping over values is as expensive as parsing, but should also help with common cases.

Does any of these sound good? Any other ideas?

Mingun:
Fixes #1495

Also that change has one positive side-effect: if tag is the first field, negative effects from #1183 is eliminated, because buffering is not used

@RReverser, feel free to try to run your benchmarks against this branch


dtolnay (changes requested):
The discussion in #1495 focuses on whether this would be worth it for the cost in compile time. The biggest thing wrong with internally tagged enums is not when the deserializer is slow but that they take significantly long to compile. Overall I would rather optimize for lowering their compile time, not on performance or features at the expense of compile time.

Would you be able to provide some measurements showing the impact of this change on the time to compile internally tagged enums?

Mingun:
I cannot agree with such a question. For me, the runtime performance is much more important thing that compile-time performance. Things are written not for the pleasure of developers, but for solving customer problems.

> Would you be able to provide some measurements showing the impact of this change on the time to compile internally tagged enums?

I will try to study how to do performance measurements, but any guidance is welcome

RReverser:
> Overall I would rather optimize for lowering their compile time, not on performance or features at the expense of compile time.

That sounds odd tbh. Compile times are affecting only developers, while runtime affects every user of the library / application, which is way more impactful. Why choose compile-time over runtime perf here when we don't do that at any other levels of development (e.g. `opt-level = 0` vs `opt-level = 2` etc.)?

Mingun:
I've made some research and there is the results. I've created a library project with 1000 types and I've measured compilation time.

I've noticed small increasing of the compilation time, about 0.01 sec per type (or 7-30%). I think it is acceptable worth for the bigger runtime performance.

<details>
<summary>Test code and raw data</summary>

[serde-perf.zip](https://github.com/serde-rs/serde/files/6095693/serde-perf.zip)

Created library cargo project with following `lib.rs` content:
```rust
use serde::{Deserialize};
macro_rules! generate {
  ($(#[$counter:meta])*) => {
    $(
      const _: () = {
        #[$counter]
        #[derive(Deserialize)]
        #[serde(tag = "tag")]
        enum Node {
          Unit,
          Struct {
            name: String,
            list: Vec<Node>,
          },
          // Uncomment for "big enum" tests
          /*
          Newtype1(std::collections::HashMap<String, String>),
          Newtype2(String),
          Newtype3(u32),
          Newtype4(f32),
          Unit1,
          Unit2,
          Unit3,
          Unit4,
          Struct1 { f1: String, f2: u32, f3: bool, f4: f64 },
          Struct2 { f1: String, f2: u32, f3: bool, f4: f64 },
          Struct3 { f1: String, f2: u32, f3: bool, f4: f64 },
          Struct4 { f1: String, f2: u32, f3: bool, f4: f64 },// */
        }
      };
    )*
  };
}
// Expanded manually for "expand" tests
generate!(
  /// ...
  /// 1000 lines
  /// ...
);
```
Tests run with command
```sh
cargo +nightly build -Ztimings
```

|OS version|Windows_NT x64 10.0.18363|
|CPUs|Intel(R) Core(TM) i7-8550U CPU @ 1.80GHz (8 x 1992)|
</details>

<details>
<summary>Summary table</summary>

### Small enum (2 variants), 1000 types

Derived both `Serialize` and `Deserialize`, types generated with `generate!` macro
|Deserialize + Serialize |master (c26101532509477132721a8c56b7024891aaf1b4) (100%)|PR (a169beee0d84fa0cad022357c974d05c223e32a0)      | Diff
|------------------------|--------------------------------------------------------|---------------------------------------------------|-------------
|all                     |47.37 = (46.96 + 47.10 + 48.06)/3                       |58.67 = (59.27 + 59.17 + 57.56)/3                  |+11.29 (+29%)
|codegen                 | 3.28 = (46.96-43.95 + 47.10-43.92 + 48.06-44.42)/3     | 4.81 = (59.27-53.93 + 59.17-53.85 + 57.56-53.80)/3| +1.53 (+10%)

Derived only `Deserialize`, types generated with `generate!` macro
|Deserialize             |master (c26101532509477132721a8c56b7024891aaf1b4) (100%)|PR (a169beee0d84fa0cad022357c974d05c223e32a0)      | Diff
|------------------------|--------------------------------------------------------|---------------------------------------------------|-------------
|all                     |39.25 = (39.16 + 39.66 + 38.94)/3                       |50.08 = (49.70 + 51.62 + 48.93)/3                  |+10.83 (+28%)
|codegen                 | 3.62 = (39.16-35.31 + 39.66-35.28 + 38.94-36.31)/3     | 4.53 = (49.70-46.17 + 51.62-45.29 + 48.93-45.21)/3| +0.91 (+25%)

Derived only `Deserialize`, types written manually
|Deserialize + expanded  |master (c26101532509477132721a8c56b7024891aaf1b4) (100%)|PR (a169beee0d84fa0cad022357c974d05c223e32a0)      | Diff
|------------------------|--------------------------------------------------------|---------------------------------------------------|-------------
|all                     |38.76 = (38.38 + 39.67 + 38.23)/3                       |50.36 = (50.34 + 50.59 + 50.14)/3                  |+11.60 (+30%)
|codegen                 | 3.24 = (38.38-35.41 + 39.67-35.65 + 38.23-35.50)/3     | 5.67 = (50.34-44.34 + 50.59-44.74 + 50.14-45.02)/3| +2.42 (+75%)

### Big enum (14 variants), 1000 types

Derived only `Deserialize`, types generated with `generate!` macro
|Deserialize             |master (c26101532509477132721a8c56b7024891aaf1b4) (100%)  |PR (a169beee0d84fa0cad022357c974d05c223e32a0)             | Diff
|------------------------|----------------------------------------------------------|----------------------------------------------------------|-------------
|all                     |241.04 = (236.94 + 239.35 + 246.84)/3                     |257.01 = (257.75 + 258.76 + 254.70)/3                     |+16.03 (+7%)
|codegen                 | 40.90 = (236.94-197.11 + 239.35-199.61 + 246.84-203.70)/3| 46.63 = (257.75-211.25 + 258.76-210.17 + 254.70-209.91)/3| +5.72 (+14%)

Derived only `Deserialize`, types written manually
|Deserialize + expanded  |master (c26101532509477132721a8c56b7024891aaf1b4) (100%)  |PR (a169beee0d84fa0cad022357c974d05c223e32a0)             | Diff
|------------------------|----------------------------------------------------------|----------------------------------------------------------|-------------
|all                     |238.41 = (238.49 + 236.58 + 240.16)/3                     |254.86 = (258.93 + 254.75 + 250.90)/3                     |+16.45 (+7%)
|codegen                 | 39.70 = (238.49-198.46 + 236.58-199.66 + 240.16-198.02)/3| 45.05 = (258.93-209.91 + 254.75-211.19 + 250.90-208.34)/3| +5.35 (+13%)

</details>

pickfire:
@Mingun What about runtime performance improvements?

Mingun:
I didn't measure it, maybe I should

jarredholman:
Would this also fix the incorrect error messages caused by the internal buffer? https://github.com/serde-rs/serde/issues/1621

Mingun:
For the optimized case yes, it should

serde/src/de/value.rs
```diff
@@ -1076,9 +1080,38 @@ where
...
+        self.deserialize_unit(visitor)
+    }
+
+    fn deserialize_newtype_struct<V>(self, _name: &str, visitor: V) -> Result<V::Value, Self::Error>
+    where
+        V: de::Visitor<'de>,
+    {
+        visitor.visit_newtype_struct(self)
+    }
+
```
Mingun, line 1111: I'm not sure, should  we change behavior of `SeqAccessDeserializer` and `MapAccessDeserializer` or introduce new private deserializers? From one hand those deserializers was created for support of various serde attributes. From the other hand, technically this is breaking change because those types are public.

Mingun (commented):
@dtolnay, @oli-obk , this PR ready for review again.

Mingun:
[As I remember](https://github.com/serde-rs/serde/pull/2445#issuecomment-2322202957), changes in this PR depends on changes in #2445.

As I already said, if you wish, it is possible to do not touch `MapAccessDeserializer` and `SeqAccessDeserializer`, but instead introduce new private deserializers.

@dtolnay, @oli-obk, please give your opinion, should I make these changes?

Mingun:
No. Any suggestions for the benchmark are welcome.

RReverser:
> No. Any suggestions for the benchmark are welcome.

It's pretty ancient by now, but in the original issue I referenced https://github.com/binast/binjs-ref/commit/22103b90287dda3bb49c89d51bee85c7d2888873 where I did a manual implementation of this optimisation just for the types we used.

For testing, I checked out a commit right before that. https://github.com/binast/binjs-ref/commit/53bd87aeb70538feb50f52ab2cdeeeb57b932f8b

The numbers before this PR:

```
test bench_parsing_reuse_parser       ... bench:  88,088,720 ns/iter (+/- 12,560,451)
```

The numbers with this PR applied:

```
test bench_parsing_reuse_parser       ... bench:  66,515,390 ns/iter (+/- 7,634,431)
```

Note that this is far from a pure JSON benchmark - it uses an external Node.js process to parse JS and produce JSON, and only then parses the output using serde-json, but in that context the -25% perf improvement is even more impressive.

It should be easy to save the JSON output and do a pure serde-json benchmark instead (in my original commit I suggested that showed 2x improvement, which seems realistic), but perhaps someone has more modern examples.

Anything touching JS AST represented as JSON (e.g. [Deserialize for ESTree Program](https://docs.rs/swc_estree_ast/latest/swc_estree_ast/struct.Program.html#impl-Deserialize%3C'de%3E-for-Program) from https://swc.rs/) should work.

serde/src/private/de.rs
```diff
@@ -2287,11 +2308,17 @@ mod content {
             )
         }
 
-        fn visit_seq<S>(self, _: S) -> Result<(), S::Error>
+        fn visit_seq<S>(self, mut seq: S) -> Result<(), S::Error>
         where
             S: SeqAccess<'de>,
         {
-            Ok(())
+            match tri!(seq.next_element()) {
```
RReverser, line 2315: This behaves quite differently from `IgnoredAny.visit_map`. I think the behaviour should be consistent, as in, iterate over the entire sequence and ignore its values.

Mingun: I tried that initially, but that failed other tests and in general not what you want. The unit / unit struct represented in sequence as nothing, so we need to ensure that sequence is empty. This is consistent with normal behavior where struct deserialization from a sequence expects exact number of values, and those fact that flattened unit / unit struct considered as equal to the struct without fields.

Mingun: Actually, the `unit_variant_with_unknown_fields` is a test that failed if consume the whole sequence here.
https://github.com/serde-rs/serde/blob/3aca38d2d3c9b5dba9bcb25be2e16bde6678dbf6/test_suite/tests/test_enum_internally_tagged.rs#L1447-L1456

RReverser: > The unit / unit struct represented in sequence as nothing

Hm but "nothing" should be pretty different conceptually from "ignored any". I'd expect a custom check just for the nothing case, whereas ignored any should be able to consume anything thrown at it silently.

Mingun: This code tries to read something, doesn't matter what. We expect an empty sequence, so if it contains some element, we fail.

RReverser: Nevermind, I'm sleepy - I thought you're changing how IgnoredAny works everywhere. I've expanded the context of the diff and I see this is a change on this one specific visitor.

Please disregard my original comment 🤦‍♂️

Although I now wonder if `visit_map` should be changed to check length as well.

Mingun: By default maps in serde allows unknown keys and when unit is flattened, all keys become unknown. But you're right -- in case of `#[serde(deny_unknown_fields)]` we should return error if map not empty. That's idea for another PR!

Mingun:
@oli-obk, how your review progressed?

Mingun:
@dtolnay, @oli-obk, any feedback? Also for other my PRs, please.

oli-obk:
Please refrain from pinging, it's already one of the only two PRs in my self review/assigned list

RReverser:
> Please refrain from pinging, it's already one of the only two PRs in my self review/assigned list

Sorry to bother, but... is this still the case? Do you need any help with maintenance / reviews (in general)?

oli-obk:
> Do you need any help with maintenance / reviews (in general)?

Mingun reviewing PRs has helped in reviewing them as a lot of things had been covered that I then didn't need to.

In general it's hard tho, one can't just hand off serde and we haven't finished landing the serde core PR that would make writing other derive crates feasible

Also a lot of reviews (not this one, this one is just me failing at prioritizing work) are simply blocked on "we don't have crater-for-serde"

RReverser:
Oh wow, it's about to hit 5 years anniversary 😅

PR #1926, opened 2020-11-24
Allow serializing of Wrapping without std

attente:
I'm not sure why, but `std::num::Wrapping` implements `Serialize` and `Deserialize`, but only when `std` is enabled. This change allows it in the `no-std` case as well.


dtolnay (approved):
Thanks!

PR #1948, opened 2021-01-12
refactor: Merge multiple FromStr visitors to a single FromStrVisitor

Marwes:
Only refactors, doesn't actually expose this visitor


dtolnay (approved):
Thanks!

PR #1955, opened 2021-01-19
Add 128-bit integer support to de::IgnoredAny

TheJokr:
This fixes the errors that occur when `IgnoredAny` is deserialized from anything containing a 128-bit integer somewhere.
As `IgnoredAny` is used in `serde_derive` to skip ignored fields in structs, these errors currently prevent deserializing structs with an ignored field containing a 128-bit integer in the input.

There's a second bug I encountered related to 128-bit integer support for untagged enums, but that one already has an issue (#1717) and a draft PR (#1719).

### DIFF
--- a/serde/src/de/ignored_any.rs
+++ b/serde/src/de/ignored_any.rs
@@ -130,10 +130,26 @@ impl<'de> Visitor<'de> for IgnoredAny {
         Ok(IgnoredAny)
     }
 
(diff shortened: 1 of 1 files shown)


dtolnay (approved):
Thanks!

dtolnay:
Published in 1.0.120.

PR #1958, opened 2021-01-20
Prevent panic when deserializing malformed Duration

### LINKED ISSUE #1933: Serde panics when deserializing malformed Duration
The documentation for `std::time::Duration::new` states that it panics if `nanos` overflows and causes `secs` to overflow.

This should cause a deserialization error instead of causing the application to panic.

jonasbb:
std::time::Duration::new can panic. There is no alternative non-panicing constructor.
Check the panic condition beforehand and return an error instead of panicing.

Fixes #1933

### DIFF
--- a/serde/src/de/impls.rs
+++ b/serde/src/de/impls.rs
@@ -1849,6 +1849,17 @@ impl<'de> Deserialize<'de> for Duration {
             }
         }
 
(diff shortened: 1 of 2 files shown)


dtolnay (approved):
Thanks!

PR #1972, opened 2021-01-26
Add alias_all attribute for containers similar to rename_all

### LINKED ISSUE #1530: Add alias_all attribute for containers similar to rename_all
The `rename_all` attribute is very handy for cases where you have fields named in ways that don't follow Rust conventions. However, there are cases where you might want to support both ways - for example, loading data with support for both naming conventions, but preferring a different naming convention for saving that data later. I propose that an `alias_all` container attribute be added that functions similarly to `rename_all`, so that a struct like this...

```rust
#[serde(alias_all = "PascalCase")]
pub struct MyStruct {
    string: String,
    bool: bool
}
```
...can be deserialized from JSON like this...
```json
{"String":"hello world","Bool":true}
```
...but will still be serialized like this:
```json
{"string":"hello world","bool":true}
```

MicroJoe:
Closes #1530.

### DIFF
--- a/serde_derive/src/internals/attr.rs
+++ b/serde_derive/src/internals/attr.rs
@@ -212,6 +212,7 @@ pub struct Container {
     deny_unknown_fields: bool,
     default: Default,
     rename_all_rules: RenameAllRules,
+    alias_all_rules: Vec<RenameRule>,
     ser_bound: Option<Vec<syn::WherePredicate>>,
     de_bound: Option<Vec<syn::WherePredicate>>,
     tag: TagType,
@@ -295,6 +296,7 @@ impl Container {
         let mut default = Attr::none(cx, DEFAULT);
         let mut rename_all_ser_rule = Attr::none(cx, RENAME_ALL);
         let mut rename_all_de_rule = Attr::none(cx, RENAME_ALL);
+        let mut alias_all_rules = VecAttr::none(cx, ALIAS_ALL);
         let mut ser_bound = Attr::none(cx, BOUND);
         let mut de_bound = Attr::none(cx, BOUND);
         let mut untagged = BoolAttr::none(cx, UNTAGGED);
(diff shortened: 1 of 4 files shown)


serde_derive/src/internals/attr.rs
```diff
@@ -345,6 +347,16 @@ impl Container {
                     }
                 }
 
+                // Parse `#[serde(alias_all = "foo")]`
+                Meta(NameValue(m)) if m.path == ALIAS_ALL => {
+                    if let Ok(s) = get_lit_str(cx, ALIAS_ALL, &m.lit) {
```
pickfire, line 352: PascalCase and camelCase differs only by the first character, maybe we can have an optimization to not check all the characters twice?

MicroJoe: I am not sure to understand, the renaming is done somewhere else and is not impacted by this change. I think this optimization is not needed and would be hard to implement given current implementation.

dtolnay (commented):
Thanks for the PR! I am closing because I would prefer not to build this new attribute into serde_derive. Over time since 1.0 I have tried to be increasingly selective about adding more attributes, approaching a moratorium on new attributes, to prevent the list from expanding forever and destroying compile times, and to curb the superlinear complexity of attributes interacting with each other. I think this one is just too uncommonly applicable to make sense adding right now.

DzenanJupic:
@dtolnay  Given, that #1530 and this PR received exclusively positive feedback (there's even a [Readdit thread](https://www.reddit.com/r/rust/comments/l67tcv/parsing_realworld_data_with_rust_introducing_the/) and a [blog post](https://www.reddit.com/r/rust/comments/l67tcv/parsing_realworld_data_with_rust_introducing_the/) covering the problem and the issue), is there a likelihood that you will reopen this PR at some point? Or is #1530 just a dead issue that will eventually be closed?

PR #1974, opened 2021-02-02
New internally tagged tests

Mingun:
Added missing tests that proves that internally tagged enums can be deserialized from all serialized forms:
- Struct representation (named list of key-value pairs)
- Map representation (unnamed list of key-value pairs)
- Seq representation (unnamed list of values)


dtolnay (approved):
Thanks!

PR #1997, opened 2021-03-06
Prevent various panics when deserializing malformed SystemTime

### LINKED ISSUE #1996: Serde panics when deserializing a malformed SystemTime
`SystemTime` is deserialized by deserializing a duration and adding it to `UNIX_EPOCH`. The documentation states that this can panic `if the resulting point in time cannot be represented by the underlying data structure`.

Also, `SystemTime` panics in the same situations as described for `Duration` in #1993.

cyang1:
`impl Add<Duration> for SystemTime` can panic. Moves to using `checked_add` instead.

Also implements #1958 for `SystemTime`, which checks the panic conditions for `std::Time::Duration::new` before calling.

Fixes #1996.

### DIFF
--- a/serde/build.rs
+++ b/serde/build.rs
@@ -76,12 +76,14 @@ fn main() {
         println!("cargo:rustc-cfg=serde_derive");
     }
 
-    // TryFrom, Atomic types, and non-zero signed integers stabilized in Rust 1.34:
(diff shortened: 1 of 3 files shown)


cyang1:
Oh, it's news to me that `SystemTime::checked_add` is new as of 1.34.0. Wonder what I can do about that.

dtolnay (approved):
Thanks!

PR #2005, opened 2021-03-20
Provide serialisation context to data structures

### LINKED ISSUE #2003: Additional data format context required to determine serde data model representation
I've come upon a situation where my data structures must be serialised into/deserialised from multiple formats that each require different representations in serde's data model:

- data elements are named differently (aliases can address this for deserialisation, but there doesn't presently appear to be any solution for serialisation)

- flattening (of elements into their container) occurs differently

- certain values might be serialised in either a "flat" representation (e.g. a string) or a "structured" representation (but context that is lost in serde's data model is required for this determination to be made)

I see that [`Serializer::is_human_readable`](https://docs.serde.rs/serde/trait.Serializer.html#method.is_human_readable) and [`Deserializer::is_human_readable `](https://docs.serde.rs/serde/trait.Deserializer.html#method.is_human_readable) exist so that data structures can adapt their serde data model representations to accommodate certain differences in the data format—but obviously this is limited only to whether the format "is human readable" or not.

In my case, similar functions to indicate whether the format "is flat" (vs "is structured") would suffice.  But perhaps there could be a more general/extensible means for the format to pass some context from its Serializer/Deserializer implementations?

eggyal:
In order for representations in the serde data model to be adaptable to
particular requirements from different serialisation formats, we provide
means for the de/serializer to be queried for additional context.

This is an escape hatch for niche situations, that increases coupling
between the data structure and the de/serializer and should therefore be
avoided wherever possible.  Neither serialisation format nor data
structure should assume/require that the other supports any particular
context, but should instead adopt sensible defaults whenever an expected
context is unsupported.

Closes #2003

### DIFF
--- a/serde/src/de/mod.rs
+++ b/serde/src/de/mod.rs
@@ -1213,6 +1213,93 @@ pub trait Deserializer<'de>: Sized {
     fn is_human_readable(&self) -> bool {
         true
     }
+
+    /// Enable `Deserialize` implementations to obtain context that determines the
+    /// form from which they should expect to deserialize data.  This is a more
+    /// generalized version of the [`is_human_readable`] method.
+    ///
+    /// Some types have multiple forms that may each have different representations
+    /// in serde's data model.  For example, one form may use different element
+    /// names from another.
+    ///
+    /// ```edition2018
+    /// # use std::ops::Add;
+    /// # use std::str::FromStr;
+    /// #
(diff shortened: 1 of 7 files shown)


dtolnay (commented):
Thanks for the PR!

I think I would prefer not to build this into the Serializer/Deserializer traits. I would recommend using a thread-local instead to expose your context to data types.

PR #2017, opened 2021-04-17
Add `#[serde(from_str)]` and `#[serde(to_string)]` attributes

jaehl:
This makes it easier to implement `Serialize` and `Deserialize` for
types that already implement `FromStr` and `ToString`, respectively.


dtolnay (commented):
Thanks for the PR!

I think I would prefer not to build this into serde_derive as new attributes, since it doesn't feel like a sufficiently widespread use case, and is only substituting for impls that are trivial to handwrite anyway:

```rust
impl Serialize for T {
    fn serialize<S: Serializer>(&self, s: S) -> Result<S::Ok, S::Error> {
        s.collect_str(self)
    }
}

impl<'de> Deserialize<'de> for T {
    fn deserialize<D: Deserializer<'de>>(d: D) -> Result<Self, D::Error> {
        let string = String::deserialize(d)?;
        string.parse().map_err(serde::de::Error::custom)
    }
}
```

If there is a project that needs such impls often enough to require a macro, generating a bunch of these impls with a macro_rules should be sufficient, or alternatively a crate like https://crates.io/crates/serde_with may be willing to accept helpers for it.

jonasbb:
For people finding this issue later, the `serde_with` crate already implements two derive macros [`serde_with::SerializeDisplay`](https://docs.rs/serde_with/1.9.4/serde_with/derive.SerializeDisplay.html) and [`serde_with::DeserializeFromStr`](https://docs.rs/serde_with/1.9.4/serde_with/derive.DeserializeFromStr.html). The derived `Deserialize` implementation is also more advanced than shown here, allowing it to work without allocations.

PR #2021, opened 2021-04-29
Allow accessing internals for reexportable derives

SergioBenitez:
So. This one's a bit of a doozy.

Rocket is integrating `serde` as a core dependency. I'd like to reexport serde's derives (and other things) from `rocket::serde` so that, in the common case, a user can derive `Serialize` and `Deserialize` without depending directly on `serde`:

```rust
use rocket::serde::Serialize;

#[derive(Serialize)]
struct Foo { .. }
```

Of course, this doesn't work because there is no `serde` crate in the root:

```rust
error[E0463]: can't find crate for `serde`
   |
 3 | #[derive(Serialize)]
   |          ^^^^^^^^^ can't find crate
```

While a user can add `#[serde(crate = "rocket::serde")]` manually to fix this, I'd like for this to _just work_. So, the idea is for Rocket to define its own `Serialize` and `Deserialize` derives that work _exactly_ like `serde`'s. This commit makes this possible. With these changes, this can be done as:

```rust
#[macro_use] extern crate quote;
#[macro_use] extern crate syn;
extern crate proc_macro;
extern crate proc_macro2;

use syn::DeriveInput;
use proc_macro::TokenStream;

const SERDE_EXPORT_PATH: &'static str = "::rocket::serde";

#[proc_macro_derive(Serialize, attributes(serde))]
pub fn derive_serialize(input: TokenStream) -> TokenStream {
    let mut input = parse_macro_input!(input as DeriveInput);
    add_default_serde_export_path(&mut input);
    serde_derive_internals::ser::expand_derive_serialize(&mut input)
        .unwrap_or_else(to_compile_errors)
        .into()
}

#[proc_macro_derive(Deserialize, attributes(serde))]
pub fn derive_deserialize(input: TokenStream) -> TokenStream {
    let mut input = parse_macro_input!(input as DeriveInput);
    add_default_serde_export_path(&mut input);
    serde_derive_internals::de::expand_derive_deserialize(&mut input)
        .unwrap_or_else(to_compile_errors)
        .into()
}

fn add_default_serde_export_path(to: &mut DeriveInput) {
    let crate_override = to.attrs.iter()
        .filter_map(|attr| attr.parse_meta().ok())
        .filter_map(|meta| match meta {
            syn::Meta::List(list) if list.path.is_ident("serde") => Some(list.nested),
            _ => None
        })
        .flat_map(|list| list.into_iter())
        .filter(|nested| match nested {
            syn::NestedMeta::Meta(meta) => meta.path().is_ident("crate"),
            _ => false
        })
        .next();

    if crate_override.is_none() {
        to.attrs.insert(0, parse_quote!(#[serde(crate = #SERDE_EXPORT_PATH)]))
    }
}

fn to_compile_errors(errors: Vec<syn::Error>) -> proc_macro2::TokenStream {
    let compile_errors = errors.iter().map(syn::Error::to_compile_error);
    quote!(#(#compile_errors)*)
}
```

Then, the `derive` above _just works_, and the user can still override the external crate to use.

As an alternative, `serde` could export some kind of `bang!` macro that does exactly what the `Serialize`/`Deserialize` impls do. Then, instead of calling the internal methods in the new derives, they can simply emit a call to the `bang!` macro:

```rust
#[macro_use] extern crate quote;
extern crate proc_macro;
extern crate proc_macro2;

use proc_macro::TokenStream;

const SERDE_EXPORT_PATH: &'static str = "::rocket::serde";

#[proc_macro_derive(Serialize, attributes(serde))]
pub fn derive_serialize(input: TokenStream) -> TokenStream {
    let input: proc_macro2::TokenStream = input.into();
    quote!(::rocket::serde::serialize!(#SERDE_EXPORT_PATH, #input)).into()
}

#[proc_macro_derive(Deserialize, attributes(serde))]
pub fn derive_deserialize(input: TokenStream) -> TokenStream {
    let input: proc_macro2::TokenStream = input.into();
    quote!(::rocket::serde::deserialize!(#SERDE_EXPORT_PATH, #input)).into()
}
```

I'd be satisfied with either solution.

### DIFF
--- a/serde_derive_internals/lib.rs
+++ b/serde_derive_internals/lib.rs
@@ -37,11 +37,29 @@
 #[macro_use]
 extern crate syn;
 
-extern crate proc_macro2;
+#[macro_use]
 extern crate quote;
 
-#[cfg_attr(serde_build_from_git, path = "../serde_derive/src/internals/mod.rs")]
-#[cfg_attr(not(serde_build_from_git), path = "src/mod.rs")]
-mod internals;
+extern crate proc_macro2;
 
-pub use internals::*;
+#[cfg_attr(serde_build_from_git, path = "../serde_derive/src/")]
+#[cfg_attr(not(serde_build_from_git), path = "./src/")]
+mod derive {
+    pub mod internals;
+
+    #[macro_use]
+    pub mod bound;
+
+    #[macro_use]
+    pub mod fragment;
+
+    pub mod de;
+    pub mod dummy;
+    pub mod pretend;
+    pub mod ser;
+    pub mod try;
+}
+
+pub use derive::internals::*;
+
+pub use derive::*;
--- a/serde_derive_internals/src
+++ b/serde_derive_internals/src
@@ -1,1 +1,1 @@
-../serde_derive/src/internals/
+../serde_derive/src


dtolnay (commented):
I think I would prefer not to make this change (nor the suggested alternative). Surely the user importing the macro from `serde::Serialize` not `rocket::serde::Serialize` should be fine right? If not, I would recommend maintaining a rocket-specific fork of serde_derive_internals containing what you need.

As soon as there is a way for macro-generated code to refer reliable to a particular crate in a way that survives re-exports, I would happily adopt that.

SergioBenitez:
> I think I would prefer not to make this change (nor the suggested alternative).

What is your reservation with the alternative? It seems like a low/no-maintenance solution (which I'd be happy to implement) that resolves this concern nicely. If the derive doesn't already do this, the only thing that would need to change is that serde should keep the last instance of a given container attribute. 

> Surely the user importing the macro from `serde::Serialize` not `rocket::serde::Serialize` should be fine right?

The idea is that `serde` becomes something that Rocket's users can benefit from right out of the box with no configuration at all. Since the release of 0.5-rc, we've already had a few requests from users hoping to remove `serde` from their dependencies and either 1) being unable to do so, or 2) lamenting the additional `#[serde(crate = "rocket::serde")]`. Most seem to be opting for the latter, so it seems that my desire to not have `serde` be an explicit dependency is shared by Rocket's users as well.

> If not, I would recommend maintaining a rocket-specific fork of serde_derive_internals containing what you need.

This is a much higher maintenance burden that the alternative I proposed above while simultaneously being really suboptimal. It duplicates code and increases compilation time if anyone _does_ depend on `serde` with the `derive` feature enabled.

> As soon as there is a way for macro-generated code to refer reliable to a particular crate in a way that survives re-exports, I would happily adopt that.

As much as I'd like to see this as well, I cannot imagine such a feature being shipped in stable, or at all, any time soon. I'd love to give Rocket's users seamless access to `serde`. This is the only missing puzzle piece.

dtolnay:
> we've already had a few requests from users hoping to remove `serde` from their dependencies and either 1) being unable to do so, or 2) lamenting the additional `#[serde(crate = "rocket::serde")]`. Most seem to be opting for the latter, so it seems that my desire to not have `serde` be an explicit dependency is shared by Rocket's users as well.

I see this as self-inflicted :slightly_frowning_face:. The 0.5-rc documentation like https://docs.rs/rocket/0.5.0-rc.1/rocket/serde/json/fn.from_str.html steers people toward `use rocket::serde::Deserialize` + `#[serde(crate = "rocket::serde")]` which basically forces everyone following this documentation to stumble on a missing crate error and debug it on their own. If the documentation showed `[dependencies] serde = "1.0"` + `use serde::Deserialize`, I am very skeptical that anyone would have an objection.

SergioBenitez:
> > we've already had a few requests from users hoping to remove `serde` from their dependencies and either 1) being unable to do so, or 2) lamenting the additional `#[serde(crate = "rocket::serde")]`. Most seem to be opting for the latter, so it seems that my desire to not have `serde` be an explicit dependency is shared by Rocket's users as well.
> 
> I see this as self-inflicted slightly_frowning_face. The 0.5-rc documentation like https://docs.rs/rocket/0.5.0-rc.1/rocket/serde/json/fn.from_str.html steers people toward `use rocket::serde::Deserialize` + `#[serde(crate = "rocket::serde")]` which basically forces everyone following this documentation to stumble on a missing crate error and debug it on their own.

You have the implication arrow backwards. I _added_ that bit of documentation to prevent confusion, and indeed queries dropped. Clearly if they were reading _these_ docs, they'd see `#[serde(crate = "rocket::serde")]` and have no issues.

We _could_ simply not reexport `Serialize` and `Deserialize`, and that might prevent issues. But again, the point is to make this seamless. Sadly, it seems we're not aligned in there being value in that. But I have no idea why. The maintenance burden on your side to expose the bang macro is about zero. The utility is obvious. I'm not the first nor will I be the last to call for this functionality in serde. With near-to-no downside and a clear upside, what is the hangup?

dtolnay:
The hangup is:

1. I am not comfortable endorsing either of the workarounds. For better or worse serde is the most used Rust library. People look here for guidance; the majority of idioms seen in proc macros are cribbed originally from serde. I would prefer instead to put focus on the language change to make exportable proc macros doable correctly, for which my outlook is less pessimistic than  yours (except if I accept your PR, in which case a language change will never happen).

2. This is a niche feature. There are a zillion feature requests less niche than this. The only way to maintain a foundational library is to draw and hold a line on how niche features to accept, because doing otherwise destroys a foundational library: you can't accept whatever is individually justifiable as making the library better. My position isn't "anti-seamlessness" but my priority also isn't doing what's best for rocket or any other use case.

PR #2026, opened 2021-05-12
Allow only `unused_extern_crates` instead of the whole `rust_2018_idioms` lint group in `serde_derive`-generated code

hyd-dev:
I found that this kind of `extern` only triggers `unused_extern_crates` warning: https://godbolt.org/z/bMxP8vGE9
Therefore, I think it's unnecessary to `#[allow]` the *whole* `rust_2018_idioms` lint group.

(It's causing some problems for me since I'm also using `#[forbid(future_incompatible)]`, and rust-lang/rust#83213 added some of the lints in `rust_2018_idioms` to `future_incompatible`, causing my program [fails to compile](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=60284f31914d92f2fc512885f0573b66).)

### DIFF
--- a/serde_derive/src/dummy.rs
+++ b/serde_derive/src/dummy.rs
@@ -23,7 +23,7 @@ pub fn wrap_in_const(
             use #path as _serde;
         },
         None => quote! {
-            #[allow(rust_2018_idioms, clippy::useless_attribute)]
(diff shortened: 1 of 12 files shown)


dtolnay (approved):
Thanks!

PR #2049, opened 2021-06-29
enable 128bit on emscripten target

### LINKED ISSUE #1995: Enable `integer128` feature for emscripten targets
It seems like 128-bits integers support has been added to the emscripten compiler (https://github.com/emscripten-core/emscripten-fastcomp/issues/169), does it mean we can enable it back [here](https://github.com/serde-rs/serde/blob/master/serde/build.rs#L58)?

extrawurst:
closes #1995

### DIFF
--- a/serde/build.rs
+++ b/serde/build.rs
@@ -52,10 +52,7 @@ fn main() {
 
     // 128-bit integers stabilized in Rust 1.26:
     // https://blog.rust-lang.org/2018/05/10/Rust-1.26.html
-    //
-    // Disabled on Emscripten targets as Emscripten doesn't
-    // currently support integers larger than 64 bits.
-    if minor >= 26 && !emscripten {
+    if minor >= 26 {
         println!("cargo:rustc-cfg=integer128");
     }
 


dtolnay (changes requested):
Please enable this only on compiler versions that support it. Old compilers probably still don't support 128 bit on emscripten right?

extrawurst:
> Please enable this only on compiler versions that support it. Old compilers probably still don't support 128 bit on emscripten right?

I would love to do that. I assume you mean the rust version check in the `build.rs`? is there a way to find out what emscripten version the rust versions are targeting?

oli-obk:
> 
> 
> > Please enable this only on compiler versions that support it. Old compilers probably still don't support 128 bit on emscripten right?
> 
> I would love to do that. I assume you mean the rust version check in the `build.rs`? is there a way to find out what emscripten version the rust versions are targeting?

https://github.com/serde-rs/serde/pull/2076 checks for Rust 1.40, which is when this got supported. I don't think the emscripten version matters

PR #2051, opened 2021-06-30
Add docs for serde_derive

### LINKED ISSUE #1999: Add docs for serde_derive
Currently serde_derive links to online docs which if the user is offline the user is not able to see it, and may have to dig through the source code available offline to see it.

Maybe we could provide something like
- https://docs.rs/dtolnay/0.0.9/dtolnay/
- https://docs.rs/egg/0.6.0/egg/tutorials/index.html

But I think it would be sufficient to document all the attributes in https://serde.rs/attributes.html and have a few examples there on how to use them. But it have the risk of requiring it being as up-to-date as the book, I think we might not have so much issues on this because the attributes probably won't be changed frequently.

pickfire:
Most of it is based on serde.rs but more towards a API reference style.
Useful for offline preview by default without additional action, user
can just `cargo doc --open`, even though it links to website it will
still retain a certain amount of information.

Fix #1999


oli-obk:
CI got stuck

serde_derive/src/lib.rs
```diff
@@ -9,9 +9,455 @@
...
 //!
-//! Please refer to [https://serde.rs/derive.html] for how to set this up.
+//! Please refer to [https://serde.rs/derive.html] for how to set this up and
+//! troubleshoot.
 //!
 //! [https://serde.rs/derive.html]: https://serde.rs/derive.html
+//!
+//! # Examples
+//!
+//! An example below while showing the differences between attributes.
```
dtolnay, line 19: This is not a grammatical sentence.

Copying the intro from https://serde.rs/attributes.html would be sufficient I think.

serde_derive/src/lib.rs
```diff
@@ -9,9 +9,455 @@
...
+//!
+//! - [Container attributes](#container-attributes) — apply to a struct or enum declaration.
+//! - [Variant attributes](#variant-attributes) — apply to a variant of an enum.
+//! - [Field attributes](#field-attributes) — apply to one field in a struct or in an enum variant.
+//!
+//! ```
+//! # extern crate serde;
+//! use serde::{Serialize, Deserialize};
+//!
+//! #[derive(Serialize, Deserialize)]
```
dtolnay, line 29: I would prefer not to start with this as the first code example. This code serves a very specific purpose in https://serde.rs/attributes.html and is not the front page of the book. The top example needs to be something that is as representative as possible of real world usage.

dtolnay (changes requested):
Thanks @pickfire &mdash; mostly looks good. I am on board with shipping this documentation with the serde_derive crate. I agree with https://github.com/serde-rs/serde/issues/1999#issue-831364535 that we are not adding or changing attributes at a rate that would make it a burden to keep this up to date.

Some minor comments below.

serde_derive/src/lib.rs
```diff
@@ -83,6 +529,7 @@ mod pretend;
 mod ser;
 mod try;
 
+/// Derive [`Serialize`]. See [`crate`] for more.
```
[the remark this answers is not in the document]
pickfire: I think there is no way to do this, I think maybe we can remove the intra doc link?

PR #2082, opened 2021-08-30
Added macro for building data formats without support for 128-bit integers

cheshirskycode:
It is vice versa for serde_if_integer128. Macro can make more convenient work for defining structures for versions younger than older than 1.26

```
use serde::{serde_if_integer128, serde_if_no_integer128};

serde_if_no_integer128! {
    enum N {
        PosInt(u64),
        NegInt(i64),
        Float(f64),
    }
}

serde_if_integer128! {
    enum N {
        PosInt(u128),
        NegInt(i128),
        Float(f64),
    }
}

impl N {
    fn is_i64(&self) -> bool {. todo!() }

    serde_if_integer128! {
        fn is_i128(&self) -> bool {. todo!() }
        fn is_u128(&self) -> bool {. todo!() }
    }

    // Etc..
}
```

### DIFF
--- a/serde/src/integer128.rs
+++ b/serde/src/integer128.rs
@@ -80,3 +80,62 @@ macro_rules! serde_if_integer128 {
 macro_rules! serde_if_integer128 {
     ($($tt:tt)*) => {};
 }
+
+/// Conditional compilation depending on whether Serde is built without support for
+/// 128-bit integers.
+///
+/// This macro may be used in the case of building data formats that may support i128 / u128 integers.
+///
+/// ```edition2018
+/// use serde::{serde_if_integer128, serde_if_no_integer128};
+///
+/// serde_if_no_integer128! {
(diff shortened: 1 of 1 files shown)


dtolnay (commented):
I would prefer not to add a new macro for this in serde. New code generally shouldn't be trying to accommodate a pre-1.26 compiler. I believe you opened this specifically for https://github.com/dtolnay/serde-yaml/pull/211, but serde-yaml already only supports 1.38+.

For code that does want this, there is almost always a way to do without a new macro. For example:

```rust
// new compilers = u128, old compilers = u64
pub type T = t::T;

mod t {
    pub use self::old_t::*;
    serde_if_integer128! {
        pub type T = u128;
    }
    mod old_t {
        pub type T = u64;
    }
}
```

or in the worst case, the downstream crate can do their own version checking.

PR #2104, opened 2021-10-17
Support std::lazy::OnceCell (unstable: RFC 2788)

### LINKED ISSUE #1952: impl Serialize for OnceCell
Like `RefCell`, it is useful if `OnceCell` can be serialized with serde.

Techcable:
Let me know if anything needs fixing or if this should have a dedicated feature flag.

Fixes #1952

### DIFF
--- a/serde/src/de/impls.rs
+++ b/serde/src/de/impls.rs
@@ -50,6 +50,32 @@ impl<'de> Deserialize<'de> for ! {
     }
 }
 
+#[cfg(feature = "unstable")]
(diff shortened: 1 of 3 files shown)


dtolnay (commented):
Would you be able to share a detailed example of a use case where this impl would be beneficial?

Unlike RefCell, storing OnceCell in a data structure seems pretty unusual. I've only ever seen them in `static`.

dtolnay (commented):
I'll leave #1952 open for now in case someone wants to chip in with use cases, but I am closing this PR as I am not convinced of the value for real-world code.

Thanks anyway for the PR!

PR #2114, opened 2021-10-26
Point to type instead of field for missing traits

### LINKED ISSUE #2103: Doc comment as snippet for missing deserialize with stable/beta rustc
Initially reported as https://github.com/rust-lang/rust/issues/89877 and it's possible that this is still in the wrong repository but I couldn't determine which would the correct.

Given the following code: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=fc1473e4afb90d6ab12751d3cc0cb6a8

```rust
fn main() {
    todo!("does not compile")
}

#[derive(Debug, serde::Deserialize)]
pub struct Root {
    /// doc comment included as the snippet
    foo: Bar,
    // this has a regular comment, not shown as snippet
    car: Bar,
}

/// this one didn't have deserialize derived by accident
#[derive(Debug)]
pub struct Bar {
    bar: u32
}
```

The manually deduplicated errors are:

```
   Compiling playground v0.0.1 (/playground)
error[E0277]: the trait bound `Bar: Deserialize<'_>` is not satisfied
    --> src/main.rs:7:5
     |
7    |     /// doc comment included as the snippet
     |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `Deserialize<'_>` is not implemented for `Bar`
     |
note: required by `next_element`
    --> /playground/.cargo/registry/src/github.com-1ecc6299db9ec823/serde-1.0.130/src/de/mod.rs:1703:5
     |
1703 | /     fn next_element<T>(&mut self) -> Result<Option<T>, Self::Error>
1704 | |     where
1705 | |         T: Deserialize<'de>,
     | |____________________________^

error[E0277]: the trait bound `Bar: Deserialize<'_>` is not satisfied
    --> src/main.rs:10:5
     |
10   |     car: Bar,
     |     ^^^ the trait `Deserialize<'_>` is not implemented for `Bar`
     |
note: required by `next_element`
    --> /playground/.cargo/registry/src/github.com-1ecc6299db9ec823/serde-1.0.130/src/de/mod.rs:1703:5
     |
1703 | /     fn next_element<T>(&mut self) -> Result<Option<T>, Self::Error>
1704 | |     where
1705 | |         T: Deserialize<'de>,
     | |____________________________^
```

However on nightly-2021-10-13 (not beta) this produces the following ~error~ correct snippet for the first field:

```
7    | /     /// doc comment included as the snippet
8    | |     foo: Bar,
     | |____________^ the trait `Deserialize<'_>` is not implemented for `Bar`
```

jonasbb:
In issue #2103 it was noted that spans for missing `Deserialize`
implementations include the doc comment. This commit changes it to
explicitly use the type of each field, as this is the behavior nightly
rust has.
This is done for both `Deserialize` and `Serialize`.

At the same time this fixes two places where no span was used, thus the
error message would point at the derive attribute instead of the
problematic field. Both cases concern the ude of `Default::default()`.

Compiletests currently only run on nightly, which does not have this issue. Therefore I'm not sure if this can be tested or needs a test.

Closes #2103

### DIFF
--- a/serde_derive/src/de.rs
+++ b/serde_derive/src/de.rs
@@ -364,7 +364,7 @@ fn deserialize_transparent(cont: &Container, params: &Parameters) -> Fragment {
     let path = match transparent_field.attrs.deserialize_with() {
         Some(path) => quote!(#path),
         None => {
-            let span = transparent_field.original.span();
+            let span = transparent_field.original.ty.span();
             quote_spanned!(span=> _serde::Deserialize::deserialize)
         }
     };
@@ -375,7 +375,10 @@ fn deserialize_transparent(cont: &Container, params: &Parameters) -> Fragment {
             quote!(#member: __transparent)
         } else {
             let value = match field.attrs.default() {
-                attr::Default::Default => quote!(_serde::__private::Default::default()),
+                attr::Default::Default => {
+                    let span = transparent_field.original.ty.span();
+                    quote_spanned!(span=> _serde::__private::Default::default())
+                },
                 attr::Default::Path(path) => quote!(#path()),
                 attr::Default::None => quote!(_serde::__private::PhantomData),
             };
@@ -658,7 +661,7 @@ fn deserialize_seq(
             let visit = match field.attrs.deserialize_with() {
                 None => {
                     let field_ty = field.ty;
-                    let span = field.original.span();
+                    let span = field.original.ty.span();
                     let func =
                         quote_spanned!(span=> _serde::de::SeqAccess::next_element::<#field_ty>);
                     quote!(try!(#func(&mut __seq)))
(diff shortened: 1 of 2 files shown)


serde_derive/src/de.rs
```diff
@@ -375,7 +375,10 @@ fn deserialize_transparent(cont: &Container, params: &Parameters) -> Fragment {
             quote!(#member: __transparent)
         } else {
             let value = match field.attrs.default() {
-                attr::Default::Default => quote!(_serde::__private::Default::default()),
+                attr::Default::Default => {
+                    let span = transparent_field.original.ty.span();
```
dtolnay, line 379: I think `transparent_field`'s span would be a misleading span to use here. This code runs if the field whose value is being initialized to `_serde::__private::Default::default()` is **not** the transparent field.

dtolnay (changes requested):
Thanks for the PR!

#2103 has since been closed because rustc changes and/or serde changes have improved that error. I am not sure which of this PR is still needed.

PR #2144, opened 2021-12-30
increase array_impls for ser,de

tklebanoff:


dtolnay (commented):
Last time I checked, these were absurdly expensive to compile, so I don't think they are going to be worth adding.

dtolnay:
It looks like on my machine this increases compile time of `cargo build` from 3.9 seconds to 6.0 seconds, which is way more than would be worth it for these impls.

tklebanoff:
right you are. that is too expensive.  what do you think about simply adding size 64 and 128? these seem to me common enough.  i hit a size 64 case today

PR #2145, opened 2021-12-30
increase array_impls for ser,de 64 case only

tklebanoff:


dtolnay (commented):
Thanks for the PR!

I am not interested in making this change, but you can check out https://github.com/serde-rs/serde/issues/631 for some approaches for working with arbitrary sized arrays.

Issue #2146, opened 2022-01-01
Improve performance using nightly features


kdy1:
# Context

I'm working to improve the performance of the deserializer. (See https://github.com/swc-project/swc/pull/3160)

I found that `serde::__private::de::Content` is created quite frequently from `__deserializer` and `__deserializer` is `serde::__private::de::ContentDeserailizer` in many cases.

# Try: specialization

So I copied code from serde, and specialized implementation of `deserialize` for `Content`.

```rust
impl<'de> Deserialize<'de> for Content<'de> {
    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
    where
        D: Deserializer<'de>,
    {
        <Content as DeserImpl<'de, D>>::deserialize_spec(deserializer)
    }
}

trait DeserImpl<'de, D>: Sized
where
    D: Deserializer<'de>,
{
    fn deserialize_spec(deserializer: D) -> Result<Self, D::Error>;
}

impl<'de, D> DeserImpl<'de, D> for Content<'de>
where
    D: Deserializer<'de>,
{
    default fn deserialize_spec(deserializer: D) -> Result<Self, D::Error> {
        // Untagged and internally tagged enums are only supported in
        // self-describing formats.
        let visitor = ContentVisitor { value: PhantomData };
        deserializer.deserialize_any(visitor)
    }
}

impl<'de, E> DeserImpl<'de, ContentDeserializer<'de, E>> for Content<'de>
where
    ContentDeserializer<'de, E>: Deserializer<'de, Error = E>,
{
    fn deserialize_spec(deserializer: ContentDeserializer<'de, E>) -> Result<Self, E> {
        Ok(deserializer.content)
    }
}
```

## Previous

```
test json_deserialize                     ... bench:     1,226,131 ns/iter (+/- 58,655)
```

## After

```
test json_deserialize                     ... bench:     621,480 ns/iter (+/- 69,766)
```

I spotted few more creation of `Content` from `serde::__private::de::ContentDeserailizer` due to `#[flatten]` and going to implement custom `#[derive(Deserialize)]` which uses my fork of `ContentDeserailizer`.

But this is useful for many users of`serde`, so I'm filing an issue.

## Using fork

To quickly apply specialization for `ContentDeserailizer` of `#[flatten]`, I created a fork (https://github.com/kdy1/serde/tree/content-specialization). Using it, I could get the result before creating a custom dervie macro.

```
[...]

PR #2147, opened 2022-01-01
Use specialization to improve performance

kdy1:
I'll undraft this after changing it to use `min_specialization`.
See https://github.com/serde-rs/serde/issues/2146 for details.

Also, as this is nightly-only, it should be feature-gated I guess?


kdy1:
Thank you!

PR #2148, opened 2022-01-01
Optimize deserialization of recursive buffered types

### LINKED ISSUE #2146: Improve performance using nightly features

dtolnay:
Closes #2146.

### DIFF
--- a/serde/src/de/mod.rs
+++ b/serde/src/de/mod.rs
@@ -1213,6 +1213,20 @@ pub trait Deserializer<'de>: Sized {
     fn is_human_readable(&self) -> bool {
         true
     }
+
+    // Not public API.
+    #[cfg(all(serde_derive, any(feature = "std", feature = "alloc")))]
+    #[doc(hidden)]
+    fn __deserialize_content<V>(
+        self,
+        _: ::actually_private::T,
+        visitor: V,
+    ) -> Result<::private::de::Content<'de>, Self::Error>
+    where
+        V: Visitor<'de, Value = ::private::de::Content<'de>>,
+    {
+        self.deserialize_any(visitor)
+    }
 }
 
 ////////////////////////////////////////////////////////////////////////////////
--- a/serde/src/lib.rs
+++ b/serde/src/lib.rs
@@ -295,3 +295,8 @@ extern crate serde_derive;
 #[cfg(feature = "serde_derive")]
 #[doc(hidden)]
 pub use serde_derive::*;
+
+#[cfg(all(serde_derive, any(feature = "std", feature = "alloc")))]
+mod actually_private {
+    pub struct T;
+}
--- a/serde/src/private/de.rs
+++ b/serde/src/private/de.rs
@@ -206,6 +206,7 @@ mod content {
     use lib::*;
 
     use __private::size_hint;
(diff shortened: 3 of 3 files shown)


dtolnay:
Benchmark:

```rust
#![feature(test)]

extern crate test;

use std::collections::BTreeMap as Map;

#[serde(tag = "type")]
enum Enum {
    Rec(Map<String, Enum>),
    End,
}

#[bench]
fn bench(b: &mut test::Bencher) {
    let j = serde_json::to_string(&obj()).unwrap();
    b.iter(|| serde_json::from_str::<Enum>(&j).unwrap());
}

fn obj() -> Enum {
    let mut map = Map::new();
    for i in 0..100 {
        let mut outer = Map::new();
        outer.insert(i.to_string(), Enum::Rec(map));
        map = outer;
    }
    Enum::Rec(map)
}
```

- Before: 834 μs/iter
- After: 49 μs/iter

PR #2160, opened 2022-01-21
de: Make flattened structs and aliases interact correctly.

### LINKED ISSUE #1504: Field aliases do not work in combination with `flatten`

jhoobergs:
Rebased version of https://github.com/serde-rs/serde/pull/1519

@dtolnay What would be needed to get this merged?

Fixes #1504

### DIFF
--- a/serde/src/de/mod.rs
+++ b/serde/src/de/mod.rs
@@ -1125,6 +1125,28 @@ pub trait Deserializer<'de>: Sized {
     where
         V: Visitor<'de>;
 
+    /// Hint that the `Deserialize` type is expecting a struct with a particular
+    /// name and fields.
+    ///
+    /// `_fields_with_aliases` includes all valid field names, including
+    /// aliases.  `fields` only includes the canonical struct fields.
+    ///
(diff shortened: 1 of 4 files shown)


jhoobergs:
@dtolnay Any thoughts?

PhotonQuantum:
So what's the progress of this PR? The related issue has been locked and there's no review activity on this one.

dtolnay (commented):
https://github.com/serde-rs/serde/pull/1519#pullrequestreview-233751469 is still the approach that I would prefer. I guess this got stuck on @emilio (or anyone else) looking into what the practical downsides of that approach might be. As far as I can tell it doesn't cause any issue with bincode.

dralley:
https://github.com/serde-rs/serde/pull/2387 claims to fix #1504 also

dtolnay (commented):
Fixed a different way by #2387.

PR #2161, opened 2022-01-22
Name equality checking

avitex:
I put this forward as an alternative to #1902 allowing for a generic method of checking a field name on deserialization over a specific implementation tied to ASCII.

The provided `name_eq` container attribute can be be used for case-insensitivity.

```rust
// Note in this example we use ASCII case checking,
// but we could just as well check unicode case.
//
// We require `value` to be a byte slice as the deserializer
// may or may not provide us validated UTF-8.
fn eq_ignore_case(name: &str, value: &[u8]) -> bool {
    name.as_bytes().eq_ignore_ascii_case(other)
}

#[derive(Deserialize)]
#[serde(name_eq = "eq_ignore_case")]
pub struct Foo {
    // ..
}
```

`NameEq`/`name_eq` might not be the best name for this, but I wanted to put forward this initial implementation before bikeshedding.


avitex:
Now that I think about it, perhaps `alias_with` would be a better name inline with `deserialize_with` and `alias` which is a loose idea of what this functionality actually provides.

avitex:
@dtolnay Any thoughts on the initial implementation or comments?

0cv:
Case insensitive keys is exactly what I've been looking for. It would be really great to get this merged if all is fine.

Issue #2045, opened 2021-06-20
Adjacently tagged enum unit variant does not allow sequence representation of size 1


Kobzol:
Hi! I noticed that there is a discrepancy in deserializing adjacently tagged unit variants. If there is an adjacently tagged unit variant, like this:
```rust
#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq)]
#[serde(tag = "type", content = "payload")]
pub enum Example {
    Variant1
}
```
`serde` will serialize this as a struct containing a single key `type` with the value `Variant1`. The concrete implementation might encode this struct either as a map or as a sequence.

During deserialization, either a map or a seq is expected. If the input is a map, the code assumes that it might only contain the `tag` field, and not the `content` field (https://github.com/serde-rs/serde/blob/master/serde_derive/src/de.rs#L1567). However, this assumption is provided only for the map representation. If the adjacently tagged enum unit variant is encoded with a seq, the code assumes that the sequence must have two elements.

For example, if you serialize the unit variant as `{"type": "Variant1"}`, during deserialization it will be okay. But if you serialize it as `["Variant1"]`, the deserialization will fail (`invalid length 1, expected adjacently tagged enum Example`). This causes problems in msgpack-rust (https://github.com/3Hren/msgpack-rust/issues/250).

Is this behaviour intended? Or would you be willing to change it so that sequences of length 1 are okay? I can send a PR if you can guide me a little bit.

Kobzol:
I tried to create a prototype implementation [here](https://github.com/Kobzol/serde/commit/5f517025ffff8a417efff5e987630a20efc8eabc). I'd be glad for any fedback.

gcoakes:
You don't want to just check for unit variants. You want to check for new type variants as well. They might have an optional type.

zeenix:
> For example, if you serialize the unit variant as `{"type": "Variant1"}`, during deserialization it will be okay. But if you serialize it as `["Variant1"]`, the deserialization will fail (`invalid length 1, expected adjacently tagged enum Example`).

It's actually worse: the Deserialize impl won't be able to handle `{"type": "Variant1", "payload": {}}` either. An easy workaround is to make `Variant1` a struct (`Variant1 {}`) but then `{"type": "Variant1"}` will fail to deserialize. 😞

Issue #2174, opened 2022-02-17
Additional enum derive variant `[tag, content]`.


gcoakes:
I'm implementing a parser for socket.io's protocol. It models each event as a list with a guaranteed length of 2 where the first element is the tag and the second is the event data. Would it be possible for an additional derive variant for enums which encodes or decodes like this?

gcoakes:
I've been attempting to understand the derive code to see if I could actually do a PR for this, but I realized that adjacent deserialization is actually already compatible with this. Looking at the documentation, I don't think this is particularly intentional, but I'll take it. For serialization though, it looks like it will always use the map serialization as opposed to sequence serialization. I think that's going to be the real request here. i.e.:

``` rust
#[derive(Deserialize, Serialize)]
#[serde(sequentially_adjacent_tag)]
enum Foo {
    // ...
}
```

It would serialize and deserialize in to a sequence of length 2 with the first being the tag and the second being the content. I would think this can use the same deserialize derive as adjacently tagged enums, but the serialize would use serialize_seq.

dtolnay:
I would prefer not to build this into serde_derive. But it would be reasonable for some other derive library to support this less common representation.

PR #2175, opened 2022-02-17
Add sequentially_adjacent_tag representation for enums.

gcoakes:
If present, this attribute will cause enums to be serialized and
deserialized into a sequence. The first element is the tag, and the
second the content. If the content is optional, then the sequence may
only be a single item.

Also, add a test to verify the functionality of this new feature.


gcoakes:
This is the implementation of #2174. I think elements of it could be taken to fix #2045.

dtolnay (commented):
Thanks for the PR!

As I mentioned in https://github.com/serde-rs/serde/issues/2174#issuecomment-1043605261, I am not interested in this feature for serde_derive, but if this representation is a common thing people want, then building a separate macro crate to support it would be useful.

PR #2199, opened 2022-04-07
Optimize few string allocations

klensy:
Few string allocating optimizations and replacing of deprecated `trim_left_matches`.

### DIFF
--- a/serde_derive/src/de.rs
+++ b/serde_derive/src/de.rs
@@ -1902,7 +1902,7 @@ fn deserialize_untagged_newtype_variant(
 }
 
 fn deserialize_generated_identifier(
-    fields: &[(String, Ident, Vec<String>)],
+    fields: &[(&str, Ident, Vec<String>)],
     cattrs: &attr::Container,
     is_variant: bool,
     other_idx: Option<usize>,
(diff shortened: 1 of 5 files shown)


klensy:
This ready to review, i guess ;-)

klensy:
One month ping, @dtolnay.

dtolnay (commented):
I probably am not going to review this. My projects get more PRs per month than can be reviewed, so I focus on the most impactful ones, and in this case from a skim I would expect zero effect on macro expansion performance. Allocators are fast and nothing here is a bottleneck part of the macro. That puts it right to the bottom for impactfulness unless there is some benchmark data to back it up.

You're free to leave it open if you want but please do not ping me.

striezel:
> I probably am not going to review this. My projects get more PRs per month than can be reviewed, so I focus on the most impactful ones, [...]

That is completely understandable. Focus is important when time is a very limited resource.

Maybe one of your co-maintainers can afford the time to review this? I'd rather see pull requests reviewed late than not reviewed at all, because otherwise one may miss potential improvements, even if they are small.

@klensy:
>  Allocators are fast and nothing here is a bottleneck part of the macro. That puts it right to the bottom for impactfulness unless there is some benchmark data to back it up.

Did you run any benchmarks before and after the changes of this PR? If so, you could provide the data and then maybe it's easier for the maintainer(s) to make a decision on the possible impact of this PR.

dtolnay (commented):
Skimming this PR again, everything touched here is in cold code that runs at most once or a low number of times per serde macro call. I don't see any way that compile performance could be sensitive to a few allocations in this part of the code. I think the original code is all right as is.

Thanks anyway for the PR!

PR #2208, opened 2022-04-27
Add skip_serializing_if_default field attribute

Gordon01:
This attribute skips field serialization if its value equals default.

### Motivation:
Imagine we have an application, where the config may be changed dynamically and needs to be saved. Or just be saved for some reason.
```
struct Config {
    #[serde(default = "default_version")]
    version: u32,
    #[serde(default = "default_author")]
    author: String,
}
```
Probably, if we have a lot of fields that just defaulted and saving them adds unnecessary noise to the config file.

We may use `skip_serializing_if` attribute and provide functions that check if the value is default:
 ```
struct Config {
    #[serde(default = "default_version", skip_serializing_if = "is_version_default")]
    version: u32,
    #[serde(default = "default_author", skip_serializing_if = "is_author_default")]
    author: String,
}

fn is_version_default(value: &u32) -> bool {
    *value == default_version()
}

fn is_author_default(value: &String) -> bool {
    *value == default_author()
}
```

This adds a lot of one-line functions and adds noise to the code.

### Solution:
Use `skip_serializing_if_default` attribute from this PR:
```
struct Config {
    #[serde(default = "default_version", skip_serializing_if_default)]
    version: u32,
    #[serde(default = "default_author", skip_serializing_if_default)]
    author: String,
}
```

It calls the field's default function, compares its current value, and if it matches with the value provided by the default function, skips serialization.

### DIFF
--- a/serde_derive/src/internals/attr.rs
+++ b/serde_derive/src/internals/attr.rs
@@ -1109,6 +1109,7 @@ pub struct Field {
     skip_serializing: bool,
     skip_deserializing: bool,
     skip_serializing_if: Option<syn::ExprPath>,
+    skip_serializing_if_default: bool,
     default: Default,
     serialize_with: Option<syn::ExprPath>,
     deserialize_with: Option<syn::ExprPath>,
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by Gordon01** [comment]
I've added a test for this feature and fixed the wrong struct length that was revealed by the test.
I don't like the code duplication, but it seems to me it's the style you prefer.

Any advice on how it should be?

Otherwise, the feature seems to be complete.

**Comment by zkldi** [comment]
This feature would be really useful. Is there any reason why this is not merged? If there's any work that needs to be done I would not mind picking this feature up and pushing it through.

**Comment by Gordon01** [comment]
I believe, we should just tag @dtolnay and see his verdict

**Comment by lorenzleutgeb** [comment]
We use `#[serde(skip_serializing_if = "is_default"]` with
```rust
fn is_default<T: Default + PartialEq>(t: &T) -> bool {
    t == &T::default()
}
```

It'd be great if this could be reconsidered.

BergLucas:
This would be particularly useful when `serde` is used with `schemars`. With the `skip_serializing_if` approach, `schemars` cannot automatically determine whether a field should not be included in the JSON schema because it is a private field or because we just don't want it serialised when it has its default value. With `skip_serializing_if_default`, it would then be possible to determine this automatically.

1Dragoon:
This feature seems kind of obvious. What are the remaining concerns for it?

PR #2209, opened 2022-04-28
Allow integer renames for struct fields

### LINKED ISSUE #1773: Allow integers (or custom types) to be used as names/keys
This is similar to #745 ("Allow integer tags for internally tagged enums") but this issue focuses on types and struct fields.

When working with structs and fields, the `Serializer` and `Deserializer` traits take `name: &'static str` parameters, and other traits like `SerializeStruct` take `key: &'static str` parameters. These parameters describe the type/field.

But not all formats use textual keys. Some, for example, uses integers for their keys. I'd like to be able to use `#[serde(rename = 42)]` (or `#[serde(rename = SOME_CONST)]` if that's not too much), which would allow me to provide serde with the integer key for that particular type or field. But doing this is problematic for two reasons:

1. It requires `Serializer`/`Deserializer`/etc. to use a different type in their methods for the `name` or `key` parameters.
2. It doesn't play well when trying to (de)serialize the same struct to/from a human readable format (e.g., JSON).

I think this could be worked around by using the `is_human_readable()` function and allowing users to provide both a human-readable and machine-readable name to Serde. For example:

```Rust
#[serde(rename = "ThisIsANameForHumans")]
#[serde(machine_name = 0x10)]
struct Foo {
    #[serde(rename = "Bar")]
    #[serde(machine_name = 0x21)]
    bar: u32,
    #[serde(rename = "Baz")]
    #[serde(machine_name = 0x8014)]
    baz: f32,
}
```

Then the `#[derive]` macros could be changed like so:

```
impl Serialize for Foo {
    where
        S: Serializer,
    {
        // The check for is_human_readable() is only added if machine_name is used.
        if serializer.is_human_readable() {
            let mut foo = serializer.serialize_struct("ThisIsANameForHumans", 2)?;
            foo.serialize_field("Bar", &self.bar)?;
            foo.serialize_field("Baz", &self.baz)?;
            foo.end()
        } else {
            let mut foo = serializer.serialize_machine_struct(0x10, 2)?;
            foo.serialize_field(0x21, &self.bar)?;
            foo.serialize_field(0x8014, &self.baz)?;
            foo.end()
        }
    }
}
```

Open questions:
- If only some of the fields have `machine_name` specified, how should the be handled? Fallback to using human names? Or fallback to using the field ordinal number? Or fail to compile?
- Not all formats use integers as keys. Should the machine_name instead take a string of bytes (i.e. a byte literal that can have non-UTF-8 contents)?

I'd be willing to take a shot at implementing this if the serde maintainers are interested.

AndrewScull:
Closes #1773 and related to PR #2056.

Extends the serde_rename and alias annotations on struct fields to support integer names. Any struct with an integer name is serialized as a map, similarly to structs with any flattened fields, avoiding the need for changes in existing serializer implementations.

Integer names are typed and serialized based on that type, but deserialization is only based on the integer value because deserializers often don't know the exact type of integers.

There are a few unrelated refactors at the start of the series as I was learning the project, then the bulk of the changes are to represent non-string renames (similar to #2056) and to handle the cases of deserialization.

### DIFF
--- a/serde/src/de/mod.rs
+++ b/serde/src/de/mod.rs
@@ -262,7 +262,9 @@ macro_rules! declare_error_trait {
             /// Raised when a `Deserialize` struct type received a field with an
             /// unrecognized name.
             #[cold]
-            fn unknown_field(field: &str, expected: &'static [&'static str]) -> Self {
+            fn unknown_field<T>(field: T, expected: &'static [&'static str]) -> Self
+            where
+                T: Display {
                 if expected.is_empty() {
                     Error::custom(format_args!(
                         "unknown field `{}`, there are no fields",
--- a/test_suite/tests/test_annotations.rs
+++ b/test_suite/tests/test_annotations.rs
@@ -644,6 +644,352 @@ fn test_unknown_field_rename_struct() {
 }
 
 #[derive(Debug, PartialEq, Serialize, Deserialize)]
+#[serde(rename = "IgnoredRename")]
+struct IntRenameStruct {
+    a1: i32,
+    #[serde(rename = 12i8)]
+    a2: i32,
+    #[serde(rename = -598i16)]
+    a3: i32,
+    #[serde(rename = 18435)]
+    a4: i32,
+    #[serde(rename = 592734i32)]
+    a5: i32,
+    #[serde(rename = 123123123123123123i64)]
+    a6: i32,
+    #[serde(rename = 73u8)]
+    a7: i32,
+    #[serde(rename = 598u16)]
+    a8: i32,
+    #[serde(rename = 26547u32)]
+    a9: i32,
+    #[serde(rename = 12312312312312312312u64)]
+    a10: i32,
+}
+
+#[derive(Debug, PartialEq, Serialize, Deserialize)]
+#[serde(rename(serialize = "IgnoredSerRename", deserialize = "IgnoredDeRename"))]
+struct IntRenameStructSerializeDeserialize {
+    a1: i32,
+    #[serde(rename(serialize = -301, deserialize = 660u16))]
+    a2: i32,
+    #[serde(rename(serialize = 618, deserialize = "a12"))]
+    a3: i32,
+    #[serde(rename(serialize = "a13", deserialize = 320u16))]
+    a4: i32,
+}
+
+#[derive(Debug, PartialEq, Deserialize)]
+#[serde(deny_unknown_fields)]
+struct IntAliasStruct {
+    a1: i32,
+    #[serde(alias = -50i64)]
+    a2: i32,
+    #[serde(alias = 77, rename = "a6")]
+    a4: i32,
+    #[serde(alias = "a7", rename = -8)]
+    a5: i32,
+}
+
+#[derive(Debug, PartialEq, Deserialize)]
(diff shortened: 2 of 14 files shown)


test_suite/tests/ui/expected-string/integer.rs
```diff
@@ -2,8 +2,12 @@ use serde_derive::Serialize;
 
 #[derive(Serialize)]
 struct S {
-    #[serde(rename = 100)]
-    integer: (),
+    #[serde(rename = 100i128)]
```
maurer, line 5: I think it might be nice to squash the test patch into this one, since it's a bit odd to have negative tests but no positive tests.

test_suite/tests/test_annotations.rs
```diff
@@ -643,6 +643,352 @@ fn test_unknown_field_rename_struct() {
     );
 }
 
+#[derive(Debug, PartialEq, Serialize, Deserialize)]
+#[serde(rename = "IgnoredRename")]
+struct IntRenameStruct {
```
maurer, line 648: I think it will work as is, but it might be nice to have a test for renames to hex/binary/octal just in case.

maurer (commented):
LGTM. I noted a couple minor things, but it seems fine as is to me.

Progdrasil:
This would be very useful, is there anyway that I could help to move this along if @AndrewScull is too busy?

frykher:
This is very useful for my specific use case, where I have an enum to be serialized with adjacent tags. The tags however, are numbers.
The data I want to serialize looks like this:
```json
{
  "op": 8,
  "d": {
    "v": 4,
    "heartbeat_interval" :13750.0
  }
}
```

This is the code that doesn't work unfortunately:
```rs
#[serde(tag = "op", content = "d")]
pub enum DiscordPayload {
    #[serde(rename = "8")]
    Hello(Hello),
}
```
It expects an identifier rather than an integer
```
Error deserializing payload: invalid type: integer `8`, expected variant identifier at line 1 column 7
```
This change does have a valid use case!

Progdrasil:
@frykher see #2056 for the MR pertaining to enums

labbott:
I'd love to see this completed. It would be very useful for CBOR specifications that use integers in maps.

Progdrasil:
As an example of working around this limitation for CBOR encoding, I made [this declarative macro](https://github.com/1Password/passkey-rs/blob/main/passkey-types/src/utils/serde_workaround.rs) to rename the fields to numbers for CTAP2 encoding. You can see it' use in [passkey-types::ctap2::make_credential](https://github.com/1Password/passkey-rs/blob/main/passkey-types/src/ctap2/make_credential.rs)

serde/src/de/mod.rs
```diff
@@ -262,7 +262,9 @@ macro_rules! declare_error_trait {
             /// Raised when a `Deserialize` struct type received a field with an
             /// unrecognized name.
             #[cold]
-            fn unknown_field(field: &str, expected: &'static [&'static str]) -> Self {
+            fn unknown_field<T>(field: T, expected: &'static [&'static str]) -> Self
```
oli-obk, line 265: I don't think this is backwards compatible

Issue #1993, opened 2021-02-28
serde triggers clippy::use_self lint on recent clippy


imp:
On recent nightly (2021-02-27) serde upsets clippy when `clippy::use_self` lint [is activated](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=8bcd4e9adcc43b71004a32bdd99a8a8c). Apparently this lint got some [attention recently](https://github.com/rust-lang/rust-clippy/issues/3909), so it might be related

dtolnay:
Could you file this against Clippy? I believe that lint should not be triggering inside of macro generated code.

MarijnS95:
It seems like this has resurfaced in Rust 1.61: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=f735206a9a21f1ec7d55730e2d6c6a17

PR #2221, opened 2022-05-19
Fix `clippy::use_self` warning in generated `derive(Serialize)` output

MarijnS95:
Rust 1.61 seems to point out `clippy::use_self` inside macros again (#1993), which may or may not be an upstream problem that gets a hotfix release; it's probably easier and more concise to remove the variable `#this` expansion for hygiene reasons too.

Perhaps more functions can/should be updated to just use `Self::`, for simplicity.

### DIFF
--- a/serde_derive/src/ser.rs
+++ b/serde_derive/src/ser.rs
@@ -128,6 +128,15 @@ impl Parameters {
     fn type_name(&self) -> String {
         self.this.segments.last().unwrap().ident.to_string()
     }
+
+    /// Path to the type when used in remote context, `Self` otherwise
+    fn self_type(&self) -> syn::Path {
+        if self.is_remote {
+            self.this.clone()
+        } else {
+            syn::Path::from(syn::PathSegment::from(format_ident!("Self")))
+        }
+    }
 }
 
 // All the generics in the input, plus a bound `T: Serialize` for each generic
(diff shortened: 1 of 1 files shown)


MarijnS95:
Submitted this to see what the CI would think, looks like this approach is already breaking the `serde(remote = "AnotherType")` feature which can't use `Self`.

dtolnay (commented):
`use_self` is categorized as a Nursery lint and shouldn't be used. I would prefer not to introduce workarounds in serde_derive to accommodate it.

MarijnS95:
@dtolnay That's pretty lame as this has been a pretty useful lint thus far, even if this is an upstream compiler issue (macro-expanded code shouldn't be linted).

If it wasn't for the `remote` feature it'd have been better to use `Self` directly in macro code anyway instead of having to pass a `this` around.

dtolnay:
It's a nursery lint because it has excessive false positives. Try again when it no longer has excessive false positives. It is not a good idea to put workarounds for that kind of lint into macro generated code because the workaround doesn't disappear after the false positive goes away, which is bad for maintainability.

MarijnS95:
I may have been misunderstood, so I'll try again. If you can use `Self` instead of coming up with some `let this = &...this;` followed by `#this` in `quote` expansion that's simpler and preferred in my eyes. But we can't because of the `remote` feature, so the second revision of this PR merely turned into a workaround instead of being - in my eyes - legitimate simplification.

aldanor:
Any chance to reanimate this? Or reimplement it if there's anything wrong with the diff itself?

(People do use nursery lints whether they're "recommended" or not, in a 50k line codebase the only 'excessive false positives' we've observed so far are those for serializing enum variants with serde, plus one or two other proc macro crates, but generally use_self is fairly useful and catches most of 'true positives'.)

Issue #2237, opened 2022-06-29
Provide public access to RenameAllRules in `serde_derive_internals`


uint:
Hi! In `serde_derive_internals` I can get access to serde attribute parsers, which is super neat.

The particular data I need though is the `rename_all` rules, and the exposed struct isn't very useful with its fields private and no way to access them.

https://github.com/serde-rs/serde/blob/b23a76841425c70926efb51d46353cccc45c1145/serde_derive/src/internals/attr.rs#L203-L206

dtolnay:
I've published #2743 in serde_derive_internals 0.29.1. https://docs.rs/serde_derive_internals/0.29.1/serde_derive_internals/attr/struct.RenameAllRules.html

Sorry that this didn't get handled more promptly.

PR #2238, opened 2022-06-29
Expose `RenameAllRules` fields for use in `serde_derive_internals`

uint:
One way to deal with #2237. Would this be okay?


uint:
I need this once again for [`jtd-derive`](https://github.com/uint/jtd-derive). Any chance we could do this?

dtolnay (commented):
Thanks for the PR!

I think `pub` on the fields would be fine here &mdash; #2743.

uint:
Nice, thanks!

PR #2240, opened 2022-06-30
Call `reserve()` in `DeserializeSeed` example

Kixunil:
This suggests calling `reserve()` in example code so that people who ~blindly copy it get faster code.


dtolnay (approved):
Thanks!

PR #2251, opened 2022-07-20
Invert build.rs cfgs in serde_derive

taiki-e:
Similar to https://github.com/serde-rs/serde/commit/55a7cedd737278a9d75a2efd038c6f38b8c38bd6, but for serde_derive.


dtolnay (approved):
Thanks!

PR #2259, opened 2022-08-12
Explicitly use `edition = "2015"`

Kijewski:
This way `cargo package` won't implictily insert a `resolver = "1"` configuration. This fixes the inadversial MSRV bump mentioned in issue [#2255].

[#2255]: https://github.com/serde-rs/serde/issues/2255

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,11 +69,11 @@ jobs:
         if: matrix.os != 'windows'
 
   msrv:
-    name: Rust 1.13.0
+    name: Rust 1.31.0
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: dtolnay/rust-toolchain@1.13.0
+      - uses: dtolnay/rust-toolchain@1.31.0
       - name: Get timestamp for cache
         id: date
         run: echo ::set-output name=yearmo::$(date +%Y%m)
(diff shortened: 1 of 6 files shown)


Kijewski:
Oh, well, this way rustc `1.27 ..= 1.30` break. But that would be a smaller MSRV bump than `1.51`. What do y'all think? Is this an acceptable solution?

HamzaYasin1:
Now it started giving this error:

              error[E0432]: unresolved import `wasm_builder_runner`
               --> runtime/build.rs:1:5
                |
              1 | use wasm_builder_runner::WasmBuilder;
                |     ^^^^^^^^^^^^^^^^^^^ maybe a missing crate `wasm_builder_runner`?
              
Kijewski:
Thank you for testing this, @HamzaYasin1! I don't know how to run cargo for older `wasm32-unknown-unknown` or `wasm32-wasi` targets. Can to paste needed `rustup toolchain add …` and/or `rustup target add …` commands for me?

HamzaYasin1:
When you run command , so it will show you all the installed version that you can use.
       rustup show
        
        Default host: x86_64-apple-darwin
        rustup home:  /Users/algocoder/Applications/.rustup
        
        installed toolchains
        --------------------
        
        stable-x86_64-apple-darwin
        nightly-2020-08-19-x86_64-apple-darwin
        nightly-2020-10-05-x86_64-apple-darwin
        nightly-2021-11-12-x86_64-apple-darwin
        nightly-x86_64-apple-darwin (default)
        1.48-x86_64-apple-darwin
        1.45.1-x86_64-apple-darwin
        
        installed targets for active toolchain
        --------------------------------------
        
        wasm32-unknown-unknown
        x86_64-apple-darwin
        
        active toolchain
        ----------------
        
        nightly-x86_64-apple-darwin (default)
        rustc 1.65.0-nightly (20ffea693 2022-08-11)

You can see this link for toolchain changes
[ switch between Rust toolchains](https://stackoverflow.com/questions/58226545/how-to-switch-between-rust-toolchains)

Yes Serde doesn't contain any runtime folder as of my knowledge.

Kijewski:
Well, `runtime/build.rs` is missing an `extern crate wasm_builder_runner;` line. But the error seems unrelated to me?

Mingun:
Probably that code was built in 2018 edition and now it is build in 2015 edition.

Kijewski:
> Probably that code was built in 2018 edition and now it is build in 2015 edition.

But *what* code? :) Afaik `edition` is not "viral". I.e. if a dependency has an older `edition`, it should not affect me. Or does it?

taiki-e:
Considering how https://github.com/rust-lang/cargo/issues/10954 was fixed, this should not be necessary and the problem should be automatically resolved when the maintainer creates a release using the nightly toolchain after nightly-2022-08-13 (tomorrow).

Kijewski:
@taiki-e, ah, I didn't see the <https://github.com/rust-lang/cargo/issues/10954>. Yes, that's the (only) correct solution.

PR #2266, opened 2022-08-23
fix: don't check skipped variant with internal tag

flisky:


dtolnay (approved):
Thanks!

test_suite/tests/test_annotations.rs
```diff
@@ -2265,6 +2265,37 @@ fn test_externally_tagged_enum_containing_flatten() {
...
+#[test]
+fn test_internally_tagged_enum_with_skipped_conflict() {
+    #[derive(Serialize, Deserialize, PartialEq, Debug)]
+    #[serde(tag = "t")]
+    enum Data {
+        A,
+	#[serde(skip)]
+	B {
+	    t: String
+	},
```
Mingun, line 2277: What mean that variant is skipped? How it behaves when you try to serialize that variant?

flisky: The variant cannot be serialized & deserialized. Serializing it will throw a runtime error `Err(Error("the enum variant Data::B cannot be serialized", line: 0, column: 0))`.

PR #2269, opened 2022-08-26
Add support of `dot.case` and `SCREAMING.DOT.CASE` rename on variant and field

qboileau:
Add support of `Dotcase` and `CapitalDotcase`  for compatibility with scala libs [`enumeratum`](https://github.com/lloydmeta/enumeratum#mixins-to-override-the-name) serialization (using [`circe`](https://circe.github.io/circe/) for example).

Idea is to add `dot.case` and `SCREAMING.DOT.CASE` to rename_all `RenameRule` for Serialize and Deserialize fields and variants.
```rust 
use serde::Serialize;

#[derive(Serialize)]
#[serde(rename_all = "dot.case")]
struct Person {
    first_name: String,
    last_name: String,
}

fn main() {
    let person = Person {
        first_name: "Graydon".to_string(),
        last_name: "Hoare".to_string(),
    };

    let json = serde_json::to_string_pretty(&person).unwrap();

    // Prints:
    //    {
    //      "first.name": "Graydon",
    //      "last.name": "Hoare"
    //    }
    println!("{}", json);
}
```
I'm not sure if that's all that is need to do or if it's missing additional changes.

### DIFF
--- a/serde_derive/src/internals/case.rs
+++ b/serde_derive/src/internals/case.rs
@@ -33,6 +33,10 @@ pub enum RenameRule {
     KebabCase,
     /// Rename direct children to "SCREAMING-KEBAB-CASE" style.
     ScreamingKebabCase,
+    /// Rename direct children to "dot.case" style.
+    DotCase,
+    /// Rename direct children to "SCREAMING.DOT.CASE" style.
+    ScreamingDotCase,
 }
 
 static RENAME_RULES: &[(&str, RenameRule)] = &[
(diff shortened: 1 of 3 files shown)


qboileau:
@dtolnay do you need something else for this PR to be merged ?

Property404:
This would be nice to have. I was going to push pretty much the same PR but I guess you beat me to it

dtolnay (commented):
Thanks for the PR!

I would prefer not to build this into serde_derive. But I would be open to someone making an attribute macro (in a different crate) that supports applying a broader set of rename rules to a serde data structure beyond the ones built into serde_derive, by expanding to multiple individual `serde(rename = "...")` attributes in the right places.

Mingun:
What about to allow specify path to the `const fn` that would perform rename?
```rust
const fn rename(name: &str) -> &str { ... }

#[serde(rename_all = path::to::rename)]
struct MyStruct { ... }
```
Then almost all needs would be covered.

PR #2284, opened 2022-09-23
Fix "deserialize" -> "deserialize_any" in Deserializer documentation

benediktwerner:
Same as it is on https://serde.rs/impl-deserialize.html


dtolnay (approved):
Thanks!

PR #2288, opened 2022-10-01
Change used serializer methods when serialize untagged enums

Mingun:
Let's consider the following enum:
```rust
#[derive(serde::Serialize)]
#[serde(untagged)]
enum Untagged {
    Unit,
    Newtype(bool),
    Tuple(),
    Struct {},
}
```

Currently (1.0.145) the following functions are used to serialize enum:
|Variant|Serialized with|Suggestion
|----------------|----------|---------------
|`Enum::Unit`|`serialize_unit()`|`serialize_unit_struct("Unit")`
|`Enum::Newtype`|delegates serialization to the inner type| |
|`Enum::Tuple`|`serialize_tuple(N)`|`serialize_tuple_struct("Tuple", N)`
|`Enum::Struct`|`serialize_struct("Enum", N)`|`serialize_struct("Struct", N)`

I suggest to use methods for named variants of those methods, as shown in column **Suggestion**. That would be useful for the formats where name of type can be stored. Because in most cases name of type is not used anyway, that change shouldn't break a many.

That change seems logical in that sense, if we would consider untagged enum as an opportunity to combine various types in the Rust code that have nothing in common in the format. Instead of trying to (de)serialize a struct, a tuple or a unit struct, we (de)serialize a one enum.

One use case -- serializing enums in [quick-xml](https://github.com/tafia/quick-xml). Usually serializer don't use the type names except one case -- when we serializes top-level type. For convenience, the root element of the document can be automatically named by the type name. Because untagged enums serialized like their content, a user can expect, that serializing untagged enum would produce a name for the root element from variant name, but in reality that not the case.

### DIFF
--- a/serde_derive/src/ser.rs
+++ b/serde_derive/src/ser.rs
@@ -681,7 +681,7 @@ fn serialize_adjacently_tagged_variant(
                 };
             }
             Style::Tuple => serialize_tuple_variant(
-                TupleVariant::Untagged,
+                TupleVariant::AdjacentlyTagged,
                 params,
                 &variant.fields,
                 &variant_name,
@@ -810,6 +810,7 @@ enum TupleVariant {
         type_name: String,
         variant_index: u32,
     },
+    AdjacentlyTagged,
     Untagged,
 }
 
@@ -821,7 +822,8 @@ fn serialize_tuple_variant(
 ) -> Fragment {
     let tuple_trait = match context {
         TupleVariant::ExternallyTagged { .. } => TupleTrait::SerializeTupleVariant,
-        TupleVariant::Untagged => TupleTrait::SerializeTuple,
+        TupleVariant::AdjacentlyTagged => TupleTrait::SerializeTuple,
+        TupleVariant::Untagged => TupleTrait::SerializeTupleStruct,
     };
 
     let serialize_stmts = serialize_tuple_struct_visitor(fields, params, true, &tuple_trait);
@@ -860,13 +862,23 @@ fn serialize_tuple_variant(
                 _serde::ser::SerializeTupleVariant::end(__serde_state)
             }
         }
-        TupleVariant::Untagged => {
+        TupleVariant::AdjacentlyTagged => {
             quote_block! {
                 let #let_mut __serde_state = try!(_serde::Serializer::serialize_tuple(
                     __serializer,
                     #len));
                 #(#serialize_stmts)*
                 _serde::ser::SerializeTuple::end(__serde_state)
+            }
+        }
+        TupleVariant::Untagged => {
+            quote_block! {
+                let #let_mut __serde_state = try!(_serde::Serializer::serialize_tuple_struct(
+                    __serializer,
+                    #variant_name,
+                    #len));
+                #(#serialize_stmts)*
+                _serde::ser::SerializeTupleStruct::end(__serde_state)
             }
         }
     }
--- a/serde_derive/src/ser.rs
+++ b/serde_derive/src/ser.rs
@@ -770,7 +770,7 @@ fn serialize_untagged_variant(
     match effective_style(variant) {
         Style::Unit => {
             quote_expr! {
-                _serde::Serializer::serialize_unit(__serializer)
+                _serde::Serializer::serialize_unit_struct(__serializer, #variant_name)
             }
         }
         Style::Newtype => {
--- a/serde_derive/src/ser.rs
+++ b/serde_derive/src/ser.rs
@@ -956,7 +956,7 @@ fn serialize_struct_variant(
             quote_block! {
                 let #let_mut __serde_state = try!(_serde::Serializer::serialize_struct(
                     __serializer,
-                    #type_name,
+                    #variant_name,
                     #len,
                 ));
                 #(#serialize_fields)*
--- a/serde_derive/src/ser.rs
+++ b/serde_derive/src/ser.rs
@@ -552,19 +552,17 @@ fn serialize_externally_tagged_variant(
             TupleVariant::ExternallyTagged {
                 type_name,
                 variant_index,
-                variant_name,
             },
             params,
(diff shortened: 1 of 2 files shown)


oli-obk:
Without the ability to run serde PRs on all of crates.io as a test, I don't think we can pursue such changes in a way I'm comfortable with. The alternative of just shipping it and yanking if it's a problem is not a great user experience.

So... I don't know how to continue here. Do you know of any (even if annoying) workarounds? Or does it just boil down to "implement Serialize manually"?

Mingun:
I see only manual implementation or use a fork of `serde_derive`.

The my preferred way would to start using semantic versioning for serde releases. Then this problem just wont arise. Alternatively, we can find all serde deserializers (or at least those listed on https://serde.rs) and check how they handle that cases.

oli-obk:
While we definitely cannot release new major versions of serde without massive disruption to the crate ecosystem, `serde_derive` does not really have the same issues. We could collect this issue under a `serde-derive 2.0` label and revisit it every year to see if we have enough strongly motivated changes under that label to actually do that. Note that this effectively means that it won't happen, judging from the similar label in rust-lang/rust.

Mingun:
> While we definitely cannot release new major versions of serde without massive disruption to the crate ecosystem

I'm not sure, why you think so. If some library crates cannot update their (serde) dependency it means that they does not evolving and not maintained and you anyway should not use them. If this is application crate that is not evolving, it still can use previous serde releases fine and will ever never see any new serde releases (where do they come from if the project is not developing?)

oli-obk:
The problem isn't people being able to update, it is that we'll have two incompatible versions of serde on crates.io and then suddenly lots of crates don't work together. This will cause an enormous amount of churn. While doable, it's even less doable than a serde-derive major version bump. And so far we have had nothing that even remotely warrants that churn. If you want a similar example of such churn, look at the `libc` crate's 0.2 bump, that was bad, and it doesn't even expose that many types that are used widely.

Mingun:
> suddenly lots of crates don't work together

This is unavoidable artifact of transition period and it will not too long. Crates will upgraded and all will be as before. And again -- if some crates would not be upgraded, then they are unmaintained; why we should prefer compatibility with unmaintained crates over needs of developing crates?

> If you want a similar example of such churn, look at the `libc` crate's 0.2 bump, that was bad, and it doesn't even expose that many types that are used widely.

And how long did this problem continue? I didn't follow it. I'll speculate -- a year at max.

> it is that we'll have two incompatible versions of serde on crates.io and then suddenly lots of crates don't work together.

You think, that having two crates instead of two version of one crate would be better? This will happen as soon as I find a way to resolve #1183 and release my version of serde (of course, if my PR for that wouldn't be merged).

> And so far we have had nothing that even remotely warrants that churn

I think fixing #1183 is a sufficient reason already; I'm pretty sure that in over 200 issues there at least 10 that also can give that reason

oli-obk:
I don't think the cost we introduce to thousands of crates is worth it for the issues we have. Considering the fix in this PR (and for other issues) lies solely within `serde_derive`, I also consider that speculation and this PR is the wrong place for that.

I'd much prefer a solution that doesn't require changing `serde_derive`. Since this is for certain (most? many? some?) untagged enums specifically, a simpler first step would be to publish a crate that modifies the type in a way that makes the derive handle your case. I don't know if that is possible with some of the attributes we have, but if it is, that seems like a solution that is achievable faster and with very contained churn.

Mingun:
> I don't think the cost we introduce to thousands of crates is worth it for the issues we have.

Are you sure, that there would be any significant costs? First, I think that different crates that implements `Deserializer` are not more than 100, second, I pretty sure, that most of them will not notice the difference. But your are right, this is off-topic of course.

> Since this is for certain (most? many? some?) untagged enums specifically, a simpler first step would be to publish a crate that modifies the type in a way that makes the derive handle your case.

Hm, what about introducing a feature flag to `serde_derive`?

oli-obk:
> But your are right, this is off-topic of course.

feel free to ping me on zulip or discord about this :)

> Hm, what about introducing a feature flag to `serde_derive`?

the problem with that is that this changes behaviour for *all* `serde_derive`s in the dependency graph if just one crate turns on that feature (cargo features are additive, which means you should only add API surface, not change behaviour). There are some details in https://doc.rust-lang.org/cargo/reference/features.html#feature-unification and below.

dtolnay (commented):
I appreciate the PR, but the original implementation is definitely my preferred behavior here.

> Because untagged enums serialized like their content, a user can expect, that serializing untagged enum would produce a name for the root element from variant name, but in reality that not the case.

The variant name is literally what the "tag" in "untagged" refers to. Serializing an untagged enum and ending up with serialized output that identifies which variant it was, by name, is not expected behavior for users of untagged enums.

I do not know much about quick-xml but I am willing to believe there is some yet other enum representation which that library would find useful, but I don't know that it should be conflated with what is currently implemented as untagged enums. Specifically regarding XML, in the past I have avoided accepting a variety of XML-motivated features because the needs seem to be so divergent from other formats that it's going to be better served by a different macro.

Mingun:
> The variant name is literally what the "tag" in "untagged" refers to. Serializing an untagged enum and ending up with serialized output that identifies which variant it was, by name, is not expected behavior for users of untagged enums.

I think, that this decision should be made by the concrete serializer, not by the serialized type. Right now it is impossible, and moreover, it is inconsistent -- compare what is used to serialize a struct variant vs. other variants (hint: look at the used struct name). JSON completely ignores type names, but the other serializers could use them (not only XML; RON, for example).

> I appreciate the PR, but the original implementation is definitely my preferred behavior here.

:( It is irrefutable to hear that a project whose description says
> Serde is a framework for <i>ser</i>ializing and <i>des</i>erializing Rust data structures efficiently and generically.

is actually so strongly tied in JSON even in cases where a change in implementation does not affect the JSON format in any way.

PR #2289, opened 2022-10-02
Fix `Deserializer` of `serde_test` to correctly deserialize unit variants from integers/strings/bytes

Mingun:
This is actually possible, but `Deserializer` of `serde_test` returns an error instead of deserialization

### DIFF
--- a/serde_test/src/de.rs
+++ b/serde_test/src/de.rs
@@ -256,7 +256,10 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
             Token::Enum { name: n } if name == n => {
                 self.next_token()?;
 
-                visitor.visit_enum(DeserializerEnumVisitor { de: self })
+                visitor.visit_enum(DeserializerEnumVisitor {
+                    de: self,
+                    optional_token: false,
+                })
             }
             Token::UnitVariant { name: n, .. }
             | Token::NewtypeVariant { name: n, .. }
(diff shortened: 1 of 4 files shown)


dtolnay (commented):
Thanks for the PR!

Would you be able to help me understand the motivation for wanting this behavior? The point of serde_test is to test the behavior of a Deserialize impl / Visitor. If we want to test how the Visitor responds to visit_enum being called, we write e.g. Token::UnitVariant to make serde_test pass a unit variant into visit_enum. Is there a use case that is not already well served by that?

Writing Token::Str would generally be for testing the behavior of a Visitor's visit_str. Silently casting the str into a unit variant would be misleading if someone were trying to test how their enum visitor responds to getting a visit_str call.

Mingun:
It is difficult to remember original motivation because so many time gone, but I will try. Most probably, I tried to create a full set of tests for quick-xml crate and looking for a common ways to testing serializer and deserializer. During that I've found, that serde itself has very disparate and weakly structured tests. It is often difficult to understand which parts are tested and which are not, some parts, on the contrary, are tested many times in different places. So I've started with structured approach and tried to implement exhaustive tests for some parts, enums particularly.

I've found, that some token streams these either produced by the enum serializers or just supported by deserializer and tested implicitly (via deserialization, see additions in [test_suite/tests/test_value.rs](https://github.com/serde-rs/serde/pull/2289/files#diff-484f5cd274f75477c7718b863aa703f2dad6c2b563df985498814c5f97a4a5f7) cannot be tested by `serde_test` deserializers because of panic in `serde_test` about unexpected data.

So I've fixed `serde_test`'s deserializer to accept tokens that de-facto accepted by real deserializers and add corresponding tests in [test_suite/tests/test_de.rs](https://github.com/serde-rs/serde/pull/2289/files#diff-8ae44c680e502021130f357de0f583c600aadb95f9e5d3b982b29d9d350c30db)).

dtolnay (commented):
> I've found, that some token streams _\[...\]_ cannot be tested by `serde_test` deserializers

I mentioned above that serde_test is meant for testing Deserialize impl / Visitor impls; testing token streams is not a right mental model for it. In other words it isn't an issue if there are token streams which cannot be tested. It would instead only be an issue if there is Visitor behavior which cannot be tested.

As far as I understand, this PR isn't making any new Visitor codepaths testable that were not already testable before?

For example in one of the test cases added by this PR:

```rust
assert_de_tokens(&Enum::Unit, &[Token::U8(0)]);
```

the objective is to test that a Visitor used for deserializing variant tags works when called with visit_u8, and that variant 0u8 is `Unit` which is a unit variant.

All of the codepaths involved here in the Visitor and Deserialize side are already testable without this PR. What one writes in the test closely follows the serde APIs being tested:

```rust
assert_de_tokens(&Enum::Unit, &[
    Token::Enum { name: "Enum" },
    Token::U8(0),
    Token::Unit,
]);
```

In general I would prefer to keep serde_test's token representations closely aligned with the serde trait APIs that they are meant to test.

Other formats, like JSON, fill in various type conversions out of necessity from the constraints of the data format. For example there is no such concept **unit variant** natively in JSON. So serde_json uses JSON **string** to represent serde unit variants, because that is the thing available natively in the data format. But serde_test is not constrained in this way: we choose the format and tailor it to the serde API. So I would prefer not to diverge from serde data model by building in JSON-inspired conversions.

Thank you anyway for the work on this PR!

PR #2290, opened 2022-10-02
Remove unused `impl` and unnecessary struct-wrapper around tuple

Mingun:
This PR does some minor cleanups in the code:
- Implementation of `DeserializeSeed` for `TaggedContentVisitor` is never used, so it is removed
- struct `TaggedContent` is used only in one place in the generated code, so it will be replaced by usual tuple

### DIFF
--- a/serde/src/private/de.rs
+++ b/serde/src/private/de.rs
@@ -797,19 +797,13 @@ mod content {
     /// Used by generated code to deserialize an internally tagged enum.
     ///
     /// Not public API.
-    pub struct TaggedContent<'de, T> {
-        pub tag: T,
-        pub content: Content<'de>,
-    }
-
-    /// Not public API.
-    pub struct TaggedContentVisitor<'de, T> {
+    pub struct TaggedContentVisitor<T> {
         tag_name: &'static str,
         expecting: &'static str,
-        value: PhantomData<TaggedContent<'de, T>>,
+        value: PhantomData<T>,
     }
 
-    impl<'de, T> TaggedContentVisitor<'de, T> {
+    impl<T> TaggedContentVisitor<T> {
         /// Visitor for the content of an internally tagged enum with the given
         /// tag name.
         pub fn new(name: &'static str, expecting: &'static str) -> Self {
@@ -821,11 +815,11 @@ mod content {
         }
     }
 
-    impl<'de, T> Visitor<'de> for TaggedContentVisitor<'de, T>
+    impl<'de, T> Visitor<'de> for TaggedContentVisitor<T>
     where
         T: Deserialize<'de>,
     {
-        type Value = TaggedContent<'de, T>;
+        type Value = (T, Content<'de>);
 
         fn expecting(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
             fmt.write_str(self.expecting)
@@ -842,10 +836,7 @@ mod content {
                 }
             };
             let rest = de::value::SeqAccessDeserializer::new(seq);
-            Ok(TaggedContent {
-                tag,
-                content: try!(Content::deserialize(rest)),
-            })
+            Ok((tag, try!(Content::deserialize(rest))))
         }
 
         fn visit_map<M>(self, mut map: M) -> Result<Self::Value, M::Error>
@@ -870,10 +861,7 @@ mod content {
             }
             match tag {
                 None => Err(de::Error::missing_field(self.tag_name)),
-                Some(tag) => Ok(TaggedContent {
-                    tag,
-                    content: Content::Map(vec),
-                }),
+                Some(tag) => Ok((tag, Content::Map(vec))),
             }
         }
     }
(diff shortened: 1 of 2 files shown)


serde/src/private/de.rs
```diff
@@ -797,19 +797,13 @@ mod content {
...
-    }
-
-    /// Not public API.
-    pub struct TaggedContentVisitor<'de, T> {
+    pub struct TaggedContentVisitor<T> {
         tag_name: &'static str,
         expecting: &'static str,
-        value: PhantomData<TaggedContent<'de, T>>,
+        value: PhantomData<T>,
     }
```
oli-obk, line 804: I'm not really convinced that this is better than the status quo, so I'd prefer to just leave it as-is.

serde_derive/src/de.rs
```diff
@@ -1336,9 +1336,7 @@ fn deserialize_internally_tagged_enum(
                 params,
                 variant,
                 cattrs,
-                quote! {
-                    _serde::__private::de::ContentDeserializer::<__D::Error>::new(__tagged.content)
-                },
+                quote!(__deserializer),
```
oli-obk, line 1339: took me a moment, but this is only ever moved into one arm, but all arms move out of it. So this makes sense

oli-obk (changes requested):

serde/src/private/de.rs
Mingun: The removed type is used only in one place for returning pair of values from one function, for that task a dedicated language construct is exist. Occam's razor says us that we should remove superfluous entity.

dtolnay: I think either way is all right and justifiable for different reasons, but I am on board with this change.

serde_derive/src/de.rs
```diff
@@ -1965,7 +1964,7 @@ fn deserialize_generated_identifier(
     quote_block! {
         #[allow(non_camel_case_types)]
         #[doc(hidden)]
-        enum __Field #lifetime {
+        enum #this_value #lifetime {
```
dtolnay, line 1967: I think the use of `#this_value` is semantically incorrect here and in all the places below in this function. `this_value` refers to the expression-ized (using turbofish) name of a potentially remote type. For example in the case of `serde(remote = "Foo<Bar>")`, we'd have this_type=`Foo<Bar>` and this_value=`Foo::<Bar>`. The turbofish makes it appropriate for use in expression position such as `let _ = #this_value::deserialize(d)`, as opposed to type position such as `let _: #this_value = ...`.

That is not consistent with how you've applied it in this function.

dtolnay (changes requested):

serde_derive/src/de.rs
Mingun: Yes, this line appeared due to rebase from earlier implementation where there was a `#this`. It was not obvious in the first place that not just name was changed. I've changed that to use semantically correct name

serde_derive/src/de.rs
```diff
@@ -1965,7 +1965,7 @@ fn deserialize_generated_identifier(
     quote_block! {
         #[allow(non_camel_case_types)]
         #[doc(hidden)]
-        enum __Field #lifetime {
+        enum #this_type #lifetime {
```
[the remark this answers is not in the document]
Mingun: > I think this is still semantically not correct because `this_type` is normally allowed to include generic parameters (the ones that the user wrote in their `remote` attribute).

Hm, did you mean, that almost all places where `#this_type` is used, is incorrect? Because in all cases except one it is written as `#this_type #ty_generics`. The only place where lone `#this_type` is used is
https://github.com/serde-rs/serde/blob/25381be0c93f32aa43ce0e399c45442965e81ecc/serde_derive/src/de.rs#L423-L426

> Overall I would prefer to keep `__Field` as originally written everywhere in this file.
Ok, I remove this commit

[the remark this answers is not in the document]
dtolnay: `#this_type #ty_generics` has a dedicated check to ensure that at most one of `this_type` or `ty_generics` actually contains any generics. https://github.com/serde-rs/serde/blob/25381be0c93f32aa43ce0e399c45442965e81ecc/serde_derive/src/internals/check.rs#L20-L40

To illustrate the difference between `#this_type #ty_generics` vs the `enum #this_type #lifetime` that was written here, consider that all of the existing usages would work if you had this_type=`Foo<()>`, whereas this usage is different and that is not a meaningful thing to write here.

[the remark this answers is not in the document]
Mingun: > To illustrate the difference between `#this_type #ty_generics` vs the `enum #this_type #lifetime` that was written here, consider that all of the existing usages would work if you had this_type=`Foo<()>`, whereas this usage is different and that is not a meaningful thing to write here.

Quite a hard to do that because I do not see obvious link between the check that you pointed out and the content of `this_type` and `ty_generics`. I will take your word that everything is in fine here.

dtolnay (approved):
Thanks!

PR #2293, opened 2022-10-07
Fix roundtrip of enum with skipped variant

djkoloski:
On enums with a skipped variant that is not the last one, the generated serialization and deserialization code map the variants of the enum to different variant indices. In `deserialize_identifier`, the variant index starts at 0 and counts up once per non-skipped variant. However, for serialization the index is always the index of the variant within the enum.

This PR corrects the deserialization code to always use the index of the variant within the enum.

This was surfaced in https://github.com/jamesmunns/postcard/issues/79.


XBagon (commented):
I patched my project with this fix and my `serde` test finally succeeded. Thank you! (Inoffical approve 😋)

Issue #2233, opened 2022-06-22
Problem deserializing adjacently tagged enum with default content


psFried:
[Example in playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=5bc6f6fceff2d2415639cd7294f79d42)

I have an adjacently tagged enum, and I'd like for the content to be optional. The enum looks like:

```rust
#[derive(Deserialize, Debug)]
#[serde(tag = "tag", content = "content")]
enum TestEnum {
    NoContent,
    OptionalContent(#[serde(default)] OptionalContent),
}

#[derive(Deserialize, Default, Debug)]
struct OptionalContent {
    _a: Option<String>,
}
```

I expected to be able to successfully deserialize the json `{"tag": "OptionalContent"}` into the `OptionalContent` variant and have the inner `OptionalContent` be populated with the default. The actual result was `Err(Error("missing field `content`", line: 1, column: 26))`.

When I change the input json to `{"tag": "OptionalContent", "content": null}`, I get a different error: `Err(Error("invalid type: null, expected struct OptionalContent", line: 1, column: 42))`.

I was able to get this to work is when the json is `{"tag": "OptionalContent", "content": {}}`, but I'd prefer to have the default work when content is either missing or `null`.

Is this a bug? Is there some other way to get this working? Appreciate whatever help or advice you're able to provide.

sciyoshi:
I believe that this does work if you use `Option<OptionalContent>`. However it would be nice if this worked automatically if the variant has a `Default`.

kangalioo:
Related to #2248

Mingun:
Probably, this could be solving by allowing `#[serde(default)]` on enum variants in ajacently tagged enums. Then in case of absence of the `content` field corresponding variant should be constructed using `Default` implementation of all it's components:

```rust
#[derive(Deserialize, Debug)]
#[serde(tag = "tag", content = "content")]
// Also should be possible to specify enum attribute instead
// of specifying attribute on each variant
// #[serde(default)]
enum AdjacentlyTagged {
  // result = Unit;
  #[serde(default)]
  Unit,

  // result = Newtype1(MustImplementDefault::default());
  #[serde(default)]
  Newtype1(MustImplementDefault),

  // result = Newtype2(d());
  #[serde(default)]
  Newtype2(#[serde(default = "d")] DoesNotImplementDefault),

  // result = Tuple(MustImplementDefault::default(), d());
  #[serde(default)]
[...]

PR #2295, opened 2022-10-11
Allow deserializing adjacently tagged enum struct variants if all fields have #[serde(default)]

sciyoshi:
Suppose you have the following enum:

```rust
#[derive(Deserialize)]
#[serde(tag = "type", content = "attrs")]
enum Node {
     Doc { #[serde(default)] title: String }
}
```

[(playground link)](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a43d37cebf7c5248393284d171c671f5)

The `Doc` variant of this enum can be constructed from empty content, i.e. when deserializing from `{"type": "Doc", "attrs": {}}`. In this case, it would be great if serde could also deserialize a value where the content is missing completely, i.e. `{"type": "Doc"}`.

This PR adds a check for struct-style variants where every field is marked with `#[serde(default)]`.

See similar issue here: https://github.com/serde-rs/serde/issues/2233


clbarnes:
Is there anything else which needs to happen to push this over the line? I'm running into this at the moment; it looks like a fairly small change to have been sitting stale for a year.

To clarify for myself, would this allow e.g.

```rust

#[derive(Deserialize)]
#[serde(tag = "type", content = "attrs")]
enum Node {
     NeedsAttrs(SomeData),
     OptionalAttrs(#[serde(default)] OtherData),
}

struct SomeData {
    a: u8,
}

#[derive(Default)]
struct OtherData {}
```

i.e. it would be able to deserialise `{"type": "OptionalAttrs"}` even though not all variants have data implementing `Default`?

sciyoshi:
@clbarnes yes, the PR as it stands should support deserializing from `{"type": "OptionalAttrs"}`. I'm not sure if this code still works as-is with the latest changes in `main`. Maybe @dtolnay can comment on whether this would be acceptable to merge?

Issue #2302, opened 2022-10-20
Serialize/deserialize tagged newtype variant containing unit


tage64:
To me, it seems that a tagged newtype variant containing `()` should be serializable/deserializable in the same way as unit variants. Specificly I have the following enum:

```rust
#[derive(Serialize, Deserialize, Debug)]
#[serde(tag = "result", rename_all = "snake_case")]
enum Response<T> {
    Success(T),
    Error {
        ...
    },
}
```

I'd expect `Success(())` to pe deserialized from and serialized to `{ "result": "success"}`, but that doesn't seem to be the case. It doesn't work to deserialize it and when I try to serialize it (with serde_json) I get the following error:

```
Error("cannot serialize tagged newtype variant Response::Success containing unit", line: 0, column: 0)
```

Is there any reason why this doesn't work, or should we fix this? I'd would be happy to make a PR.

dtolnay:
It would be fine to make this work I think.

tage64:
The PR is: #2303

PR #2303, opened 2022-10-20
Serialize and deserialize a tagged newtype variant over unit () as if it was a unit variant

tage64:
Let's say we have the following enum:
```rust
#[serde(tag = "result", rename_all = "snake_case")]
enum Response<T> {
    Success(T),
    Error {
        ...
    },
}
```
This PR makes it possible to serialize and deserialize `Response::Success(())` into and from `{"result": "success"}`. Previously, this was not possible when the type parameter `T` was `()`.

Note that this is the same behaviour as if `Response::Success` would be a unit variant, but now it can be used in a more general context.

See [This part of the docs](https://serde.rs/enum-representations.html) for a background to internally tagged enums.

It's my first time contributing to Serde, so please appologise any mistakes regarding the contribution process.

References #2302

### DIFF
--- a/serde/src/private/de.rs
+++ b/serde/src/private/de.rs
@@ -1272,7 +1272,6 @@ mod content {
                 //
                 // We want {"result": "Success"} to deserialize into `Response<T>`.
                 Content::Map(ref v) if v.is_empty() => visitor.visit_unit(),
-                Content::Seq(ref v) if v.is_empty() => visitor.visit_unit(),
                 _ => Err(self.invalid_type(&visitor)),
             }
         }
--- a/serde/src/private/de.rs
+++ b/serde/src/private/de.rs
@@ -1262,6 +1262,17 @@ mod content {
         {
             match self.content {
                 Content::Unit => visitor.visit_unit(),
+
+                // As a special case, allow deserializing newtype variant containing unit. E.G:
+                // #[derive(Deserialize)]
+                // #[serde(tag = "result")]
+                // enum Response<T> {
+                //     Success(T),
+                // }
+                //
+                // We want {"result": "Success"} to deserialize into `Response<T>`.
+                Content::Map(ref v) if v.is_empty() => visitor.visit_unit(),
+                Content::Seq(ref v) if v.is_empty() => visitor.visit_unit(),
                 _ => Err(self.invalid_type(&visitor)),
             }
         }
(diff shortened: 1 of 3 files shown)


test_suite/tests/test_annotations.rs
```diff
@@ -2302,6 +2302,34 @@ fn test_internally_tagged_enum_containing_flatten() {
...
+    );
+    assert_ser_tokens(
+        &Data::A(()),
+        &[
+            Token::Map { len: Some(1) },
+            Token::Str("t"),
+            Token::Str("A"),
+            Token::MapEnd,
+        ],
+    );
```
dtolnay, line 2330: This same input is already tested by `assert_tokens` above.

```suggestion
```

serde/src/private/de.rs
```diff
@@ -1262,6 +1262,17 @@ mod content {
...
+                // As a special case, allow deserializing newtype variant containing unit. E.G:
+                // #[derive(Deserialize)]
+                // #[serde(tag = "result")]
+                // enum Response<T> {
+                //     Success(T),
+                // }
+                //
+                // We want {"result": "Success"} to deserialize into `Response<T>`.
+                Content::Map(ref v) if v.is_empty() => visitor.visit_unit(),
+                Content::Seq(ref v) if v.is_empty() => visitor.visit_unit(),
```
dtolnay, line 1275: I think I see the `Content::Map` case corresponds to the `serialize_map` that has been added to serde/src/private/ser.rs in this PR. But what is `Content::Seq` handling?

dtolnay (changes requested):
Thanks!

serde/src/private/de.rs
tage64: When I first read your comment, I thought that The line with `Content::Seq` should not be there and I had made a mistake. I read the code for unit structs at the function below and somehow just thought that I should copy both lines, with `Content::Map` and `Content::Seq`. However, that would mean that my example enum above could be deserialized from `["success"]`, which may not be what we want. And I made a commit to fix this.

But after a bit more investigation, I see that the behaviour of unit structs seem to be just that. I.E. if we have a unit struct `SomeStruct` we could deserialize `Response::Success(SomeStruct)` from `["success"]`. So for consistency this should perhaps also be allowed for units so that `()` and unit structs behave in the same way.

What do you think? Should we allow the sequence notation or not? If we decide to allow it, I will just revert my last commit.

dtolnay: I think we should land it like this and find out whether anyone needs the `Seq` case in practice.

Thanks!

dtolnay (approved):

PR #2319, opened 2022-11-09
Add Safety comments to unsafe blocks

erickt:
This makes some small tweaks to the unsafe blocks to make them easier to audit:

* It changes `serde::de::Format::Buf` to be a private type since it's already not publicly exposed.
* It changes `Buf::as_str()` to be unsafe, since it relies on the caller to only write UTF-8 into the buffer.
* It adds `// Safety: ...` comments to the two callsites in `de::Visitor` that uses `Buf` to write an error message with `i128` and `u128` types.
* It converts a comment into an explicit `// Safety: ...` comment.

### DIFF
--- a/serde/src/de/format.rs
+++ b/serde/src/de/format.rs
@@ -1,19 +1,19 @@
 use lib::fmt::{self, Write};
 use lib::str;
 
-pub struct Buf<'a> {
+pub(super) struct Buf<'a> {
     bytes: &'a mut [u8],
     offset: usize,
 }
 
 impl<'a> Buf<'a> {
-    pub fn new(bytes: &'a mut [u8]) -> Self {
+    pub(super) fn new(bytes: &'a mut [u8]) -> Self {
         Buf { bytes, offset: 0 }
     }
 
-    pub fn as_str(&self) -> &str {
+    pub(super) unsafe fn as_str(&self) -> &str {
         let slice = &self.bytes[..self.offset];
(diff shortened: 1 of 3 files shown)


serde/src/de/format.rs
```diff
@@ -1,19 +1,19 @@
...
 }
 
 impl<'a> Buf<'a> {
-    pub fn new(bytes: &'a mut [u8]) -> Self {
+    pub(super) fn new(bytes: &'a mut [u8]) -> Self {
         Buf { bytes, offset: 0 }
     }
 
-    pub fn as_str(&self) -> &str {
+    pub(super) unsafe fn as_str(&self) -> &str {
```
dtolnay, line 14: I don't see why this needs to be unsafe.

dtolnay: > - It changes `Buf::as_str()` to be unsafe, since it relies on the caller to only write UTF-8 into the buffer.

The API enforces that the caller only wrote UTF-8 into the buffer.

erickt: Hi dtolnay! Oh you are totally right. I’ll fix it.

dtolnay: I put up #2329 to be able to merge just your `Buf` visibility change and your `// Safety: ...` comment change in src/ser/impls.rs, without the `as_str()` safety change. We can follow up with a separate PR to document the justification for how `as_str()` is safe.

Thanks!

dtolnay (approved):

Issue #2262, opened 2022-08-17
Testing `DeserializeSeed` using `serde_test`


Anders429:
Is there any way to test types that implement `DeserializeSeed`, as opposed to `Deserialize`, using the `serde_test` crate? I am looking for some equivalent of `assert_de_tokens`, but that would accept the seed as well to test stateful deserialization. Perhaps something like a `assert_de_seed_tokens` or similar?

I would love to know if such a thing exists or if there are any workarounds that can be used to test `DeserializeSeed` types with the current tools.

Shatur:
Encountered the same issue and tried to deserialize into tokens manually. But [de](https://github.com/serde-rs/serde/blob/master/serde_test/src/de.rs) module is private despite that [Deserializer](https://github.com/serde-rs/serde/blob/d208762c81883a181e8c6a9ca3f303e040105c7d/serde_test/src/de.rs#L11) itself is `pub`. Maybe make this module public?

dtolnay:
I notice https://github.com/Anders429/serde_assert now exists. Thanks!

PR #2325, opened 2022-11-23
Add asserts for DeserializeSeed

Shatur:
Equivalent asserts that accept a seed to test stateful deserialization. Was requested in #2262.

### DIFF
--- a/serde_test/src/assert.rs
+++ b/serde_test/src/assert.rs
@@ -228,30 +228,30 @@ where
 /// ```edition2018
 /// # use serde::{de::{DeserializeSeed, Visitor}, Deserializer};
 /// # use serde_test::{assert_de_seed_tokens, Token};
-/// 
+///
 /// #[derive(Debug, PartialEq)]
 /// struct Example {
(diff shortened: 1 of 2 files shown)


dtolnay (changes requested):
Could you please add a usage example to the documentation of each of the new functions?

I'm not quite sure whether I would want to grow the API of serde_test in this way. The existing interface is already pretty clumsy and limited. But nothing in there is using any private internals of serde, so perhaps it would be best to let others experiment with different designs in a separate crate.

Shatur:
> Could you please add a usage example to the documentation of each of the new functions?

@dtolnay Sure, added.

> But nothing in there is using any private internals of serde

Deserializer is `pub`, but it's a private module, I can't use it in a third-party crate.

dtolnay (commented):
I gave this another look, and I think I would prefer not to have this in this crate.

Thanks anyway for the PR though!

I would be interested to see someone pursue a library for testing serde impls in a better way as a different crate.

Issue #2327, opened 2022-11-27
Deriving Serialize for a struct from external crate which uses generics


AntoineBalaine:
Hello, 
I'm trying to implement Serialize for [nom_locate's](https://docs.rs/nom_locate/latest/nom_locate/) [`LocatedSpan`](https://docs.rs/nom_locate/latest/nom_locate/struct.LocatedSpan.html). It's a struct that encapsulates a token with some location information.  can't figure out how to implement Serialize for it. 

I created a `WrappedSpan` to build the bridge with `LocatedSpan`, but the derive macro doesn't chew up the passed-in generic. Is there a way around this? 

```rust
#[derive(Serialize, Deserialize)]
#[serde(remote = "LocatedSpan<T>")] // <= This <T> produces an un-parseable token
struct WrappedSpan<T: AsBytes>{
    #[serde(getter = "LocatedSpan::location_offset")]
    offset: usize,
    #[serde(getter = "LocatedSpan::location_line")]
    line: u32,
    #[serde(getter = "LocatedSpan::fragment")]
    fragment: T,
};

// Provide a conversion to construct the remote type.
impl <T: AsBytes>From<WrappedSpan<T>> for LocatedSpan<T> {
    fn from(def: WrappedSpan<T>) -> LocatedSpan<T> {
        LocatedSpan{
            offset: def.offset,
            line: def.line,
            fragment: def.fragment,
            extra: (),
        }
    }
}
```

AntoineBalaine:
Thank you for this update!

dtolnay:
Fixed in serde_derive 1.0.148. You need to use `serde(remote = "LocatedSpan")`.

AntoineBalaine:
Could you be a bit more specific ? I can't seem to figure it out with the update:
```rust
#[derive(Serialize, Deserialize, PartialEq, Eq)]
#[serde(remote = "LocatedSpan")]
struct WrappedSpan<T: AsBytes> {
    #[serde(getter = "LocatedSpan::location_offset")]
    offset: usize,
    #[serde(getter = "LocatedSpan::location_line")]
    line: u32,
    #[serde(getter = "LocatedSpan::fragment")]
    fragment: T,
}

// Provide a conversion to construct the remote type.
impl<T: AsBytes> From<WrappedSpan<T>> for LocatedSpan<T> {
    fn from(def: WrappedSpan<T>) -> LocatedSpan<T> {
        LocatedSpan::new(def.fragment)
    }
}

#[test]
[...]

PR #2331, opened 2022-11-28
Improve error message on remote derive duplicate generics

dtolnay:
Related to #2327.


test_suite/tests/ui/remote/double_generic.rs
```diff
@@ -0,0 +1,17 @@
+use serde_derive::{Deserialize, Serialize};
+
+mod remote {
+    pub struct Struct<T, U> {
```
martin-g, line 4: This new struct seems not used.

PR #2333, opened 2022-12-03
Remove some Serialize trait bounds

jonasbb:
Containers for the most part do not have any trait requirements for
iterating over them. So these bounds are unnecessary when Serializing
only.


dtolnay (approved):
Thanks!

Issue #1579, opened 2019-07-17
Build failure


fdubois1:
We have a dependency on serde 1.0 so we moved to 1.0.95 and now, we have this issue in our build for Android (It builds for other platforms) : 

```
  error[E0432]: unresolved imports `std::sync::atomic::AtomicI64`, `std::sync::atomic::AtomicU64`
     --> /home/user/.cargo/registry/src/github.com-1ecc6299db9ec823/serde-1.0.95/src/lib.rs:221:33
  
  221 |     pub use std::sync::atomic::{AtomicI64, AtomicU64};
                                   ^^^^^^^^^  ^^^^^^^^^ no `AtomicU64` in `sync::atomic`
                                   |
                                   no `AtomicI64` in `sync::atomic`
  help: a similar name exists in the module
  
  221 |     pub use std::sync::atomic::{AtomicI8, AtomicU64};
                                   ^^^^^^^^
  help: a similar name exists in the module
  
  221 |     pub use std::sync::atomic::{AtomicI64, AtomicU8};
                                              ^^^^^^^^
  
  error: aborting due to previous error

```

I thought I had an old version of rust, but I'm using 1.35.0 so I don't understand why AtomicI64 is not found. Any hint ? Thanks for your help

tymcauley:
I just ran into this exact same issue. It appears this is new in 1.0.95 (I can't reproduce it in 1.0.94). I've only noticed this issue for the mips-unknown-linux-gnu and mips-unknown-linux-musl targets (works fine on x86_64 targets). You can reproduce this by creating the serde test crate from this repo's homepage (the simple JSON serde example), and running these commands:

```
rustup target add mips-unknown-linux-gnu
rustup target add mips-unknown-linux-musl
cargo build --target mips-unknown-linux-gnu
cargo build --target mips-unknown-linux-musl
```

fdubois1:
In my case, I didn't specify the exact target, but it is --target arm-linux-androideabi

Roguelazer:
Hm. The documentation does include the following line:

> MIPS and PowerPC do not support AtomicU64, and "more esoteric" platforms like armv5te, thumv6m, etc, have little-to-no support.

It looks like the relevant `cfg` flag never got merged (https://github.com/rust-lang/rust/issues/56753).

For now, do y'all want to just gate this feature on x86/x86_64 in the build.rs script so we stop breaking MIPS and ARM?

fdubois1:
[...]

PR #2337, opened 2022-12-11
Use `target_has_atomic` on Rust 1.60+ to enable atomic (de)serialization

badboy:
I recently ran into the issue that serde didn't implement `Deserialize` for atomics when compiled for Android on Arm (`armv7-linux-androideabi`).
This was once reported in #1579 and might have been addressed back then. 
[Current version doesn't enable it for arm ever](https://github.com/serde-rs/serde/blob/3aec2a96a86d5299bd6cc57ea4840cf12f176250/serde/build.rs#L95-L111), so it's still an issue regardless.

`target_has_atomic` has become stable in Rust 1.60 (see https://github.com/rust-lang/rust/issues/32976#issuecomment-1035065937), so now it is possible to use it, at least when compiled for 1.60 and above.

I implemented it the quick way, just duplicating the code when a different cfg is enabled.
I might be able to merge those two into a single macro if that's desired.

I tested it locally by building the test suite for Android (requires proper set up of the Android NDK, etc.):

```
cargo test --target armv7-linux-androideabi -p serde_test_suite --no-run
```

### DIFF
--- a/serde/build.rs
+++ b/serde/build.rs
@@ -95,19 +95,23 @@ fn main() {
     // Whitelist of archs that support std::sync::atomic module. Ideally we
     // would use #[cfg(target_has_atomic = "...")] but it is not stable yet.
     // Instead this is based on rustc's compiler/rustc_target/src/spec/*.rs.
-    let has_atomic64 = target.starts_with("x86_64")
-        || target.starts_with("i686")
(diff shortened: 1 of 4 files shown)


dtolnay (approved):
Thanks!

Issue #741, opened 2017-02-03
Make some of the Content api public


dtolnay:
Follow-up to #739. After we get a chance to iterate on the implementation of the `untagged` and `tag` attributes, we can start to polish and expose pieces of the implementation because they are definitely useful in other use cases.

oli-obk:
We can get rid of the `E` parameter: https://github.com/serde-rs/serde/commit/6277079152bb933b8ea2a3d6f3d1af5720dac31b

dtolnay:
I think the remaining tasks before we can make this public are:

- Adjacently tagged enums #758
- Integer tags #745
- Error messages #773

Not all of these require obvious breaking changes to Content but I wouldn't want to limit our solution to any of these by committing to a flawed API.

dtolnay:
One nuance here is that it is helpful to have slightly different Content enums when serializing vs when deserializing.

- [`ser::Content`](https://github.com/serde-rs/serde/blob/v0.9.9/serde/src/ser/content.rs#L90-L127)
- [`de::Content`](https://github.com/serde-rs/serde/blob/v0.9.9/serde/src/de/content.rs#L31-L58)

dtolnay:
Let's continue to track this but from the opposite direction -- I filed https://github.com/arcnmx/serde-value/issues/16 to absorb serde-value into the public API of Serde when they are ready for it. I think it will be more productive to polish and iterate on serde-value publicly than to try to get our own implementation right on the first try.

slinkydeveloper:
Hi, any updates on this?

Sytten:
@dtolnay Has the stance changed on that? The currently private `Content` and `ContentRefDeserializer` are very useful. `serde-value` is not really a replacement since the deserializer takes an owned `Value` not a ref.
Is it something you are still looking to make public?

dtolnay:
The fact that no one has made a better serde-value crate is some indication that the use case is not that important. Do you have insight into why that hasn't happened?

Regarding my stance, no change since https://github.com/serde-rs/serde/issues/741#issuecomment-325523577.

Sytten:
Its really only useful for untagged enums so it is probably niche but when you hit the problem there is not good solution except using the private api which I have seen used around since its not really private and people are lazy.

murl-digital:
[...]

PR #2348, opened 2022-12-18
Proposal for public `de:Content` API.

koehlma:
This PR is a proposal for a public `de::Content` API. It partially addresses issue #741.

The existing private `Content` structure is currently used for buffering when deserialising, e.g., tagged enums. However, it would also be useful outside of that context. I have two concrete use cases in mind: (1) Deserialization of structures where the exact type of certain parts is dynamically determined and (2) implementations of `Deserialize` which are not generated via the derive macros (e.g., by another code generator or manually written).

In fact, I am currently working on [a tool](https://oss.silitics.com/sidex/) which generates implementations of `Deserialize` from a schema. Unfortunately, replicating the functionality of `Content` in a third-party crate in an efficient way does not seem possible. Internally, Serde relies on `__deserialize_content` for an efficient implementation of `Content`.

Instead of exposing the internal details of `Content`, this PR introduces an API which mostly implements already public APIs (`Deserializer`, `IntoDeserializer` and `Visitor`) for opaque newtype wrappers around private structures. In theory, this should allow for `Content` to evolve while being committed to only a very thin public API most of which is already public anyway. At the same time, the provided API should be enough for both use cases.

I guess a similar approach should be possible for `ser::Content` as well.

I am looking forward to your thoughts and feedback regarding this proposal. I am willing to see this (or any alternative suggestion) through. In any case, I believe a public `de::Content` API would be valuable.

### DIFF
--- a/serde/src/de/content.rs
+++ b/serde/src/de/content.rs
@@ -1,24 +1,343 @@
-use crate::{de, private, Deserialize, Deserializer};
+//! Data structures for buffering self-describing formats.
+//!
+//! ```
+//! # use serde::de::{Deserialize, value, IntoDeserializer, content::Content};
+//! let content = Content::from(32);
+//! let deserializer = IntoDeserializer::<value::Error>::into_deserializer(content);
+//! assert_eq!(u32::deserialize(deserializer).unwrap(), 32);
+//! ```
 
-/// An efficient buffer for arbitrary deserialized *content*.
-///
-/// ℹ️ Note that [`Content`] can only be constructed by deserialization.
+use std::fmt;
+
+use crate::{de, private, Deserialize};
+
+use super::Visitor;
+
+/// An efficient buffer for arbitrary self-describing *content*.
 #[derive(Clone)]
 #[repr(transparent)]
 pub struct Content<'de>(private::de::Content<'de>);
 
-impl<'de> Content<'de> {
-    /// Turns the content into a deserializer.
-    pub fn into_deserializer<E: de::Error>(self) -> impl Deserializer<'de, Error = E> {
-        private::de::ContentDeserializer::new(self.0)
-    }
-}
-
 impl<'de> Deserialize<'de> for Content<'de> {
+    #[inline]
     fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
     where
         D: crate::Deserializer<'de>,
     {
         Ok(Self(private::de::Content::deserialize(deserializer)?))
+    }
+}
+
+impl<'de, E: de::Error> de::IntoDeserializer<'de, E> for Content<'de> {
+    type Deserializer = ContentDeserializer<'de, E>;
+
+    fn into_deserializer(self) -> Self::Deserializer {
+        ContentDeserializer(private::de::ContentDeserializer::new(self.0))
+    }
+}
+
+macro_rules! impl_from_for_content {
+    ($($type:ty => $constructor:ident ,)*) => {
+        $(
+            impl<'de> From<$type> for Content<'de> {
+                fn from(value: $type) -> Self {
+                    Self(private::de::Content::$constructor(value))
+                }
+            }
+        )*
+    };
+}
+
+impl_from_for_content! {
+    bool => Bool,
+
+    u8 => U8,
+    u16 => U16,
+    u32 => U32,
+    u64 => U64,
+
+    i8 => I8,
+    i16 => I16,
+    i32 => I32,
+    i64 => I64,
+
+    f32 => F32,
+    f64 => F64,
+
+    char => Char,
+
+    String => String,
+    Vec<u8> => ByteBuf,
+}
+
+impl<'de> From<&'de str> for Content<'de> {
+    fn from(value: &'de str) -> Self {
+        Self(private::de::Content::Str(value))
+    }
+}
+
+impl<'de> From<&'de [u8]> for Content<'de> {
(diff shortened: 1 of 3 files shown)


dtolnay (commented):
I am on board with this approach. Some thoughts:

- From your description of your use case, it's not clear where all those `From` impls become needed. If they are not needed I would prefer not to have them.

- Is it possible to deduplicate some of the trait impls related to `Content` between `serde::de::content` and `serde::private::de`? If `serde::de::content::Content` has a `Deserializer` impl, and a `Visitor`, it doesn't seem like `serde::private::de::Content` also needs to have one, if all the code currently needing it can switch to the public one.

- I would want someone to brainstorm names other than `Content`. It's not clear this conveys the purpose of the type in buffering and deferring deserialization, particularly as the content itself is only opaque.

koehlma:
Thanks for the quick response! I am glad you are on board with the general approach.

> From your description of your use case, it's not clear where all those `From` impls become needed.

They are not needed as it is always possible to construct the respective instances of `Content` via the visitor. In fact, I see them as mere syntactic sugar for the visitor, i.e., for `ContentVisitor::new().visit_$type(value)`. As I see them as mere syntactic sugar, I feel that they do not add any additional commitments in terms of the API while adding a bit of  convenience for constructing `Content` (e.g., in something like `TagOrContentVisitor`). However, as you hinted at, my use case does not depend on that, so not having them is also perfectly fine with me.

> Is it possible to deduplicate some of the trait impls related to `Content` between `serde::de::content` and `serde::private::de`?

Yes, I think, this will be possible.

> I would want someone to brainstorm names other than `Content`.

How do you feel about `Buffer` (or alternatively `DeBuffer` for *deserialization buffer*)? I think `Buffer` would convey the purpose of the structure quite well – as you said, it is about buffering.

Another option maybe something with `Any`. However, it is not actually true that `Content` can be *anything* which could be deserialized if the concrete type were known because it is somewhat limited to self-describing formats. Yet another option maybe something with `Value` – however, `de::value` is already taken. Furthermore, both these options make it seem as if the structure is intended to be used for more than just buffering.

koehlma:
@dtolnay I finally got the time to address your points.

- `Content` is now named `Buffer` and placed in `de::buffer`.
- Everything `Buffer` specific from `private::de::content` has been moved to `de::buffer` preventing duplication.

Unfortunately, I am unable to figure out how to fix the issues with Rust version `<= 1.25`. I will leave a few comments indicating other issues which we need to address before this can be merged.

serde_derive/src/de.rs
```diff
@@ -2143,16 +2141,16 @@ fn deserialize_identifier(
...
+                let __value = _serde::de::buffer::BufferVisitor::new().visit_borrowed_str::<_serde::de::value::Error>(__value).unwrap();
             }),
             Some(quote! {
-                let __value = _serde::__private::de::Content::ByteBuf(__value.to_vec());
+                let __value = _serde::de::buffer::BufferVisitor::new().visit_byte_buf::<_serde::de::value::Error>(__value.to_vec()).unwrap();
             }),
             Some(quote! {
-                let __value = _serde::__private::de::Content::Bytes(__value);
+                let __value = _serde::de::buffer::BufferVisitor::new().visit_borrowed_bytes::<_serde::de::value::Error>(__value).unwrap();
             }),
```
koehlma, line 2154: Using the `BufferVisitor` here is a kind of hack. This is where the `From` implementations would be useful.

serde/src/de/buffer.rs
```diff
@@ -0,0 +1,1806 @@
...
+    // function in `private::de` instead.
+    pub fn as_str(&self) -> Option<&str> {
+        match self.0 {
+            BufferInner::Str(x) => Some(x),
+            BufferInner::String(ref x) => Some(x),
+            BufferInner::Bytes(x) => str::from_utf8(x).ok(),
+            BufferInner::ByteBuf(ref x) => str::from_utf8(x).ok(),
+            _ => None,
+        }
+    }
```
koehlma, line 37: This function is used by code generated by `serde_derive`. How should we deal with that?

serde/src/de/buffer.rs
```diff
@@ -0,0 +1,1809 @@
...
+}
+
+#[derive(Debug, Clone)]
+pub(crate) enum BufferInner<'de> {
+    Bool(bool),
+
+    U8(u8),
+    U16(u16),
+    U32(u32),
+    U64(u64),
```
jonasbb, line 76: Is it possible to extend the support to `u128` and `i128`. Not all old compiler versions support it, but it is supported on all targets starting from Rust 1.40. Given that this is now a private type, that should not be a problem, as all the depending code can also be cfg-gated.

koehlma: Good point! I modified the PR accordingly.

koehlma:
After giving the `as_str` method and `From` implementations a bit more thought, I think they are worthwhile. I updated the first post with an explanation as to why I think this is the case. From my side, this PR is ready to be merged now.

I am looking forward to your feedback.

koehlma:
I just tried to resolve the conflicts so that this PR is mergeable again. However, now the tests fail because `trybuild` appears to use another, now incompatible, version of `serde_derive` – which I do not know how to fix. 😕 I would like to avoid adapting the PR to new changes over and over again.

What is the best way for moving forward with this PR? Is there anything I can do?

PR #2355, opened 2022-12-30
Remove appendix from LICENSE-APACHE

Niki4tap:
All rust repos are removing this appendix, not sure why serde still has it, for more info see:
- https://github.com/rust-lang/rust/pull/67734
- https://github.com/rust-lang/miri/pull/1846

and [this comment](https://github.com/rust-lang/rust/pull/67734#issuecomment-576030746) specifically.


dtolnay (approved):
Thanks!

PR #2357, opened 2023-01-02
Use `const` generics for fixed-size arrays

jun-sheaf:
This PR implements `const` generics for fixed-size arrays.

- For deserializing, some manual memory management is used. However, in theory, the compiler should optimize the code back to the original code due to loop unrolling and memory optimizations.
- Because the size parameter is now generic, `T` must always be serializable, even for empty tuple types.

### DIFF
--- a/serde/build.rs
+++ b/serde/build.rs
@@ -94,6 +94,12 @@ fn main() {
         println!("cargo:rustc-cfg=no_relaxed_trait_bounds");
     }
 
+    // Constant generics stabilized in Rust 1.59:
+    // https://blog.rust-lang.org/2022/02/24/Rust-1.59.0.html#const-generics-defaults-and-interleaving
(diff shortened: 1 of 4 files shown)


dtolnay (commented):
This looks substantially identical to #1860, so I'll go ahead and close. The next step for anyone interested in progress on this would be contributing to the resolution of https://github.com/rust-lang/rust/issues/61415, most likely via lang RFC or compiler MCP.

jun-sheaf:
From looking at the issue you posted, it seems your problem with this PR is w.r.t. the `[T;0]` case and the poor optimization of uninit. Outside of optimization, is `[T;0]` a deal breaker?

dtolnay:
Yes.

jun-sheaf:
Would you accept this under a feature flag?

PR #2361, opened 2023-01-09
Add exhaustive match arm in `#[derive(Serialize)]` expansion

### LINKED ISSUE #2360: Add wildcard arm when deriving `Serialize` for non exhaustive enums
I am trying to derive (de)serialize for an enum containing a `std::io::ErrorKind`
```rust
use std::io;
enum IoErrorKind {
    FileNotFound,
    PermissionDenied,
    Other(io::ErrorKind),
}
```

This complains with `std::io::ErrorKind` does not impl `Deserialize`. Having read #1480, I understand that that won't happen.
Trying to work around this with
```rust
enum IoErrorKind {
    FileNotFound,
    PermissionDenied,
    Other(#[serde(with = "_StdIoErrorKindDef")] io::ErrorKind),
}

#[serde(remote = "io::ErrorKind")]
pub enum _StdIoErrorKindDef {
    NotFound,
    PermissionDenied,
    ConnectionRefused,
    ConnectionReset,
    #[cfg(io_error_more)]
    HostUnreachable,
    #[cfg(io_error_more)]
    NetworkUnreachable,
    // ...
}
```
fails, because the derive macro expansion matches exhaustively against all variants, but the compiler still doesn't like that it doesn't have a `_ => todo!()` arm.
To fix this, a `_ => unreachable!()` arm could be added to the match statement in the expansion.
I am fairly sure `unreachable!()` is appropriate, because the macro matches exhaustively against all the variants, so it actually would be unreachable. If it isn't, I don't really know what would be.

mrlegohead0x45:
Closes #2360

### DIFF
--- a/serde_derive/src/ser.rs
+++ b/serde_derive/src/ser.rs
@@ -408,13 +408,18 @@ fn serialize_enum(params: &Parameters, variants: &[Variant], cattrs: &attr::Cont
 
     let self_var = &params.self_var;
 
-    let arms: Vec<_> = variants
+    let mut arms: Vec<_> = variants
         .iter()
         .enumerate()
         .map(|(variant_index, variant)| {
             serialize_variant(params, variant, variant_index as u32, cattrs)
         })
         .collect();
+
+    // unreachable because all the variants are matched exhaustively above
+    // this is here to stop the compiler complaining about trying to serialize
+    // a non-exhaustive enum
+    arms.push("_ => unreachable!()".parse().unwrap());
 
     quote_expr! {
         match *#self_var {


serde_derive/src/ser.rs
```diff
@@ -408,14 +408,19 @@ fn serialize_enum(params: &Parameters, variants: &[Variant], cattrs: &attr::Cont
...
-    let arms: Vec<_> = variants
+    let mut arms: Vec<_> = variants
         .iter()
         .enumerate()
         .map(|(variant_index, variant)| {
             serialize_variant(params, variant, variant_index as u32, cattrs)
         })
         .collect();
 
+    // unreachable because all the variants are matched exhaustively above
```
jonasbb, line 419: The code is matching here on the remote enum, i.e., the one which does not implement `Serialize`. In your original issue that is the `io::ErrorKind` enum which is marked as `non_exhaustive`. So this reasoning does not work here. You never know for sure that you covered all enum variants, since any dependency update could change it.

It might be possible to add a catch-all variant to the def enum. All enum variants which match will be translated 1-to-1. The catchall arm `_` then goes to a separate enum variant.

mrlegohead0x45: Apologies if I still misunderstand you, but looking at the expansion of `#[derive(Serialize)]` on `_StdIoErrorKindDef` using vscode rust-analyzer expand macro recursively it looks like this
```rust
impl _StdIoErrorKindDef {
    fn serialize<__S>(
        __self: &io::ErrorKind,
        __serializer: __S,
    ) -> _serde::__private::Result<__S::Ok, __S::Error>
    where
        __S: _serde::Serializer,
    {
        // don't know what this bit does
        match _serde::__private::None::<&_StdIoErrorKindDef> {
            _ => {}
        }
        match _serde::__private::None {
            _serde::__private::Some(()) => {
                let _ = _StdIoErrorKindDef::NotFound;
            }
            _ => {}
        }
        // repeats for other variants

        // then
        match *__self {
            io::ErrorKind::NotFound => _serde::Serializer::serialize_unit_variant(
                __serializer,
                "_StdIoErrorKindDef",
                0u32,
                "NotFound",
            ),
            io::ErrorKind::PermissionDenied => _serde::Serializer::serialize_unit_variant(
                __serializer,
                "_StdIoErrorKindDef",
                1u32,
                "PermissionDenied",
            ),
            // for other variants
        }
    }
}
```
So it looks to me like it is matching against `std::io::ErrorKind`

mrlegohead0x45: Sorry, I misread your comment, so disregard my comment. I think though it would be fine as long as one kept the Def enum up to date with `std::io:ErrorKind`

mrlegohead0x45: Could we make it like this
```rust
#[serde(remote = "io::ErrorKind", wildcard_variant = "__Unrecognised")]
enum _StdIoErrorKindDef {
    // ...
    __Unrecognised
}
```
and then do `_ => _StdIoErrorKindDef::__Unrecognised`, and
```rust
#[serde(remote = "io::ErrorKind", wildcard = "unreachable")]
enum _StdIoErrorKindDef {
    // ...
}
```
for `_ => unreachable!()`, although I don't know how the first one would affect deserialisation.

PR #2367, opened 2023-01-20
Remove `#[doc(hidden)]` from `deserialize_in_place`.

### LINKED ISSUE #2204: Consider unhiding `deserialize_in_place`
I was surprised to discover recently the existence of a hidden `Deserialize::deserialize_in_place` while reading the serde source code. In the docs it's explained:

> This method is stable and an official public API, but hidden from the documentation because it is almost never what newbies are looking for. Showing it in rustdoc would cause it to be featured more prominently than it deserves.

While I can understand the urge to push people away from methods that are likely not what they need, I believe that this swings way too far in the other direction. Just today I [answered](https://stackoverflow.com/a/71961037/864406) a question on stack overflow from someone who was looking for exactly this functionality, and it's not clear to me how they ever would have discovered it, given that it's explicitly omitted from the rustdocs and doesn't appear to be mentioned in any other serde documentation.

Given this, and given that `#[doc(hidden)]` is fairly uniformly understood to be used for removing public items from the crate's public API (eg, for macro implementations, a pattern that serde itself makes extensive use of), I propose that `#[doc(hidden)]` be removed from this method, and it instead be simply documented as "likely not what you need." It could even be hidden instead behind a cargo feature flag to force users to opt into it.

Lucretiel:
As proposed in #2204, this PR removes `#[doc(hidden)]` from `deserialize_in_place`, replacing it with a notice that `deserialize` is usually preferred. That issue contains the full justification for this change.

Fixes #2204

### DIFF
--- a/serde/src/de/mod.rs
+++ b/serde/src/de/mod.rs
@@ -554,17 +554,14 @@ pub trait Deserialize<'de>: Sized {
     /// If you manually implement this, your recursive deserializations should
     /// use `deserialize_in_place`.
     ///
-    /// This method is stable and an official public API, but hidden from the
-    /// documentation because it is almost never what newbies are looking for.
-    /// Showing it in rustdoc would cause it to be featured more prominently
-    /// than it deserves.
-    #[doc(hidden)]
+    /// It is almost always preferable to use [`deserialize`] instead of
+    /// [`deserialize_in_place`]. Make sure to benchmark before using it.
     fn deserialize_in_place<D>(deserializer: D, place: &mut Self) -> Result<(), D::Error>
     where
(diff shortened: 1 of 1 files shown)


Lucretiel:
Note that the alternative I proposed in #2204, which was to hide this method behind a `cfg` flag, is unfortunately impossible without breaking backwards compatibility. The method is currently unconditionally part of the public API, despite its absence from `docs.rs`, which means that feature-gating it would remove it from anyone who's using serde with `no-default-features`.

rkuris:
Although not critical for me now that I know about this, please save the next engineer by committing this change.

PR #2376, opened 2023-02-18
When a serde(untagged) enum can't deser, show all the reasons why

cbeck88:
I was writing some code recently that uses `serde(untagged)` for an enum when deserializing certain JSON schema, but I was unhappy with the quality of the error messages I was getting, because it doesn't really tell you why each possibility doesn't work.

I noticed that there is a TODO item in the `serde_derive` proc macro code that generates this deserialization.

I decided that trying to improve the error messages upstream in serde-derive is easier than trying to change how I'm using serde, so I took a stab at implementing this TODO, and updating the tests that test error messages, and writing some more tests.

I have tried to follow the patterns and conventions that I have seen elsewhere in the serde-derive source code, and I think that the code gen is good in that it uses `format_args!` like other parts of serde error handling and avoids making a new dynamic memory allocation. When untagged deserialization fails, the errors are collected on the stack rather than in a new dynamically-sized container.

Let me know if you think this is a good direction, I'm happy to iterate if this patch is interesting to you. Thanks for building serde, it's great.

### DIFF
--- a/serde_derive/src/de.rs
+++ b/serde_derive/src/de.rs
@@ -1693,9 +1693,9 @@ fn deserialize_untagged_enum(
         num_variants += 1;
      }
 
-     // We need two copies of this iterator, and it's not cloneable
+     // We need two copies of this iterator
      let err_identifiers1 = (0..num_variants).map(|idx| format_ident!("_err{}", idx));
-     let err_identifiers2 = (0..num_variants).map(|idx| format_ident!("_err{}", idx));
+     let err_identifiers2 = err_identifiers1.clone();
 
      quote_block! {
         let __content = try!(<_serde::__private::de::Content as _serde::Deserialize>::deserialize(__deserializer));
--- a/serde_derive/src/de.rs
+++ b/serde_derive/src/de.rs
@@ -1674,28 +1674,44 @@ fn deserialize_untagged_enum(
             ))
         });
 
-    // TODO this message could be better by saving the errors from the failed
-    // attempts. The heuristic used by TOML was to count the number of fields
-    // processed before an error, and use the error that happened after the
-    // largest number of fields. I'm not sure I like that. Maybe it would be
-    // better to save all the errors and combine them into one message that
-    // explains why none of the variants matched.
     let fallthrough_msg = format!(
         "data did not match any variant of untagged enum {}",
         params.type_name()
     );
     let fallthrough_msg = cattrs.expecting().unwrap_or(&fallthrough_msg);
 
-    quote_block! {
+    // If all variants cannot be deserialized, we will try to write an error
+    // message explaining why it failed for each one
+    // The format string we are building will have the following structure:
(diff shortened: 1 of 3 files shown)


appetrosyan:
Why isn't this PR merged?

serde_derive/src/de.rs
```diff
@@ -1674,28 +1674,44 @@ fn deserialize_untagged_enum(
...
+    for var in variants.iter().filter(|variant| !variant.attrs.skip_deserializing()) {
+        err_format_string.push_str("\n");
+        err_format_string.push_str(&var.ident.to_string());
+        err_format_string.push_str(": {}");
+        num_variants += 1;
+     }
+
+     // We need two copies of this iterator
+     let err_identifiers1 = (0..num_variants).map(|idx| format_ident!("_err{}", idx));
+     let err_identifiers2 = err_identifiers1.clone();
```
oli-obk, line 1698: While not strictly necessary, as you could probably do the iteration over the `attempts` directly in the `format_args!` arguments, this seems better to avoid any question about the order of evaluation.

serde_derive/src/de.rs
```diff
@@ -1674,28 +1674,44 @@ fn deserialize_untagged_enum(
...
+    // The format string we are building will have the following structure:
+    // "data did not match any variant of untagged enum{}\nvar1: {}\nvar2: {}\nvar3: {}"
+    let mut err_format_string = fallthrough_msg.to_owned();
+    let mut num_variants = 0usize;
+    for var in variants.iter().filter(|variant| !variant.attrs.skip_deserializing()) {
+        err_format_string.push_str("\n");
+        err_format_string.push_str(&var.ident.to_string());
+        err_format_string.push_str(": {}");
+        num_variants += 1;
+     }
```
oli-obk, line 1694: I'm not confident this will render well in all the error libraries out there, but we can adjust it once we have seen some of these messages in the wild. The Display/Debug output isn't something we provide back compat guarantees for anyway.

oli-obk (approved):
I'll discuss this PR with dtolnay next time we chat. May take a few weeks, but if I don't comment here again in the next month, please ping me.

oli-obk:
We had a chat, and are both worried about the diagnostic's usefulness. It will mostly appear to users of your library that have given bad input, and whether it helps them actually solve the problem is not clear. We feel like while this solves the immediate issue you are seeing, we can do better than just patching a diagnostic.

Thus we have an alternate proposal:

Add new examples to https://github.com/serde-rs/serde-rs.github.io/tree/master/_src that explains how to use untagged enums.
* e.g. via a custom visitor for a custom Deserialize impl to produce a single unified diagnostic instead of a generic line per variant.
* You could also ask https://github.com/jonasbb/serde_with if they'd want to add helpers for automatically creating such visitors
* Or create a helper crate for easily writing visitors (e.g. with a builder pattern)

Thanks for your work. I'm going to close this PR in my attempt to get a hold of our PR queue, but feel free to use this PR to ask follow up questions.

PR #2377, opened 2023-02-18
Allow bytes for adjacently tagged enums

mfro:
This pull request will add support for `Bytes` as the tag and content field name in adjacently tagged enums. This addresses the issue raised here: https://github.com/BurntSushi/rust-csv/issues/278

Code changes are in `TagContentOtherFieldVisitor` as suggested by the investigation in the above issue, and a unit test is added to cover this case. The unit test currently fails on serde/master.

### DIFF
--- a/serde/src/private/de.rs
+++ b/serde/src/private/de.rs
@@ -990,6 +990,19 @@ mod content {
                 Ok(TagContentOtherField::Other)
             }
         }
+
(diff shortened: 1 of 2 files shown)


dtolnay (approved):
Thanks!

dtolnay:
Published in serde 1.0.162.

PR #2379, opened 2023-02-21
Resolve FIXME now that we support Rust 1.19

AtariDreams:
Replace the old code with that that newer Rust supports


dtolnay (changes requested):
Thanks! Could you look into why this makes tests fail?

PR #2387, opened 2023-03-05
Make #[serde(alias)] works in #[serde(flatten)] context

### LINKED ISSUE #1504: Field aliases do not work in combination with `flatten`

bebecue:
fix #1504

### DIFF
--- a/test_suite/tests/test_annotations.rs
+++ b/test_suite/tests/test_annotations.rs
@@ -2659,6 +2659,68 @@ fn test_flatten_any_after_flatten_struct() {
 }
 
 #[test]
+fn test_alias_in_flatten_context() {
+    #[derive(Debug, PartialEq, Deserialize)]
+    struct Outer {
+        #[serde(flatten)]
+        a: AliasStruct,
+        b: i32,
+    }
+
+    assert_de_tokens(
+        &Outer {
+            a: AliasStruct {
+                a1: 1,
+                a2: 2,
+                a4: 4,
+            },
+            b: 7,
+        },
+        &[
+            Token::Struct {
+                name: "Outer",
+                len: 4,
+            },
+            Token::Str("a1"),
+            Token::I32(1),
+            Token::Str("a2"),
+            Token::I32(2),
+            Token::Str("a5"),
+            Token::I32(4),
+            Token::Str("b"),
+            Token::I32(7),
+            Token::StructEnd,
+        ],
+    );
+
+    assert_de_tokens(
+        &Outer {
+            a: AliasStruct {
+                a1: 1,
+                a2: 2,
+                a4: 4,
+            },
+            b: 7,
+        },
+        &[
+            Token::Struct {
+                name: "Outer",
+                len: 4,
+            },
+            Token::Str("a1"),
+            Token::I32(1),
+            Token::Str("a2"),
+            Token::I32(2),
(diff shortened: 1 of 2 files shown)


oli-obk:
Please also add a few tests for a struct like the one from the issue.

test_suite/tests/test_annotations.rs
```diff
@@ -780,7 +780,7 @@ fn test_rename_enum() {
             Token::StructVariant {
                 name: "AliasEnum",
                 variant: "sailor_moon",
-                len: 3,
```
jonasbb, line 783: This seems wrong. The struct still only has a length of 3,since three fields are emitted. Not 5 just because there are five field names and aliases.

dtolnay: In the serde data model, the struct definitely has 5 fields (some of which are optional, some subsets of which are mutually exclusive).

The exact number of Rust fields that the deserialized Rust representation ends up storing doesn't seem relevant. If we wrote a Deserialize impl that stored none of them, it doesn't make a difference to the Deserializer. Or compare with how `#[serde(transparent)] struct Struct(i32);` deserializes using deserialize_i32, not deserialize_newtype_struct, even with the deserialized Rust representation `Struct` being a struct not an int -- that isn't relevant.

bebecue:
> Please also add a few tests for a struct like the one from the issue.

Added in https://github.com/serde-rs/serde/pull/2387/commits/01da3f79c9a41220174e8b7e89c3e836325e97ff

oli-obk (approved):

dtolnay (commented):
Thanks!

In https://github.com/serde-rs/serde/pull/2160#pullrequestreview-1143343662 I was hopeful that somebody would look into what possibility of breakage this has for existing formats, but in the absence of anyone having been willing to do that, I think it will be all right to reassess based on any reported consequences.

janik-cloudflare:
Thanks a ton @dtolnay! Just ran into this and it's fantastic to see that a fix is already out. Really appreciate your work on serde :)

banool:
> I was hopeful that somebody would look into what possibility of breakage this has for existing formats, but in the absence of anyone having been willing to do that, I think it will be all right to reassess based on any reported consequences.

I know I'm well and truly late to the party hahah but this change did break some of our code.

At a high level you can see the issue like this:
```
git clone git@github.com:aptos-labs/aptos-core.git
# Change serde version in Cargo.toml to 1.0.153
cd testsuite/generate-format
cargo run -- --corpus aptos
```

With 1.0.152 this code works without error.

I tried the latest version of serde (1.0.188 at the time of writing) and the issue seems to persist.

I'm still trying to figure out how this change actually breaks our code, I just wanted to report in with an instance of a breakage.

banool:
btw I looked into some issues in this repo. I know serde complies with semver but I also know that that only applies to public APIs. I'm not 100% sure if this qualifies as a breaking change. I'll try figure that out and open a new issue if so.

banool:
Moving to this issue now: https://github.com/serde-rs/serde/issues/2627.

PR #2400, opened 2023-03-14
Use explicit re-export of `serde_derive` to give rustc more info

Nilstrieb:
rustc will start looking behind `#[cfg(FALSE)]` items to start giving better diagnostics. By using an explicit re-export instead of a glob export, we tell rustc that `Deserialize` and `Serialize` exist here.

I'm not sure whether it's worth it to open this PR before upstream has been merged, but since this adds minimal maintenance effort and may even make the code nicer I opened it already, feel free to wait if you prefer.

Upstream PR: rust-lang/rust#109005


dtolnay (approved):
Thanks!

PR #2403, opened 2023-03-16
Allowed Enum variants to be individually marked as untagged

### LINKED ISSUE #1402: #[serde(flatten)] attribute support for enum variants
Hi!

I have the following use-case for a feature similar to `#[serde(flatten)]`, but for enum variants rather than struct fields. Generally happy to dive into `serde_derive` myself in case you think this would be worth having, but it could also be that I'm trying to model this the wrong way.

I'm trying to parse JSON values in the [estree format](https://github.com/estree/estree). The format specifies an ECMAScript syntax tree using an inhertance relationship, where generally only the nodes at the bottom of the hierarchy are tagged.

For example, one type of node is `Expression`, but there doesn't exist any node with the "Expression" tag, because the tag is assigned to the "concrete" expression nodes such as `Identifier` or `Literal`. The representation of the `Expression` node is basically just a union of all possible JSON representations of its descendant node types.

So far, I've been modelling it like this, which has been working fine for the most part:

```rust
#[derive(Deserialize)]
#[serde(tag = "type")]
pub enum Expression {
    Identifier { name: String },
    Literal { value: Value },
    ...
}
```

Now, the difficulty I've been having comes in when I try to mix types that are at the bottom of the node hierarchy with node types that have sub-types (such as `Expression`):

```rust
#[derive(Deserialize)]
#[serde(tag = "type")]
pub enum ForStatementInit {
    VariableDeclaration { kind: String, ... },
    Expression(Expression),
}
```

In this case, what I really want is for the `ForStatementInit` type to be the result of "adding" the node of type `VariableDeclaration` to the set of representations for `Expression`. What ends up happening though of course is that it expects a value tagged with either `VariableDeclaration` or `Expression`.

I can work around this by either nesting `VariableDeclaration` into a single-variant enum, and using `#[serde(untagged)]` on the `ForStatementInit` enum, or by manually "flattening" the expression types, like so:

```rust
#[derive(Deserialize)]
#[serde(tag = "type")]
pub enum ForStatementInit {
    VariableDeclaration { kind: String, ... },
    // Expression
    Identifier { name: String },
    Literal { value: Value },
}
```

However, what I think would be really useful is to be able to do something like this instead, which would apply the same "flattening" transformation in the example above:

```rust
#[derive(Deserialize)]
#[serde(tag = "type")]
pub enum ForStatementInit {
    VariableDeclaration { kind: String, ... },
    #[serde(flatten)]
    Expression(Expression),
}
```

Hope this makes somewhat sense, keen to hear if this is something that you think is worth solving in serde itself!

dewert99:
This should hopefully resolve #1402.

This can be thought of as applying the following transformation
```rust
#[derive(Serialize, Deserialize)
enum Test {
  A(u32, u32),
  B(u32),
  #[serde(untagged)]
  C(u32),
  #[serde(untagged)]
  D(u32, u32),
}
```
to
```rust
#[derive(Serialize, Deserialize)
enum TestTagged {
  A(u32, u32),
  B(u32),
}

#[derive(Serialize, Deserialize)
#[serde(untagged)]
enum TestUntagged {
  Tagged(TestTagged),
  C(u32),
  D(u32, u32)
}

#[derive(Serialize, Deserialize)
#[serde(from = "TestUntagged")]
#[serde(into = "TestUntagged")]
enum Test {
  ...
}
```
The actual implementation generates the implementation directly allowing it to also work efficiently for recursive enums.

The current version includes a restriction that all variants marked as `untagged` appear at the end of the enum (after the tagged variants) in order to keep the implementation strategy consistent with an alternate strategy of attempting to deserialize each variant in order.

Deserialize could also be implemented to accept all of the variants with tags, only falling back to untagged if none of them match, I'm not sure if this would be preferred.

### DIFF
--- a/serde_derive/src/de.rs
+++ b/serde_derive/src/de.rs
@@ -1166,6 +1166,29 @@ fn deserialize_enum(
     variants: &[Variant],
     cattrs: &attr::Container,
 ) -> Fragment {
+    match variants
+        .iter()
+        .enumerate()
+        .find(|(_, var)| var.attrs.untagged())
+    {
+        Some((variant_idx, _)) => {
+            let (tagged, untagged) = variants.split_at(variant_idx);
+            let tagged_frag = Expr(deserialize_homogenious_enum(params, tagged, cattrs));
+            let tagged_frag = |deserializer| Expr(quote_block! {
+                let __deserializer = #deserializer;
+                #tagged_frag
+            });
+            deserialize_untagged_enum_after(params, untagged, cattrs, Some(tagged_frag))
+        }
+        None => deserialize_homogenious_enum(params, variants, cattrs),
+    }
+}
+
+fn deserialize_homogenious_enum(
+    params: &Parameters,
+    variants: &[Variant],
+    cattrs: &attr::Container,
+) -> Fragment {
     match cattrs.tag() {
         attr::TagType::External => deserialize_externally_tagged_enum(params, variants, cattrs),
         attr::TagType::Internal { tag } => {
@@ -1661,6 +1684,18 @@ fn deserialize_untagged_enum(
     variants: &[Variant],
     cattrs: &attr::Container,
 ) -> Fragment {
+    deserialize_untagged_enum_after(params, variants, cattrs, None::<fn(_) -> _>)
+}
+
+fn deserialize_untagged_enum_after(
+    params: &Parameters,
+    variants: &[Variant],
+    cattrs: &attr::Container,
+    first_attempt: Option<impl FnOnce(TokenStream) -> Expr>,
+) -> Fragment {
+    let deserializer = || quote!(
+        _serde::__private::de::ContentRefDeserializer::<__D::Error>::new(&__content)
+    );
     let attempts = variants
         .iter()
         .filter(|variant| !variant.attrs.skip_deserializing())
@@ -1669,12 +1704,10 @@ fn deserialize_untagged_enum(
                 params,
                 variant,
                 cattrs,
-                quote!(
-                    _serde::__private::de::ContentRefDeserializer::<__D::Error>::new(&__content)
-                ),
+                deserializer(),
             ))
         });
-
+    let attempts = first_attempt.map(|f| f(deserializer())).into_iter().chain(attempts);
     // TODO this message could be better by saving the errors from the failed
     // attempts. The heuristic used by TOML was to count the number of fields
     // processed before an error, and use the error that happened after the
--- a/serde_derive/src/de.rs
+++ b/serde_derive/src/de.rs
@@ -1176,12 +1176,12 @@ fn deserialize_enum(
             let (tagged, untagged) = variants.split_at(variant_idx);
             let tagged_frag = Expr(deserialize_homogeneous_enum(params, tagged, cattrs));
             let tagged_frag = |deserializer| {
-                Expr(quote_block! {
+                Some(Expr(quote_block! {
                     let __deserializer = #deserializer;
                     #tagged_frag
-                })
+                }))
             };
-            deserialize_untagged_enum_after(params, untagged, cattrs, Some(tagged_frag))
+            deserialize_untagged_enum_after(params, untagged, cattrs, tagged_frag)
         }
         None => deserialize_homogeneous_enum(params, variants, cattrs),
     }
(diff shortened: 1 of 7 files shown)


oli-obk (commented):
Please also add some tests for non-recursive examples like in https://github.com/serde-rs/serde/issues/1402#issuecomment-435197821 or generic examples like https://github.com/serde-rs/serde/issues/1402#issuecomment-435444299

test_suite/tests/test_annotations.rs
```diff
@@ -2395,6 +2395,33 @@ fn test_untagged_enum_containing_flatten() {
...
+fn test_partially_untagged_enum() {
+    #[derive(Serialize, Deserialize, PartialEq, Debug)]
+    enum Exp {
+        Lambda(u32, Box<Exp>),
+        #[serde(untagged)]
+        App(Box<Exp>, Box<Exp>),
+        #[serde(untagged)]
+        Var(u32),
+    }
+    use Exp::*;
```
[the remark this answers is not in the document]
dewert99: That would be difficult for this specific test since it's recursive, but I did this for a non-recursive example.

dewert99:
Thanks for all the feedback. I've tried to implement the suggestions.

serde_derive/src/de.rs
```diff
@@ -1669,12 +1706,12 @@ fn deserialize_untagged_enum(
...
                 variant,
                 cattrs,
-                quote!(
-                    _serde::__private::de::ContentRefDeserializer::<__D::Error>::new(&__content)
-                ),
+                deserializer.clone(),
             ))
         });
-
+    let attempts = first_attempt(deserializer.clone())
```
oli-obk, line 1712:
```suggestion
    let attempts = first_attempt(deserializer)
```

the last use can consume it.

oli-obk (changes requested):
one thing I just realized that we may want to reject is marking all enum variants as untagged, as then you could instead mark the entire enum as untagged.

lgtm otherwise. I'll ping a few folk from the issue so they can try out your PR and see if it works for their use cases.

serde_derive/src/de.rs
dewert99: I tried this, but `deserializer` is actually captured by the closure in `map` producing an iterator over the rest of the attempts.

dewert99:
I wasn't sure what the policy was on warning-like cases. Other cases to consider might be:
- marking a variant as `untagged` when the enum itself is already `untagged`
- marking a variant with `rename`/`alias` as well as `untagged` (I don't think this has a different effect than just using `untagged`)
- marking a variant with `rename`/`alias` when the enum is `untagged` (This is currently allowed, so changing it would be a breaking change)

serde_derive/src/internals/ast.rs
```diff
@@ -153,8 +154,12 @@ fn enum_from_ast<'a>(
                 fields,
                 original: variant,
             }
-        })
-        .collect()
+        }).inspect(|variant| {
+            if !variant.attrs.untagged() && seen_untagged {
+                cx.error_spanned_by(&variant.ident, "all variants with the #[serde(untagged)] attribute must be placed at the end of the enum")
```
gavrilikhin-d, line 159: Why only allowing `#[serde(untagged)]` to be at the end?

gavrilikhin-d:
@oli-obk Hi! This feature is really usefull, but sadly there is no activity on PR more than a month. The PR itself looks quite finished except warnings. Can it be merged in this state?

serde_derive/src/internals/ast.rs
dewert99: This keeps the implementation more consistent with the deserialisation strategy of going though each variant in order and choosing the first variant that matches.

oli-obk:
Unfortunately no one on the issue has so far tested out this branch to see if it works for their use cases.

gavrilikhin-d:
@oli-obk I tried to check it, but don't know how to replace serde dependency in serde_json

oli-obk:
Someone was so kind and wrote instructions at https://github.com/serde-rs/serde/issues/1402#issuecomment-1565751691

gavrilikhin-d:
@oli-obk My case was covered too. https://github.com/serde-rs/serde/issues/1402#issuecomment-1574550592

dtolnay (approved):
Thanks!

PR #2420, opened 2023-03-27
Add `std::any::Provider` support to `Serializer`

bryanburgers:
Turn a `Serializer` into a [`std::any::Provider`] in a backward compatible way.

This allows specialization between a `Serializer` and a `Serialize` in a generic way.

The one particular use case I have in mind is for [`serde_dynamo`]. DynamoDB supports lists of generic items – which is what `serde_dynamo` currently uses to serialize e.g. a `Vec<String>`. However, DynamoDB also supports specialized types like [String Set] that it would be nice to be able to opt in to.

Using [`std::any::Provider`], this could look something like this:

```rust
/// A specialized serializer
///
/// This would most likely be used with `#[serde(with = "string_set")]`
struct StringSet<T>(T);

impl<T> Serialize for StringSet<T> where T: Serialize {
    where
        S: serde::Serializer,
    {
        if let Some(string_set_serializer) = std::any::request_value::<serde_dynamo::StringSetSerializer>(&serializer.provider()) {
            self.0.serialize(string_set_serializer)
        } else {
            self.0.serialize(serializer)
        }
    }
}

/// In user code, a struct that wants `strings` to be serialized
/// using a DynamoDb string set *if* the serializer is
/// `serde_dynamo`
#[derive(Debug, Serialize, Deserialize)]
struct Subject {
    strings: StringSet<Vec<String>>,
}
```

With this set up, `Subject.strings` would be serialized as normal using `serde_json`

```json
{
    "strings": ["one", "two"]
}
```

but when using `serde_dynamo`, `Subject.strings` would use the specific `StringSetSerializer` to get custom behavior.

This is a very specific example where `serde_dynamo` would likely provide both the `StringSet` wrapper and the `StringSetSerializer`.

There *may* also be situations where the ecosystem would develop common practices. One example is the `Serializer::is_human_readable` function. If this existed before that was introduced, it's *possible* the serializers could have provided `struct IsHumanReadable(bool)` and serializers could have used that to determine how to display timestamps. It's possible that this alleviates some of the need for https://github.com/serde-rs/serde/pull/455, but I haven't looked at every use case there.

Links regarding [`std::any::Provider`]:

* `std::any::Provider` RFC: https://rust-lang.github.io/rfcs/3192-dyno.html
* `std::any::Provider` tracking issue: https://github.com/rust-lang/rust/issues/96024

Implementation notes:

* ~I opted for getting this out for discussion and review quickly, which means I didn't spend too much time on documentation and zero time on tests. If this is an approach the team is interested in, I'll be happy to revisit those.~ I've added doc tests to `provide` and `provider`.
* ~There doesn't necessarily need to be a `Serializer::provider` convenience method. We could use documentation to push implementors to use `SerializerProvider::wrap(&serializer)` when they need a provider.~ Based on doc tests and some experimenting, I think having the default `provider` method is valuable.
* ~This is most useful to my use case on `Serializer`, but I wonder if it also would be useful on the `Serialize*` sub-traits of `Serializer`, and/or on `Deserialize` traits.~ I've added this to `Deserialize` as well.

[`std::any::Provider`]: https://doc.rust-lang.org/nightly/core/any/trait.Provider.html
[`serde_dynamo`]: https://docs.rs/serde_dynamo
[String Set]: https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.NamingRulesDataTypes.html

### DIFF
--- a/serde/src/de/mod.rs
+++ b/serde/src/de/mod.rs
@@ -1220,6 +1220,93 @@ pub trait Deserializer<'de>: Sized {
         true
     }
 
+    /// Provide generic, type-driven access to data tied to this deserializer, using the machinery
+    /// in [`std::any::Provider`] and [`std::any::Demand`].
+    ///
+    /// ```edition2018
+    /// # #![feature(provide_any)]
+    /// use serde::{Deserialize, Deserializer, de::Visitor, forward_to_deserialize_any};
+    /// # use serde::__private::doc::Error;
+    ///
+    /// /// An enum shared between both the Deserializer and Deserialize that communicates which
+    /// /// distance unit is used when deserializing.
+    /// #[derive(Copy, Clone)]
+    /// enum PreferredUnit {
+    ///     Millimeters,
+    ///     Meters,
+    /// }
+    ///
+    /// /// A deserializer that knows which distance unit is used
+    /// struct MyDeserializer {
+    ///     preferred_unit: PreferredUnit,
+    /// }
+    ///
+    /// impl<'de> Deserializer<'de> for MyDeserializer {
+    ///     type Error = Error;
+    ///
+    ///     fn deserialize_any<V>(self, visitor: V) -> Result<V::Value, Self::Error>
+    ///     where V: Visitor<'de>,
+    ///     {
+    ///         unimplemented!()
+    ///     }
+    ///
+    ///     forward_to_deserialize_any! {
+    ///         bool i8 i16 i32 i64 i128 u8 u16 u32 u64 u128 f32 f64 char str string
+    ///         bytes byte_buf option unit unit_struct newtype_struct seq tuple
+    ///         tuple_struct map struct enum identifier ignored_any
+    ///     }
+    ///
+    ///     fn provide<'a>(&'a self, demand: &mut std::any::Demand<'a>) {
+    ///         // If asked about this serializer's preferred distance unit, provide it!
+    ///         demand.provide_value(self.preferred_unit);
+    ///     }
+    /// }
+    ///
+    /// /// A distance that can be deserialized based on distance unit preference
+    /// pub struct Distance {
+    ///     millimeters: f64,
+    /// }
+    ///
+    /// impl Distance {
+    ///     pub fn from_millimeters(millimeters: f64) -> Self { Self { millimeters } }
+    ///     pub fn from_meters(meters: f64) -> Self { Self { millimeters: meters * 1000.0 } }
+    /// }
+    ///
+    /// impl<'de> Deserialize<'de> for Distance {
+    ///     fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    ///     where
+    ///         D: Deserializer<'de>,
+    ///     {
+    ///         let provider = ProviderAdapter(&deserializer);
+    ///         if let Some(preferred_unit) = std::any::request_value::<PreferredUnit>(&provider) {
+    ///             // If the deserializer knows what distance unit was used to serialize
+    ///             let value = f64::deserialize(deserializer)?;
+    ///             match preferred_unit {
+    ///                 PreferredUnit::Millimeters => Ok(Distance::from_millimeters(value)),
+    ///                 PreferredUnit::Meters => Ok(Distance::from_meters(value)),
+    ///             }
+    ///         } else {
+    ///             // Assume millimeters
+    ///             let value = f64::deserialize(deserializer)?;
+    ///             Ok(Distance::from_millimeters(value))
+    ///         }
+    ///     }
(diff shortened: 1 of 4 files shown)


RReverser:
This would be extremely useful for things like serde_json::RawValue and similar!

RReverser:
Although it would need to apply to Deserialize as well.

bryanburgers:
I can certainly add it for Deserialize, too!

RReverser:
FWIW I'm just a drive-by reader, maintainer of Serde, although I could very much make use of such feature if it ever lands to avoid hacks for communication between deserializer & custom type.

oli-obk:
I don't know about the timeline on std::any::Provider stabilization, but if/when it lands this would indeed be a great solution to the serializer-specific specialization problem.

> I think having the default provider method is valuable.

While it is valuable for everyone who needs it, it also adds a just-for-convenience method to an already extremely big trait. Afaict even the `SerializerProvider` type is not needed in `serde` itself, right?

bryanburgers:
> I don't know about the timeline on std::any::Provider stabilization, but if/when it lands this would indeed be a great solution to the serializer-specific specialization problem.

🥳

Yeah, I'm aware that `std::any::Provider` isn't stabilized, and I'm in no hurry.
 
> > I think having the default provider method is valuable.
> 
> While it is valuable for everyone who needs it, it also adds a just-for-convenience method to an already extremely big trait. Afaict even the `SerializerProvider` type is not needed in `serde` itself, right?

I suppose there's a balance here between smaller API surface and more documentation burden.

* `Serializer::provider` method + `SerializerProvider`
  * Adds an unnecessary method to the trait
  * Documentation says "Use `serializer.provider()`"
* `SerializerProvider` only
  * No unnecessary method on the trait
  * Documentation says "Use `SerializerProvider::wrap(&serializer)`" and users need to know about `SerializerProvider`
* Neither
  * No unnecessary method on the trait
  * No extra `SerializerProvider` and `DeserializerProvider` structs
  * Documentation has at least one example of creating your own Provider adapter and then using it

`serde` is a huge part of the Rust ecosystem, so I'm going to defer to you if we fall at different points on that balance.

I'll be glad to take that stuff out and update the documentation.

Mingun:
This would be very useful for deserialization to add an ability to provide spans of deserialized objects to the `Deserialize` implementation aware of spans. Unfortunately, deserializer is consumed during deserialization so providing safe interface and implementation could be impossible (it could not be possible to ask deserializer of spans after deserialization because it is consumed).

I was able to implement such a concept by extending `Deserializer` trait with safe interface, but with `unsafe` implementation, a sketch:
```rust
trait PositionProvider {
  type Position;

  fn current_position() -> Self::Position;
}
trait Deserializer<'de> {
  type PositionProvider: PositionProvider;

  fn position_provider(&self) -> Self::PositionProvider;
  ...
}

// impl
impl<'de, 'a> Deserializer<'de> for &'a mut MyDeserializer {
  type PositionProvider = MyPositionProvider;

  fn position_provider(&self) -> Self::PositionProvider {
    MyPositionProvider(self as _)
  }
}

// Cannot use reference here because of borrow checker
// and that fact that deserializer is consumed during deserialization
struct MyPositionProvider(*const MyDeserializer);
impl PositionProvider for MyPositionProvider {
  type Position = usize;

  fn current_position(&self) -> usize {
    let de: &MyDeserializer = unsafe { &*self.0 };
    de.current_position();
  }
}
```

Then any type can be wrapped into
```rust
struct Spanned<T> {
  value: T,
  span: Range<usize>,
}
impl Deserialize<'de> for Spanned<T>
where
  T: Deserialize<'de>,
{
  fn deserialize<D>(deserializer: D) -> Result<Self, D::Error> {
    let provider = deserializer.position_provider();
    let start = provider.current_position();
    let value = T::deserialize(deserializer)?;
    let end = provider.current_position();

    Ok(Self { value, span: (start..end).into() })
  }
}
```

This trick is only worked for deserializers that implements `Deserializer` for a reference to the deserializer struct (but it seems most of the deserializers implemented like that)

RReverser:
@Mingun You could just make `position_provider` return an `Rc<Cell<...>>` or something. It doesn't have to hold a reference to the Deserializer - instead, Deserializer could own position provider and hand out its `Rc`-clones to anyone who wants them, making the code safe.

RReverser:
Looks like the `Provider` feature has been removed from nightly Rust in favour of just extending the `Error` trait, so this PR is sadly no longer viable.

Mingun:
Actually, the `Provider` trait was needed only to use `std::any::provide_value` helper method. It is still possible to implement that method in serde for use with `Serializer` / `Deserializer`.

Issue #2415, opened 2023-03-22
Possible semver breaking change in serde 1.0.157


MidasLamb:
When pinning the version of `serde` in Cargo.toml to `1.0.156`, my crate compiles, switching to `1.0.157` results in an error in my `influxdb2` dependency:
```bash
error: unexpected end of input, unexpected token in nested attribute, expected ident
  --> /home/midas/.cargo/registry/src/github.com-1ecc6299db9ec823/influxdb2-0.3.5/src/models/ast/dialect.rs:16:13
   |
16 |     #[serde()]
   |             ^

   Compiling opentelemetry-otlp v0.11.0
error[E0277]: the trait bound `Dialect: buckets::_::_serde::Serialize` is not satisfied
    --> /home/midas/.cargo/registry/src/github.com-1ecc6299db9ec823/influxdb2-0.3.5/src/models/query.rs:10:35
     |
10   | #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
     |                                   ^^^^^^^^^ the trait `buckets::_::_serde::Serialize` is not implemented for `Dialect`
...
20   |     /// Dialect
     |     ----------- required by a bound introduced by this call
     |
     = help: the following other types implement trait `buckets::_::_serde::Serialize`:
               &'a T
               &'a mut T
               ()
               (T0, T1)
               (T0, T1, T2)
               (T0, T1, T2, T3)
               (T0, T1, T2, T3, T4)
               (T0, T1, T2, T3, T4, T5)
             and 208 others
     = note: required for `std::option::Option<Dialect>` to implement `buckets::_::_serde::Serialize`
note: required by a bound in `buckets::_::_serde::ser::SerializeStruct::serialize_field`
    --> /home/midas/.cargo/registry/src/github.com-1ecc6299db9ec823/serde-1.0.157/src/ser/mod.rs:1901:12
     |
1901 |         T: Serialize;
     |            ^^^^^^^^^ required by this bound in `buckets::_::_serde::ser::SerializeStruct::serialize_field`

error[E0277]: the trait bound `Dialect: buckets::_::_serde::Deserialize<'_>` is not satisfied
    --> /home/midas/.cargo/registry/src/github.com-1ecc6299db9ec823/influxdb2-0.3.5/src/models/query.rs:22:18
     |
22   |     pub dialect: Option<crate::models::ast::Dialect>,
[...]

PR #2421, opened 2023-03-28
test: test for empty serde args

CGQAQ:
# Potential regression

Today I came acrossed with this error:
image

And I found [it uses `#[serde()]`][l1] all over the place, and I also found [Possible semver breaking change in serde 1.0.157][l2], , is this the expected behavier? if so, I added a test for it, if not, I will change the test and fix it

ref: oxidecomputer/progenitor/issues/394

[l1]: https://github.com/oxidecomputer/third-party-api-clients/blob/main/github/src/types.rs#L257
[l2]: https://github.com/serde-rs/serde/issues/2415

### DIFF
--- a/test_suite/tests/ui/unexpected-literal/empty_args.rs
+++ b/test_suite/tests/ui/unexpected-literal/empty_args.rs
@@ -1,0 +1,7 @@
+use serde_derive::Serialize;
+
+#[derive(Serialize)]
+#[serde()]
+struct S;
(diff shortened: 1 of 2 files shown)


dtolnay (commented):
It was not an intended change. I have put up a fix in #2422.

If it's simple for someone to remove these useless attributes though, that may be a good idea regardless of the fix.

PR #2425, opened 2023-04-07
Implement serialization and deserialization of 'core::convert::Infallible'

DanikVitek:
As the usage of analogous implementations for `!` (`never` type) is currently unstable, I suppose, it would be good to have these implementations for the `Infallible` type, as it could be pretty helpful sometimes, but own `Infallible` struct is boilerplate.

For example, consider the app, written using the Tauri framework. It needs to have a command, that returns a plain value, but for a reason, it is not possible to create such a command if it is async and has references as arguments. The only way currently is to return a `Result` type, but if the action is infallible, it may be redundant. Currently, the only way to make this happen is to put as an error type `()` or own implementation if `Infallible`

```rust
#[tauri::command]
pub async fn is_logged_in(login_info: State<'_, Mutex<Option<LoginInfo>>>) -> Result<bool, ()> {
    Ok(login_info.lock().await.is_some())
}
```

I think there may be more examples like this. Storing Results in logs or something.

### DIFF
--- a/serde/src/de/impls.rs
+++ b/serde/src/de/impls.rs
@@ -428,6 +428,17 @@ serde_if_integer128! {
 
 ////////////////////////////////////////////////////////////////////////////////
 
+impl<'de> Deserialize<'de> for core::convert::Infallible {
+    fn deserialize<D>(_: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        Err(D::Error::custom("cannot deserialize `Infallible`"))
+    }
+}
+
+////////////////////////////////////////////////////////////////////////////////
+
 struct CharVisitor;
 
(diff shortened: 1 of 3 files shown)


dtolnay (commented):
I think error types in general are not expected to be serializable. For example serde does not provide (or plan to provide) trait impls for std::io::Error, std::fmt::Error, std::num::ParseIntError, etc, and other libraries also do not typically provide serde impls for their errors, not even serde_json::Error, or something like anyhow::Error.

The standard library's intention for `Infallable` is limited to being an error type for infallible TryFrom conversions, such as TryFrom&lt;&amp;str&gt; for String.

So I don't think I would want to provide these two impls.

The use case for `!` is more general than just an error type, so that one probably still makes sense to support in serde, but of course it's unstable for now.

Thanks anyway for the PR!

PR #2426, opened 2023-04-11
Make derived serializer/deserializer internals `doc(hidden)`

compiler-errors:
This adds a handful of `#[doc(hidden)]` attrs to make sure we don't document all of the derive serialization/deserialization internals when we use `--document-private-items`, since rustdoc now documents things inside of `const _: () = { /* ... */ };` blocks due to rust-lang/rust#104889 (which was reverted in rust-lang/rust#107083, and later re-applied in [rust-lang/rust#107000](https://github.com/rust-lang/rust/pull/107000#issuecomment-1405407485) -- so the behavior is only present on 1.69 and above).

Before:

image

After:

image


dtolnay (approved):
Thanks!

PR #2427, opened 2023-04-15
Ignore deprecation warnings for derived `Serialize`/`Deserialize` impls

MrGunflame:
Fixes the depreciation warnings when deriving `Serialize` or `Deserialize` on a depreciated item.

For example the following code currently emits a bunch of warnings, which come from the derived `Serialize`/`Deserialize` impls:
```rust
#[derive(Debug, Serialize, Deserialize)]
#[deprecated]
struct UserId(u64);
```

Note that the `Debug` does not emit any warning.

### DIFF
--- a/serde_derive/src/de.rs
+++ b/serde_derive/src/de.rs
@@ -54,6 +54,7 @@ pub fn expand_derive_deserialize(
 
         quote! {
             #[automatically_derived]
+            #[allow(deprecated)]
             impl #de_impl_generics #serde::Deserialize<#delife> for #ident #ty_generics #where_clause {
                 fn deserialize<__D>(__deserializer: __D) -> #serde::__private::Result<Self, __D::Error>
                 where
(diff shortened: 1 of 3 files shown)


dtolnay (commented):
This also would suppress deprecation warning in the following case though, right?

```rust
use serde::{Deserialize, Serialize, Serializer};

pub struct Struct {
    #[serde(serialize_with = "deprecated_serializer")]
    thing: i32,
}

#[deprecated]
fn deprecated_serializer<S>(_thing: &i32, _serializer: S) -> Result<S::Ok, S::Error>
where
    S: Serializer,
{
    unimplemented!()
}
```

If so, that is not desirable. Could you come up with any other way to handle the issue?

MrGunflame:
It would be possible to revert to `#[warn(deprecated)]` after the impl statement. That would of overwrite any previous `allow/deny` lints. Aside from that I'm not aware of any other way to only affect the impl statement.

dtolnay:
All right -- I would prefer to get this fixed in rustc then. A data structure marked `deprecated` should not be treated as deprecated inside of code generated by a `derive` macro on that same data structure. Could you check whether an issue exists already and otherwise file one?

MrGunflame:
Yep, I opened an issue here: https://github.com/rust-lang/rust/issues/110374

Issue #1584, opened 2019-07-21
Unable to fetch struct name if it contains flattened fields


luojia65:
Recently I wrote a serializer with serde for a given format. There is one thing different in my format is that it requires a name for every formatted files; or if regarded as trees, every tree root must be named. I chose to fetch the struct name for root name, but there is one condition where I'm unable to fetch them. Consider this:
```rust
#[derive(Serialize)]
#[serde(rename = "book")] // <- the struct is renamed here
struct Book {
    resolved: Option<i8>,
    #[serde(flatten)]  // <- serde flatten attribute 
    extra: Option<Extra>,
}

#[derive(Serialize)]
struct Extra {
    generation: i32,
    author: &'static str,
    title: &'static str,
}
``` 
When the root struct `Book` contains a flattened struct, I expect the root struct would still be regarded as a struct. But (if I am right) it is actually procceeded as a map instead of a struct. Meanwhile in my serializer, the `serialize_map` function holds this signature as `serde::Serializer` trait requires:
```rust
fn serialize_map(self, _len: Option<usize>) -> Result<Self::SerializeMap>
```
which (unlike `serialize_struct`) does not have a `name` field, meaning the function is unable to capture the root struct name. That means, if I want to use `#[serde(flatten)]` attribute, I have to think of another way to pass the root name; if I need to use the struct name directly as root name, I have no way to use serde flatten attribute which could be useful in my applications.

Are there any solution for this situation?

To reproduce this bug, I put my code into [this](https://github.com/luojia65/coruscant) repository. Run `cargo run --example nbt-option-flatten` command in the cloned project root, the actual output contains:
```
Compound ''
  Byte 'resolved' 1
  Int 'generation' 0
  String 'author' luojia65
  String 'title' hello
EndCompound
```
in the first line, the root name follows default as an empty string, instead of `Compound 'book'` as I expected. I changed the type of `extra` from `Option<Extra>` to `Extra`, but the output is same with an empty root name.

---

Edit: I found a serde code snip [here](https://github.com/serde-rs/serde/blob/e9cd73f78e4d8dc03189ef11afa28df3f4f7000b/serde_derive/src/ser.rs#L292). Should this struct-to-map be a feature or a bug?

luojia65:
Maybe produced by this change: https://github.com/serde-rs/serde/commit/58d52e784b1a6eb71bbd05695fa97b24638ec094 . Is this a feature or a bug? @mitsuhiko

[...]

Issue #1346, opened 2018-08-05
Structs with flattened fields are serialized and deserialized as maps


torkleyy:
This causes problems for formats that use different syntax for structs and maps. Is there some way to fix it?

dtolnay:
Rust structs with flattened fields are not "structs" by Serde's definition of "struct". A struct in Serde means that the fields are known at compile time. This is enforced for example by [`Deserializer::deserialize_struct`](https://docs.serde.rs/serde/de/trait.Deserializer.html#tymethod.deserialize_struct) which requires a `&'static [&'static str]` of the field names. But in a Rust struct with flattened fields, the Serialize impl of the flattened type can potentially insert any fields it wants at runtime.

RON could work around this by allowing deserializing Serde maps from RON structs, or I would be willing to consider a PR to support a compile-time declared set of what fields the flattened type is allowed to insert:

```rust
#[derive(Deserialize, Serialize)]
struct Main {
    #[serde(flatten(first, second))]
    required: Required,
    #[serde(flatten(third))]
    optional: Optional,
    some_other_field: u32,
}
```

dtolnay:
Right, this would be for flattening types that have a compile-time knowable set of fields.

RReverser:
Just ran into the same issue with [serdebug](https://github.com/RReverser/serdebug).

@dtolnay I see there is already `deserialize_identifier`. Could we just add a counterpart for it in `Serializer` and for flattened structs continue to use `serialize_map` but use that new `serialize_identifier` for keys? I imagine that would be easier to use than enumerating flattened fields by hand.

RReverser:
Ideally, I would also add `Token::Ident` to distinguish these from regular strings, but that would be a breaking change, so I guess not an option.

dtolnay:
Considering this issue again, I would prefer to simply leave this as a known issue affecting a subset of data formats. There are many different choices for how a flatten attribute could be implemented, the most powerful of which might be requiring some flatten-specific trait to be implemented for any type which you want to be able to flatten, but I believe the current implementation built around just the Serialize and Deserialize traits is making the correct tradeoffs.

Mingun:
Whether it is possible to call hinted deserialising methods if `serde(flatten)` is applied only to structures (but not to `HashMap`, for example, when the count of fields can be unknown that how I understand, is the main motivation when deserialising such structures as map)? It can be implemented also as in this case we will make manually -- to implement deserialising for the generated intermediate structure
```rust
[...]

PR #2431, opened 2023-04-23
add `serialize_map_struct` and `deserialize_map_struct`

### LINKED ISSUE #2350: `serde_derive`: flatten causes struct serializer to lose access to struct name
I'm working on a serializer for a format I came across in the wild which represents all structs as externally-tagged, and as a result I need access to the struct name during serialization. The format looks like this:

```text
Parent {
    a = b

    Child {
        enabled = True
        type = SomeValue
        SomeValueField = 2
    }

    Child {
        enabled = False
        type = OtherValue
        OtherValueField = True
    }
}
```

The way I'd like to write my Rust structs is as follows:

```rust
#[derive(serde::Serialize)]
struct Parent {
    a: String,
    children: Vec<Child>,
}

#[derive(serde::Serialize)]
struct Child {
    enabled: bool,

    #[serde(flatten)]
    variant: ChildVariant,
}

#[derive(serde::Serialize)]
#[serde(tag = "type")]
enum ChildVariant {
    SomeValue {
        SomeValueField: u64,
    },
    OtherValue {
        OtherValueField: bool,
    },
}
```

However with `#[serde(flatten)]` present, `serde_derive` calls `serialize_map` directly when serializing `Child`, meaning that I don't have access to the struct name anymore and must use a fallback value &mdash; `Child` formats itself as:

```text
AnonymousMap {
    enabled = True
    type = SomeValue
    SomeValueField = 2
}
```

The root of this behavior appears to be that structs with flattened members are serialized as if they were maps: https://github.com/serde-rs/serde/blob/2062a3c16d44a611558e79a6eaeb7ea1e718db9c/serde_derive/src/ser.rs#L301-L305

While this is understandable (as `serialize_struct` expects a list of fields, and we can't compute this statically if we're flattened with a `HashMap`), I don't think there's an a priori reason to lose access to the struct name in this case.

Perhaps we could add a `Serializer::serialize_flattened_struct(self, name: &'static str, len: usize) -> SerializeFlatMap` with a default impl pointing at `Serializer::serialize_map`? I'd be happy to PR this if you're amenable.

clouds56:
This might fixes #2350, #1346 and #1584.

**This PR should not break any exist crates/code.**

## Problem
When a struct `S` has a field with attr `flatten`, the struct would be treated as map
```rust
struct S {
  name: String,
  #[serde(flatten)]
  others: HashMap<String, i32>,
}

```
This approach works well widely in format that doesn't care with type name (like `serde_json`), while sometimes it's annoying that we would like to treat map and struct differently (like [ron](https://github.com/ron-rs/ron/issues/115), [envy](https://github.com/softprops/envy/issues/26)).

## This PR
add `serialize_map_struct` and `deserialize_map_struct`
```rust
    fn deserialize_map_struct<V>(
        self,
        name: &'static str,
        fields: &'static [&'static str],
        visitor: V,
    ) -> Result<V::Value, Self::Error>
    where
        V: Visitor<'de>,
    {
        let _ = name;
        let _ = fields;
        self.deserialize_map(visitor)
    }

    fn serialize_map_struct(
        self,
        name: &'static str,
        len: Option<usize>,
    ) -> Result<Self::SerializeMap, Self::Error> {
        let _ = name;
        let _ = len;
        self.serialize_map(len)
    }
```
1. which simply delegates to `serialize_map` and `deserialize_map`, so nothing would be broken
2. pass `name` (as well as `fields` in deserialize), so your serializer could tell it is a struct other than map
3. add `serialize_field` to `SerializeMap` in order to get field names if you interest.
4. consist with other interface
   * we have `serialize_tuple`, `serialize_tuple_struct` and `serialize_tuple_variant` all handle with tuple, but with different parameters to indicate different semantics,
   * the same story goes with `serialize_unit`, `serialize_newtype_struct`, `serialize_struct`,
   * now it went to `serialize_map`. `serialize_map_struct` is also a map, just come with additional information for struct.

## Alternatives
https://github.com/ron-rs/ron/issues/115#issuecomment-1519121167 shows a workaround that used in [serde-starlark](https://github.com/dtolnay/serde-starlark)
```rust
#[derive(Serialize)]
pub struct RustLibrary {
    // ...
    #[serde(flatten)]
    pub common: CommonAttrs,
}

#[derive(Serialize)]
#[serde(untagged)]
pub enum Starlark {
    // ...
    #[serde(serialize_with = "serialize::rust_library")]
    RustLibrary(RustLibrary),
}

// serialize::rust_library
pub fn rust_library<S>(rule: &RustLibrary, serializer: S) -> Result<S::Ok, S::Error>
where
    S: Serializer,
{
    FunctionCall::new("rust_library", rule).serialize(serializer)
}
```
It wrapped with [`FunctionCall`](https://docs.rs/serde_starlark/latest/serde_starlark/struct.FunctionCall.html) which gives `RustLibrary` struct name, as well as force child struct to be parsed as struct,
```rust
impl<'a, S> Serializer for FunctionCallSerializer<'a, S>
where
    S: Serializer,
{
    // ...
    fn serialize_map(self, len: Option<usize>) -> Result<Self::SerializeMap, Self::Error> {
        let len = len.unwrap_or(2);
        let mut delegate = self.delegate.serialize_struct("(", len)?;
        delegate.serialize_field("", self.function)?;
        Ok(FunctionCallArgs { delegate })
    }
}
```
This approach could handle things well for custom struct, but it is common that the `flatten` structs are import from some crates (and maybe deeply nested), so users had no control on this.

### DIFF
--- a/serde/src/de/mod.rs
+++ b/serde/src/de/mod.rs
@@ -1113,6 +1113,22 @@ pub trait Deserializer<'de>: Sized {
     where
         V: Visitor<'de>;
 
+    /// Hint that the `Deserialize` type is expecting a map of key-value pairs.
+    /// Additional name and fields information is provided
+    fn deserialize_map_struct<V>(
+        self,
+        name: &'static str,
+        fields: &'static [&'static str],
+        visitor: V,
+    ) -> Result<V::Value, Self::Error>
+    where
+        V: Visitor<'de>,
+    {
+        let _ = name;
+        let _ = fields;
+        self.deserialize_map(visitor)
+    }
+
     /// Hint that the `Deserialize` type is expecting a struct with a particular
     /// name and fields.
     fn deserialize_struct<V>(
--- a/serde/src/ser/mod.rs
+++ b/serde/src/ser/mod.rs
@@ -1171,6 +1171,64 @@ pub trait Serializer: Sized {
     /// ```
     fn serialize_map(self, len: Option<usize>) -> Result<Self::SerializeMap, Self::Error>;
 
+    /// Begin to serialize a map. This call must be followed by zero or more
+    /// calls to `serialize_key` and `serialize_value`, then a call to `end`.
+    ///
+    /// The argument is the number of elements in the map, which may or may not
+    /// be computable before the map is iterated. Some serializers only support
+    /// maps whose length is known up front.
+    ///
+    /// ```edition2018
+    /// # use std::marker::PhantomData;
+    /// #
+    /// # struct HashMap<K, V>(PhantomData<K>, PhantomData<V>);
+    /// #
+    /// # impl<K, V> HashMap<K, V> {
+    /// #     fn len(&self) -> usize {
+    /// #         unimplemented!()
+    /// #     }
+    /// # }
+    /// #
+    /// # impl<'a, K, V> IntoIterator for &'a HashMap<K, V> {
+    /// #     type Item = (&'a K, &'a V);
+    /// #     type IntoIter = Box<Iterator<Item = (&'a K, &'a V)>>;
+    /// #
+    /// #     fn into_iter(self) -> Self::IntoIter {
+    /// #         unimplemented!()
+    /// #     }
+    /// # }
+    /// #
+    /// use serde::ser::{Serialize, Serializer, SerializeMap};
+    ///
+    /// struct Config {
+    ///     name: String,
+    ///     others: HashMap<String, String>,
+    /// }
+    ///
+    /// impl Serialize for Config {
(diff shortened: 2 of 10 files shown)


dtolnay (commented):
Thanks for the PR!

Overall I don't think the new functionality that would be unlocked by this for data formats is compelling enough to warrant having new methods on Deserializer, Serializer, SerializeMap for this. I would prefer if the issue with RON were addressed in some other way (like the way which works well for serde-starlark that was pointed out in the RON issue).

In general it is not a goal for the serde traits to natively support all functionality in arbitrary formats, just some of the most ubiquitous commonalities across a range of formats.

Issue #2432, opened 2023-04-25
More permissive support for sequence -> struct deser


alexheretic:
It is possible to deserialize sequence data into a struct. E.g.
```rust
#[derive(Debug, Deserialize)]
struct A {
    number: i32,
    text: String,
    _dont: IgnoredAny,
    _really: IgnoredAny,
    _care: IgnoredAny,
}

let json = r#"[1234,"hello",[],345.234,true]"#;

let data: A = serde_json::from_str(json).unwrap();
// A { number: 1234, text: "hello", _dont: IgnoredAny, _really: IgnoredAny, _care: IgnoredAny }
```

However, the number of fields are required to match the sequence length. This has a couple of consequences:
* Each sequence item I'm not interested in must be explicitly ignored.
* If the sequence API changes and adds another item this will break deserialization.

Ideally I'd like a way of ignoring the tail.
```rust
#[derive(Debug, Deserialize)]
struct A {
    number: i32,
    text: String,
    _tail: IgnoredSequenceTail,
}
```
And json sequences `[1234,"hello",[],345.234,true]` & `[1234,"hello",[],345.234,true,123]` would work.

<details>

<summary>Example: Implementing the trait for this behaviour is fairly verbose</summary>

```rust
#[derive(Debug)]
struct A {
    number: i32,
    text: String,
}

impl<'de> Deserialize<'de> for A {
    fn deserialize<D>(deserializer: D) -> Result<A, D::Error>
    where
        D: Deserializer<'de>,
    {
        struct SeqVisitor;

        impl<'de> Visitor<'de> for SeqVisitor {
            type Value = A;

            fn expecting(&self, formatter: &mut Formatter) -> std::fmt::Result {
                formatter.write_str("An array [number, text, ...]")
            }

            fn visit_seq<SeqAccessor>(
                self,
                mut seq: SeqAccessor,
            ) -> Result<Self::Value, SeqAccessor::Error>
            where
                SeqAccessor: SeqAccess<'de>,
            {
                let number = seq
                    .next_element::<i32>()?
                    .ok_or_else(|| serde::de::Error::missing_field("number"))?;
                let text = seq
                    .next_element::<String>()?
                    .ok_or_else(|| serde::de::Error::missing_field("text"))?;
                while seq.next_element::<IgnoredAny>()?.is_some() {}
                Ok(A { number, text })
            }
        }

        deserializer.deserialize_seq(SeqVisitor)
    }
}
```

</details>

[...]

PR #2436, opened 2023-04-29
Allow to flatten `IgnoredAny`

Mingun:
This can be useful for use-case described in #2432 (if other problems will be solved) and if you have a generic struct with flatten field and want to pass an `IgnoredAny` for whatever reason.

I also implement `PartialEq` for `IgnoredAny` so it can be tested using `assert_de_tokens`.

### DIFF
--- a/serde/src/de/ignored_any.rs
+++ b/serde/src/de/ignored_any.rs
@@ -108,7 +108,7 @@ use de::{
 /// #     Ok(())
 /// # }
 /// ```
-#[derive(Copy, Clone, Debug, Default)]
+#[derive(Copy, Clone, Debug, Default, PartialEq)]
 pub struct IgnoredAny;
 
 impl<'de> Visitor<'de> for IgnoredAny {
--- a/serde/src/private/de.rs
+++ b/serde/src/private/de.rs
@@ -2811,6 +2811,13 @@ where
         visitor.visit_unit()
     }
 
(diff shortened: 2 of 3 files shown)


dtolnay (changes requested):
Independent of #2437, I am not sure whether this still has a use case. Could you let me know whether this PR is still needed?

For deserializing struct containing flattened fields from a map, I expect people would not need to write `#[serde(flatten)] inner: IgnoredAny` because ignoring all other fields is already the default behavior.

Thanks anyway for the PR!

dtolnay:
Do we know of any code that uses or has wanted to use `flatten` in that manner with a generic type?

In general "least astonishment" is not compelling to me as it's in tension with "least bloat" and the intersection between features can be disproportionately costly to support for the sake of something that literally nobody might want in the first place.

Mingun:
The point is not whether we know or not. I believe that the library should not limit the user to strange, not explained restrictions.

> can be disproportionately costly to support 

I think, that here definitely not this case. In fact, this PR _decreases_ cost for support, because it removes special case and adds an explicit test for it. In contrast, in current situation you actually supports invariant that `IgnoredAny` cannot be type of the `#[serde(flatten)]` field, but you do that without a test that check that this invariant is hold.

dtolnay (approved):
Sounds good. Thanks for the PR!

PR #2437, opened 2023-04-30
Take into account `deny_unknown_fields` attribute when deserialzie struct from sequence

### LINKED ISSUE #2432: More permissive support for sequence -> struct deser

Mingun:
This PR expands scope of `deny_unknown_fields` container attribute to sequences, making behaviour consistent when struct deserialized from a map or from a sequence:

```rust
#[derive(Deserialize)]
struct A {
    number: i32,
    text: String,
}
```
could be deserialized from
```json
{
    "number": 1234,
    "text": "hello",
    "dont": 345.234,
    "really": [],
    "care": true
}
```
and from
```json
[
    1234,
    "hello",
    345.234,
    [],
    true
]
```

The
```rust
#[derive(Deserialize)]
#[serde(deny_unknown_fields)]
struct A {
    number: i32,
    text: String,
}
```
could be deserialized only from
```json
{
    "number": 1234,
    "text": "hello"
}
```
and from
```json
[
    1234,
    "hello"
]
```

Closes #2432

### DIFF
--- a/serde_derive/src/de.rs
+++ b/serde_derive/src/de.rs
@@ -509,7 +509,7 @@ fn deserialize_tuple(
     };
 
     let all_skipped = fields.iter().all(|field| field.attrs.skip_deserializing());
-    let visitor_var = if all_skipped {
+    let visitor_var = if all_skipped && cattrs.deny_unknown_fields() {
         quote!(_)
     } else {
         quote!(mut __seq)
@@ -574,7 +574,9 @@ fn deserialize_tuple_in_place(
         None
     };
 
-    let visit_seq = Stmts(deserialize_seq_in_place(params, fields, cattrs, expecting));
+    let visit_seq = Stmts(deserialize_seq_in_place(
+        params, fields, false, cattrs, expecting,
+    ));
 
     let visitor_expr = quote! {
         __Visitor {
@@ -596,7 +598,7 @@ fn deserialize_tuple_in_place(
     };
 
     let all_skipped = fields.iter().all(|field| field.attrs.skip_deserializing());
-    let visitor_var = if all_skipped {
+    let visitor_var = if all_skipped && cattrs.deny_unknown_fields() {
         quote!(_)
     } else {
         quote!(mut __seq)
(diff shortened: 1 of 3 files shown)


dtolnay (commented):
Thanks for the PR!

> ```rust
> #[derive(Deserialize)]
> struct A {
>     number: i32,
>     text: String,
> }
> ```
>
> ...
>
> could be deserialized from
>
> ```json 
> [
>     1234,
>     "hello",
>     345.234,
>     [],
>     true
> ]
> ```

It is not clear to me that this is desirable, or the behavior that most users would expect to get here from derived Deserialize impls by default.

I left a suggestion in https://github.com/serde-rs/serde/issues/2432#issuecomment-1535584280 -- I think I would prefer if this were pursued in a different library.

PR #2442, opened 2023-05-05
Remove build script from serde_derive

taiki-e:
The current serde_derive's MSRV is 1.56, and both underscore consts and ptr::addr_of! are always available.


dtolnay (approved):
Thanks!

PR #2443, opened 2023-05-06
Simplify code in deserialize_in_place_struct and implement #2387 for in-place case

Mingun:
This PR makes a small refactoring which, I think, improves undertanding of the code that will be generated. During this work I found, that #2387 did not change code generation for in-place case, so the last commit fixes that. The strange thing, that tests does not catch this. I tried to see the generated code using
```console
serde\test_suite> cargo expand --test test_annotations --all-features
```
and it shows that `deserialize_in_place` method doesn't generated. I tried
- add explicit dependency to the `serde_derive` here:
  https://github.com/serde-rs/serde/blob/25381be0c93f32aa43ce0e399c45442965e81ecc/test_suite/Cargo.toml#L11-L13
- add explicit requirement for the test:
  ```toml
  [[test]]
  name = "test_annotations"
  required-features = ["serde_derive/deserialize_in_place"]
  ```
- add a new feature `deserialize_in_place = ["serde_derive/deserialize_in_place"]` here:
  https://github.com/serde-rs/serde/blob/25381be0c93f32aa43ce0e399c45442965e81ecc/test_suite/Cargo.toml#L8-L10
- combine the previous two attempts

Nothing of this is work. So I do not know how to test those changes. I would be appreciate if someone point me out how to correctly test this.

### DIFF
--- a/serde_derive/src/de.rs
+++ b/serde_derive/src/de.rs
@@ -1171,22 +1171,14 @@ fn deserialize_struct_in_place(
         })
         .collect();
 
-    let fields_stmt = {
-        let field_names = field_names_idents.iter().map(|(name, _, _)| name);
-        quote_block! {
-            #[doc(hidden)]
-            const FIELDS: &'static [&'static str] = &[ #(#field_names),* ];
-        }
-    };
-
     let field_visitor = deserialize_generated_identifier(&field_names_idents, cattrs, false, None);
 
     let visit_seq = Stmts(deserialize_seq_in_place(params, fields, cattrs, expecting));
     let visit_map = deserialize_map_in_place(params, fields, cattrs);
 
     let field_visitor = Stmts(field_visitor);
-    let fields_stmt = Stmts(fields_stmt);
     let visit_map = Stmts(visit_map);
+    let field_names = field_names_idents.iter().map(|(name, _, _)| name);
 
     let visitor_expr = quote! {
         __Visitor {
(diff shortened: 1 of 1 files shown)


Mingun:
Is there something that prevents merging this PR? I now working on proper testcases for enums (there a many pitfalls there) and having simplifications that I made here would be very useful to fix them

oli-obk:
The PR mixes functional changes with pure refactorings, requiring a larger amount of contiguous time to review it. I still have 20+ reviews on rustc that are higher priority, so this one keeps getting bumped down. I have no timeline to tell you when I'll be able to review it.

Mingun:
I understood. Usually, refactoring is required to see the need of the actual change, and in some cases unavoidable. Most changes are pretty obvious, although. Just review commit-by-commit and follow hints when "ignore whitespace changes" is recommended. I hope you will find time to review it, it shouldn't take more than a half-hour of your time

jplatte (commented):
I've had a look at all the commits. They mostly make sense to me but it looks like they are almost all independent of each other. Maybe it would make sense to split this PR into multiple smaller ones that can be independently prioritized / reviewed / discussed / merged?

dtolnay (approved):
Thank you! This looks great.

Regarding the question in the PR description: I can't tell what might have been going wrong. For me, `cargo expand --test test_annotations --all-features` correctly generates all the `deserialize_in_place` functions I would expect (27 of them as of current master). This is fewer than the total number of times `derive(Deserialize)` is used on a struct in that file, but that's expected because there is a set of conditions where a `deserialize_in_place` is not supposed to be generated.

`test_suite` unconditionally enables serde_derive's "deserialize_in_place" feature here:

https://github.com/serde-rs/serde/blob/0c367838cc442333fdc6d5a523c3d43c6a2ca751/test_suite/Cargo.toml#L19

PR #2444, opened 2023-05-06
Simplify code for generation of struct deserializers

Mingun:
The first 7 commits of this PR eliminates various duplications from the code, making it more easely to understand, as I hope.

The last 3 commits makes more obvious which code is generated for different combinations of arguments. Because not all combinations are possible, after the changes impossible variants are not representable in the code.

### DIFF
--- a/serde_derive/src/de.rs
+++ b/serde_derive/src/de.rs
@@ -914,8 +914,6 @@ fn deserialize_struct(
     deserializer: Option<TokenStream>,
     untagged: &Untagged,
 ) -> Fragment {
-    let is_enum = variant_ident.is_some();
-
     let this_type = &params.this_type;
     let this_value = &params.this_value;
     let (de_impl_generics, de_ty_generics, ty_generics, where_clause) =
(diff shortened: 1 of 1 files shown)


serde_derive/src/de.rs
```diff
@@ -932,83 +956,74 @@ fn deserialize_struct(
...
     };
     let expecting = cattrs.expecting().unwrap_or(&expecting);
 
-    let visit_seq = Stmts(deserialize_seq(
-        &type_path, params, fields, true, cattrs, expecting,
+    let field_names_idents: Vec<_> = fields
+        .iter()
+        .enumerate()
+        // Skip fields that shouldn't be deserialized or that is flatten
+        // so them doesn't present in the storage in their literal form
```
oli-obk, line 979:
```suggestion
        // Skip fields that shouldn't be deserialized or that were flattened,
        // so they don't appear in the storage in their literal form
```

oli-obk (approved):

dtolnay (approved):
Thanks! This is a great cleanup. And thanks Oli for the review.

> *#2444 looks good (modulo a comment). It is a pure refactoring in well-reviewable individual commits. I checked that it does not change behaviour. The main part that needs your review is the actual change (moving from multiple somewhat connected boolish or option arguments to a single enum that separates all the different cases into explicit variants)*

PR #2445, opened 2023-05-07
Add ability to deserialize enums from `SeqAccessDeserialzier`

Mingun:
Motivation of this PR is following: when I rebase #1922 I noticed, that test, introduced in the first commit of #2303 failed after the change. That failure appeared after changing
```rust
// seq: A where A: SeqAccess
let rest = SeqAccessDeserializer::new(seq);
let content = try!(Content::deserialize(rest));
let deserializer = ContentDeserializer::<A::Error>::new(content);
// use `deserializer`
```
to
```rust
// seq: A where A: SeqAccess
let deserializer = SeqAccessDeserializer::new(seq);
// use `deserializer`
```
i. e. after removing intermediate bufferisation step. Because code for the `MapAccessDeserializer` is the same and it does not suffer from such problem, I checked the difference and realized, that `MapAccessDeserializer` has a special handling fo the `Deserializer::deserialize_enum` hint:
https://github.com/serde-rs/serde/blob/25381be0c93f32aa43ce0e399c45442965e81ecc/serde/src/de/value.rs#L1474-L1484
which is missing for the `SeqAccessDeserializer`.

Because `SeqAccessDeserializer` is used in such context when `deserialize_enum` could be requested from it, I think, it should provide the special implementation for that.

This implementation is following: enum is represented as:
- two elements, where the first element is a tag and the second is a content
- one element: a special case for unit variants, where the second element for content can be missed

Additionally, this PR:
- extends testcases to cover all possible enum variants (Unit, Newtype, Tuple and Struct)
- extends testcases to cover negative scenarios (wrong tag, empty map / seq, missing content for seq)
- changes the error message for the empty map for `MapAccessDeserializer` from `invalid type: map, expected enum` to `invalid length: 0, expected enum`, because actually map *is* expected, but it should not be empty
- document behavior of `Deserializer::deserialize_enum` for `MapAccessDeserializer` and `SeqAccessDeserializer`


dtolnay (commented):
Thanks for the PR!

I am not convinced that this is the behavior that would be expected by users. As far as I can tell the motivation explained in the PR description is to make the following:

```rust
let deserializer = SeqAccessDeserializer::new(seq);
```

behave more like the following:

```rust
let rest = SeqAccessDeserializer::new(seq);
let content = try!(Content::deserialize(rest));
let deserializer = ContentDeserializer::<A::Error>::new(content);
```

But I tried that code, and `ContentDeserializer` also does not deserialize enums from sequences. The current `SeqAccessDeserializer` is already consistent with this.

```rust
use serde::de::value::SeqAccessDeserializer;
use serde::de::{Deserialize, Deserializer, SeqAccess, Visitor};
use std::fmt;

#[derive(Deserialize, Debug)]
enum Enum {
    Variant(usize),
}

struct MyVisitor;

impl<'de> Visitor<'de> for MyVisitor {
    type Value = Enum;
    fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
        formatter.write_str("repro")
    }
    fn visit_seq<A>(self, seq: A) -> Result<Self::Value, A::Error>
    where
        A: SeqAccess<'de>,
    {
        let rest = SeqAccessDeserializer::new(seq);
        let content = serde::__private::de::Content::deserialize(rest).unwrap();
        let deserializer = serde::__private::de::ContentDeserializer::<A::Error>::new(content);
        Enum::deserialize(deserializer)
    }
}

fn main() {
    let mut de = serde_json::Deserializer::from_str(r#" ["Variant", 1] "#);
    let e = de.deserialize_any(MyVisitor).unwrap();
    println!("{:?}", e);
}
```

I'm left unclear about what the motivation for this PR is.

Mingun:
I cannot remember exactly, why I decided to make `SeqAccessDeserializer` behaves like `MapAccessDeserializer` when working with enums. The whole situation with test from #2303 looked strange for me but I remember that I got deal with it. I need to restore the whole chain of reasoning, this can take some time.

Mingun:
I remembered the chain of reasoning. To make sure #1922 doesn't break existing behavior, I decided to add the missing tests. One of the tests is the deserialization of `ExternallyTagged` enums that are inside of `InternallyTagged` enums:

```rust
#[derive(Deserialize)]
enum ExternallyTagged {
  Unit,
}

#[derive(Deserialize)]
#[serde(tag = "tag")]
enum InternallyTagged {
  StructWithEnum { f: ExternallyTagged },
}
```
Internally tagged enums can be represented either as maps or as sequences:
```json
[
  { "tag": "StructWithEnum", "f": "Unit" },
  ["StructWithEnum", "Unit"],
]
```
[You can make sure that on playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=7b7ef512c922c6d26bf154a4e81488b6).

Currently internally tagged enums always use bufferisation and `ExternallyTagged` enum will be deserialized from `ContentDeserializer`:
https://github.com/serde-rs/serde/blob/3aca38d2d3c9b5dba9bcb25be2e16bde6678dbf6/serde_derive/src/de.rs#L1399-L1406

After optimisation in #1922 in the optimized case `ExternallyTagged` will be deserialized either from `MapAccessDeserializer` or `SeqAccessDeserializer`, because intermediate bufferisation step will be skipped here:
https://github.com/serde-rs/serde/blob/3aca38d2d3c9b5dba9bcb25be2e16bde6678dbf6/serde/src/private/de.rs#L853-L894

Without changes in this PR the test from playground failed for #1922 (I've put it to serde's `test_suite/regression/issue2445.rs` for check):
```
---- regression::issue2445::seq stdout ----
thread 'regression::issue2445::seq' panicked at test_suite\tests\regression\issue2445.rs:28:71:
called `Result::unwrap()` on an `Err` value: Error("invalid type: sequence, expected internally tagged enum InternallyTagged", line: 1, column: 0)
```

Then I realized that deserialization tests of externally tagged from `MapAccessDeserializer` are incomplete and completely missing for `SeqAccessDeserializer`. So I've added them. Because both deserializers used in the same circumstances it looks reasonable that they should behave the same.

PR #2446, opened 2023-05-08
Eliminate some duplicated code

Mingun:
The first commit removes the one copy of the `FlatMapAccess` struct.

The third commit removes superfluous constructors of `FlatMapAccess` and `FlatStructAccess`. Since `FlatInternallyTaggedAccess` did without it, it seems to me that this is acceptable. Besides, I found that without constructors that used only in one place to construct a struct is is a little easier to understand the code.

The last two commits extracts repeated code into functions.

### DIFF
--- a/serde/src/private/de.rs
+++ b/serde/src/private/de.rs
@@ -2738,11 +2738,7 @@ where
     where
         V: Visitor<'de>,
     {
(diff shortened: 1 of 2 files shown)


dtolnay (approved):
Thanks!

Issue #2186, opened 2022-03-09
Avoid lossy buffering in #[serde(flatten)]


sfackler:
The `Deserialize` implementation of a type with a field annotated `#[serde(flatten)]` currently buffers up all of the unknown map entries in memory, and then deserializes the flattened value from them at the end. This approach has a couple of downsides:

1. It causes the deserializer to allocate when that would normally not be the case.
2. It confuses "smart" deserializers - things like serde_ignored_value or serde_json's error position tracking don't work properly.
3. It can lose out on special casing in deserializers based on hinting.

Fortunately, there's alternate approach that avoids all of these issues! Instead of buffering, we create a custom `MapAccess` which intercepts fields that we want to handle in the top-level type, and then returning the remainder to the nested deserialize implementation: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=96546d7ba882b70042268a0b48de6286

jonasbb:
This is a very interesting idea. 
I have a couple of remarks / incompatibility observations. Maybe some of them can be resolved.

Currently, the `Foo` type ultimately decides if a field name is known or unknown. This leads to an incompatibility currently, if `deny_unknown_fields` is added to `Foo`. With the normal `flatten` extra fields will be ignored, since `ExtendedFoo` allows them.
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=4e249bb1961ff44945606ed6bdcdab7d

I think that can be fixed by adding an `Ignore` variant to `ExtendedFooFields`, similar to the current derive. An unknown field error from the seed of `ExtendedFooFieldsVisitor` must be converted into the `Ignore` variant, but all other error likely should remain. Inspecting the errors might be complicated, since they are very opaque.

It is unclear to me how this could be extended to more than one flattened field.
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=7451b4980894d2ddda33b7ce3fc0aa9c
An additional complication here is the fact, that the same value can be deserialized multiple times. A map visitor (i.e., `Value`) will not consume the fields from the internal buffer, but a struct visitor (i.e., `Sub`) will. This leads the second `Value` to share all fields with the first except those consumed by the `Sub` visitor.
<details>
<summary>Output</summary>

```text
[src/main.rs:27] d = Data {
    f0: Object({
        "x": String(
			"foo",
        ),
        "y": String(
            "bar",
        ),
        "z": String(
            "baz",
        ),
    }),
    a: 123,
[...]

Issue #1173, opened 2018-03-11
Add a rustc version badge


dtolnay:
Copied from the [`num`](https://github.com/rust-num/num) readme:

https://img.shields.io/badge/rustc-1.13+-red.svg

![rustc 1.13+](https://img.shields.io/badge/rustc-1.13+-red.svg)

PR #2453, opened 2023-05-13
Add a module with adapters for conversion Visitor to Deserializer

Mingun:
Several times I have noticed that it is necessary to be able to save the value passed to `Visitor` and then send it further through `Deserializer`. In most cases, there is a suitable `Deserializer` (and even an `IntoDeserializer` implementation) for this, but not for all types, and moreover, sometimes this implementation calls another visitor method.

So here I add a dedicated module that guaranties that deserializer will pass the value only to one `Visitor` method in all cases. The `value` module re-export those deserializers that have the same behavior, as deserializers that were defined in it before. Additionally, all deserializers have been documented what they exactly does.

One example usage can be [found](https://github.com/Lucretiel/serde-bufferless/blob/a00fb677f4d01d385b3bde45c0662866cf0cc5dd/src/private/flatten.rs#L285) in POC implementation of one case for #2186, and my own implementation for #1183 also could require it (if I remember correctly).

### DIFF
--- a/serde/src/de/adapters.rs
+++ b/serde/src/de/adapters.rs
@@ -1,0 +1,725 @@
+//! Building blocks for conversion `Visitor` into `Deserializer`.
+//!
+//! Those deserializers can be used to temporary save the argument of a `Visitor`
+//! method and pass it to the same method in which it was captured, later.
+
+use lib::*;
+
+use de::{self, Deserializer, EnumAccess, IntoDeserializer, MapAccess, SeqAccess, Visitor};
+
+primitive_deserializer!(
+    /// A deserializer holding a `bool`.
+    ///
+    /// This deserializer will call [`Visitor::visit_bool`] for all requests.
+    pub BoolDeserializer, bool, visit_bool
+);
+
+primitive_deserializer!(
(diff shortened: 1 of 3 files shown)


dtolnay (commented):
Thanks for the PR!

I can see a use case for these, but this is a pretty large amount of bloat to put into the main `serde` crate. This increases the number of functions in serde by 14.5%, and increases compile time by 6.75% on my machine. I think that's probably not the right tradeoff for something which much less than 0.1% of users (or 0.01%?) will ever use.

What do you think of pursuing maintaining a separate crate containing just the adapters?

Mingun:
That seems strange, because most implementations are shared. Only 6 implementations are new + `AsVisitor`. What if you drop the last commit with `AsVisitor` implementations? Does it then fit into appropriate bounds?

> What do you think of pursuing maintaining a separate crate containing just the adapters?

That would be possible. The main motivation to put it to serde in that only few (6) deserializers does the special things and when I've testing different ways to solve some pain issues (#1173 mostly) I've run into necessity to catch some values in `Visitor` and then pass them unchanged to another deserializer.

PR #2460, opened 2023-05-20
Show deserialized value in the `assert_de_tokens_error` message

Mingun:
This PR improves error message when the type in `assert_de_tokens_error` deserialized successful when it shouldn't.

For example, the call
```rust
assert_de_tokens_error::<Enum>(
    &[
        Token::Map { len: Some(2) },
        Token::Str("a"),
        Token::I32(1),
        Token::Str("a"),
        Token::I32(42),
        Token::MapEnd,
    ],
    "duplicate field `a`",
);
```
could return the following result:
```
---- untagged::duplicate_field_in_struct stdout ----
thread 'untagged::duplicate_field_in_struct' panicked at 'assertion failed: `(left == right)`
  left: `"duplicate field `a`"`,
 right: `"StructWithSkipped { a: NotDeserializable }"`: expected error (left) but tokens deserialized successfully (right)', test_suite\tests\test_de_enum.rs:506:9
```
compared to this (current):
```
---- untagged::duplicate_field_in_struct stdout ----
thread 'untagged::duplicate_field_in_struct' panicked at 'tokens deserialized successfully', test_suite\tests\test_de_enum.rs:506:9
```

Unfortunately, this is breaking change, but:
- usually the types that you test also tested with `assert_de_tokens` which requires `Debug`. So there are high probability that `Debug` already implemented for your types
- this is testing code which should not be many. [According to lib.rs](https://lib.rs/crates/serde_test/rev) there are few dependers, so it can be acceptable to make a small breaking change to improve usability

### DIFF
--- a/serde_test/src/assert.rs
+++ b/serde_test/src/assert.rs
@@ -206,11 +206,15 @@ where
 #[cfg_attr(not(no_track_caller), track_caller)]
 pub fn assert_de_tokens_error<'de, T>(tokens: &'de [Token], error: &str)
 where
-    T: Deserialize<'de>,
+    T: Deserialize<'de> + Debug,
 {
     let mut de = Deserializer::new(tokens);
     match T::deserialize(&mut de) {
-        Ok(_) => panic!("tokens deserialized successfully"),
+        Ok(x) => assert_eq!(
+            error,
+            format!("{:?}", x),
+            "expected error (left) but tokens deserialized successfully (right)"
+        ),
         Err(e) => assert_eq!(e, *error),
     }
 
--- a/test_suite/tests/test_annotations.rs
+++ b/test_suite/tests/test_annotations.rs
@@ -2769,19 +2769,19 @@ fn test_alias_in_flatten_context() {
 
 #[test]
 fn test_expecting_message() {
-    #[derive(Deserialize, PartialEq, Debug)]
+    #[derive(Deserialize, Debug)]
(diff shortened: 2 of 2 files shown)


oli-obk:
> * this is testing code which should not be many. [According to lib.rs](https://lib.rs/crates/serde_test/rev) there are few dependers, so it can be acceptable to make a small breaking change to improve usability

this makes me wish we had a serde-crater to test if a change affects any of the reverse dependencies. I'd expect most callers to provide a type that implements `Debug`, so it would break very few people's code.

Without being able to ascertain the effect of this PR, I don't think we can land it.

dtolnay (commented):
Thanks for the PR!

I agree with Oli, this is not a serde_test 1.x change. I am open to considering this for 2.x as long as it is batched up with other desirable changes. On its own it's obvious to me that this would not motivate a 2.x of serde_test.

Feel free to open an issue in https://github.com/serde-rs/test to track the suggestion.

Mingun:
> On its own it's obvious to me that this would not motivate a 2.x of serde_test.

Why not? Now, when serde_test release cycle is detached from serde release cycle there is no objective reasons to not release version 2.0 as soon as it will required.

Ok, if you don't like to release 2.0 soon (which I totally don't understand because numbers are cheap and nobody force you to maintain two versions, just focus on the last), then just adding new function `assert_de_tokens_error2` with additional boundary would be acceptable?

Issue #2105, opened 2021-10-18
`#[serde(skip)]` is ignored on a 1-field tuple struct


kpreid:
With serde v1.0.130, applying `#[serde(skip)]` to the sole field of a tuple struct is silently ignored, so that the field is serialized (and hence required to be serializable). This seems like a bug: it should result in the struct being treated the same as one which actually has no fields, or if there is some reason this should not be the interpretation, there should be an error or warning.

```rust
use serde::{Serialize, Deserialize};

#[derive(Serialize, Deserialize)]
struct Skip1(#[serde(skip)] i32);

#[derive(Serialize, Deserialize)]
struct Skip2(#[serde(skip)] i32, i32);

#[derive(Serialize, Deserialize)]
struct ActuallyEmpty();

fn main() {
    println!(
        "{}",
        serde_json::to_string(&(
            Skip1(101),
            Skip2(201, 202),
            ActuallyEmpty(),
        )).unwrap()
    );
}
```

Expected output: `[[], [202], []]`

Actual output: `[101, [202], []]`

Issue #1740, opened 2020-02-13
Warn when defining untagged enum with multiple unit variants


tpdickso:
Currently, an enum such as this will compile, but will silently result in one of the two variants not being deserializable:
```rust
#[derive(Serialize, Deserialize)]
#[serde(untagged)]
enum T {
    A,
    B,
    C(f32)
}
```
This behaviour is potentially unintuitive, as the intent of the programmer writing this enum is likely that `T::A` and `T::B` would be deserialized from the strings `"A"` and `"B"`. A warning in this case could be helpful, particularly if a link could be provided to an entry on the "Examples" section of the documentation with a recipe for doing what the programmer intended. This warning could potentially have some sort of annotation to silence it, but the easiest solution would just be to mark all but one of the unit variants as `#[serde(skip_deserializing)]`, which does not change the semantics of the enum and makes it clearer that only the unit variant that isn't marked as skipped will ever be deserialized.

In practice this warning could apply to any untagged variant whose valid representations are a subset of the union of previous variants, but that's fairly impractical to warn for, since deserialize impls can be arbitrary. But I think it would be sufficiently useful to have a warning for this particular case, since it's easy to detect and I feel like it could be a common first attempt at the pattern "One of several named constants, or a fully-described value", such as the use case where I discovered this:
```rust
#[derive(Deserialize)]
#[serde(untagged)]
enum Color {
    White,
    Black,
    Transparent,
    Custom(Vector4<f32>)
}
```

tpdickso:
N.B. There is support for emitting warnings from procedural macros, but it's currently an unstable API. I think could be worth adding to nightly, though, in anticipation of stabilization. https://doc.rust-lang.org/stable/proc_macro/struct.Diagnostic.html#method.warning

Inspirateur:
> the intent of the programmer writing this enum is likely that T::A and T::B would be deserialized from the strings "A" and "B"

I'm having the exact same problem, any workaround for deserializing like that ?

PR #2465, opened 2023-05-28
Allow to deserialize struct variants of untagged and adjacently tagged enums from seq

Mingun:
The artificial limitation on use of `visit_seq` method was introduced in serde-rs/serde#1058 without any rationale. I suggest to remove it. The consequences are:

- Struct variants of adjacently tagged enums now can be deserialized using `visit_seq` method
- Struct variants of untagged enums now can be deserialized using `visit_seq` method
  - Because each variant in untagged enum are tried in the declaration order that means that tuple variants always should be defined before struct variants in enum declaration. We can check that during compilation time and issue a warning about that (we can also issue some other warnings in that stage (#1740 related), for example, structs with similar fields that cannot be distinguished or can be incorrectly distinguished, that is why I didn't include that code in that PR).

This change has a sense, because you already can do the same by wrapping an ordinary struct into a Newtype variant, but I think, that change from
```rust
enum Enum {
  Struct {
    // fields
  },
}
```
to
```rust
struct Struct {
  // fields
}
enum Enum {
  Struct(Struct),
}
```
should not change the behavior because this situation can arise during refactoring and changing behavior can be surprising.

I didn't include tests in that PR, because correct test suite should cover all possible situations, and many of them have certain problems. I plan to include a full testsuite in my upcoming PR for fixing serde-rs/serde#2105.

Right now you can see what changed in the generated code by running the following commands:
- place [skip.txt](https://github.com/serde-rs/serde/files/11585014/skip.txt) into `test_suite/tests` folder and change its extension to `.rs`
- run the following command in master and this PR and compare the generated output:
   ```console
   ...\serde\test_suite> cargo expand --all-features --test skip
   ```

Closes serde-rs/serde#2473, closes serde-rs/json#1049

### DIFF
--- a/serde_derive/src/de.rs
+++ b/serde_derive/src/de.rs
@@ -1008,29 +1008,28 @@ fn deserialize_struct(
 
     // untagged struct variants do not get a visit_seq method. The same applies to
     // structs that only have a map representation.
-    let visit_seq = match form {
-        _ if cattrs.has_flatten() => None,
-        _ => {
-            let mut_seq = if field_names_idents.is_empty() {
-                quote!(_)
-            } else {
-                quote!(mut __seq)
-            };
+    let visit_seq = if cattrs.has_flatten() {
+        None
+    } else {
+        let mut_seq = if field_names_idents.is_empty() {
+            quote!(_)
+        } else {
+            quote!(mut __seq)
+        };
 
-            let visit_seq = Stmts(deserialize_seq(
-                &type_path, params, fields, true, cattrs, expecting,
-            ));
+        let visit_seq = Stmts(deserialize_seq(
+            &type_path, params, fields, true, cattrs, expecting,
+        ));
 
-            Some(quote! {
-                #[inline]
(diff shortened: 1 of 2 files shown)


dtolnay (commented):
_"that means that tuple variants always should be defined before struct variants in enum declaration"_ &mdash; I think this kind of consideration is part of why this was implemented the current way. When writing Rust data structures to deserialize a particular pattern of input data into, usually it's relatively harmless if serde contains various implicit conversions which don't apply to the data that you expect to receive. Untagged enums are one of the main exceptions to this: at some point each additional implicit behavior makes it _less_ likely that a data structure has the behavior that the programmer was after, not more likely.

When combined with the other consideration of increasing the compile time bloat of untagged enums further, and the rarity of compact binary formats that try to support untagged enums in the first place, and the existence of a reasonable workaround for those few formats to apply (https://github.com/serde-rs/serde/issues/2473#issuecomment-1586239202), my sense is this ends up being a net negative change. I can see the rationale in favor of it but I'd like to keep the original behavior in this case.

Thanks anyway for the PR!

PR #2466, opened 2023-05-28
Fix incorrect count of fields passed to tuple deserialization methods

Mingun:
This count should mean the number of fields expected in the serialized form, so if some fields are skipped, they shouldn't be counted.

Methods affected:
- Deserializer::deserialize_tuple
- Deserializer::deserialize_tuple_struct
- VariantAccess::tuple_variant

After this PR number of fields passed to deserializer methods will be the same as number, passed to the serializer methods.

I didn't include tests in that PR, because correct test suite should cover all possible situations, and many of them have certain problems. I plan to include a full testsuite in my upcoming PR for fixing #2105.

Right now you can see what changed in the generated code by running the following commands:
- place [skip.txt](https://github.com/serde-rs/serde/files/11585014/skip.txt) into `test_suite/tests` folder and change its extension to `.rs`
- run the following command in master and this PR and compare the generated output:
   ```console
   ...\serde\test_suite> cargo expand --all-features --test skip
   ```

### DIFF
--- a/serde_derive/src/de.rs
+++ b/serde_derive/src/de.rs
@@ -445,13 +445,18 @@ fn deserialize_tuple(
     cattrs: &attr::Container,
     deserializer: Option<TokenStream>,
 ) -> Fragment {
+    assert!(!cattrs.has_flatten());
+
+    let field_count = fields
+        .iter()
+        .filter(|field| !field.attrs.skip_deserializing())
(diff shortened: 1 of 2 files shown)


oli-obk (approved):

dtolnay (approved):
Thanks!

Issue #1713, opened 2020-01-11
impl Serialize/Deserialize for RangeTo, RangeFrom, RangeFull


kgv:
@dtolnay 
Are there any restrictions by which this has not yet been implemented?
I can create PR if have not any restrictions.

kgv:
Oh, sorry, found answer:
https://github.com/serde-rs/serde/pull/1466

PR #2471, opened 2023-06-09
Add `Serialize`/`Deserialize` impls for `Range{From,Full,To}`

tbu-:
CC #796
CC #1466
CC #1713


tbu-:
I have `RangeFrom` and `RangeTo` in my data structures as a way to specify that I'm not talking about a specific index, but a range from the start or till the end.

dtolnay (approved):
Thanks!

I think `RangeFrom` and `RangeTo` impls are all right. I am deleting the `RangeFull` ones in a followup commit (e6a4a3772e7263cebd8e0f38ec08f5b3a9958f7b) because you have not mentioned those being relevant to your use case. I prefer not to accept code that I don't know of anybody wanting to use.

PR #2475, opened 2023-06-12
Add `Deserializer::deserialize_identifier` support for adjacently tagged enums

Baptistemontan:
Adjancently tagged enum should behave like a struct with 2 elements, the tag and the content such as

```rust 
struct Foo<'a, T> {
  tag: &'a str,
  content: T
}
```

This is the case for most format until you come across a format that serialize structs with fields identifier other than `&str`. for exemple auto derived struct handle the case where the field identifier is an unsigned number, or a `&[u8]`, not just `&str`. 

This is due to the fact that  `TagOrContentFieldVisitor` and `TagContentOtherFieldVisitor` directly calls `Deserializer::deserialize_str` and only handle cases for `&str`, any other format are rejected.

This PR enable `TagOrContentFieldVisitor` and `TagContentOtherFieldVisitor` to call `Deserializer::visit_identifier` and handle the case for `&str`, `&[u8]` and unsigned numbers.

This modifications should only allow more flexibility and should'nt conflict with flattened adjacently tagged enums, as those serialize as a map and not a struct.

### DIFF
--- a/test_suite/tests/test_macros.rs
+++ b/test_suite/tests/test_macros.rs
@@ -1086,6 +1086,34 @@ fn test_adjacently_tagged_enum() {
         ],
     );
 
+    // unit with no content (integer tag)
+    assert_de_tokens(
+        &AdjacentlyTagged::Unit::<u8>,
+        &[
+            Token::Struct {
+                name: "AdjacentlyTagged",
+                len: 2,
+            },
+            Token::U64(0),
+            Token::Str("Unit"),
+            Token::StructEnd,
+        ],
+    );
+
+    // unit with no content (bytes tag)
+    assert_de_tokens(
+        &AdjacentlyTagged::Unit::<u8>,
+        &[
+            Token::Struct {
+                name: "AdjacentlyTagged",
+                len: 2,
+            },
+            Token::Bytes(b"t"),
+            Token::Str("Unit"),
+            Token::StructEnd,
+        ],
+    );
+
     // unit with tag first
     assert_de_tokens(
         &AdjacentlyTagged::Unit::<u8>,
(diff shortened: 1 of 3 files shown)


test_suite/tests/test_macros.rs
```diff
@@ -1086,6 +1086,34 @@ fn test_adjacently_tagged_enum() {
...
 
+    // unit with no content (integer tag)
+    assert_de_tokens(
+        &AdjacentlyTagged::Unit::<u8>,
+        &[
+            Token::Struct {
+                name: "AdjacentlyTagged",
+                len: 2,
+            },
+            Token::U64(0),
```
Mingun, line 1097: I would add comments to tokens that represents tag and content, for clarify
```suggestion
            Token::U64(0), // tag
```

dtolnay (commented):
Thanks for the PR!

May I ask whether there is a specific publicly-available data format library you are working with where this makes a difference? Could you share a short program I could run to reproduce the error that happens under the old behavior?

Baptistemontan:
I'm working on a compact binary format that is'nt meant to be publicly used (at least for now?), and it serialize struct as sequences.

My tests looks like this, some simple round trip checks:

```
#[derive(Debug, Serialize, Deserialize, PartialEq)]
#[serde(tag = "t", content = "c")]
enum AdjTaggedEnum {
    NewType(String),
    Struct { num: usize },
}

#[test]
fn test_serialize_deserialize_adj_tagged_enum_variant1() {
    let value = AdjTaggedEnum::NewType("john".into());

    let mut v: Vec<u8> = Vec::new();
    ser::to_writer(&value, &mut v).unwrap();

    let res: AdjTaggedEnum = de::from_bytes(&v).unwrap();

    assert_eq!(value, res);
}

#[test]
fn test_serialize_deserialize_adj_tagged_enum_variant2() {
    let value = AdjTaggedEnum::Struct { num: 12 };

    let mut v: Vec<u8> = Vec::new();
    ser::to_writer(&value, &mut v).unwrap();

    let res: AdjTaggedEnum = de::from_bytes(&v).unwrap();

    assert_eq!(value, res);
}
```

But I get this error:

>tests::test_serialize_deserialize_adj_tagged_enum_variant1' panicked at 'called `Result::unwrap()` on an `Err` value: Message("invalid type: integer `0`, expected \"t\", \"c\", or other ignored fields")

I use a work around for struct-like types that don't implement `visit_seq` where I give the field identfier as a u64 index, as auto derived field visitor implement `visit_u64`. This works great until adjacently tagged enums.

This PR is'nt to fix my own library, I probably can find works around or can settle to serialize the field name, this is just to have consistency, it's a niche case that makes you rework a big portion of the format for no apparent reason.

[here is the repo](https://github.com/Baptistemontan/serde-bin), it's highly WIP and I put a very generic name that will be changed, if I publish it. You can find the tests in `src/any/mod.rs`

dtolnay (approved):
Thanks!

Issue #2231, opened 2022-06-21
Optional tag for internally tagged enum


adamspofford-dfinity:
I have an internally tagged enum where one variant should be considered the 'default'. If the tag is not present, the default variant should be selected and deserialized from the other fields. Unfortunately serde does not seem to support this case. The fields of the variants overlap, so `serde(untagged)` cannot be used; the tag should be validated and an error should be raised if it's an incorrect tag, so `serde(other)` cannot be used either as it accepts any tag. It would be great if the tag could be specified as optional with a variant to default to. Two ways I can think of doing that are a `serde(default)` variant attribute, or a `serde(default_tag = "tag_name")` container attribute.

mmun:
I have this use case at work as well.

kangalioo:
I think a `#[serde(default)]` variant attribute is the better choice. It aligns with Rust's built-in `#[default]` attribute on enum variants: https://doc.rust-lang.org/std/default/trait.Default.html#enums

ayngling:
I also have this use case at work.

I agree with the previous poster about `#[serde(default)]` being the better choice.

pront:
Hey, this would be a very useful for us (https://github.com/vectordotdev).

Issue #1799, opened 2020-05-05
#[serde(default)] as a variant attribute for when the tag is not present


agausmann:
I have a scenario where I'd like to infer the variant of an enum where the tag is not present in an internally-tagged representation.

Consider this enum, which deserializes either an item type or a fluid type for an ingredient of a crafting recipe in a video game, as specified by a game or mod developer in the game data. The developer may omit the `type` member and it will be assumed to be an `Item`.

```
#[derive(Serialize, Deserialize)]
#[serde(tag = "type", rename_all = "kebab-case")]
enum Ingredient {
    Item {
        name: String,
        amount: u16,
    }
    Fluid {
        name: String,
        amount: f64,
        #[serde(default)] 
        temperature: Option<f64>,
    }
}
```

The variants are so similar in structure, but so different in semantic meaning, that `untagged` would be inappropriate here. I can not allow a fluid type to be accidentally parsed as item, or vice versa, but because of the ambiguity of numeric types in the data format, it is possible for that to happen.

What I'd like to do is add a `#[serde(default)]` attribute to the `Item` variant, which would infer that it is an `Item` in the case where the `type` tag is not present.

This is somewhat related to #912, but there is an important difference in the expected behavior when the tag is present but not recognized. In this case, unrecognized tags should still be an error, but in #912 the feature is specifically about handling unrecognized tags.

The semantics here are more similar to the existing container/field attribute `#[serde(default)]`, which uses some default value when a field is missing from the input, which is why I propose using this as the syntax.

(Example adapted from https://wiki.factorio.com/Types/IngredientPrototype)

agausmann:
Possible duplicate of #1410

agausmann:
I've implemented a prototype of this feature at [`agausmann/default_variant`](https://github.com/serde-rs/serde/compare/master...agausmann:default_variant)

agausmann:
Here's one possible alternative using a tagged enum nested within an untagged enum, though it isn't nice to manually manage for larger enums.

[(playground)](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=17755bf6b83d89b1866ad307e3074cdd)

```rust
use serde::{Serialize, Deserialize};

#[derive(Debug, Serialize, Deserialize)]
#[serde(from = "MaybeTaggedFoo")]
pub enum Foo {
    Bar {
        name: String,
    },
    Baz {
        name: String,
    },
}

#[derive(Deserialize)] 
#[serde(untagged)]
[...]

Issue #1410, opened 2018-10-12
Default tag.


dpc:
I hava an `enum` with internal `tag`. However, for brevity, it's important for me to skip the tag if it has a default value. It goes both ways - if the tag is missing during deserialization, it means that the default variant should be used.

This is quite easily doable with normal fields, but does not seem possible with `#[tag = somefieldname]`, since I would have to use `default` and `skip_serializing_if` on `somefieldname`, which doesn't exist.

dtolnay:
This is not well supported now. It is possible to capture this behavior in a handwritten Serialize and Deserialize impl but would be a lot of code. I would be interested in suggestions for how to support this better.

dpc:
Naively speaking: allow `default_tag` and `skip_serializing_tag_if` on `enum`, and then make them work the same as `default` and `skip_serializing_if` on a normal field. :)

dpc:
Actually, maybe `default_tag` should be on a variant itself?

dtolnay:
I am going to close this issue because I don't believe it is a sufficiently common use case to justify adding two new serde attributes. It should be possible to make this work with a handwritten Serialize and Deserialize impl. If you get it working, please link to the commit to help anyone else looking for this pattern in the future.

PR #2476, opened 2023-06-14
Add support for a default variant when an internally tagged enum is untagged

harrisonturton:


serde_derive/src/de.rs
```diff
@@ -1344,6 +1344,25 @@ fn deserialize_internally_tagged_enum(
 ) -> Fragment {
     let (variants_stmt, variant_visitor) = prepare_enum_variant_enum(variants, cattrs);
 
+    let mut default_tag = None;
+    for (i, variant) in variants.iter().enumerate() {
+        if variant.attrs.default() {
+            default_tag = Some(quote!{
+                Some(__Field::#{field(i)})
+            });
```
harrisonturton, line 1352: TODO return error if more than one variant is marked with `#[serde(default)]`

harrisonturton:
These changes are causing `cargo test --features derive` to fail with the error:

```
error[E0308]: mismatched types
   --> test_suite/tests/test_gen.rs:456:25
    |
456 |     #[derive(Serialize, Deserialize)]
    |                         ^^^^^^^^^^^
    |                         |
    |                         expected `Option<__Field>`, found `None`
    |                         arguments to this function are incorrect
    |
    = note: expected enum `std::option::Option<test_gen::_::<impl Deserialize<'de> for InternallyTaggedVariantWith>::deserialize::__Field>`
             found struct `None`
note: associated function defined here
   --> /Users/harryt/Documents/patches/serde/serde/src/private/de.rs:816:16
    |
816 |         pub fn new(name: &'static str, default_variant: Option<T>, expecting: &'static str) -> Self {
    |                ^^^

```

in the `test_gen.rs` test:

https://github.com/serde-rs/serde/blob/2fe69d9ac8d69acb31ceb95a82b470bfa853be77/test_suite/tests/test_gen.rs#L456-L479

I believe this has something to do with custom serialisers and deserialisers, but not completely sure. Digging into this...

serde_derive/src/de.rs
harrisonturton: Done. Added checks in `enum_from_ast` in `serde_derive/src/internals/ast.rs`.

test_suite/tests/test_macros.rs
```diff
@@ -812,6 +812,117 @@ fn test_internally_tagged_enum() {
     );
 }
 
+#[test]
+fn test_internally_tagged_enum_with_default_variant() {
```
harrisonturton, line 816: Is this the best way to test the macro?

I had a dig around the codebase, and I felt it made sense, but unsure.

harrisonturton:
Ah, I've just found that a request for this feature [was rejected last year](https://github.com/serde-rs/serde/issues/1799#issuecomment-1019416025) because it's an uncommon use-case.

We ran into this recently (hence this patch), and others ran into it in #1799, #2231 and #1410. This feature is very helpful for evolving a message into a sum type in a backwards compatible way.

For example, allowing `GetUserRequestV2` to replace `GetUserRequestV1` without any compatibility issues:

```rust
#[derive(Deserialize)]
struct GetUserRequestV1 {
  user_id: String
}

#[derive(Deserialize)]
#[serde(tag = "type")]
enum GetUserRequestV2 {
  #[serde(default)]
  GetUserById { user_id: String },
  GetUserByEmail { email: String },
}
```

This is possible today, but afaict it requires workarounds that need [quite a bit of boilerplate](https://github.com/serde-rs/serde/issues/1799#issuecomment-624978919), rely on an [undocumented feature](https://stackoverflow.com/a/61219284), or use a custom derive implementation.

Is this worth reconsidering?

harrisonturton:
@dtolnay would love to help get this feature over the line. WDYT of these changes/approach?

MaxwellBo:
Hi @dtolnay, I'd be really interested in utilizing the behavior implemented by this PR - it really simplifies a large scale API migration I need to perform (and it seems like functionality a lot of people have been clamoring for). Was just wondering if you'd seen this PR, I noticed it hadn't had any activity in over a month?

serde/src/private/de.rs
```diff
@@ -885,7 +892,13 @@ mod content {
...
             }
             match tag {
-                None => Err(de::Error::missing_field(self.tag_name)),
+                None => match self.default_variant {
+                    Some(default) => Ok(TaggedContent {
+                        tag: default,
+                        content: Content::Map(vec),
+                    }),
+                    None => Err(de::Error::missing_field(self.tag_name)),
+                },
```
Mingun, line 901: Same here. `tag.or(self.default_variant)`

serde_derive/src/internals/ast.rs
```diff
@@ -156,6 +156,18 @@ fn enum_from_ast<'a>(
...
 
+    let mut has_default = false;
+    for variant in &variants {
+        if !variant.attrs.default() {
+            continue;
+        }
+        if has_default {
+            cx.error_spanned_by(&variant.ident, "only one variant can be marked as default");
+            break;
+        }
```
Mingun, line 167: Do not break after the first error so all errors can be found at once

serde/src/private/de.rs
```diff
@@ -851,11 +857,12 @@ mod content {
...
-            let tag = match try!(seq.next_element()) {
+            let tag = match seq.next_element()? {
                 Some(tag) => tag,
-                None => {
-                    return Err(de::Error::missing_field(self.tag_name));
-                }
+                None => match self.default_variant {
+                    Some(variant) => variant,
+                    None => return Err(de::Error::missing_field(self.tag_name)),
+                },
```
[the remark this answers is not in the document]
harrisonturton: You're right, that's much nicer :)

MaxwellBo:
@harrisonturton what happens when an unrecognized tag is provided? Does it resolve to the default, or does it error? _Could_ it error?

harrisonturton:
Closing because this change hasn't received attention in 2 years and I'm not actively working on it anymore. Would love for someone else to drive it! The feature would still be valuable.

PR #2488, opened 2023-06-30
make the explanation show up on GitHub and crates.io

szabgab:
For a newbie it is rather challenging to find out what needs to be added to the Cargo.toml file in order for the example to work. The information was already in the readme file, but it was hidden both in the default display of the README.md on GitHub and on [crates.io](https://crates.io/crates/serde)

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,6 @@ You may be looking for:
 
 ## Serde in action
 
-<details>
 <summary>
 Click to show Cargo.toml.
 <a href="https://play.rust-lang.org/?edition=2018&gist=72755f28f99afc95e01d63174b28c1f5" target="_blank">Run this code in the playground.</a>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -24,7 +24,6 @@ You may be looking for:
 
 ## Serde in action
 
-<details>
 <summary>
```
tavianator, line 27: I think you need to remove the `<summary>` too

README.md
```diff
@@ -24,11 +24,8 @@ You may be looking for:
 
 ## Serde in action
 
-<details>
-<summary>
 Click to show Cargo.toml.
```
CohenArthur, line 27:
```suggestion
```
I believe this line was used to summarize the `details` button/link thingy. At the moment it just shows up like regular text in front of the playground link, so it should probably go away

Issue #2497, opened 2023-07-09
Invalid Deserialize impl generated for unit struct with const generic parameter


dtolnay:
```rust
#[derive(serde::Deserialize)]
struct Generic<const N: u8>;
```

```console
error[E0107]: missing generics for struct `Generic`
 --> src/main.rs:2:8
  |
2 | struct Generic<const N: u8>;
  |        ^^^^^^^ expected 1 generic argument
  |
note: struct defined here, with 1 generic parameter: `N`
 --> src/main.rs:2:8
  |
2 | struct Generic<const N: u8>;
  |        ^^^^^^^ -----------
help: add missing generic argument
  |
2 | struct Generic<N><const N: u8>;
  |               +++
```

The generated code contains:

```rust
impl<'de, const N: u8> serde::Deserialize<'de> for Generic<N> {
    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
    where
        D: serde::Deserializer<'de>,
    {
        struct __Visitor;
        impl<'de> serde::de::Visitor<'de> for __Visitor {
            type Value = Generic;
            fn expecting(&self, formatter: &mut std::fmt::Formatter) -> std::fmt::Result {
                std::fmt::Formatter::write_str(formatter, "unit struct Generic")
            }
            fn visit_unit<E>(self) -> Result<Self::Value, E>
            where
                E: serde::de::Error,
            {
                Ok(Generic)
            }
        }
        serde::Deserializer::deserialize_unit_struct(deserializer, "Generic", __Visitor)
    }
}
```

In contrast, the generated impl for `struct Generic<const N: u8> {}` compiles successfully, and has the correct Visitor:

```rust
impl<'de, const N: u8> serde::de::Visitor<'de> for __Visitor<'de, N> {
    type Value = Generic<N>;
    ...
}
```

PR #2500, opened 2023-07-09
Allow `[derive(serde::Deserialize)]` for generic unit structs

Baptistemontan:
Allow to auto derive `Deserialize` generic unit structs such as 

```rust
#[derive(serde::Deserialize)]
struct Generic<const N: u8>;
```

see #2497


dtolnay (approved):
Thanks!

PR #2501, opened 2023-07-10
Eliminate 'de lifetime from visitors of const generic unit struct

dtolnay:
In #2500 I was trying to figure out whether the two PhantomData in `__Visitor` are necessary i.e. if there is any case that compiles with them and breaks without them. I came up with this one:

```rust
#[derive(Deserialize)]
#[serde(bound(deserialize = "[&'de str; N]: Copy"))]
struct GenericUnitStruct<const N: usize>;
```

```console
error[E0401]: can't use generic parameters from outer item
 --> src/main.rs:4:29
  |
3 | #[derive(Deserialize)]
  |          ----------- lifetime parameter from outer item
4 | #[serde(bound(deserialize = "[&'de str; N]: Copy"))]
  |                             ^^^^^^^^^^^^^^^^^^^^^ use of generic parameter from outer item
  |
  = note: for more information on higher-ranked polymorphism, visit https://doc.rust-lang.org/nomicon/hrtb.html
help: consider making the bound lifetime-generic with a new `'de` lifetime
  |
4 | #[serde(bound(deserialize = for<'de> "[&'de str; N]: Copy"))]
  |                             ++++++++
help: consider introducing lifetime `'de` here
  |
5 | struct GenericUnitStruct<'de, const N: usize>;
  |                          ++++

error: impl method assumes more implied bounds than the corresponding trait method
 --> src/main.rs:3:10
  |
3 | #[derive(Deserialize)]
  |          ^^^^^^^^^^^
  |
  = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
  = note: for more information, see issue #105572 <https://github.com/rust-lang/rust/issues/105572>
  = note: `#[deny(implied_bounds_entailment)]` on by default
```

### DIFF
--- a/serde_derive/src/de.rs
+++ b/serde_derive/src/de.rs
@@ -410,28 +410,17 @@ fn deserialize_unit_struct(params: &Parameters, cattrs: &attr::Container) -> Fra
     let this_type = &params.this_type;
     let this_value = &params.this_value;
     let type_name = cattrs.name().deserialize_name();
-    let (de_impl_generics, de_ty_generics, ty_generics, where_clause) =
-        split_with_de_lifetime(params);
-    let delife = params.borrowed.de_lifetime();
+    let (impl_generics, ty_generics, where_clause) = params.generics.split_for_impl();
+    let de_impl_generics = DeImplGenerics(params);
 
     let expecting = format!("unit struct {}", params.type_name());
     let expecting = cattrs.expecting().unwrap_or(&expecting);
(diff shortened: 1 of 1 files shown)


Baptistemontan:
For #2500 I reused as much code as I could from the code for "normal" struct (not unit, not tuple, not newtype), I not that much confident in my variance skills so I left the lifetime untouched.

Issue #2496, opened 2023-07-09
Revisit the representation of variant id for adjacently tagged enums


dtolnay:
Currently the `Serialize` impl for an untagged enum such as:

```rust
#[derive(Serialize, Deserialize)]
#[serde(tag = "t", content = "c")]
enum Enum {
    Variant(T),
    /* ... */
}
```

looks like this:

```rust
impl Serialize for Enum {
    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
    where
        S: Serializer,
    {
        match self {
            Enum::Variant(__field0) => {
                let mut __struct = serializer.serialize_struct("Enum", 2)?;
                __struct.serialize_field("t", "Variant")?;
                __struct.serialize_field("c", __field0)?;
                __struct.end()
            }
            /* ... */
        }
    }
}
```

i.e. the variant id is simply serialized using `impl Serialize for str`.

During deserialization, that "Variant" is deserialized by this type:

```rust
enum __VariantIdentifier {
    Variant,
    /* ... */
}

struct __VariantIdentifierVisitor;

const __VARIANTS: &'static [&'static str] = &["Variant", /* ... */];

impl<'de> Visitor<'de> for __VariantIdentifierVisitor {
    type Value = __VariantIdentifier;

    fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
        formatter.write_str("variant identifier")
    }

    fn visit_u64<E>(self, value: u64) -> Result<Self::Value, E>
    where
        E: serde::de::Error,
    {
        match value {
            0u64 => Ok(__VariantIdentifier::Variant),
            /* ... */
            _ => Err(E::invalid_value(
                serde::de::Unexpected::Unsigned(value),
                &"variant index 0 <= i < 1",
            )),
        }
    }

    fn visit_str<E>(self, value: &str) -> Result<Self::Value, E>
    where
        E: serde::de::Error,
    {
        match value {
            "Variant" => Ok(__VariantIdentifier::Variant),
            /* ... */
            _ => Err(E::unknown_variant(value, __VARIANTS)),
        }
    }

    fn visit_bytes<E>(self, value: &[u8]) -> Result<Self::Value, E>
    where
        E: serde::de::Error,
    {
        match value {
            b"Variant" => Ok(__VariantIdentifier::Variant),
            /* ... */
            _ => Err(E::unknown_variant(value, __VARIANTS)),
        }
    }
}

[...]

PR #2505, opened 2023-07-11
Revisit of the representation of adjacently tagged enums tag

Baptistemontan:
This PR is a follow on #2496.

This is a proof of concept to evaluate if such feature would be interesting. 

I've run some tests with `serde_json` and the output is the same, but still allow some format to take advantage of `variant_index` during serialization.

If it looks promising I can continue to work on it

### DIFF
--- a/serde_derive/src/de.rs
+++ b/serde_derive/src/de.rs
@@ -1510,6 +1510,14 @@ fn deserialize_adjacently_tagged_enum(
         }
     };
 
+    let variant_seed = quote! {
+        _serde::__private::de::AdjacentlyTaggedEnumVariantSeed::<__Field> {
+            tag: #tag,
+            variants: &VARIANTS,
+            fields_enum: _serde::__private::PhantomData
+        }
+    };
+
     let mut missing_content = quote! {
         _serde::__private::Err(<__A::Error as _serde::de::Error>::missing_field(#content))
     };
@@ -1557,6 +1565,10 @@ fn deserialize_adjacently_tagged_enum(
         _serde::de::MapAccess::next_key_seed(&mut __map, #tag_or_content)?
     };
 
+    let variant_from_map = quote! {
+        _serde::de::MapAccess::next_value_seed(&mut __map, #variant_seed)?
+    };
+
     // When allowing unknown fields, we want to transparently step through keys
     // we don't care about until we find `tag`, `content`, or run out of keys.
     let next_relevant_key = if deny_unknown_fields {
@@ -1602,11 +1614,11 @@ fn deserialize_adjacently_tagged_enum(
 
     let finish_content_then_tag = if variant_arms.is_empty() {
         quote! {
-            match _serde::de::MapAccess::next_value::<__Field>(&mut __map)? {}
+            match #variant_from_map {}
         }
     } else {
         quote! {
-            let __ret = match _serde::de::MapAccess::next_value(&mut __map)? {
+            let __ret = match #variant_from_map {
                 // Deserialize the buffered content now that we know the variant.
                 #(#variant_arms)*
             }?;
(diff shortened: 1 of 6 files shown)


serde_derive/src/de.rs
```diff
@@ -1313,6 +1313,56 @@ fn prepare_enum_variant_enum(
     (variants_stmt, variant_visitor)
 }
 
+fn prepare_adjacently_tagged_enum_variant_enum(tag: &str) -> Fragment {
+    let expecting = format!("enum {}", tag);
+    quote_block! {
+        #[doc(hidden)]
+        struct __FieldVariant(pub __Field);
```
dtolnay, line 1320: I would expect this whole type to go in `serde::__private::de` instead of emitting a whole new type and set of trait impls into every single use of adjacently tagged enums.

```rust
pub struct AdjacentlyTaggedEnumVariant<F> {
    tag: &'static str,
    variants: &'static [&'static str],
    fields_enum: PhantomData<F>,
}

impl<'de, F> DeserializeSeed<'de> for AdjacentlyTaggedEnumVariant<F>
where
    F: Deserialize<'de>,
{
    type Value = F;
    ...
}
```

dtolnay (commented):
I can confirm this looks promising. Thanks!

serde_derive/src/de.rs
Baptistemontan: that's what I wanted to do in the first place, but I just looked at `Deserialize` and because it don't take `self` I just moved on  and did not thought of DeserializeSeed, I did something like this for serialization

Baptistemontan:
I did not realised that `DeserializeSeed` was imported only with `feature = "std"` so I had some trouble fixing the import.

Does something like this seams more appropriate ?

Baptistemontan:
Is there a crate implementing a binary format that use `variant_index` to serialize enums and support adjacently tagged enums deserialization ?

I've looked into `postcard` but it does'nt support `deserialize_identifier` so adjacently tagged enums are not supported (maybe #2475 broke it?).
And I looked at `ciborium` but it serialize enums with the variant name.

dtolnay:
> Is there a crate implementing a binary format that use `variant_index` to serialize enums and support adjacently tagged enums deserialization ?

I don't know of any. But if there is, then yes, we would have heard about it breaking from #2475.

dtolnay:
This PR is still marked as draft. Are there concrete changes planned (what are they?), or was this supposed to be ready for review?

Baptistemontan:
@dtolnay I just forgot to open it for review, there is no changes planned

Baptistemontan:
There is conflicts needed to be resolved, I'm gonna need some time to understand why thoses changes has been made and what can I do to not break anything.

Baptistemontan:
All conflicts were related to the removal of `try!` in #2539

dtolnay (approved):
Thank you!

Mingun:
This PR has some breaking changes that results in failing tests in quick-xml (https://github.com/tafia/quick-xml/issues/630)

jost-s:
Since this is a breaking change, why was it released as a patch instead of a minor version?

### v1.0.180
In my case, an enum
```rust
#serde(tag = "type", content = "data")]
enum Message {
  A(i32),
}

let m = Message::A(100);
```

was rmp_serded to
```rust
[130, 164, 116, 121, 112, 101, 161, 65, 164, 100, 97, 116, 97, 100]
```
which the JS msgpack package converts to JSON as
```
{
  "type": "A",
  "data": 100
}
```
with serde v1.0.180.

### v1.0.181
Same enum and attributes, rmp_serde spits out
```rust
[130, 164, 116, 121, 112, 101, 129, 161, 65, 192, 164, 100, 97, 116, 97, 100]
```

and JS msgpack converts to
```json
{
  "type": {
    "A": null
  },
  "data": 100
}
```

### serde_json

What is further confusing this is that converting this to JSON with serde_json results in
```json
{\"type\":\"A\",\"data\":100}
```
regardless of the serde version.

Can you please shed light on this?

PR #2509, opened 2023-07-13
Enable the use of `#[serde(default)]` and `#[serde(skip_serialize_if)]` for newtype variants in adjacently tagged enums

### LINKED ISSUE #2508: Improve #[serde(default, skip_serializing_if = "xxx")] behavior on Enum variant value.
Currently, #[serde(default, skip_serializing_if = "is_default"] below has no effects.

```rust
use anyhow::Result;
use serde_derive::{Serialize, Deserialize};

#[serde(tag = "type", content = "settings", rename_all = "lowercase")]
pub enum TypeWithSettings {
    Foo(
        #[serde(default, skip_serializing_if = "is_default")]
        FooSettings
    ),
    Bar(
        #[serde(default, skip_serializing_if = "is_default")]
        BarSettings
    ),
}

#[derive(Serialize, Deserialize, Debug, Default, PartialEq)]
pub struct FooSettings {
    value: String,
}

#[derive(Serialize, Deserialize, Debug, Default, PartialEq)]
pub struct BarSettings {
    value: u32,
}

fn main() {
    println!("calling de():");
    if let Err(e) = de() {
        println!("Error: {:?}", e);
    }

    println!("calling ser():");
    if let Err(e) = ser() {
        println!("Error: {:?}", e);
    }
}

fn de() -> Result<()> {
    let json = r#"
    {
        "type": "foo"
    }"#;
    let settings: TypeWithSettings = serde_json::from_str(json)?;
    println!("{:?}", settings);
    Ok(())
}

fn ser() -> Result<()> {
    let settings = TypeWithSettings::Foo(FooSettings::default());
    println!("{}", serde_json::to_string_pretty(&settings)?);
    Ok(())
}

pub fn is_default<T: Default + PartialEq>(t: &T) -> bool {
    t == &T::default()
}
```

What i expect the code is to deserialize the json string with a default FooSettings instance as the Foo varant value, and serialize the enum without "settings" field.

But currelty the output is:
```
calling de():
Error: missing field `settings` at line 4 column 5
calling ser():
{
  "type": "foo",
  "settings": {
    "value": ""
  }
}
```

Baptistemontan:
This PR enable the use of `default` and `skip_serialize_if` for newtype variants in adjacently tagged enums when deriving `Serialize` and `Deserialize`.

close #2508.

### DIFF
--- a/serde_derive/src/de.rs
+++ b/serde_derive/src/de.rs
@@ -1525,10 +1525,23 @@ fn deserialize_adjacently_tagged_enum(
                     _serde::__private::Ok(#this_value::#variant_ident)
                 },
                 Style::Newtype if variant.attrs.deserialize_with().is_none() => {
-                    let span = variant.original.span();
-                    let func = quote_spanned!(span=> _serde::__private::de::missing_field);
-                    quote! {
-                        #func(#content).map(#this_value::#variant_ident)
+                    let field = &variant.fields[0];
+                    match field.attrs.default() {
+                        attr::Default::Default => {
+                            let span = field.original.span();
+                            let func = quote_spanned!(span=> _serde::__private::Default::default);
+                            quote!(Ok(#this_value::#variant_ident(#func())))
+                        }
+                        attr::Default::Path(path) => {
+                            quote!(Ok(#this_value::#variant_ident(#path())))
+                        }
+                        attr::Default::None => {
+                            let span = variant.original.span();
+                            let func = quote_spanned!(span=> _serde::__private::de::missing_field);
+                            quote! {
+                                #func(#content).map(#this_value::#variant_ident)
+                            }
+                        }
                     }
                 }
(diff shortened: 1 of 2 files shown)


serde_derive/src/de.rs
```diff
@@ -1525,10 +1525,23 @@ fn deserialize_adjacently_tagged_enum(
                     _serde::__private::Ok(#this_value::#variant_ident)
                 },
                 Style::Newtype if variant.attrs.deserialize_with().is_none() => {
-                    let span = variant.original.span();
-                    let func = quote_spanned!(span=> _serde::__private::de::missing_field);
-                    quote! {
-                        #func(#content).map(#this_value::#variant_ident)
+                    let field = &variant.fields[0];
```
Mingun, line 1528: I think, this will not work for skipped fields

Issue #1833, opened 2020-06-10
Add `#[serde(ref_into = "RefIntoType")]` where `&T: Into<RefIntoType>, RefIntoType: Serialize`


timotree3:
I suggest adding a container-level attribute `ref_into`. With this feature, the following two examples would be equivalent:
```rust
// Example one

pub struct Foo;

impl Serialize for Foo {
    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error> where S: Serializer {
        let val: RefIntoType = self.into();
        val.serialize(serializer)
    }
}
```
```rust
// Example two

#[derive(Serialize)]
#[serde(ref_into = "RefIntoType")]
pub struct Foo;
```
This feature would be applicable whenever `&Foo` implements `Into<RefIntoType>` and `RefIntoType` implements `Serialize`.

There's already a very similar shortcut [`#[serde(into = "IntoType")]`](https://serde.rs/container-attrs.html#into) but it clones the value before converting and therefore would be wasteful in cases like this where conversion only requires a reference.

As always, the exact chosen name `ref_into` is subject to bikeshedding.

Mingun:
This should be easy to implement, need just introduce and use special `RefInto` trait: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=ace054282891ad2e26a184f6487dc88a

JakobDegen:
I also have a use case for this, but I don't think restricting it to `&T: Into` or such makes sense. Just having `into_with = "path"` where `path` is some `fn(&Self) -> T` where `T: Serialize` seems like a more general solution.

dtolnay:
I think I would prefer to leave this use case to handwritten Serialize impls, instead of introducing new dedicated `ref_into` or `into_with` attributes.

As you showed above, these tiny Serialize impls are trivial to handwrite.

PR #2511, opened 2023-07-15
Add `try_into`, `borrowed_into`, and `borrowed_try_into` attributes

oblique:
For the `borrowed_*` the user needs to implement `impl<'a> Into<IntoType> for &'a Type`. With this the cloing of the value will not happen.


dtolnay (commented):
Thanks for the PR!

An equivalent attribute was proposed in the past in #1833, and I would prefer not to add it.

oblique:
@dtolnay what about `try_into`? I can cherry-pick that in another PR

dtolnay:
I feel the same about all of them &mdash; handwritten Serialize impl is fine for these situations.

PR #2513, opened 2023-07-18
Do not write skipped fields in deserialize_in_place()

### LINKED ISSUE #2512: deserialize_in_place fills skipped fields with default
Hi, thanks for this awesome library! I have some questions regarding in-place deserialization.

I'm wondering why we are filling skipped fields in `deserialize_in_place()`. Skipped fields are ignored for maps but not for sequences. Is there a reason for it or should the filling with default be removed? This is the relevant code that writes default values into the in-place struct:

https://github.com/serde-rs/serde/blob/03da66c805499a9e79a5b6cb11fa21ef6da636eb/serde_derive/src/de.rs#L805-L810

My use case is the following: I have a struct (`MixedContent` below) which contains fields that I want to serialize and fields that cannot be serialized and also don't implement `Default`. Something like this:

```rust
// serde_derive with Feature deserialize_in_place
struct MixedContent {
    #[serde(skip, default = "panic")]
    non_serializable: NonSerializable,

    serializable: Serializable,
}

fn panic() -> NonSerializable {
    panic!("deserialize not implemented")
}

struct NonSerializable {
    // ...
}

struct Serializable {
    // ...
}
```

Since I cannot implement `Default` for the `NonSerializable` but for `Serializable`, my idea was to use the `deserialize_in_place` functionality with skipped fields. With that I can create a custom constructor with more fancy logic and then use `deserialize_in_place()` afterwards on the struct. I think, using some `DeserializeSeed` is too complicated because it seems I need to implement it manually. Also custom home-grown derive macros doesn't feel right to achieve my scenario.

What I need to adjust from basic serde usage is

1. Generate a proper `deserialize_in_place`: Easy, just enable the feature flag
2. Disable the normal `deserialize`: Can be done via a `serde(default = ...)` attribute which never returns anything but fulfills the trait and makes the compiler happy
3. Deactivate all struct field writes of skipped fields: Doesn't work because the sequence code generation still contains them, see above

I'm willing to create a PR to remove the skipped field filling in sequences but wanted to make sure, nothing else gets broken by it (Chesterton's Fence). Do you have suggestions?

h3ndrk:
This PR actually skips fields in sequences in `deserialize_in_place()` as explained in #2512.

Fixes #2512

Formatting didn't work in this file so I formatted manually...


h3ndrk:
I rebased to the latest `master`. What's the state? Any chance to get this merged?

twvd:
I'm also very interested in having this merged, for the same reasons as stated in https://github.com/serde-rs/serde/issues/2512

PR #2514, opened 2023-07-18
Add experiment to produce precompiled builds of serde_derive

dtolnay:
```toml
# Cargo.toml

[dependencies]
serde = "1"  # no features=["derive"]
serde_derive-x86_64-unknown-linux-gnu = "1.0.171-alpha.0"
```

```rust
use serde_derive::{Deserialize, Serialize};

pub struct MyStruct {
    /* ... */
}
```

If this works, it needs to be productionized further using any of the following techniques:

- [Cosmopolitan Libc](https://github.com/jart/cosmopolitan) / αcτµαlly pδrταblε εxεcµταblε
- `[target.x86_64-unknown-linux-gnu.dependencies]` to make the macro forward to a different crate on different targets
- Precompile multiple different targets into the same crate, and detect the current platform during macro expansion time

Screenshot from 2023-07-18 13-44-10

### DIFF
--- a/precompiled/proc-macro2/src/watt/mod.rs
+++ b/precompiled/proc-macro2/src/watt/mod.rs
@@ -1,0 +1,196 @@
+pub mod buffer;
+pub mod bytecode;
+
+use crate::watt::buffer::{InputBuffer, OutputBuffer};
+use crate::watt::bytecode::Bytecode;
+use crate::{Delimiter, Group, Ident, Literal, Punct, Spacing, Span, TokenStream, TokenTree};
+use std::str::FromStr;
+
+pub enum Kind {
+    Group(Delimiter),
+    Ident,
+    Punct(Spacing),
+    Literal,
+}
+
+pub enum Identity {}
+
+impl Identity {
+    pub const RESPANNED: u32 = 1 << 31;
+    pub const NOVEL: u32 = u32::MAX;
+}
+
+impl Span {
+    fn is_call_site(&self) -> bool {
+        self.lo == 0 && self.hi == 0
+    }
+}
+
+fn post_increment(counter: &mut u32) -> impl FnMut() -> u32 + '_ {
+    || {
+        let value = *counter;
+        *counter += 1;
+        value
+    }
+}
+
+pub fn load(buf: &mut InputBuffer) -> TokenStream {
+    let mut span_counter = 0;
+    let mut next_span = post_increment(&mut span_counter);
+    let mut next_span = || {
+        let next = next_span();
+        Span { lo: next, hi: next }
+    };
+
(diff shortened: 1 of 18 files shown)


dtolnay:
### Expansion time benchmark

I tested against https://github.com/dtolnay/clang-ast/blob/0.1.19/tests/exhaustive.rs to quantify the overhead of process spawning and serialization. That integration test contains 286 calls to `derive(Deserialize)`.

```diff
 [dev-dependencies]
-serde_derive = "1.0.171"
+serde_derive-x86_64-unknown-linux-gnu = "1.0.171-alpha.1"
```

Here is the command I measured:

```console
$ time cargo expand --ugly --test exhaustive >expand.rs
```

On my machine, the macro expansion takes 1.06 seconds with `serde_derive` and 1.91 seconds with `serde_derive-x86_64-unknown-linux-gnu`. (They produce the same code.)

That is about 3 milliseconds per macro call.

Mingun:
Interesting. Could you explain, what is improved, because in you last comment there is slowdown!? Some statistics also would be interesting.

dtolnay:
That specifically only measures macro expansion. Overall build time for serde_derive goes from 9.2 seconds to 0.8 seconds (at `-j1`) or 4.9 seconds to 0.48 seconds (at `-j8`) which is 10&times; improvement.

safinaskar:
> Cosmopolitan Libc

@dtolnay, if this feature will be ever re-enabled, then, please, don't use Cosmopolitan. Cosmopolitan is just a hack. It has a lot of problems:
- Despite advertised as universal, it seems to work on x86 only on some limited amount of platforms. So it is no better than, say, 3 separate statically linked binaries for very same platforms
- It relies on Thompson shell hack, i. e. shell scripts without shebang. So if you run the binary using `execve` you are probably out of lack
- Sometimes these binaries are flagged by Windows antiviruses

Overall, this is very hacky solution for a non-problem (i. e. what is wrong with just creating some amount of separate binaries?)

precompiled/proc-macro2/src/watt/mod.rs
```diff
@@ -0,0 +1,196 @@
...
+
+    let [mut group_counter, mut ident_counter, mut punct_counter, mut literal_counter] = [0; 4];
+    let mut next_group = post_increment(&mut group_counter);
+    let mut next_ident = post_increment(&mut ident_counter);
+    let mut next_punct = post_increment(&mut punct_counter);
+    let mut next_literal = post_increment(&mut literal_counter);
+
+    let mut trees = Vec::new();
+    while !buf.is_empty() {
+        match match buf.read_u8() {
```
martin-g, line 53: double match

PR #2515, opened 2023-07-18
Add `#[serde(with)]` for containers

oblique:
Why this is useful:

* It will enable `#[serde_with::serde_as(as = "SomeSerdeWithUtil")]` to be applied on a container.

* It can be easily used with code-generators such as [prost-build](https://docs.rs/prost-build). For example:

```rust
    prost_build::Config::new()
        .type_attribute(
            "my_messages.MyMessageType",
            "#[derive(Serialize, Deserialize)] #[serde(with = \"my_message_type\")]",
        )
        .complie_protos(&["my_messages.proto"])
        .unwrap();
```


dtolnay (commented):
Thanks for the PR!

I would prefer not to build this into serde_derive. But if it's helpful to have reusable code for producing these 1-line impls, someone could factor the needed parts out into its own macro crate.

PR #2524, opened 2023-07-23
Add wasm_precomp feature

jsgf:
Add wasm_precomp feature

This uses wasm32-wasi for the pre-compiled helper process, which means
- it's platform independent, and
- it's sandboxed so that there's a stronger guarantee that the process
  can't do anything other than read/write stdin/stdout.

The downsides are:
- ~~somewhat slower for proc-macro processing (2x?). This is mitigated by making the helper process persist for multiple macro invocations~~
- needs wasmtime installed in the path

There doesn't seem to be a performance impact from this. Using a persistent process, using wasm seems to
have a minimal performance difference from native code:

Native x86:
```
$ hyperfine ' cargo expand --ugly --test exhaustive >expand.rs'
Benchmark 1:  cargo expand --ugly --test exhaustive >expand.rs
  Time (mean ± σ):      1.404 s ±  0.175 s    [User: 0.543 s, System: 0.097 s]
  Range (min … max):    1.195 s …  1.820 s    10 runs
```

Wasm:
```
$ hyperfine ' cargo expand --ugly --test exhaustive >expand.rs'
Benchmark 1:  cargo expand --ugly --test exhaustive >expand.rs
  Time (mean ± σ):      1.554 s ±  0.341 s    [User: 0.547 s, System: 0.087 s]
  Range (min … max):    1.247 s …  2.429 s    10 runs
```

(This is stacked on the "persistent" branch/PR https://github.com/serde-rs/serde/pull/2523, but is logically independent from it.)

### DIFF
--- a/precompiled/.gitignore
+++ b/precompiled/.gitignore
@@ -1,1 +1,1 @@
-/serde_derive/serde_derive-x86_64-unknown-linux-gnu
+/serde_derive/serde_derive-*
--- a/precompiled/bin/main.rs
+++ b/precompiled/bin/main.rs
@@ -24,22 +24,36 @@ unsafe impl GlobalAlloc for MonotonicAllocator {
 }
 
 fn main() {
-    let mut buf = Vec::new();
-    io::stdin().read_to_end(&mut buf).unwrap();
-
-    let mut buf = InputBuffer::new(&buf);
-    let derive = match buf.read_u8() {
-        0 => serde_derive::derive_serialize,
-        1 => serde_derive::derive_deserialize,
-        2 => {
-            serde_derive::DESERIALIZE_IN_PLACE.store(true, Ordering::Relaxed);
-            serde_derive::derive_deserialize
+    let mut stdin = io::stdin();
+    let mut stdout = io::stdout();
+    loop {
+        let mut len_le32 = [0u8; 4];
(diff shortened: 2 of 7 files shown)


jsgf:
That clang-ast test is probably the best case for wasm if most of the cost is in process startup. It shows that the actual performance of the proc-macro logic itself isn't dominant (unless wasmtime is actually native-code equiv).

dtolnay (commented):
Thanks for the PR!

I would prefer not to release this, because I can't see this being something that anyone would want to opt in to (making their build dependent on users having `wasmtime` installed in $PATH, without even the upside of eliminating serde_derive's dependency on syn).

But it is terrific to have the benchmarks that validate Wasm as a viable implementation for macros.

The next step I would like on precompiled macros is for the Rust distribution to ship a Wasm runtime as rustup component &mdash; realistically wasmtime today, but abstracted to a minimal API that just lets us feed in a Wasm blob and collection of host fns. If that's available and somehow selectable in `cfg` expressions, I would be open to serde_derive using it.

jsgf:
Yes, agreed, this shouldn't be merged. It's just intended to be a proof-of-concept so I could get a sense of the relative performance, esp in the context of #2523.

PR #2525, opened 2023-07-23
UPDATED: Integer/boolean tags for internally/adjacently tagged enums

Astavie:
this is basically just #2056 but synced with the latest serde master


7sDream:
@dtolnay 

Is there still any blocking point for this PR after update and pass CI? 
Using integers as tags is the feature I am most looking forward to and find useful.

Astavie:
The new adjacently tagged enum code really put a wrench into this. It effectively now serializes tags into a unit enum like:

```rust
enum EnumTag {
    A,
    B,
    // ...
}
```

This way some (de)serializers can use the enum index instead of the name for smaller sizes. However there is no (de)serialize method that takes a variant index and a generic data type. The closest is `serialize_enum_newtype_variant` but that wraps the newtype.

A new (de)serialize method could be created for this, something like `serialize_enum_unit_newtype_variant`. However that would be a breaking change (unless given a default impl I suppose).

Astavie:
@dtolnay is making a new base serialization method a good solution here? It could even expand enum tags to more than booleans and integers later on and add support for non-string tags for unit enums.

dtolnay (commented):
I would not accept a new Serializer method.

I have not gotten a chance to look over the code in the PR, but in response to the last 2 comments: I don't immediately see why recent adjacently tagged enum changes (referring to #2505/#2496) would be an obstacle. Enums that use all integer tags need to serialize the tag using the appropriately sized integer type's Serialize impl, and deserialize using that same integer type. Similarly enums that use all boolean tags need to serialize using serialize_bool and deserialize using deserialize_bool. Enums that use a mix of integer and boolean and string tags, if this is supported, need to deserialize using deserialize_any. It's not clear to me that the solution would have been any different prior to #2505.

Astavie:
It not necessarily an obstacle, but it does mean I need to do a slight rework of the (de)serialization of tags. Your implementation suggestion actually makes a lot of sense, I'll work on that one.
Also yes, a mix is currently supported.

Astavie:
Alright, that should be up-to-date again now

serde/src/private/de.rs
```diff
@@ -2587,6 +2587,28 @@ where
...
+    E: Error,
+{
+    type Deserializer = <i64 as IntoDeserializer<'de, E>>::Deserializer;
+
+    fn from(self) -> Self::Deserializer {
+        self.into_deserializer()
+    }
+}
+
+impl<'de, E> IdentifierDeserializer<'de, E> for bool
```
dtolnay, line 2601: If I understand correctly (based on what test fails if I delete this impl), this is to support the following?

```rust
#[derive(Deserialize)]
#[serde(field_identifier)]
enum Field {
    Aaa,
    Bbb,
    Other(bool),
}
```

Is that a necessary part of this PR? I'd prefer if this PR had no impact on `field_identifier` generated code, since integer/boolean continue to not be supported as a field name, and do not have a compelling justification as enum tags do.

In general I would like this PR to avoid introducing support for anything other than what is required for `rename = 1` and `rename = true` on adjacently tagged enum variants and internally tagged enum variants. That includes no `serde(variant_identifier)` changes for now; those can be independently motivated if someone would like.

serde_derive/src/internals/ast.rs
```diff
@@ -36,6 +36,42 @@ pub struct Variant<'a> {
...
+
+impl VariantMix {
+    pub fn from_single(variant: &Variant) -> Self {
+        match variant.attrs.name().serialize_name() {
+            attr::VariantName::String(_) => Self::OnlyStrings,
+            attr::VariantName::Integer(_) => Self::OnlyIntegers,
+            attr::VariantName::Boolean(_) => Self::OnlyBooleans,
+        }
+    }
+    pub fn from_multiple(variants: &[Variant]) -> Self {
```
dtolnay, line 56: Neither of these is used from outside of this file. Please remove `pub`.

serde_derive/src/internals/ast.rs
```diff
@@ -66,7 +102,10 @@ impl<'a> Container<'a> {
         let mut attrs = attr::Container::from_ast(cx, item);
 
         let mut data = match &item.data {
-            syn::Data::Enum(data) => Data::Enum(enum_from_ast(cx, &data.variants, attrs.default())),
+            syn::Data::Enum(data) => Data::Enum(
+                VariantMix::Any,
+                enum_from_ast(cx, &data.variants, attrs.default()),
+            ),
```
dtolnay, line 108: The code below this is only applying rename_all rules. That does not change the kind of a variant tag. The correct `VariantMix` can be set right here. We do not need to use `Any` as a placeholder.

Alternatively it may be more convenient to not even store `VariantMix` and avoid needing all those `_` placeholders throughout the rest of serde_derive. After checking has happened, a correct `VariantMix` can be computed efficiently from just the Vec\<Variant\> by looking at only the first variant's name.

serde_derive/src/internals/ast.rs
```diff
@@ -36,6 +36,42 @@ pub struct Variant<'a> {
...
+pub enum VariantMix {
+    OnlyStrings,
+    OnlyIntegers,
+    OnlyBooleans,
+    Any,
+}
+
+impl VariantMix {
+    pub fn from_single(variant: &Variant) -> Self {
+        match variant.attrs.name().serialize_name() {
```
dtolnay, line 50: This is using `serialize_name()`. That is not the right name to pay attention to during generating a Deserialize impl.

`#[serde(rename(serialize = 1, deserialize = false))]`

serde/src/private/ser.rs
```diff
@@ -1356,6 +1356,25 @@ where
...
+}
+
+impl Serialize for VariantName {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        match self {
+            VariantName::String(s) => serializer.serialize_str(s),
+            VariantName::Integer(i) => serializer.serialize_i64(*i),
```
dtolnay, line 1372: I am not thrilled about using i64 as the one true type of integer enum tags. For example if someone writes `serde(rename = 100u16)` it would be weird for that to serialize using i64. I'd like to consider respecting the integer suffix, and for enums where all the integers are unsuffixed, use the smallest type that fits all the given values. We do this for extern enums in https://github.com/dtolnay/cxx and it seems to match what people expect.

serde_derive/src/internals/attr.rs
```diff
@@ -129,25 +129,71 @@ impl<'c, T> VecAttr<'c, T> {
...
+
+    pub fn as_bool(&self) -> Option<&bool> {
+        match self {
+            VariantName::Boolean(b) => Some(b),
+            _ => None,
+        }
+    }
+}
+
+impl ToString for VariantName {
```
dtolnay, line 162: https://doc.rust-lang.org/1.71.0/std/string/trait.ToString.html: _"ToString shouldn’t be implemented directly"_

serde_derive/src/internals/attr.rs
```diff
@@ -192,6 +238,13 @@ impl Name {
     }
 }
 
+pub type VariantNames = Names<VariantName>;
+pub type Name = Names<String>;
```
dtolnay, line 242: Is there a more harmonious pair of names to use here? What is `Name` used for &mdash; field names? If so, `FieldNames`?

serde_derive/src/internals/attr.rs
```diff
@@ -1462,6 +1528,88 @@ fn get_lit_str(
...
+fn get_variant_name(
+    cx: &Ctxt,
+    attr_name: Symbol,
+    meta: &ParseNestedMeta,
+) -> syn::Result<Option<VariantName>> {
+    get_variant_name2(cx, attr_name, attr_name, meta)
+}
+
+// meta_item_name is unused but is present to comply with the contract for
+// get_ser_and_de.
```
dtolnay, line 1576: Looks like it is used below, after all.

serde_derive/src/internals/attr.rs
```diff
@@ -1462,6 +1528,88 @@ fn get_lit_str(
...
+    meta_item_name: Symbol,
+    meta: &ParseNestedMeta,
+) -> syn::Result<Option<VariantName>> {
+    let expr: syn::Expr = meta.value()?.parse()?;
+    let mut value = &expr;
+    while let syn::Expr::Group(e) = value {
+        value = &e.expr;
+    }
+    if let Ok(lit) = try_get_lit_str(value) {
+        Ok(Some(VariantName::String(lit.value())))
```
dtolnay, line 1589: This is missing a check for `lit.suffix().is_empty()`.

It can go in `try_get_lit_str`.

serde_derive/src/de.rs
```diff
@@ -1,5 +1,5 @@
 use crate::fragment::{Expr, Fragment, Match, Stmts};
-use crate::internals::ast::{Container, Data, Field, Style, Variant};
+use crate::internals::ast::{Container, Data, Field, Style, Variant, VariantMix};
 use crate::internals::{attr, replace_receiver, ungroup, Ctxt, Derive};
```
dtolnay, line 3:
```suggestion
use crate::internals::attr::{self, VariantName};
use crate::internals::{replace_receiver, ungroup, Ctxt, Derive};
```

serde_derive/src/ser.rs
```diff
@@ -505,7 +513,13 @@ fn serialize_externally_tagged_variant(
     cattrs: &attr::Container,
 ) -> Fragment {
     let type_name = cattrs.name().serialize_name();
-    let variant_name = variant.attrs.name().serialize_name();
+
+    let variant_name = match variant.attrs.name().serialize_name() {
+        VariantName::String(name) => name,
+        // An externally tagged variant with a non-string name will fail to
+        // check, so this branch _should_ be unreachable.
```
dtolnay, line 520: If you are confident about this, it should come across that way in the comment. If you are not confident about this, then this code would need to be changed.

serde_derive/src/internals/check.rs
```diff
@@ -408,6 +409,43 @@ fn check_transparent(cx: &Ctxt, cont: &mut Container, derive: Derive) {
...
+    let details = match &cont.data {
+        Data::Enum(_, _) => match cont.attrs.tag() {
+            TagType::Adjacent { .. } | TagType::Internal { .. } => return,
+            TagType::External => "externally tagged enums",
+            TagType::None => "untagged enums",
+        },
+        Data::Struct(_, _) => return,
+    };
+
+    if let Data::Enum(_, variants) = &cont.data {
```
dtolnay, line 423: This `if` is always true.

It would be better to capture the required `variants` in the previous match above.

serde_derive/src/internals/check.rs
```diff
@@ -408,6 +409,43 @@ fn check_transparent(cx: &Ctxt, cont: &mut Container, derive: Derive) {
...
+        for v in variants {
+            let name = v.attrs.name();
+            let ser_name = name.serialize_name();
```
dtolnay, line 426: This is only looking at serialize name.

serde_derive/src/ser.rs
```diff
@@ -599,11 +614,12 @@ fn serialize_internally_tagged_variant(
 
     match effective_style(variant) {
         Style::Unit => {
+            let variant_name = serialize_variant_name(&variant_name);
             quote_block! {
                 let mut __struct = _serde::Serializer::serialize_struct(
                     __serializer, #type_name, 1)?;
                 _serde::ser::SerializeStruct::serialize_field(
-                    &mut __struct, #tag, #variant_name)?;
+                    &mut __struct, #tag, &#variant_name)?;
```
dtolnay, line 622: This should not need to go through `_serde::__private::ser::VariantName`. You can directly serialize `&true` or `&1` as the value here.

Same comment may apply elsewhere in this file. Please use `VariantName` only where needed for `serialize_tagged_newtype` calls.

test_suite/tests/ui/rename/non_string_on_externally_tagged.stderr
```diff
@@ -0,0 +1,6 @@
+error: #[serde(rename)] must use a string name in externally tagged enums
+ --> $DIR/non_string_on_externally_tagged.rs:5:5
+  |
+5 | /     #[serde(rename = 1)]
+6 | |     A,
+  | |_____^
```
dtolnay, line 6: It would be better to highlight the `rename = 1` attribute for this error.

```console
          #[serde(rename = 1)]
                  ^^^^^^^^^^
```

test_suite/tests/ui/rename/non_string_on_untagged_enum.stderr
```diff
@@ -0,0 +1,6 @@
+error: #[serde(rename)] must use a string name in untagged enums
+ --> $DIR/non_string_on_untagged_enum.rs:6:5
+  |
+6 | /     #[serde(rename = 1)]
+7 | |     A,
+  | |_____^
```
dtolnay, line 6: Same here, please put the span on `rename = 1`.

dtolnay (changes requested):
Thank you for getting this moving.

I have not looked at any deserialization code yet, but a few comments on the rest.

Astavie:
I've now done most of the simple changes. The rest of the requested changes will be completed on a later date

xamgore:
@Astavie, that's an amazing piece of work, well done! May we help you somehow?

Astavie:
@xamgore 
There still needs to be support for different type of integer tags (see https://github.com/serde-rs/serde/pull/2525#discussion_r1285288903). This requires a bit of a refactor on top of the refactor i've already done. If you're able to help with that it's much appreciated as I've had less time to devote to this.
For the rest, tidying up some error messages as discussed before.

Astavie:
Support for marking specific integer types has been added. Specifically: `i8`, `i16`, `i32`, `i64`, `u8`, `u16`, `u32`, `u64`

Rules for unmarked integers:
- If all variants are unmarked integers, the smallest fitting integer type is chosen
- If there exist variants with one integer type and the rest unmarked, that type will be chosen
- If there is a mix of integer types or there are string variants or boolean variants, it will default to `i64`

jenr24-architect:
Is there ongoing work for this PR? I can help push this one over the line, as these features are highly valuable to my company. I've pulled @Astavie 's branch, merged master, and ran the test suite thus far. Any guidance would be greatly appreciated @dtolnay

Astavie:
iirc, the only work required now are moving the highlight spot with the errors mentioned by @dtolnay above, though they also mentioned not having reviewed the deserialization code yet

IriaSomobu:
Is it broken?

Cargo.toml
```toml
[package]
name = "serde_tagged_enum"
version = "0.1.0"
edition = "2021"

[dependencies]
serde = { git = "https://github.com/Astavie/serde", branch = "integer-tags-for-enums", features = ["derive"] }
serde_json = "1"
```

main.rs
```rust

#[derive(Debug, Default, Serialize, Deserialize)]
pub struct AttachmentUploadEntry {
    pub id: u32,
    pub upload_filename: String,
    pub upload_url: String,
}

fn main() {
    let data: String = "{}".to_string();
    let mut json: AttachmentUploadEntry = serde_json::from_str(&data).unwrap();
}
```

Error:
```
error[E0277]: the trait bound `AttachmentUploadEntry: serde::de::Deserialize<'_>` is not satisfied
    --> src/main.rs:12:43
     |
12   |     let mut json: AttachmentUploadEntry = serde_json::from_str(&data).unwrap();
     |                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `serde::de::Deserialize<'_>` is not implemented for `AttachmentUploadEntry`
     |
     = note: for local types consider adding `#[derive(serde::Deserialize)]` to your `AttachmentUploadEntry` type
     = note: for types from other crates check whether the crate offers a `serde` feature flag
     = help: the following other types implement trait `serde::de::Deserialize<'de>`:
               &'a Path
               &'a [u8]
               &'a str
               ()
               (T,)
               (T0, T1)
               (T0, T1, T2)
               (T0, T1, T2, T3)
             and 142 others
note: required by a bound in `serde_json::from_str`
    --> /home/somobu/.cargo/registry/src/index.crates.io-6f17d22bba15001f/serde_json-1.0.120/src/de.rs:2676:8
     |
2674 | pub fn from_str<'a, T>(s: &'a str) -> Result<T>
     |        -------- required by a bound in this function
2675 | where
2676 |     T: de::Deserialize<'a>,
     |        ^^^^^^^^^^^^^^^^^^^ required by this bound in `from_str`
```

Astavie:
@IriaSomobu tried this myself, it's because serde_json uses the regular version of serde and not my fork, leading to serde_json not thinking the struct has a Deserialize implementation. you can fix this using `patch.crates-io` in Cargo.toml instead:

```toml
[package]
name = "serde_tagged_enum"
version = "0.1.0"
edition = "2021"

[dependencies]
serde = { version = "1", features = ["derive"] }
serde_json = "1"

[patch.crates-io]
serde = { git = "https://github.com/Astavie/serde.git", branch = "integer-tags-for-enums" }
```

note, cargo might complain about not finding a version for serde but just unlocking the dependencies by removing cargo.lock fixed that for me

IriaSomobu:
I see. Sorry for bothering you :)

FancyQian:
please make it as possible, it very helpful since rust enum can contain different type.

Astavie:
Still waiting on a new review by @dtolnay

yuto0214w:
Any progress on this PR?
As of now, I'm still relying on the patched version...

dnlsndr:
@yuto0214w I think @dtolnay still needs to review this

IriaSomobu:
@dtolnay please can you review&merge this PR?

scuggo:
any progress on this PR?

r4v3n6101:
@Astavie sorry for bothering you, but could you do update if you would like to?

ByteAlex:
Coming back to the project 3-4 years later and this topic is still not resolved, a new PR has emerged, at least one review was made, but again ghosted for years. I guess adjacently tagged enums will never be supported 😅

jobafr:
So in the meanwhile, this PR has conflicts with master. It seems that, someone did a mid-sized refactor regarding how the enum code generation is structured, splitting code into new files, renaming functions, etc.

e.g.
https://github.com/serde-rs/serde/commit/19956e6b8e57188c9d94a5aad0dbe0279ed5513e

PR #2533, opened 2023-07-26
Move serde_test out to serde-rs/test

dtolnay:
There is a good reason for `serde` and `serde_derive` to be versioned together, but no reason for `serde_test`, other than the historical effect of this being originally private code that was extracted out of serde's test suite to be published.

I generated https://github.com/serde-rs/test/commit/0a45bbf603037894809a8f8201481eafe3a5eed3 using `git filter-repo --path serde_test` at https://github.com/serde-rs/serde/commit/6140b6f527f93e718ecf05e51bdc33829d3e95f5.

Since #2441, serde's test suite no longer relies on nonpublic APIs of `serde_test`.


Cargo.toml
```diff
@@ -3,6 +3,8 @@ members = [
     "serde",
     "serde_derive",
     "serde_derive_internals",
-    "serde_test",
     "test_suite",
 ]
+
+[patch.crates-io]
+serde = { path = "serde" }
```
Mingun, line 10: This PR breaks any attempts to experiment with changing serde public API. I think, it is due to this line. Is it possible to do without it? What is it for?

PR #2536, opened 2023-07-27
Update error span for attribute / data kind mismatches

jplatte:
Follow-up to https://github.com/serde-rs/serde/pull/1695#discussion_r1275423042.

There are a few more in line 770 and below, but there it wasn't quite as clear what to set the span to.

### DIFF
--- a/test_suite/tests/ui/default-attribute/enum_path.stderr
+++ b/test_suite/tests/ui/default-attribute/enum_path.stderr
@@ -1,5 +1,5 @@
 error: #[serde(default = "...")] can only be used on structs with named fields
- --> tests/ui/default-attribute/enum_path.rs:5:1
+ --> tests/ui/default-attribute/enum_path.rs:4:9
   |
-5 | enum E {
-  | ^^^^
+4 | #[serde(default = "default_e")]
+  |         ^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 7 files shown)


test_suite/tests/ui/default-attribute/enum_path.stderr
```diff
@@ -1,5 +1,5 @@
 error: #[serde(default = "...")] can only be used on structs with named fields
- --> tests/ui/default-attribute/enum_path.rs:5:1
+ --> tests/ui/default-attribute/enum_path.rs:4:9
   |
-5 | enum E {
-  | ^^^^
+4 | #[serde(default = "default_e")]
+  |         ^^^^^^^
```
dtolnay, line 5: I am not sure whether to prefer underlining just the ident or the whole nested attribute. Do you have an opinion?

```console
#[serde(before, default = "...", after)]
                ^^^^^^^
```

```console
#[serde(before, default = "...", after)]
                ^^^^^^^^^^^^^^^
```

I think the second behavior is implementable with `cx.syn_error(meta.error(msg))`.

dtolnay (approved):
Thanks!

test_suite/tests/ui/default-attribute/enum_path.stderr
jplatte: I guess the latter is a bit nicer, thanks for the impl hint.

dtolnay (approved):
Looks good! Thank you.

Issue #2366, opened 2023-01-20
Nitpick: use `?` instead of `try!` internally and in generated code


Lucretiel:
Much of serde's implementation uses `try!`, which I assumed was for backwards compatibility reasons. However, while support for `Option` was only added in Rust 1.22, the `?` itself was added in Rust 1.13, a much earlier version than serde's claimed MSRV of 1.19.

I'd need to bench to be sure, but my presumption is that `?`, a built-in operator, is faster for the compiler than expanding the `try!` macro.

dtolnay:
The last time I checked (~1 year ago) `?` still produced larger binaries than `try!`, compiled slower, and had worse performance at runtime.

Lucretiel:
Surprising but fair enough.

Lucretiel:
Supposedly this was fixed last August, as of 1.52: https://github.com/rust-lang/rust/issues/37939

dtolnay:
Oh nice. I think that issue was only in regard to runtime performance though? I tried to check compile time using this diff in serde_derive:

```diff
         macro_rules! try {
             (#dollar __expr:expr) => {
-                match #dollar __expr {
-                    _serde::__private::Ok(__val) => __val,
-                    _serde::__private::Err(__err) => {
-                        return _serde::__private::Err(__err);
-                    }
-                }
+                #dollar __expr?
             }
         }
```

and using the following commands:

- In https://github.com/serde-rs/json-benchmark:
    - `touch src/lib.rs; time CARGO_INCREMENTAL=0 cargo check --no-default-features --features all-files,parse-struct,lib-serde`
    - `touch src/lib.rs; time CARGO_INCREMENTAL=0 cargo build --release --no-default-features --features all-files,parse-struct,lib-serde`
- In https://github.com/gluon-lang/lsp-types:
    - `touch src/lib.rs; time CARGO_INCREMENTAL=0 cargo check`
    - `touch src/lib.rs; time CARGO_INCREMENTAL=0 cargo build`

and these all compile consistently 6.5&ndash;7.5% slower on my machine with `?`, using current nightly. That's the identical ratio noted 5 years ago in https://github.com/serde-rs/serde/issues/1121#issuecomment-383310311. `?` just generates more inference variables, and function calls / copies / layers of trait abstractions that need to get compiled or inlined.

The gap would be larger upon using `?` in serde_json too, since that currently uses its own `tri!` macro, for the same reason.

[...]

PR #2539, opened 2023-07-28
Replace `try!` with `?` in serde_derive

dtolnay:
#2366

As of today, this regresses compile time of serde's generated code by 6.5–7.5% which is too bad. However `try!` has been disastrous for readability of the generated code coming out of `cargo expand`. I was expecting the performance problems in rustc to be fixed a long time ago but at this point it doesn't seem like that is going to happen.


Mingun:
I see, that not all `try!` was replaced by `?`. I would say, most of them. Is it so conceived?

Baptistemontan:
So what about https://github.com/serde-rs/serde/commit/91ec1c290f73777a49d200c880a3e87d86067a94 ? The crate itself should suffer from readability but it goes backward for generated code ?

Mingun:
Yes, that is... very strange decision :)

dtolnay:
It's specifically `cargo expand` readability that is the motivation.

People are not looking at `cargo expand` of the serde crate. They are looking at expanded derive macros.

Issue #812, opened 2017-03-09
Unstable functionality in Serde


dtolnay:
#### Specialization https://github.com/rust-lang/rust/issues/31844

- The default collect_seq and collect_map specialize for ExactSizeIterator.

#### Deserialize for Box\<CStr> https://github.com/rust-lang/rust/issues/40380

- Requires [`CString::into_boxed_c_str`](https://doc.rust-lang.org/nightly/std/ffi/struct.CString.html#method.into_boxed_c_str)

#### Serialize for RangeInclusive https://github.com/rust-lang/rust/issues/28237

nox:
Feature `zero_one` just died. Can we have new versions on both `0.9` and `1.0`?

SimonSapin:
Submitted https://github.com/serde-rs/serde/pull/898 for `zero_one`.

Techcable:
Thanks so much for the clarification, I didn't realize I was using an outdated nightly!

dtolnay:
I updated the list to reflect functionality that was stabilized in Serde 1.0.60 and 1.0.61. These are no longer behind the `"unstable"` cfg.

dtolnay:
Added `never_type` in #1302.

dtolnay:
Added `error_in_core` in #2344.

gootorov:
Would it be appropriate to add support for `Vec<T, A: Allocator>`/`Box<T, A: Allocator>` under this feature flag?
If so, I can submit a PR in some near future.

jdygert-spok:
Error in core has been stabilized. Would it be appropriate to make `StdError = core::error::Error` when `minor >= 81`?

PR #2540, opened 2023-07-28
Fix `unstable` build

mina86:
Before:

    $ cargo +nightly build --no-default-features --features unstable
       Compiling serde v1.0.177 (/srv/mpn/serde/serde)
    error[E0432]: unresolved import `std_error`
       --> serde/src/de/mod.rs:134:9
    134 | pub use std_error::Error as StdError;
        |         ^^^^^^^^^ maybe a missing crate `std_error`?

After:

    $ cargo +nightly build --no-default-features --features unstable
       Compiling serde v1.0.177 (/srv/mpn/serde/serde)
        Finished dev [unoptimized] target(s) in 1.28s

Note that `unstable` and non-`std` tests are still failing.

Issue: https://github.com/serde-rs/serde/issues/812

### DIFF
--- a/serde/src/de/mod.rs
+++ b/serde/src/de/mod.rs
@@ -126,10 +126,6 @@ mod utf8;
 
 pub use self::ignored_any::IgnoredAny;
 
-#[cfg(feature = "std")]
-#[doc(no_inline)]
-pub use std::error::Error as StdError;
-#[cfg(not(feature = "std"))]
 #[doc(no_inline)]
 pub use std_error::Error as StdError;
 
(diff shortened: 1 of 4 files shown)


dtolnay (changes requested):
Thanks for the PR!

Unconditionally using `std_error` makes the rendered documentation less clear. I opened #2541 with a different implementation that follows the one of serde::ser::StdError.

Before this PR:
Screenshot from 2023-07-28 16-02-07

After this PR:
Screenshot from 2023-07-28 16-02-10

PR #2553, opened 2023-08-03
Allow `#[serde(default)]` on tuple structs

Mingun:
Right now you cannot place `#[serde(default)]` at tuple struct, although it is possible to mark individual fields as `#[serde(default)]`:

```rust
#[derive(Deserialize)]
struct Tuple(
  // possible, but make no sense in that case, because if
  // field will be set to default value that means that sequence
  // is exhausted and the next field will fail to deserialize
  #[serde(default)] u8,
  u8,
);
#[derive(Deserialize)]
#[serde(default)] // impossible, but make sense
struct Tuple(
  #[serde(default)] u8,
  u8,
);
```
This PR fixes that and also guaranties that tuple structs will always have a trail of defaulted non-skipped fields as soon as first defaulted field is appear.

### DIFF
--- a/serde_derive/src/de.rs
+++ b/serde_derive/src/de.rs
@@ -723,19 +723,11 @@ fn deserialize_seq(
                     })
                 }
             };
-            let value_if_none = match field.attrs.default() {
-                attr::Default::Default => quote!(_serde::__private::Default::default()),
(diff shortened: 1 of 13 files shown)


dtolnay (approved):
Thanks!

PR #2555, opened 2023-08-06
Refactor code that generates `__Field` enums

Mingun:
This PR makes a small refactoring in order to remove complexity in `deserialize_generated_identifier` about determining `ignore_variant` and `fallthrough` values. Because both values depends only on input parameters, I moved they determination from `deserialize_generated_identifier` to the calling functions. Thus, it is easier to understand which pairs of parameters are possible


dtolnay (approved):
Nice!

PR #2557, opened 2023-08-06
Deduplicate similar methods of `ContentDeserializer` and `ContentRefDeserializer`

Mingun:
This PR moves helper functions that is implemented identically in both content deserializers to the `Content` itself, which would guarantee that they are always have the same implementation.

The last commit also express `deserialize_float` method in terms of `deserialize_integer` method. While that is good for readability, maybe compiler will not inline it and it can be slightly slower. If you do not like it, you can drop the last commit from that PR

### DIFF
--- a/serde/src/private/de.rs
+++ b/serde/src/private/de.rs
@@ -288,6 +288,63 @@ mod content {
                 Content::Map(_) => Unexpected::Map,
             }
         }
+
+        #[cold]
+        fn invalid_type<E>(&self, exp: &Expected) -> E
+        where
+            E: de::Error,
+        {
+            de::Error::invalid_type(self.unexpected(), exp)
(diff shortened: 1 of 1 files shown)


dtolnay (changes requested):
Thank you!

This looks good except for the "Merge unexpected() into invalid_type()" commit. I think those are intentionally separate. `unexpected` is a relatively big chunk of logic that shouldn't need to be re-instantiated downstream for all the different `E` types that Content is used with. It is supposed to be instantiated once in the serde crate.

Mingun:
Makes sense. I removed this commit

Mingun:
The correct way is completely remove any attempt to return error from the deserializer itself. Closed in preference to #2569

PR #2558, opened 2023-08-07
Improve error reporting about mismatched signature in `with` and `default` attributes

Mingun:
This PR changes the errors reported by `rustc` when signature of function specified in
- `#[serde(with = "...")]` 
- `#[serde(deserialize_with = "...")]`
- `#[serde(serialize_with = "...")]`
- `#[serde(default = "...")]`

is wrong. With this PR error will point to the `"..."`.

### DIFF
--- a/serde_derive/src/de.rs
+++ b/serde_derive/src/de.rs
@@ -371,7 +371,11 @@ fn deserialize_transparent(cont: &Container, params: &Parameters) -> Fragment {
         } else {
             let value = match field.attrs.default() {
                 attr::Default::Default => quote!(_serde::__private::Default::default()),
-                attr::Default::Path(path) => quote!(#path()),
+                // If #path returns wrong type, error will be reported here (^^^^^).
+                // We attach span of the path to the function so it will be reported
+                // on the #[serde(default = "...")]
+                //                          ^^^^^
+                attr::Default::Path(path) => quote_spanned!(path.span()=> #path()),
                 attr::Default::None => quote!(_serde::__private::PhantomData),
             };
             quote!(#member: #value)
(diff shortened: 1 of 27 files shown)


Mingun:
Strange, but order of implementations shown in [failed run](https://github.com/serde-rs/serde/actions/runs/5784133307/job/15674235243?pr=2558) is different from my (my removed, PR added):

```diff
    = help: the following other types implement trait `Serializer`:
-             FlatMapSerializer<'a, M>
-             &'a mut Formatter<'b>
              _::_serde::__private::ser::content::ContentSerializer<E>
+             FlatMapSerializer<'a, M>
+             &'a mut Formatter<'b>
```
This can create difficulties in local testing.

Mingun:
Probably the order of suggested implementations is heavy dependent on rustc version. With rustc 1.82.0-nightly (fbccf5053 2024-07-27) the local output is the same as at GitHub.

oli-obk (approved):
These are great improvements. It seems like there is still space for rustc to provide more diagnostics attributes, but nothing that we can fix ok this end

Veetaha:
@oli-obk looks like the merge of this PR broke master [CI](https://github.com/serde-rs/serde/actions/runs/11447410021). The CI on this PR was green long time ago, and it broke with time

dtolnay:
The failing ui test was from a non-textual conflict with https://github.com/serde-rs/serde/commit/8e1ae68569db46251a88997e5e212392d4ad83ea. Fixed in #2839. (Also #2840).

It is unfortunate that:

```console
   = help: the following other types implement trait `Serializer`:
             &mut Formatter<'a>
             FlatMapSerializer<'a, M>
             _::_serde::__private::ser::content::ContentSerializer<E>
```

is the set of impls chosen to be suggested by rustc. Is there something we can do with `doc(hidden)` to exclude nonpublic APIs from this error message? If not, this needs to be reported as a rustc bug.

dtolnay:
Reported https://github.com/rust-lang/rust/issues/132024.

oli-obk:
> is the set of impls chosen to be suggested by rustc. Is there something we can do with `doc(hidden)` to exclude nonpublic APIs from this error message? If not, this needs to be reported as a rustc bug.

We'll get `diagnostics::do_not_suggest` or sth. Iirc it's being worked on or works on nightly already

Turbo87:
it looks like this change might have unintended side effects. https://github.com/rust-lang/crates.io/pull/9720 upgrades crates.io from v1.0.210 to v1.0.211 and we are now seeing various "not found in this scope" errors that were not there before.

PR #2559, opened 2023-08-07
Remove misleading `?Sized` requirement - `Deserialize` requires `Sized`

Mingun:
`Deserialize` trait requires `Sized` types, so there no sense to declare `?Sized` bounds:
https://github.com/serde-rs/serde/blob/05a5b7e3c6de502d45597cbc083f28bc1d4f4626/serde/src/de/mod.rs#L537

### DIFF
--- a/serde/src/de/impls.rs
+++ b/serde/src/de/impls.rs
@@ -1814,7 +1814,7 @@ where
 ///
 /// [`"rc"`]: https://serde.rs/feature-flags.html#-features-rc
 #[cfg(all(feature = "rc", any(feature = "std", feature = "alloc")))]
-impl<'de, T: ?Sized> Deserialize<'de> for RcWeak<T>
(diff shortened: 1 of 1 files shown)


Mingun:
Implemented in 89139e2c11c9e975753ebe82745071acb47ecb03.

dtolnay (commented):
Thank you &mdash; sorry I missed that there was already a PR for this. I noticed the redundant bounds as part of resolving `clippy::multiple_bound_locations` lints (c9cab490f7a3c1038fb5d5bc93d87737228300f8).

Would you want to follow up with a Clippy feature request to detect this case with `?Sized`?

Mingun:
Probably this should be compile time error instead of a lint error. Actually, this case can be reduced to `Sized + ?Sized` boundary and compiler should not silently assume any concrete behavior here.

PR #2560, opened 2023-08-07
Use correct tokens in test to represent an enum variant

Mingun:
The test that is fixing by this PR was introduces in #933 and has a purpose to reach the `Content::Str` case here that was fixed by that PR:
https://github.com/serde-rs/serde/blob/05a5b7e3c6de502d45597cbc083f28bc1d4f4626/serde/src/private/de.rs#L1446

The `Content::Str` case here produced using `visit_borrowing_str` here (that the only case when `Content::Str` can be created):
https://github.com/serde-rs/serde/blob/05a5b7e3c6de502d45597cbc083f28bc1d4f4626/serde/src/private/de.rs#L424-L429

which is called as a response to `Token::BorrowedStr`:
https://github.com/serde-rs/test/blob/d1294b3ad549874d5035752ac62b4eb75cee5060/src/de.rs#L129

However, unit variants of externally tagged enum cannot be deserialized from the string token by itself. The one of following combinations of tokens should be used to correctly represent enum variant:
- `[xxxVariant { .. }]`
- `[Enum { .. }, xxxVariant { variant, .. }]`
- `[Enum { .. }, String(variant), <variant content>]`
- `[Enum { .. }, Str(variant), <variant content>]`
- `[Enum { .. }, BorrowedStr(variant), <variant content>]`

Because we want to reach `Content::Str` case, the only possible representation which we can use is `[Enum { name: "Level" }, BorrowedStr("Info"), Unit]` (other variants will be captured as `Content::String`).

### DIFF
--- a/serde/src/private/de.rs
+++ b/serde/src/private/de.rs
@@ -313,6 +313,8 @@ mod content {
         }
     }
 
+    /// Used to capture data in [`Content`] from other deserializers.
+    /// Cannot capture externally tagged enums, `i128` and `u128`.
     struct ContentVisitor<'de> {
         value: PhantomData<Content<'de>>,
     }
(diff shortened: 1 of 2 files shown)


Mingun:

dtolnay (approved):
Thank you!

PR #2520, opened 2023-07-21
Fix incorrect representation of tuples with skipped fields

Mingun:
This PR fixes #2105 and give some insights about the current situation with skipped fields.

You can see what changed in the generated code by running the following commands:
- place [skip.txt](https://github.com/serde-rs/serde/files/12131031/skip.txt) into `test_suite/tests` folder and change it extension to `.rs`
- run the following command:
   ```console
   ...\serde\test_suite> cargo expand --all-features --test skip > skip-N.rs
   ```
   where `N` is some number.

When commit message contains words "Changes in generated code" that means that output of the command above was changed. I recommend to diff it against the previous result and explore the difference. There will be 4 files -- baseline and three changed.

Overall changes over all commits:
- `Tuple1as0`, `Tuple1as0Default`, `Tuple1as0With`:
  - removed `visit_newtype_struct`
  - changed `*_newtype_struct` -> `*_tuple_struct(0)`
- `Tuple2as1`, `Tuple2as1Default`, `Tuple2as1With`:
  - added `visit_newtype_struct`

The table below summarizes the current behavior and the new behavior, introduced in this PR. Empty cells in the last column means that the behavior wasn't changed.

The *behavior* columns shows which methods are used to serialize or deserialize a type. The number in parenthesis is a count of fields which passed to the corresponding method.

<table>
<tr><th>Code</th><th>Current behavior</th><th>Fixed behavior</th></tr>

<tr><th colspan="3">Unit representations</th></tr>
<tr><td>

```rust
struct Unit;
```
</td><td>

`serialize_unit_struct`  
`deserialize_unit_struct`  
`visit_unit`  
</td><td></td></tr>

<tr><th colspan="3">Tuple representations</th></tr>
<tr><td>

```rust
struct Tuple0();
```
</td><td>

`serialize_tuple_struct(0)`  
`deserialize_tuple_struct(0)`  
`visit_seq`  
</td>
<td>

:information_source: It would be better to use Unit representation [(see below)](#list-of-inconsistencies)</td></tr>

<tr><td>

```rust
// Can be named
// Newtype(u8)
struct Tuple1(u8);
```
</td><td>

`serialize_newtype_struct`  
`deserialize_newtype_struct`  
`visit_newtype_struct`  
`visit_seq`  
</td><td></td></tr>

<tr><td>

```rust
struct Tuple2(u8, u8);
```
</td><td>

`serialize_tuple_struct(2)`  
`deserialize_tuple_struct(2)`  
`visit_seq`  
[...]

PR #2561, opened 2023-08-08
Disallow tuple variants with incorrect order of `#[serde(default)]` attributes

Mingun:
Continuation of work from #2553, but expanded to tuple variants.

Actually, we should also disallow using `#[serde(default)]` on elements of newtype structs and newtype variants, but this is highly coupled with [skip problems](https://github.com/serde-rs/serde/pull/2520), so I wait before the relevant PRs will be merged

### DIFF
--- a/serde_derive/src/internals/check.rs
+++ b/serde_derive/src/internals/check.rs
@@ -18,35 +18,49 @@ pub fn check(cx: &Ctxt, cont: &mut Container, derive: Derive) {
     check_from_and_try_from(cx, cont);
 }
 
+fn check_default_on_tuple(cx: &Ctxt, cont: &Container) {
+    match &cont.data {
+        Data::Enum(variants) => {
+            for variant in variants {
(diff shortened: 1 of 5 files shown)


Mingun:

ahl:
I hit this in the wild with a type like this:

```rust
#[derive(Deserialize)]
enum E {
    V(#[serde(default)] u32, #[String),
}
```

It compiles, but does nothing. For tuple struct, it fails during compilation:

```rust
#[derive(Deserialize)]
struct S(#[serde(default)] u32, String);
```
```
error: field must have #[serde(default)] because previous field 0 has #[serde(default)]
 --> src/main.rs:6:33
  |
6 | struct S(#[serde(default)] u32, String);
  |                                 ^^^^^^
```

I tested this patch and it applied cleanly and worked for me. Is the hesitancy because it might break compiling (but slightly wrong) code out there today?

PR #2562, opened 2023-08-08
Add checks for conflicts for aliases

### LINKED ISSUE #2308: No unreachable warning when duplicate field names on enum after rename attributes
Looking into the rename feature I discovered that rustc does not gives an unreachable warning with serde rename collisions with `enums` whereas it does on `structs` as mentioned in #754 

Would you still be open to a PR to make clashing renames an error? If so I'm happy to give it a go.

## Example
```rust
use serde::{Deserialize};

#[derive(Deserialize)]
enum Message {
    #[serde(rename = "Response")]
    Request { id: String},
    #[serde(rename = "Response")]
    Response { id: String},
}

fn main() {
    let json = "{\"Response\": {\"id\": \"...\"}}";
    
    let parsed: Message = match serde_json::from_str(&json) {
        Ok(contact) => contact,
        Err(err) => {
            println!("{:?}", err);
            unimplemented!()
        }
    };

    match parsed {
        Message::Request { id } => println!("request {}", id),
        Message::Response { id } => println!("response {}", id)
    }
    
}
```
### Output
`request ...`
with no compiler warnings

playgrounds link https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=c6a787d51f1290af999a0e36b9a6d366

### LINKED ISSUE #2551: Field/variant aliases are not checked for uniqueness
[The code](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=0551945af3b0581fefd8c0c9684e4182)
```rust
use serde::Deserialize; // 1.0.171;
use serde_json; // 1.0.102;

#[derive(Deserialize, Debug)]
#[serde(deny_unknown_fields)]
pub struct Thing {
    pub w: u8,

    #[serde(alias = "z", alias = "x")]
    pub y: u8,

    #[serde(alias = "same", alias = "other", alias = "same", alias = "x", alias = "y")]
    pub same: u8,
}

fn main() {
    let j = r#" {"j":null} "#;
    println!("{}", serde_json::from_str::<Thing>(j).unwrap_err());
}
```
gives the following output:
```
unknown field `j`, expected one of `w`, `x`, `z`, `y`, `other`, `same`, `x`, `y` at line 1 column 5
```

Mingun:
- Check that alias is not the same as name of other field / variant (it still can be the name of owning field / variant)
- Check that aliases are unique, i. e. two different fields / variants does not use the same alias

It would be better, if error points to corresponding alias, but currently alias spans are not stored anywhere, so I left this to another PR.

Fixes #2308, fixes #2551

### DIFF
--- a/serde_derive/src/internals/check.rs
+++ b/serde_derive/src/internals/check.rs
@@ -1,6 +1,8 @@
-use crate::internals::ast::{Container, Data, Field, Style};
+use crate::internals::ast::{Container, Data, Field, Style, Variant};
 use crate::internals::attr::{Default, Identifier, TagType};
 use crate::internals::{ungroup, Ctxt, Derive};
+use std::collections::btree_map::Entry;
+use std::collections::{BTreeMap, BTreeSet};
 use syn::{Member, Type};
 
 // Cross-cutting checks that require looking at more than a single attrs object.
@@ -16,6 +18,7 @@ pub fn check(cx: &Ctxt, cont: &mut Container, derive: Derive) {
     check_adjacent_tag_conflict(cx, cont);
     check_transparent(cx, cont, derive);
     check_from_and_try_from(cx, cont);
+    check_name_conflicts(cx, cont, derive);
 }
 
 // If some field of a tuple struct is marked #[serde(default)] then all fields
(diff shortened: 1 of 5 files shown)


Mingun:

oli-obk:
Annoying that we can't crater it first, but this also catches really subtly broken code, so I guess we can Yolo it.

dtolnay (commented):
I do not think this can be released as is. This is a pretty blatant breaking change in existing working code.

I would recommend emitting something like a `unreachable_patterns` warning on duplicated names.

oli-obk:
The code would never have worked correctly. I don't think anyone would ship this for long without users complaining.

dtolnay:
I do not agree that this PR is analogous to rustc emitting a new error on an unsound code pattern. It is analogous to rustc deciding that unreachable patterns are a hard error now, because their previous behavior cannot possibly be what the person who wrote the unreachable pattern intended. For code in which deserializing to the first variant with the same alias was the desired behavior (hence users not complaining) I don't understand why it's better for a new version of serde to break them as opposed to a warning that says to remove the irrelevant alias.

Issue #760, opened 2017-02-12
Serialize extra constant fields


RReverser:
Serde allows to use `tag = "..."` on enums, but I have a format in which there are also "tags" on structs - basically extra constant fields that must be present in serialized form to pass validations, but as Rust is typed, they don't matter in Rust representation.

For the sake of example:

```rust
pub struct Program {
  // type: "Program"
  pub body: Vec<Stmt>
}
```

This must be serialized as `{ "type": "Program", "body": [...] }`.

Another example, as part of enum:

```rust
#[serde(tag = "type")]
pub enum Expr {
  ...
  #[serde(rename = "Literal")]
  NullLiteral // must be serialized as `{ "type": "Literal", "value": null }` not as `{ "type": "Literal" }`
}
```

Of course, I can add extra meaningless `type: &'static str` to `Program` and `value: ()` to `NullLiteral` or implement custom `Serialize` / `serialize_with` for each of these cases, but as it seems to be a quite common repetitive task, would it make sense to add support for constant fields in Serde itself via a separate attribute?

For deserialize implementation, those fields would be just checked for existence (like `tag` on enums), and for serialize they would be implicitly added.

dtolnay:
We are tracking a similar idea in #271.

Can you go into more detail about how you would expect validation to work during deserialization?

RReverser:
@dtolnay That one won't help with "add extra field to this one enum variant" case, and generally feels like overkill for simple compile-time constants.

As for validation - just meant reading value as its original type and comparing with defined constant. If they don't match, that's unexpected-value kind of error.

lucab:
I think I have hit the same usecase with [appc pod manifest](https://github.com/appc/spec/blob/master/spec/pods.md#pod-manifest-schema), which has an internal constant field `acKind` to identify its own type.

This may be seen somehow as a degenerate case of an internally tagged enum with a single variant. 

Perhaps the `tag` container-attribute could be just extended to structs as well, allowing:
```
#[serde(tag = "acKind", rename="PodManifest")]
pub struct AppcPod {
  // All other real fields, but not acKind...
}
```

Regarding validation, I would just expect this to discriminate based on (possibly renamed) struct name.

I've also seen somwehre else a kind-of similar case of an "adjacently tagged struct" which I think can also be addressed via `tag`+`rename`+`content` on structs:
```
{
  "kind": "foo_type",
  "value: { "foo": true }
}

#[serde(tag = "kind", rename="foo_type", content="value")]
[...]

PR #2563, opened 2023-08-08
feat(derive): add custom field for struct

saiintbrisson:
Hi! It is not infrequently that I find myself needing to compute a custom
field that is only present on serialization and is calculated off of another
one of the struct's fields. Usually, I'll either implement `Serialize` myself,
but this gets tiring if the struct is too large and is prone to errors, or create
another struct with the field I want just before serializing.

I hereby propose creating a new container attribute `field` that allows for
the creation of fields on the go at the moment of serialization:

```rust
#[derive(Serialize)]
#[serde(
    field(name = "b", getter = "Self::b_getter"),
    field(name = "c", with = "c_serialize_with")
)]
struct CustomFields {
    a: i32
}

impl CustomFields {
    fn b_getter(&self) -> i32 {
        self.a * 2
    }
}

fn c_serialize_with<S: Serializer>(v: &CustomFields, ser: S) -> Result<S::Ok, S::Error> {
    (v.a * 3).serialize(ser)
}
```
So a `CustomFields { a: 10 }` results in:
```json
{
  "a": 10,
  "b": 20,
  "c": 30
}
```

I'm still not happy with the attributes' names, it may make more sense to
call it `serialize_field` or something along those lines?


dtolnay (commented):
Thanks for the PR.

This was previously discussed and rejected in #760.

PR #2564, opened 2023-08-10
Do not generate unique `__Field` for each struct

Mingun:
As suggested in https://github.com/serde-rs/serde/pull/2250#pullrequestreview-1564212707, this PR replaces `__Field` generation with common structs inside `serde` crate.

There are three different forms of a `__Field` struct, each one has its own struct:
- `Field` + `FieldSeed` -- for structs without flatten fields and without `#[serde(deny_unknown_fields)]`
- `FieldStrong` + `FieldStrongSeed` -- for structs without flatten fields and with `#[serde(deny_unknown_fields)]`
- `FieldFlatten` + `FieldFlattenSeed` -- for structs with flatten fields


Mingun:
Hm... I don't understand why build is failed. Everything is OK when run `cargo build` inside the root directory, but mysteriously failed when run from `./serde` sub-directory. I don't understand why `__private` module cannot be found? It is not exist when `no_serde_derive` is specified, but that should happens only on Rust < 1.56 and I try to build on Rust 1.72

Mingun:
Here is some results: [comparison.zip](https://github.com/serde-rs/serde/files/12325396/comparison.zip).

Measured commits v1.0.183 (05a5b7e3c6de502d45597cbc083f28bc1d4f4626) and df12e82af225d6de9af8ae54d6a290a380211f31, which removes all possible variants of `__Field` for struct field names and `__Field` for representing variants of externally tagged enums.

In both cases `serde` was injected using `[patch]` section in root `Cargo.toml` of both projects:
```toml
[patch.crates-io]
serde = { path = "../serde/serde" }
```

The following commands was used:
- json-benchmark:
  ```powershell
  cargo clean
  $env:CARGO_INCREMENTAL=0; cargo check --timings --no-default-features --features all-files,parse-struct,lib-serde
  cargo clean
  $env:CARGO_INCREMENTAL=0; cargo build --timings --release --no-default-features --features all-files,parse-struct,lib-serde
  ```
  PR is about 0.3 sec. faster on checking / building of json-benchmark lib itself. native jemalloc library was not found on my machine, so I cannot provide runtime impact right now
- lsp-types:
  ```powershell
  cargo clean
  $env:CARGO_INCREMENTAL=0; cargo check --timings
  cargo clean
  $env:CARGO_INCREMENTAL=0; cargo build --timings
  ```
  PR is about 3 sec. faster on checking / building of lsp-types itself

@nnethercote perhaps you are willing to test impact of this PR on your cases?

PR #2566, opened 2023-08-10
Show variant aliases in error message

Mingun:
Before that PR an error message of such enum did not include aliases of possible variants:
```rust
#[derive(Deserialize)]
enum Enum {
  #[serde(alias = "C")]
  A,
  B
}
```
Now the error message will mention `C`

### DIFF
--- a/serde_derive/src/de.rs
+++ b/serde_derive/src/de.rs
@@ -968,13 +968,7 @@ fn deserialize_struct(
         // Skip fields that shouldn't be deserialized or that were flattened,
         // so they don't appear in the storage in their literal form
         .filter(|&(_, field)| !field.attrs.skip_deserializing() && !field.attrs.flatten())
-        .map(|(i, field)| {
-            (
-                field.attrs.name().deserialize_name(),
-                field_i(i),
-                field.attrs.aliases(),
-            )
-        })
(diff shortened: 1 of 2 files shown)


Mingun:

oli-obk:
Not in the near future. The only PR I'm giving attention to right now is the serde_core one

juntyr:
This change breaks serde-reflection, which inspects the slice of known variants to get their numeric indices, assuming that the mapping from names to indices is bijective.

https://github.com/zefchain/serde-reflection/blob/d1a1ff0494fa3396af0423950cd43b7141890045/serde-reflection/src/de.rs#L394-L443

@Mingun @oli-obk do you have ideas for how the previous functionality could be regained?

dtolnay:
Serde-reflection was assuming that if an enum has variant names {"A", "B", "B1"} then it should expect to deserialize them all using the same number of distinct variant indices {0, 1, 2}. This isn't correct because two different variant names can share the same index in the case of aliases, and there might not exist any variant with index 2 or 1. I implemented a better approach in https://github.com/zefchain/serde-reflection/pull/53 by using variant names to deserialize instead of indices, and then comparing the discriminant of each of the deserialized values to unify aliases.

PR #2569, opened 2023-08-12
Internally tagged tests

### LINKED ISSUE #2187: Support internally tagged enums with non-self-describing formats.
Right now deserialization of internally tagged enums rely on a call to `serde::Deserializer::deserialize_any` which often doesn't work well with non-self-describing formats. Ideally `serde::Deserializer::deserialize_struct` would be called instead so that the `Deserializer` knows what to do.

https://github.com/serde-rs/serde/blob/7e19ae8c9486a3bbbe51f1befb05edee94c454f9/serde_derive/src/de.rs#L1340-L1342

Mingun:
This PR starts as a refactor of internally tagged enums tests in preparation to updating #1922, but in the process of development, I found some interesting consequences.

The first is that `ContentDeserializer` and `ContentRefDeserializer` do what they should not do, namely, determine the validity of certain data when calling deserialization methods. All `Deserializer` methods are just a hints what data is expected from the deserializer, and the deserializer is not obliged to follow them. It is free to call any `Visitor` method it sees fit. It is `Visitor` responsibility to detect if called `visit_*` method appropriate or not. So I removed all decisions from `ContentDeserializer` and `ContentRefDeserializer` about validity of content in certain methods.

This change will allow to make the next step and eliminate all calls to `deserialize_any` from internally tagged enums. Because this method does not called anymore, this would allow to support internally tagged enums in non-self-describing formats. I've replaced `deserialize_any` with `deserialize_map` because:
- internally tagged enums serialized using `serialize_map` (for newtype variants) / `serialize_struct` (for unit and struct variants), that means that deserialization also should expect map
- deserialization of structs looks the same -- calling `Deserializer::deserialize_map` or `Deserializer::deserialize_struct` and provide a visitor with `visit_map` and `visit_seq`. This is what both `TaggedContentVisitor` and `InternallyTaggedUnitVisitor` provides

This PR depends on the https://github.com/serde-rs/test/pull/31 and I temporary add a `[patch]` section to use patched version to make tests passable. That is why this PR in a draft stage.

Closes #2187
Replaces #2557


max-heller:
> ... eliminate all calls to `deserialize_any` from internally tagged enums. Because this method does not called anymore, this would allow to support internally tagged enums in non-self-describing formats. I've replaced `deserialize_any` with `deserialize_map` because:
> 
> * internally tagged enums serialized using `serialize_map` (for newtype variants) / `serialize_struct` (for unit and struct variants), that means that deserialization also should expect map

As mentioned in https://github.com/jamesmunns/postcard/issues/125#issuecomment-3139800272, I don't think this is compatible with non-self-describing formats--at least postcard--because maps and structs are serialized differently and cannot be distinguished during deserialization:

> Postcard serializes structs as a sequence of values (for each field) with no length prefix or field identifiers, whereas it serializes maps as a length prefix followed by key-value pairs. If the Serialize and Deserialize impls for the internally tagged enum both used serialize_map(), I think deserialization would work.

For instance, if variant `A` is serialized with `serialize_struct()`, deserialization will fail when it tries to use `serialize_map()` because there is no length prefix and there are no field identifiers in the serialized data. Worse, actually, it might "successfully" deserialize into garbage output.

PR #2570, opened 2023-08-14
Treat unmatched non-exhaustive remote variant as serde(skip)

### LINKED ISSUE #2360: Add wildcard arm when deriving `Serialize` for non exhaustive enums

dtolnay:
Serde already supports a `serde(skip)` attribute to not serialize a particular variant, returning an error instead.

```rust
#[derive(Serialize)]
enum Enum {
    Foo,
    #[serde(skip)]
    Bar,
}
```

This PR gives equivalent behavior to #\[non_exhaustive\] remote enums: it's as if every other possible variant were listed with a `serde(skip)` on it.

```rust
#[serde(remote = "std::io::ErrorKind")]
#[non_exhaustive]
enum ErrorKindDef {
    NotFound,
    PermissionDenied,
    // ...
}
```

```rust
assert!(ErrorKindDef::serialize(&ErrorKind::OutOfMemory, serializer).is_err());
```

Closes #2360. Closes #2361.

### DIFF
--- a/serde/src/private/ser.rs
+++ b/serde/src/private/ser.rs
@@ -1370,3 +1370,16 @@ impl Serialize for AdjacentlyTaggedEnumVariant {
         serializer.serialize_unit_variant(self.enum_name, self.variant_index, self.variant_name)
     }
 }
+
+// Error when Serialize for a non_exhaustive remote enum encounters a variant
+// that is not recognized.
+pub struct CannotSerializeVariant<T>(pub T);
+
+impl<T> Display for CannotSerializeVariant<T>
+where
+    T: Debug,
+{
+    fn fmt(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+        write!(formatter, "enum variant cannot be serialized: {:?}", self.0)
+    }
+}
--- a/serde_derive/src/internals/attr.rs
+++ b/serde_derive/src/internals/attr.rs
@@ -221,6 +221,7 @@ pub struct Container {
     is_packed: bool,
     /// Error message generated when type can't be deserialized
     expecting: Option<String>,
+    non_exhaustive: bool,
 }
 
 /// Styles of representing an enum.
@@ -306,9 +307,12 @@ impl Container {
         let mut variant_identifier = BoolAttr::none(cx, VARIANT_IDENTIFIER);
         let mut serde_path = Attr::none(cx, CRATE);
         let mut expecting = Attr::none(cx, EXPECTING);
+        let mut non_exhaustive = false;
 
         for attr in &item.attrs {
             if attr.path() != SERDE {
+                non_exhaustive |=
+                    matches!(&attr.meta, syn::Meta::Path(path) if path == NON_EXHAUSTIVE);
                 continue;
             }
 
@@ -587,6 +591,7 @@ impl Container {
             serde_path: serde_path.get(),
             is_packed,
             expecting: expecting.get(),
+            non_exhaustive,
         }
     }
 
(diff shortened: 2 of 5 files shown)


serde/src/private/ser.rs
```diff
@@ -1370,3 +1370,16 @@ impl Serialize for AdjacentlyTaggedEnumVariant {
         serializer.serialize_unit_variant(self.enum_name, self.variant_index, self.variant_name)
     }
 }
+
+// Error when Serialize for a non_exhaustive remote enum encounters a variant
+// that is not recognized.
+pub struct CannotSerializeVariant<T>(pub T);
```
Mingun, line 1376: Why not doc comment? Even if this type is internal this is not the reason to keep internals undocumented, of course if your aim is not to make life of other developers harder

dtolnay: 1. Rust doc comments have significant syntactic ceremony that makes them less readable than non-doc comments. For example a normal, readable comment like:

    ```rust
    // Cow<str> and Cow<[u8]> never borrow by default:
    //
    //     impl<'de, 'a, T: ?Sized> Deserialize<'de> for Cow<'a, T>
    //
    // A #[serde(borrow)] attribute enables borrowing that corresponds
    // roughly to these impls:
    //
    //     impl<'de: 'a, 'a> Deserialize<'de> for Cow<'a, str>
    //     impl<'de: 'a, 'a> Deserialize<'de> for Cow<'a, [u8]>
    ```

    would need to be made into something like the following, which has added a bevy of backticks and other noise that does not contribute any value to this as documentation.

    ```rust
    /// `Cow<str>` and `Cow<[u8]>` never borrow by default:
    ///
    /// ```ignore
    /// impl<'de, 'a, T: ?Sized> Deserialize<'de> for Cow<'a, T>
    /// ```
    ///
    /// A `#[serde(borrow)]` attribute enables borrowing that corresponds
    /// roughly to these impls:
    ///
    /// ```ignore
    /// impl<'de: 'a, 'a> Deserialize<'de> for Cow<'a, str>
    /// impl<'de: 'a, 'a> Deserialize<'de> for Cow<'a, [u8]>
    /// ```
    ```

2. Most contributors touching an internal-facing "doc comment" on a private function will never bother figuring the flags for getting it rendering through rustdoc to validate it renders in the way they meant. So this adds pointless review overhead to audit whether their markdown is appropriate. For example in the snippet above, needing to know that unescaped \<str\> and \[u8\] would screw up rustdoc.

Mingun: 1. Having doc comments helps to investigating internals, because editors is able to show documentation on items. They cannot show usual comments in the same way. I disagree that markdown syntax makes docs less readable, it's just a matter of habit
2. You can use `cargo doc --document-private-items` and all syntax mistakes will be catch even in private documentation

PR #2580, opened 2023-08-19
Provide opt-out `feature=from_source` to use `serde_derive` source

### LINKED ISSUE #2538: using serde_derive without precompiled binary

pinkforest:
Closes #2538
Supercedes #2579 

Follows my recommendation: https://github.com/serde-rs/serde/pull/2579#issuecomment-1684934840

**Current choice: Opt-Out** by `feature=from_source` (current HEAD commit)

Triggers build from source if used with feature `from_source` via either `serde` or `serde_derive`

**Alternative choice: Opt-Out** by `cfg(serde_derive_build = "source")` ([commit #6bccee](https://github.com/serde-rs/serde/commit/6bccee788ad3d566945640756deebf4a754af0d2))

Override via `cfg(serde_derive_build = "source")` to build from source

Has flaw re: cross compile as per @bjorn3 at https://github.com/serde-rs/serde/pull/2580#issuecomment-1685246113 due to Cargo limitation - but can be addressed via adocumentation as has been done via [commit #23059f7](https://github.com/serde-rs/serde/pull/2580/commits/23059f70c687ad8a461097441f35c5f06312bf81)

**Alterative: Opt-In** - My recommendation ([commit #8d544](https://github.com/serde-rs/serde/pull/2580/commits/8d54439c49fb7ade1b16cbd5ac72d57ea51de0b6))

Would make opt-in choice to use the precompiled binary with `cfg(serde_derive_build = "precompiled")` and leaving it to the top-level binary to use the precompiled blob if it explicitly opts in to - by default people will compile from source always.

**cfg Approach**

This also demonstrates the [cfg based approach](https://github.com/pinkforest/cfg_lottery_demo) that frees dependency chain from messing up with the features in the middle leaving it to the top level binary as informed choice.

This is used elsewhere e.g.: https://github.com/dalek-cryptography/curve25519-dalek/tree/main/curve25519-dalek#manual-backend-override - that was result of this discussion: https://github.com/dalek-cryptography/curve25519-dalek/issues/414 providing background why / how this approach has been used elsewhere.

Cc/ @sagudev I cherry picked


pinkforest:
@simonsan Could add a feature in addition but cfg is the easiest to handover the top-level binary the choice.

Please note that this PR makes it opt-in only where the top-level binary would need to want the precompiled one.

By default you would get it built from the source if this PR is merged.

kayabaNerve:
I will repeat my comment from #2579.

> I don't believe this adequately resolves https://github.com/serde-rs/serde/issues/2538. This would still cause unknown, unsigned, and unverifiable executables to placed on any systems which use serde as a dependency. That's still a notable security risk.

While I'm obviously in favor of this, as it's a massive improvement, this still leaves an unverifiable executable sitting on the system.

I'd state the binary should only be downloaded on opt-in.

pinkforest:
@kayabaNerv it is opt-in ? it doesn't download or does it ? Or did I miss the download somewhere ? let me check

oli-obk:
> serde-derive-bin which has the binary

That would certainly make it more explicit and easier to reject in a crate graph. But it doesn't actually solve too much, as afaik it will still be downloaded, just never built. Or at least show up in a Cargo.lock

andyquinterom:
As someone who works in the medical field with Rust I want to express my support for this

oli-obk:
Two things

1. I don't think changing the default to opt-in will be something that we'll land right now. I rather think that starting out with making it opt out will be something we'll be able to land while figuring out the entire situation.
2. Please document how to use this in the readme.

CryZe:
> I rather think that starting out with making it opt out will be something we'll be able to land while figuring out the entire situation.

Why? This feature didn't even exist until recently, is considered experimental and is widely considered a problem. This should be opt-in until the Rust community agrees as a whole that it is good (safe, reproducible, ...) enough to be opt-out.

oli-obk:
> Why?

Because this crate is owned by dtolnay and he may do with it as he wishes. My personal opinions on the topic are irrelevant. While I would prefer fundamental crates to have entire (funded!) teams working on them, that's not the reality.

I am reviewing this from the perspective of getting the PR landed.

novacrazy:
> Because this crate is owned by dtolnay and he may do with it as he wishes.

Serde has nearly 200 Million downloads on crates.io, and is a staple of the entire Rust ecosystem. It's downright disrespectful to the _entire_ Rust community to say their wants and needs are less important just because the one and only benevolent dictator has the publish rights on crates.io. There's a good argument that this crate should be adopted by the Rust org to avoid such bottlenecks. At this scale the need to appeal to someone's feelings is irrelevant. We need this to be opt-in until the day things upstream are safe enough to guarantee prebuilt binaries will be safe, perhaps using projects like Watt.

kayabaNerve:
@pinkforest While I agree anyone being able to include any file is a distinct issue, it being in a separate crate which isn't actually downloaded would solve that. I do see a comment claiming optional dependencies, even if unused, will appear in the crate graph _and be downloaded_ though, which may void my comment's validity.

oli-obk:
> It's downright disrespectful to the _entire_ Rust community to say their wants and needs are less important just because the one and only benevolent dictator has the publish rights on crates.io

I am not saying that. I am saying what the reality is. Not what I wish reality were.

pinkforest:
@oli-obk ok - I will add another commit to opt-out - can always re-base to previous commit if opt-in is chosen instead so choice will be with maintainers - I will also add doc how to use it - thanks

pinkforest:
New behaviour with `cfg(serde_derive_build = "source")`

**Default** - host x86_64-linux-gnu
```
$ cargo build
   Compiling serde_derive v1.0.183 (/home/foobar/code/serde/precompiled/serde_derive)
    Finished dev [unoptimized + debuginfo] target(s) in 0.16s
```

**Override** - host x86_64-linux-gnu
```
$ env RUSTFLAGS='--cfg serde_derive_build="source"' cargo build
   Compiling proc-macro2 v1.0.66
   Compiling unicode-ident v1.0.11
   Compiling quote v1.0.33
   Compiling syn v2.0.29
   Compiling serde_derive v1.0.183 (/home/foobar/code/serde/precompiled/serde_derive)
    Finished dev [unoptimized + debuginfo] target(s) in 3.17s
```

newpavlov:
@simonsan 
>How would I specify that I don't want the precompiled binary when I only pull in `serde` with the `serde_derive` feature like this?

The most flexible option is to use:
```toml
serde = { version = "1.0.x, <1.0.172", features = ["derive"] }
```
where `x` is minimum version required by your project.

@oli-obk 
>But it doesn't actually change too much, as afaik it will still be downloaded, just never built. Or at least show up in a Cargo.lock

IIRC unused crates do not get downloaded during compilation, but, yes, they are get mentioned in `Cargo.lock`.

From my experience, having a separate crate for pre-compiled version of the crate will be quite useful when dependencies get vendored. Even though it will be mentioned in `Cargo.lock`, we will not have a semi-random blob as part of our vendored source code.

bjorn3:
It is really easy to accidentally override the rustflags and thus opt back into the precompiled version. `RUSTFLAGS` will override `~/.cargo/config.toml` and a per-project `.cargo/config.toml` will override the global one. As such putting the opt-out once in your `~/.cargo/config.toml` may accidentally use the precompiled binary again if the project you build uses `.cargo/config.toml` or you or the project needs `RUSTFLAGS` (eg in rust's build system)

bjorn3:
Using rustflags as opt-out makes it easy to accidentally stop opting out. `RUSTFLAGS` and the per-project `.cargo/config.toml` take precedence over `~/.cargo/config.toml` and `RUSTFLAGS` takes presedence over the per-project `.cargo/config.toml`, so if you or the project needs to use either the opt-out will not be passed by cargo. This is for example the case in rust's build system.

README.md
```diff
@@ -73,6 +73,25 @@ fn main() {
 }
 ```
 
+## Note about serde_derive Binary
+
+When using `serde_derive` either directly or via `serde` with `derive` feature.
+
+Please note that `serde_derive` may be distributed from within pre-compiled binary
```
alerque, line 80:
```suggestion
Please note that `serde_derive` is distributed with a pre-compiled binary used by default on x86_64 platforms.
```

dvtkrlbs: that is not entirely true though this binary only exists on x86_64-unknown-linux-gnu . Not other targets. Maybe something like this be better `Please not that serde_drive may be distributed from within pre-compiled binary (currently on x86_64-unknown-linux-gnu)` and this might turn to a table if more targets start to distribute with the pre-compiled binary.

precompiled/serde_derive/src/lib.rs
```diff
@@ -15,8 +15,18 @@
...
+    serde_derive_build = "source",
+    target_arch = "x86_64",
+    target_os = "linux",
+    target_env = "gnu"
+))]
 include!("lib_from_source.rs");
 
-#[cfg(all(target_arch = "x86_64", target_os = "linux", target_env = "gnu"))]
+#[cfg(all(
+    not(serde_derive_build = "source"),
```
Kijewski, line 27: `cfg(all(not(…` → `cfg(not(all(…`

README.md
pinkforest: Rust is expressive and you can read the code for that level of detail. This is implementation detail of the decision.

precompiled/serde_derive/src/lib.rs
schoenenberg: @Kijewski this is not correct!

pinkforest: Hey chill up. There seems to be some difficulty I'm debugging with environment variables.

pinkforest:
I've added a simplified gating but then I've noticed when testing there is a bug related to cross compliation

The cross-compilation bug is here:
- #2582 

It seems this is always building on host `TARGET` configuration for some reason or maybe it's just me ?

precompiled/serde_derive/build.rs
oli-obk, line 1: Isn't this causing the build to slow down again? And change cargo's scheduling?

pinkforest: Probably - it would but I was debugging the cross-compilation problem in the meantime.
I'll get rid of it after I tackle the cross-compile issue. Let me also bench it.

pinkforest: Yeah I will revert this and fix the gating with build testing from dependency PoV - it would be good to have it in CI

pinkforest: Sigh this does not work
`[target.'cfg(any(serde_derive_build = "source", all(target_arch = "x86_64", target_os = "linux", target_env = "gnu")))'.dependencies]`
Debugging ... some problem with `cfg(any(.., all())` pattern

precompiled/serde_derive/Cargo.toml
```diff
@@ -21,7 +21,7 @@ deserialize_in_place = []
 [lib]
 proc-macro = true
 
-[target.'cfg(not(all(target_arch = "x86_64", target_os = "linux", target_env = "gnu")))'.dependencies]
+[target.'cfg(any(serde_derive_build = "source", all(target_arch = "x86_64", target_os = "linux", target_env = "gnu")))'.dependencies]
```
sagudev, line 24:
```suggestion
[target.'cfg(any(serde_derive_build = "source", not(all(target_arch = "x86_64", target_os = "linux", target_env = "gnu"))))'.dependencies]
```

this must be same as `include!("lib_from_source.rs");` cfg

pinkforest: Ahhh that was it! nice catch :heart_eyes: - I was in the weeds of getting this thing into CI and got stuck in the red herring

precompiled/serde_derive/build.rs
pinkforest: Okay thanks @sagudev - seems I missed the not in one place - this is why build .rs would have made it simler but granted it would slow down :heart:

precompiled/serde_derive/Cargo.toml
pinkforest: Oh except this still does not work :thinking: 

```
$ rm -rf target ; cargo build --target wasm32-unknown-unknown
   Compiling whatever v0.1.0 (/home/foobar/code/whatever)
   Compiling serde_derive v1.0.183 (/home/foobar/code/serde/precompiled/serde_derive)
warning: "wasm32-unknown-unknown"
    Finished dev [unoptimized + debuginfo] target(s) in 0.32s

```

sagudev:
Current CI errors are github fault as they fail on checkout.

precompiled/serde_derive/Cargo.toml
pinkforest: This is still the cross-compile error as per #2582 - it now works in host (I think)

pinkforest:
Upon closer inspection from dependency PoV - I've seemingly replicated the cross-compilation issue from dep side:
- https://github.com/serde-rs/serde/issues/2582#issuecomment-1685226535

(But this should work as-is regardless when compiling on host and GitHub gets itself together)

precompiled/serde_derive/Cargo.toml
sagudev: It works: https://github.com/sagudev/serde/actions/runs/5916401936/job/16043642831

bjorn3:
I just remember another issue with `RUSTFLAGS` opt-outs. Cargo ignores `RUSTFLAGS` for host dependencies like proc-macros when cross-compiling.

pinkforest:
@bjorn3 which level / version ? In stable / nightly the `[target.'cfg(..)'.dependencies]` gating seems to work for both manifest: https://github.com/serde-rs/serde/pull/2580#issuecomment-1685227828 and relaying to lib.rs ?

pinkforest:
CI :heavy_check_mark: :tada:

oli-obk:
I'll be able to try out the PR locally in 8 days when I'm back to having something more powerful than a phone 😅

I'll give it another code review some time later today

pinkforest:
This seems to have a fatal flaw as I both found and finally realized the hard way after @bjorn3 pointed out namedly `RUSTFLAGS` gets ignored to proc-macros upon cross-compiling :woman_facepalming: 

So it would have to be by feature (as much as difficulties those pose) per #2579 (+ documentation) :thinking: :thinking:  ?

bjorn3:
> Do you remember what the issue# was ?

https://github.com/rust-lang/cargo/issues/4423

pinkforest:
Ok this is now by feature as option when used with `from_source`

**x86_64-linux-gnu Host default**
```
$ rm -rf target ; cargo build --features from_source
   Compiling proc-macro2 v1.0.66
   Compiling unicode-ident v1.0.11
   Compiling quote v1.0.33
   Compiling syn v2.0.29
   Compiling serde_derive v1.0.183 (/home/foobar/code/serde/precompiled/serde_derive)
    Finished dev [unoptimized + debuginfo] target(s) in 3.18s
```


**wasm32-u-u cross-compile on x86_64-linux-gnu Host default**
```
[foobar@localhost serde_derive]$ rm -rf target ; cargo build --target wasm32-unknown-unknown
   Compiling serde_derive v1.0.183 (/home/foobar/code/serde/precompiled/serde_derive)
    Finished dev [unoptimized + debuginfo] target(s) in 0.29s
```

**wasm32-u-u cross-compile on x86_64-linux-gnu Host - Using from_source Feature**
```
[foobar@localhost serde_derive]$ rm -rf target ; cargo build --features from_source --target wasm32-unknown-unknown
   Compiling proc-macro2 v1.0.66
   Compiling unicode-ident v1.0.11
   Compiling quote v1.0.33
   Compiling syn v2.0.29
   Compiling serde_derive v1.0.183 (/home/foobar/code/serde/precompiled/serde_derive)
    Finished dev [unoptimized + debuginfo] target(s) in 3.46s
```

pinkforest:
@bjorn3 How do you feel this approach ? `feature=from_source` would be unified

oli-obk:
There's no env var, but maybe you can use the config settings described in https://doc.rust-lang.org/cargo/reference/config.html#buildrustflags

`target.CFG.rustflags` may apply to the host?

oli-obk:
> @bjorn3 @oli-obk How do you feel this approach ? `feature=from_source` would be unified

How does this work if used as a dependency?

pinkforest:
> How does this work if used as a dependency?

Would need to use the dependency with optional feature included - cargo adds them together so one dependency adding it would make it build from source in whole build as a sum.

But I will test the [build.rustflags](https://doc.rust-lang.org/cargo/reference/config.html#buildrustflags) w/ `CARGO_ENCODED_RUSTFLAGS` - that might be an option too to provide.

precompiled/serde_derive/Cargo.toml
```diff
@@ -17,10 +17,18 @@ rust-version = "1.56"
...
 [lib]
 proc-macro = true
 
+# By feature source compilation
+[dependencies]
+proc-macro2 = { version = "1", optional = true }
+quote = { version = "1", optional = true }
+syn = { version = "2.0.28", optional = true }
+
+# Sourc compiling when no precompiled available
```
oli-obk, line 31:
```suggestion
# Compile from source when no precompiled binary available
```

bjorn3:
If serde_derive is used for a build dependency afaik with the v2 resolver it wouldn't get feature unified with serde_derive used by a non-build dependency or the root crate.

README.md
```diff
@@ -73,6 +73,11 @@ fn main() {
 }
 ```
 
+## Note about serde_derive Binary
```
pinkforest, line 76: @oli-cbk The approach with either `CARGO_ENCODED_RUSTFLAGS` or given neither these .cargo/config work:

build.target
```toml
[build]
target = "wasm32-unknown-unknown"

[target.wasm32-unknown-unknown]
rustflags = ['--cfg=serde_derive_build="source"']
```

build.rustflags
```toml
[build]
rustflags = ['--cfg=serde_derive_build="source"']
```

Nemo157: To get rustflags into build script builds you need to use `-Zhost-config` and

```toml
[host]
rustflags = [...]
```

precompiled/serde_derive/Cargo.toml
pinkforest: thanks :sweat_smile:

README.md
pinkforest: Oh we have a winner! Thanks @Nemo157 :heart:

```
$ cargo -Zhost-config build
error: the -Zhost-config flag requires the -Ztarget-applies-to-host flag to be set
[foobar@localhost whatever]$ cargo -Ztarget-applies-to-host -Zhost-config build
   Compiling proc-macro2 v1.0.66
   Compiling unicode-ident v1.0.11
   Compiling whatever v0.1.0 (/home/foobar/code/whatever)
warning: "x86_64-unknown-linux-gnu"
   Compiling quote v1.0.33
   Compiling syn v2.0.29
   Compiling serde_derive v1.0.183 (/home/foobar/code/pushpush/serde/precompiled/serde_derive)
    Finished dev [unoptimized + debuginfo] target(s) in 2.84s
[foobar@localhost whatever]$ cargo -Ztarget-applies-to-host -Zhost-config build --target wasm32-unknown-unknown
   Compiling proc-macro2 v1.0.66
   Compiling unicode-ident v1.0.11
   Compiling whatever v0.1.0 (/home/foobar/code/whatever)
warning: "wasm32-unknown-unknown"
   Compiling quote v1.0.33
   Compiling syn v2.0.29
   Compiling serde_derive v1.0.183 (/home/foobar/code/pushpush/serde/precompiled/serde_derive)
    Finished dev [unoptimized + debuginfo] target(s) in 3.43s
[foobar@localhost whatever]$ cat .cargo/config
[host]
rustflags = ['--cfg=serde_derive_build="source"']
```

When were these -Z introduced ? :thinking: so we can document it as required version ?

bjorn3: I think using an unstable option for the opt-out is wrong. Also it still has the same issue of getting overriden by other ways to specify rustflags rather than appending to it.

najtin:
Hey, initially an opt-in version was pitched. Is opt-in off the table now?

Piturnah:
> Hey, initially an opt-in version was pitched. Is opt-in off the table now?

You're right, that was the initial pitch. The discussion around why it was changed has been marked as "not relevant" (not at all sure why) but you can find it in this thread by expanding those comments.

README.md
oli-obk: Thanks for trying it out

pinkforest: Could we live with RUSTFLAGS= for host builds and if someone wants to cross-build they can opt-in to nightly -Z's ?
I mean just looking at the minimum mergeable solution here :/

oli-obk: It's fragile (or rather: easy to get wrong). Is there some solution to the resolver=2 problem?

pinkforest: Good question need to build a test-case for it and prove it works I'll do that shortly.

LuckyTurtleDev:
> Hey, initially an opt-in version was pitched. Is opt-in off the table now?

I personally would also prefer an opt-in solution. But if any crate you used has opt-in to use the binary, you can not disable it anymore. So opt-out would probably the better solution.

bjorn3:
Opt-in through RUSTFLAGS doesn't have that problem.

README.md
pinkforest: Yeah so I've replicated the resolver problem. Maybe we can just document this ?

pinkforest: This was the repro for it: https://github.com/pinkforest/serde_optout_unification_proof

pinkforest:
@bjorn3 I just tested using [commit #585482](https://github.com/serde-rs/serde/pull/2580/commits/5854820c21b9c1de194f48ab442c230632a87421) the resolver="2" and it seems to work regardless ?

Repro here: https://github.com/pinkforest/serde_optout_unification_proof

**Source** by opt-out
```
top-bin]$ rm -rf ../target ; cargo build --features from_source
   Compiling proc-macro2 v1.0.66
   Compiling unicode-ident v1.0.11
   Compiling quote v1.0.33
   Compiling syn v2.0.29
   Compiling serde_derive v1.0.183 (/home/foobar/code/serde/precompiled/serde_derive)
   Compiling some-lib v0.1.0 (/home/foobar/code/serde_optout_unification_proof/some-lib)
   Compiling top-bin v0.1.0 (/home/foobar/code/serde_optout_unification_proof/top-bin)
    Finished dev [unoptimized + debuginfo] target(s) in 2.97s
```

**Default** precompiled
```
top-bin]$ rm -rf ../target ; cargo build
   Compiling serde_derive v1.0.183 (/home/foobar/code/serde/precompiled/serde_derive)
   Compiling some-lib v0.1.0 (/home/foobar/code/serde_optout_unification_proof/some-lib)
   Compiling top-bin v0.1.0 (/home/foobar/code/serde_optout_unification_proof/top-bin)
    Finished dev [unoptimized + debuginfo] target(s) in 0.33s
```

Anything I did wrong here ?

README.md
oli-obk: Alright with me. Is there something that can be done to ensure build dependencies also have the feature enabled?

pinkforest: It seems I didn't clear the target properly between my repro previously

and it seems I can't replicate the problem with feature unification

e.g. I don't see anything wrong with the feature approach ?

Nemo157: That's because it's a proc-macro crate, which is its own kind of dependency similar to a build-dependency, which is built for a single target: the host; so both the build-dep and normal-dep refer to the same proc-macro-dependency and their features get unified in it. A more accurate `cargo tree` would look like

```
top-bin v0.1.0 (/tmp/tmp.mNq6xeO1Ms/serde_optout_unification_proof/top-bin)
└── some-lib v0.1.0 (/tmp/tmp.mNq6xeO1Ms/serde_optout_unification_proof/some-lib)
    [proc-macro-dependencies]
    └── serde_derive v1.0.183 (proc-macro) (https://github.com/serde-rs/serde?rev=cb23de92e7f479139add329e8629eb3b179109f6#cb23de92)
        ├── proc-macro2 v1.0.66
        │   └── unicode-ident v1.0.11
        ├── quote v1.0.33
        │   └── proc-macro2 v1.0.66 (*)
        └── syn v2.0.29
            ├── proc-macro2 v1.0.66 (*)
            ├── quote v1.0.33 (*)
            └── unicode-ident v1.0.11
[build-dependencies]
└── some-build-lib v0.1.0 (/tmp/tmp.mNq6xeO1Ms/serde_optout_unification_proof/some-build-lib)
    [proc-macro-dependencies]
    └── serde_derive v1.0.183 (proc-macro) (https://github.com/serde-rs/serde?rev=cb23de92e7f479139add329e8629eb3b179109f6#cb23de92) (*)
```

AlexTMjugador:
I'd like to mention that the decision of using precompiled binaries for Linux x64 platforms is already causing significant ecosystem fragmentation. That might be useful to keep in mind for reviewing this PR: https://github.com/time-rs/time/discussions/611

jnicholls:
IMHO Opt-in for the precompiled binary should be the desired outcome. It is backwards compatible with the ecosystem's current serde dependency specifications, and its expectations before learning about the precompiled binary.

GoWeasel:
As followed this issue, just mention to **care about** upcoming cyber resillience act, 
only/or by default **just** binary would drop out serde of projects.

bjorn3:
> @bjorn3 fwiw - I just tested parked at [commit #585482](https://github.com/serde-rs/serde/pull/2580/commits/5854820c21b9c1de194f48ab442c230632a87421) the resolver="2" and it seems to work regardless ?

Maybe it only applies to non-proc-macro deps? In any case great that it works.

joshka:
@AlexTMjugador wrote:
> I'd like to mention that the decision of using precompiled binaries for Linux x64 platforms is already causing significant ecosystem fragmentation. That might be useful to keep in mind for reviewing this PR: [time-rs/time#611](https://github.com/time-rs/time/discussions/611)

Regarding this - do you think it would be worth putting forward a PR that fully reverts the precompiled PR until this or #2579 land (or whatever is the real fix)? To me that approach seems like it will give some breathing room to work out the correct fix.

AlexTMjugador:
> Regarding this - do you think it would be worth putting forward a PR that fully reverts the precompiled PR until this or #2579 land (or whatever is the real fix)? To me that approach seems like it will give some breathing room to work out the correct fix.

I'd support that, but I'm not a `serde` repository maintainer to decide.

joshka:
> I'd support that, but I'm not a `serde` repository maintainer to decide.

Yep - I understand that. I read your reply over in the time issue and figured you might have a reasonable perspective on this one way or another.

faulesocke:
`from_source` should really be on by default, making it *not* include the binary in the default feature set.

dtolnay (changes requested):
This code is no longer used since serde_derive 1.0.184 &mdash; would you be interested in sending a different PR to delete it? The `precompiled` directory and anything having to do with `cfg(precompiled)` in the rest of the code.

Thanks, I appreciate your diligence on this PR, and the dozens of commenters who helped brainstorm the possible approaches for an opt-out and opt-in implementation.

pinkforest:
Thanks will do - appreciate looking over - I can just override this existing branch in this PR no problem

dtolnay:
Nice. I'd prefer a new PR to preserve the final working approach that was settled on in this one for opt-out.

pinkforest:
Ok great no worries :heart:

PR #2587, opened 2023-08-20
fix shebang in build.sh

### LINKED ISSUE #2583: change shebang to be more compatible
This

https://github.com/serde-rs/serde/blob/0676673ca5efef667d10a7b63b5b078a8d64e43a/precompiled/build.sh#L1

should be

```bash
#!/usr/bin/env bash
````

in order to maximize compatibility. On many more nieche systems `/bin/bash` does not exist (`/bin/sh` would be fine!), but `/usr/bin/env` and `env` would be able to find `bash` on most if not all systems _if_ `bash` is present at all.

wucke13:
This fixes #2583.

### DIFF
--- a/precompiled/build.sh
+++ b/precompiled/build.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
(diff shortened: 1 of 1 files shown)


oli-obk:
Thanks!

PR #2588, opened 2023-08-21
guarantee `serde` is in lockstep with `serde_derive`

soqb:
# Objective
- Implement #2584.
- Pave the way to removing the compile-time bottleneck in `serde_derive`.

# Solution
- Implement the [associated proc macro pattern](https://github.com/matklad/macro-dep-test/), discovered by the excellent @matklad, to embed `serde_derive` in `Cargo.lock` files for projects that include `serde` without `serde_derive` being forced as a dependency, so users can do the following in their `Cargo.toml`:
```toml
[dependencies]
serde = "1"
serde_derive = "1" # these two crates are guaranteed to have the same version.
```
- I also added a job to CI that verifies `serde` and `serde_derive` have the same version in a new dummy crate, though this does introduce some dev-dependencies which could be split out.


soqb:
its going to take a better `cargo` wizard than me to figure out how to fix CI. i'm going to bed!

soqb:
i don't seem to be able to get the "Outdated" check to pass, since the new CI check has to rely on an old version of `cargo-lock`, otherwise dependency resolution will fail on rust 1.36

dtolnay (approved):
Thanks!

PR #2590, opened 2023-08-21
Phase out precompiled

pinkforest:
Following consensus on: https://github.com/serde-rs/serde/pull/2580#pullrequestreview-1586105370

This PR phases out the precompiled per final consensus made in #2580


dtolnay (commented):
To clarify, I don't actually think #2588 has anything to do with the ability to undo precompiled. Serde\_derive is going to continue to be slow to compile after this. The version equality constraint thing is an independent improvement that's useful for cases that already depend on serde / serde_derive separately to parallelize the (still slow) builds (https://github.com/niklasmohrin/clircle/pull/12, https://github.com/pfernie/cookie_store/pull/37, https://github.com/dtolnay/noisy-clippy/commit/3838953bf5ab95b6564dcf8e51fe44da34fbcf63, &hellip;).

Could you unstack this from #2588 if it includes anything from there as indicated by your PR description? I was hoping this one could land first. For the other one it may turn out to be wise to wait several weeks in case there are some unsatisfiable constraints floating around like `serde = "1.0.177", serde_derive = "=1.0.171"` in the near term, which that PR would break. I might need to write tooling to detect that.

pinkforest:
yup done unstacked

dtolnay (approved):
Thanks!

pinkforest:
Sorry bad old habit squashing my branch :grimacing: :sweat_smile:

pinkforest:
CI :green_heart: :tada:

delbonis:
Sorry for the confusion but what's the final conclusion on the default behavior?  How should binaries decide to pick their desired build process?

kayabaNerve:
The binary has been deleted entirely so there is no choice, only the from source option.

PR #2594, opened 2023-08-21
Fix: Use `&[T]` instead of `&Vec<T>` wherever possible

Uzaaft:
Found a few places where `&Vec<T>` is used and could be swapped with `&[T]`.


oli-obk:
I think there's a clippy lint for that. Maybe we need to activate it manually?

dtolnay (approved):
Thanks!

Issue #2181, opened 2022-02-26
[Suggestion] Split out serde into an additional serde_core crate


PoignardAzur:
There's been some activity recently around [the benchmark suite published by lqd](https://github.com/lqd/rustc-benchmarking-data). Those benchmarks are a trove of useful data on how rust crates are built and where the time is spent.

Looking at the benchmarks, I noticed that in multi-core builds, [quite](https://lqd.github.io/rustc-benchmarking-data/results/round-12-cargo-timing-debug-j8/cargo-timing-cargo_metadata-0.14.1-debug-j8.html) [a lot](https://lqd.github.io/rustc-benchmarking-data/results/round-12-cargo-timing-debug-j8/cargo-timing-csv-1.1.6-debug-j8.html) [of](https://lqd.github.io/rustc-benchmarking-data/results/round-12-cargo-timing-debug-j8/cargo-timing-assert-json-diff-2.0.1-debug-j8.html) [crates](https://lqd.github.io/rustc-benchmarking-data/results/round-12-cargo-timing-debug-j8/cargo-timing-jsonrpc-client-transports-18.0.0-debug-j8.html) had a bottleneck around serde and serde_derive.

Because serde reexports serde_derive's macro, the compilation order is always `syn -> serde_derive -> serde`, sequentially. In the benchmarks, this usually accounts for around 9s that cargo fails to parallelize; rustc can do some parallelizing of its own during the codegen phase, but it's still a lot of CPU usage left on the table.

Instead I suggest the crate structure should be:

* serde
  * serde_core
  * serde_derive

With the serde crate being a thin wrapper re-exporting types from the other two. That way the compilation pipeline would be `syn -> serde_derive` on one side, and `serde_core` on the other.

Another advantage is that crates could import serde_core instead of serde if they don't care about the macros; that way, they wouldn't be blocked on the compilation of syn + serde even if some other crate in the tree imports serde with the `derive` feature.

Looking at the benchmarks, it looks like doing so could shave 2 or 3 seconds off total build time for multi-core builds. The benchmarks may not be representative of real projects (they test lib crates, not bins), though. Overall, the smaller the project, the more likely serde is to be in its critical path, so the more likely it is to see improvement from a refactoring.

## Possible issues

* Backwards compatibility (though it *should* be fine?)
* Documentation: items can either appear as `serde::Serialize` or `serde_core::Serialize`. Not sure how much of a problem it actually is.

dtolnay:
[...]

Issue #1070, opened 2017-10-22
Remove serialize and deserialize impls for Result


dtolnay:
I don't know of these impls ever being useful, and they lead to confusion when the user writes too many or too few `?` or `unwrap`.

```rust
extern crate serde;
extern crate serde_json;

use serde::de::DeserializeOwned;

fn parse_t<T>() -> Result<T, String>
    where T: DeserializeOwned
{
    let j = r#" {} "#;

    // Compiles, but fails at runtime to deserialize `Result<T, String>`
    serde_json::from_str(j).unwrap()
}
```

PR #2608, opened 2023-09-04
feat: Extract serde_core out of serde crate

osiewicz:
This PR implements approach suggested in #2584 by @epage and @soqb. Also https://github.com/serde-rs/serde/issues/2181
There were two changes initially that were since reverted during the lifetime of this PR:
- I've changed serde_derive to use `serde_core` instead of `serde`. This has since been reverted as it would be a breaking change. 
- Code examples in serde_core aliased serde_core as serde so that the docs reexported in serde refer to `serde` and not `serde_core`. This however was not necessary following commit 2258bb4

Before (11.4s for a clean release build with derive feature, 8 cores):
obraz

After (7.8s for a clean release build with derive feature, 8 cores):
serde build graph after change (down to 8s)


osiewicz:
Ugh, seems like there's a bunch of CI failures on `set up job` stage.

epage:
Besides the gains here, this allows serde_json and friends to depend on serde_core, preventing packages using serde w/ derive from serializing things.

serde/src/lib.rs
```diff
@@ -92,6 +92,13 @@
 //! [Hjson]: https://github.com/Canop/deser-hjson
 //! [CSV]: https://docs.rs/csv
 
+<<<<<<< HEAD
```
NobodyXu: Note that this is a merge conflict

NobodyXu:
There're merge conflicts on this branch again which needs to be resolved.

osiewicz:
Hey, the merge conflicts on this PR are usually trivial to resolve (the PR itself doesn't change anything about the code in serde, just it's structure) and they'll pop up with each new serde version released; I am not sure if it makes sense to massage it all the time given that this PR has went almost unnoticed for 2 months now. I'd say it should be fine to resolve them once this PR gets some attention. =)

NobodyXu:
Hmmm I thought this PR is blocked because merge conflicts and would get some attentions once it is resolved, but it turns out that it just doesn't get much attention regardless of merge conflicts.

This PR would really help improve compilation time of serde with derive enabled and I was looking forward to it, it's sad that it has been blocking for so long was no review from maintainer.

osiewicz:
Yeah, I think so too. On the other hand, it was not asked for so no hard feelings there. 
I would also love to see it merged.

But yeah, the conflicts will always be there whenever a new version of serde is released, as this PR moves half of the files into other directory (which `git` seems to be able to resolve conflicts for on it's own) + the version number changes. Either way, if there's a consensus to merge it, we can just rebase the changes onto the current master and that should do it.

oli-obk:
Two things:

* Rebases are preferred over merges. A rebase that only conflicts due to file moves is usually resolved trivially without user input
* nightly is broken, you may need to run with nightly, too and adjust some imports
* I'm not going to push for this in the next two months as I'm going on leave soon
* I can't count 😆

osiewicz:
Hey, should we come back to this PR whenever you're available then? By that I also mean whether I should fix it up in the meantime or is it fine to just come back to it in 2 months and rebase it then?

oli-obk:
I think it would be fine to just not rebase it. Either @dtolnay will pick it up at whatever state it is in while I'm gone and ask you to rebase, or I will when I'm back.

epage:
As a real-world example of how this can be beneficial, I was looking at the timings chart for ripgrep on the [parallel frontend blog post](https://blog.rust-lang.org/2023/11/09/parallel-rustc.html).  I did [some analysis](https://www.reddit.com/r/rust/comments/17rd8ww/faster_compilation_with_the_parallel_frontend_in/k8igjlg/) on it.  I'm guessing this change on its own would take ripgrep build times on a 24-core machine from 9s to 6-7s.

dtolnay (commented):
Thanks for working on this!

Some things I would want to see further consideration of before deciding how to move forward with this design are:

- the effect on compiler diagnostics seen by users, or other potential friction from this change, particularly when it comes to users who may be new to Rust

- implications for serde 2.0 when it comes to the private helper types that serde_derive relies on

- implications for serde 2.0 when it comes to https://github.com/serde-rs/serde/labels/wishlist%20for%202%2E0

(Maybe @oli-obk can work with you to flesh these out.)

### Compiler diagnostics

After this change, what are the situations in which diagnostics that used to look like _"the trait bound \`T: serde::ser::Serialize\` is not satisfied"_ will instead say _"the trait bound \`T: serde_core::ser::Serialize\` is not satisfied"_? &mdash; these occur often when you are using a data format library like serde_json together with a data structure library whose Serde trait impls are behind a feature.

I am not familiar with the heuristics rustc uses to choose whether to display the trait path in these errors as `serde::ser::Serialize` or `serde_core::ser::Serialize` or `serde_core::Serialize` or just `Serialize`.

For errors that refer to `serde`, users have been able to unblock themselves well enough by finding that the crate's documentation mentions "serde support" behind a feature, or noticing in `cargo add` it printed that there is a serde feature, or noticing that the relevant `impl serde::Serialize for ...` in the crate's source code has a conditional compilation attribute.

A crate is unlikely to document that it has "serde_core support". To some users it will be obvious that a `serde` trait impl is interchangeable with a `serde_core` trait impl, but this isn't the case in general. For example `thiserror` and `thiserror_core` are unrelated in this way, `serde_json` and `serde_json_core` are unrelated, etc. Maybe consider whether `serde_traits`, or something else, would be better naming.

I think this is probably inevitable given that moving the traits to a different crate that is a dependency of serde is most likely required for ever publishing serde_derive 2.0. (Although [semver trick](https://github.com/dtolnay/semver-trick) is another option here, which `log` used successfully in the past.) So this one is a nonblocking concern. Nevertheless it would be good to understand if there is a need for any kind of followup in rustc. Serde is commonly one of the very first crates that someone new to Rust would first use, so there is a lot of leverage to scare users off of Rust with hours of frustration, likely greater than the cost of 3 seconds in a first-time build.

What mitigation options exist, and how do they compare? Is setting `[lib] name = "serde"` on serde_core better or worse?

### Private helper types

Currently this PR has moved all of `serde::__private` into serde_core.

That may not fly. It means that someone is able to write:

```toml
serde = { package = "serde_core", version = "1" }
serde_derive = "1"
```

which then means all of serde_derive 1.0's helper types need to be present in serde_core in perpetuity, even all the ones serde_derive 2.0 does not use, even if serde 1.0 is not in the build graph. I have not measured recently, but these helper types account for a nontrivial part of the compile time of serde.

What alternative options exist, and how do they compare? Keep helper types in serde for now? Put helper types in a third crate somehow? Some feature voodoo? Progress on https://github.com/rust-lang/rust/issues/54363 would be valuable here too.

### Items from 2.0 wishlist

Relocating traits into a different crate is a one-time deal. When we do serde_derive 2.0 (in which I'd like to remove support for many attributes that are too niche, and some changes like `default = $expr` should take an expression instead of a string literal) and serde 2.0 to go with it, it needs to be written against serde_core 1.0 because the cost–benefit from improving the macro is a lot better than the cost–benefit of breaking the ecosystem with incompatible traits.

That means some of what might be on the table for serde 2.0 needs to be accounted for now. In particular, this would be the time to figure out what(whether) to do about https://github.com/serde-rs/serde/issues/1070. Do the Serialize and Deserialize impl for `Result` need to be behind a feature that is enabled only by serde 1.0 and not 2.0? Missing that now would be a blunder because it cannot be fixed later.

Hopefully that one is the only 2.0 item needing consideration, but I am not sure.

serde/Cargo.toml
```diff
@@ -16,6 +15,7 @@ rust-version = "1.31"
 
 [dependencies]
 serde_derive = { version = "1", optional = true, path = "../serde_derive" }
+serde_core = {version = "1", path = "../serde_core", default-features = false }
```
dtolnay, line 18: This needs a "=1.0.Z" constraint because otherwise the serde_derive constraint below on line 40 is ineffective. A version of serde_derive could get mixed with a nonmatching version of serde_core that contains the wrong helper types.

serde_core/Cargo.toml
```diff
@@ -0,0 +1,56 @@
+[package]
+name = "serde_core"
+version = "1.0.188" # remember to update html_root_url and serde_derive dependency
+authors = ["Erick Tryzelaar <erick.tryzelaar@gmail.com>", "David Tolnay <dtolnay@gmail.com>"]
+build = "build.rs"
+categories = ["encoding", "no-std", "no-std::no-alloc"]
+description = "A generic serialization/deserialization framework"
+documentation = "https://docs.rs/serde"
```
dtolnay, line 8: Does it make a difference whether this is serde or serde_core?

serde_core/Cargo.toml
```diff
@@ -0,0 +1,56 @@
...
+edition = "2018"
+homepage = "https://serde.rs"
+keywords = ["serde", "serialization", "no_std"]
+license = "MIT OR Apache-2.0"
+readme = "crates-io.md"
```
dtolnay, line 13: This file does not exist.

serde_core/Cargo.toml
```diff
@@ -0,0 +1,56 @@
+[package]
+name = "serde_core"
+version = "1.0.188" # remember to update html_root_url and serde_derive dependency
+authors = ["Erick Tryzelaar <erick.tryzelaar@gmail.com>", "David Tolnay <dtolnay@gmail.com>"]
+build = "build.rs"
+categories = ["encoding", "no-std", "no-std::no-alloc"]
+description = "A generic serialization/deserialization framework"
```
dtolnay, line 7: Please update this description to concisely convey the intended usage of this crate in contrast to the serde crate.

serde_core/Cargo.toml
```diff
@@ -0,0 +1,56 @@
...
+homepage = "https://serde.rs"
+keywords = ["serde", "serialization", "no_std"]
+license = "MIT OR Apache-2.0"
+readme = "crates-io.md"
+repository = "https://github.com/serde-rs/serde"
+rust-version = "1.31"
+
+[dev-dependencies]
+serde_derive = { version = "1", path = "../serde_derive" }
+serde = {version = "1", path = "../serde"}
```
dtolnay, line 19:
```suggestion
serde = { version = "1", path = "../serde" }
```

serde_core/Cargo.toml
```diff
@@ -0,0 +1,56 @@
...
+
+[lib]
+doc-scrape-examples = false
+
+[package.metadata.playground]
+features = ["derive", "rc"]
```
dtolnay, line 25: There is no `derive` feature on this crate.
```suggestion
features = ["rc"]
```

serde_core/Cargo.toml
```diff
@@ -0,0 +1,56 @@
...
+
+[package.metadata.docs.rs]
+features = ["derive"]
```
dtolnay, line 28: There is no `derive` feature on this crate.

serde_core/LICENSE-APACHE
```diff
@@ -0,0 +1 @@
+../LICENSE-APACHE
```
dtolnay, line 1: Could you make sure this is correctly marked as a symlink? I think it isn't; GitHub's review ui normally makes that clear.

dtolnay (changes requested):

epage:
> After this change, what are the situations in which diagnostics that used to look like "the trait bound `T: serde::ser::Serialize` is not satisfied" will instead say "the trait bound `T: serde_core::ser::Serialize` is not satisfied"? — these occur often when you are using a data format library like serde_json together with a data structure library whose Serde trait impls are behind a feature.

Not tested with `serde_core` (since more changes are needed) but forced some serde compiler errors to see what I'd get.

Using a type in a `#[derive(Deserialize)]` struct's field that doesn't implement `Deserialize`.  Notice a lack of `serde::`.   I even removed `use serde::Deserialize` to double check that wasn't the reason why.
```
error[E0277]: the trait bound `StringA: Deserialize<'_>` is not satisfied
    --> crates/cargo-util-schemas/src/manifest.rs:30:25
     |
30   |     pub cargo_features: Option<Vec<StringA>>,
     |                         ^^^^^^^^^^^^^^^^^^^^ the trait `Deserialize<'_>` is not implemented for `StringA`
     |
     = help: the following other types implement trait `Deserialize<'de>`:
               bool
               char
               isize
               i8
               i16
               i32
               i64
               i128
             and 192 others
     = note: required for `Vec<StringA>` to implement `Deserialize<'_>`
     = note: 1 redundant requirement hidden
     = note: required for `std::option::Option<Vec<StringA>>` to implement `Deserialize<'_>`
note: required by a bound in `next_element`
    --> /home/epage/.cargo/registry/src/index.crates.io-6f17d22bba15001f/serde-1.0.196/src/de/mod.rs:1726:12
     |
1724 |     fn next_element<T>(&mut self) -> Result<Option<T>, Self::Error>
     |        ------------ required by a bound in this associated function
1725 |     where
1726 |         T: Deserialize<'de>,
     |            ^^^^^^^^^^^^^^^^ required by this bound in `SeqAccess::next_element`
```

Trait not in scope.  Note the suggestion of `use crate::manifest::_::_serde::Deserialize` rather than `serde::Deserialize`.  I suspect the compiler is picking up on an import generated by `serde_derive`.
```
error[E0599]: no function or associated item named `deserialize` found for struct `TomlInheritedField` in the current
scope
   --> crates/cargo-util-schemas/src/manifest.rs:292:37
    |
292 |                 TomlInheritedField::deserialize(mvd).map(InheritableField::Inherit)
    |                                     ^^^^^^^^^^^ function or associated item not found in `TomlInheritedField`
...
444 | pub struct TomlInheritedField {
    | ----------------------------- function or associated item `deserialize` not found for this struct
    |
note: if you're trying to build a new `TomlInheritedField`, consider using `TomlInheritedField::new` which returns `To
mlInheritedField`
   --> crates/cargo-util-schemas/src/manifest.rs:449:5
    |
449 |     pub fn new() -> Self {
    |     ^^^^^^^^^^^^^^^^^^^^
    = help: items from traits can only be used if the trait is in scope
help: the following trait is implemented but not in scope; perhaps add a `use` for it:
    |
8   + use crate::manifest::_::_serde::Deserialize;
```

epage:
> What alternative options exist, and how do they compare? Keep helper types in serde for now? Put helper types in a third crate somehow? Some feature voodoo? Progress on https://github.com/rust-lang/rust/issues/54363 would be valuable here too.

My assumption for this is the helpers would be in `serde` and we keep existing `derive` users working the same way we do today. 
- No breaking change for now
- Isolate the semver compatibility guarentees

Maybe with `serde@2`, we could switch to `serde_derive` + `serde_derive_macro` where `serde_derive` has the helpers but doing that is dependent on a lot of different factors that don't seem worth discussing atm.

serde_core/Cargo.toml
osiewicz: I think this should point to `serde_core`. `serde` API matches that of `serde_core` verbatim, but it may not always be the case.

osiewicz:
Just to add to second example provided by @epage; it looks like rustc may sometimes pick up a different path to (De)serialize trait, based on whether user has serde in non-macro scope. 
Baseline code:
```
use serde;

fn main() {
    let _ = u32::deserialize("");
}
```
<details><summary>Stdout with `use serde`</summary>
<p>

> error[E0599]: no function or associated item named `deserialize` found for type `u32` in the current scope
>  --> src/main.rs:7:23
>   |
> 7 |     let _: u32 = u32::deserialize(());
>   |                       ^^^^^^^^^^^ function or associated item not found in `u32`
>   |
>   = help: items from traits can only be used if the trait is in scope
> help: the following trait is implemented but not in scope; perhaps add a `use` for it:
>   |
> 1 + use serde::Deserialize;
>   |
> 
> For more information about this error, try `rustc --explain E0599`.
> warning: `throwaway_serde` (bin "throwaway_serde") generated 2 warnings
> error: could not compile `throwaway_serde` (bin "throwaway_serde") due to previous error; 2 warnings emitted

</p>
</details> 
<details><summary>Stdout with `use serde_core`</summary>
<p>
> error[E0599]: no function or associated item named `deserialize` found for type > `u32` in the current scope
>  --> src/main.rs:7:23
>   |
> 7 |     let _: u32 = u32::deserialize(());
>   |                       ^^^^^^^^^^^ function or associated item not found in> `u32`
>   |
>   = help: items from traits can only be used if the trait is in scope
> help: the following trait is implemented but not in scope; perhaps add a `use` > for it:
>   |
> 2 + use serde_core::Deserialize;
>   |
> 
> For more information about this error, try `rustc --explain E0599`.
> warning: `throwaway_serde` (bin "throwaway_serde") generated 1 warning
> error: could not compile `throwaway_serde` (bin "throwaway_serde") due to > previous error; 1 warning emitted
</p>
</details> 

When both serde and serde_core are used, the one that's imported first is the one that's suggested in diagnostics.

serde_core/src/lib.rs
```diff
@@ -0,0 +1,242 @@
...
+
+#[doc(inline)]
+pub use crate::de::{Deserialize, Deserializer};
+#[doc(inline)]
+pub use crate::ser::{Serialize, Serializer};
+
+// Used by generated code and doc tests. Not public API.
+#[doc(hidden)]
+#[path = "private/mod.rs"]
+pub mod __private;
```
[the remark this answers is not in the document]
osiewicz: It looks like to keep `private` in `serde`, we'd have to figure out what to do about `__deserialize_content` in `Deserializer` trait. `private::de` is a substantial contributor to `private`'s line count, so I guess having some of `private` in `serde` and then rest in `serde_core` wouldn't quite cut it.
Instead, I've added a new feature to serde_core (`private`) that enables compilation of `serde_core::private::{de, ser}` module (kinda like `no_serde_derive` rustc cfg). `serde_core::private` is then enabled with `serde::derive`.

I've also updated other benchmarks, as one in PR description was taken on a relatively old machine I don't have access to right now.
Master (derive=true, 3.8s)
image
This branch, commit 49d028d, derive=true, 2.4s:
image
This branch, patch 
[private_feature.patch](https://github.com/serde-rs/serde/files/14448278/private_feature.patch), derive=false:
image

So all in all, getting rid of `private` in serde_core would net us about 200ms out of 1400ms on M1 Max.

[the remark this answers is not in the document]
epage: My concern about having `private` in `serde_core` is that it requires continued strict coupling between `serde_core` and `serde_derive`.  For example, I suspect this would make having a `serde_derive` v2 more difficult because you couldn't run it in parallel to `serde_derive` v1 without much finagling.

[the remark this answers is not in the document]
oli-obk: I understood it as "moving private to serde" is a net gain, so that's what I meant with "let's do that":

Let's move it to serde and out of serde_core entirely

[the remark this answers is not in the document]
osiewicz: Hey,
Just to give an update, I've encountered an issue with non-public API of Deserialize (`__deserialize_content`); should it still be a part of `serde_core`? If so, `Content` type would have to be in `serde_core` as well, which isn't ideal due to [reasons outlined by dtolnay](https://github.com/serde-rs/serde/pull/2608#pullrequestreview-1899765113).  I've tried numerous angles to remove `__deserialize_content` from `Deserialize` (and somehow move it's implementation into `serde`), without luck.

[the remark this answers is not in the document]
oli-obk: I guess we'll have to revert https://github.com/serde-rs/serde/pull/2148 for now and reopen the issue it closed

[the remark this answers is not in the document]
osiewicz: Alrighty, `private` has been moved into `serde`.

oli-obk:
> When both serde and serde_core are used, the one that's imported first is the one that's suggested in diagnostics.

That's fine. When people are actually directly depending on serde_core, it's fine to suggest importing from it. Almost no one will be directly depending on serde_core. The fact that the import suggestions and other diagnostics always refer to `serde` is a good sign.



> Using a type in a `#[derive(Deserialize)]` struct's field that doesn't implement `Deserialize`. Notice a lack of `serde::`. I even removed `use serde::Deserialize` to double check that wasn't the reason why.

There are possibly other diagnostics that use the full path instead of just the reduced path, especially on older compiler versions. We've since moved diagnostics to refer to imported paths, but there may be some cases left where we report the full path (and that may then refer to serde_core). That's rare and in almost all cases a compiler bug, so we can ignore that.

oli-obk:
also next time you merge, maybe rebase and squash instead?

osiewicz:
Outside of `Result`, I find impls for `Mutex` pretty suspicious; at the very least, I've never really found myself needing that + it's pretty subtle that the lock is essentially hidden away from you in a library code. That's just my 2 cents though.

osiewicz:
> Besides the gains here, this allows serde_json and friends to depend on serde_core, preventing packages using serde w/ derive from serializing things.

I wonder, is there a clear migration path for crates such as serde_json to start using `serde_core` in place of `serde`? Say the user depends on `serde` explicitly (using `1.0` as version constraint) - unless they decide to `cargo update` their lockfile, will current resolver prefer to have just one copy of serde that is `serde_core`-compliant, or is it possible that an user will end up with both jumbo serde (as a direct dependency of their code) and `serde_core` as a dependency of serde_json et al?

epage:
> I wonder, is there a clear migration path for crates such as serde_json to start using serde_core in place of serde? Say the user depends on serde explicitly (using 1.0 as version constraint) - unless they decide to cargo update their lockfile, will current resolver prefer to have just one copy of serde that is serde_core-compliant, or is it possible that an user will end up with both jumbo serde (as a direct dependency of their code) and serde_core as a dependency of serde_json et al?

They could have a "never matching" `cfg` dependency on a new-enough serde that uses `serde_core`.

Mingun:
So what status of this? It is finished or not? If I understand correctly, the only thing remains is to add bunch of features, one for every manual `Serialize` / `Deserialize` implementation that will be in `serde_core`?

osiewicz:
I don't have anything to add from my side (sans https://github.com/serde-rs/serde/pull/2608#issuecomment-2204657621), so I'd expect it to be in a merge-able state.

serde/src/private/de.rs
```diff
@@ -298,7 +302,7 @@ mod content {
             // Untagged and internally tagged enums are only supported in
             // self-describing formats.
             let visitor = ContentVisitor { value: PhantomData };
-            deserializer.__deserialize_content(actually_private::T, visitor)
+            deserializer.deserialize_any(visitor)
```
Mingun, line 305: Another way to deal with that is to remove `actually_private::T` from the method signature which is needed only to prevent calling this method from the other crates. The method with the similar intention, `Visitor::__private_visit_untagged_option` does not have such seal and survived separation without problems.

serde/src/lib.rs
```diff
@@ -263,28 +233,14 @@ mod lib {
     };
     #[cfg(all(feature = "std", no_target_has_atomic, not(no_std_atomic64)))]
```
oli-obk, line 234: Why are these imports still here, but not the ones with some of the cfgs disabled?

serde/src/lib.rs
```diff
@@ -263,28 +233,14 @@ mod lib {
...
-    #[cfg(not(no_core_num_saturating))]
-    pub use self::core::num::Saturating;
 }
 
 // None of this crate's error handling needs the `From::from` error conversion
 // performed implicitly by the `?` operator or the standard library's `try!`
 // macro. This simplified macro gives a 5.5% improvement in compile time
 // compared to standard `try!`, and 9% improvement compared to `?`.
+#[doc(hidden)]
+#[macro_export]
```
oli-obk, line 243: Why is this exported?

serde/src/private/de.rs
oli-obk: This sounds like a change we could do on master and consider in isolation from this PR

serde/src/private/de.rs
```diff
@@ -1048,7 +1052,7 @@ mod content {
         E: de::Error,
     {
         #[cold]
-        fn invalid_type(self, exp: &Expected) -> E {
+        fn invalid_type(self, exp: &dyn Expected) -> E {
```
oli-obk, line 1055: Does this cause msrv issues?

oli-obk (changes requested):
Please test that everything still works with msrv and a few key versions

serde/src/private/de.rs
osiewicz: I don't think so, as dyn traits were stabilized in 1.27 (vs. the MSRV of 1.31 that serde currently has)

oli-obk: I've recently added `dyn` everywhere, so this is resolved indeed

serde/src/lib.rs
osiewicz: Prolly a rebase artifact, akin to https://github.com/serde-rs/serde/pull/2608#discussion_r1809420075. I've redone the whole branch from scratch.

serde_core/crates-io.md
oli-obk, line 1: I think we should write docs explaining what serde_core is here instead of duplicating serde's docs. Almost everyone will still want to go to serde itself. `serde_core` is for derive authors and format authors.

serde_core/src/lib.rs
```diff
@@ -0,0 +1,318 @@
+//! # Serde
```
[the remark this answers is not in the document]
oli-obk: I unresolved this again. The main serde_core rustdocs need something specific to serde_core

oli-obk (changes requested):
needs another rebase now that the `actually_private::T` is gone

modulo documentation this now looks good to me

serde_core/LICENSE-APACHE
oli-obk, line 1: this needs to become a symlink again (just like `LICENSE-MIT`)

oli-obk:
> Outside of `Result`, I find impls for `Mutex` pretty suspicious; at the very least, I've never really found myself needing that + it's pretty subtle that the lock is essentially hidden away from you in a library code. That's just my 2 cents though.

the change of having `Result` behind a feature gate got lost in the redo.

osiewicz:
One extra "nit": `build.rs` for `serde` right now is just about detecting whether we're on a Rust version that is supported by serde_derive. If so, we enable extra items (regardless of the state of `derive` feature on serde itself).
We could get rid of this build script if we gated these items behind `derive` feature instead; that'd make it so that `serde_derive` could not be used with `serde` without `derive` feature enabled, which is a breaking change.

serde_core/crates-io.md
oli-obk, line 1: the crates-io text also needs to be changed to point ppl to the main serde crate

serde/src/private/de.rs
```diff
@@ -1496,14 +1496,6 @@ mod content {
...
             visitor.visit_unit()
         }
-
-        fn __deserialize_content<V>(self, visitor: V) -> Result<V::Value, Self::Error>
-        where
-            V: Visitor<'de, Value = Content<'de>>,
-        {
-            let _ = visitor;
-            Ok(self.content)
-        }
```
Mingun, line 1506: This is not needed anymore because you can use it in another crate and keep this optimization

osiewicz: I'm a bit unclear on that tbh; won't we have to keep `Content` in `serde_core` to make this work?

martin-g:
> that'd make it so that `serde_derive` could not be used with `serde` without `derive` feature enabled, which is a breaking change.

Using `serde_derive` directly has the benefit of faster compile times - https://github.com/serde-rs/serde/issues/2907#issuecomment-2936782977

osiewicz:
That benefit would largely be superseded by this PR though, because the format crates (`serde_json` etc) can just depend on `serde_core` directly. They will not have `serde_derive` in their build graph, even if it is enabled by a feature on `serde`.

serde/src/private/de.rs
osiewicz: @dtolnay brought up https://github.com/serde-rs/serde/pull/2912 when I last spoke to him, but I guess this still requires `Content` to live in `serde_core`.
I may be clinging onto `Content` removal for no good reason, though to me it seems like a private implementation detail of derive machinery, which I guess we don't want to live in core? (per: https://github.com/serde-rs/serde/pull/2608#pullrequestreview-1899765113)

oli-obk (approved):

PR #2626, opened 2023-10-10
Adds Serialize with AsRef or ToString and Deserialize with FromStr

wyatt-herkamp:
Adds

- Serialize `#[serde(as_ref = "Type")` Calls AsRef<Type> and serializes the value
- Serialize `#[serde(to_string)` Calls ToString and serializes the String
- Deserialize `#[serde(from_str)` Calls FromStr for deserialize.


dtolnay (commented):
Thanks for the PR!

The `from_str` and `to_string` part of this are a duplicate of #2017. I would prefer not to build that into this crate but they are available in `serde_with`. The implementation there is more powerful than the one in this PR because deserialization works for data formats that do not expose borrowed data, such as deserializing from a std::io::Read.

For `as_ref`, I would similarly prefer not to build that into this crate. Like `to_string`, this is substituting for a pretty trivial Serialize impl that one can go ahead and handwrite without needing macro support; I don't see the use case occurring commonly enough to warrant dedicating an attribute (and compile time) to.

PR #2632, opened 2023-10-18
SERDE-1 - Warning on unused attributes

### LINKED ISSUE #2491: Warning on unused attributes
When using container attributes `into` all fields attributes `serialize_with` (and most probably others) or effectively ignored. 

Would be nice to provide a warning or an error in those cases. If issue is validated I can work on a PR 😊

See example and its [link to the playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=cafc324548686b35cc2142e629cfdfc0):

```
use serde::Serialize; // 1.0.164

#[derive(Serialize, Clone)]
#[serde(into = "MyThroughStruct")]
pub struct MySerializableStruct {
    #[serde(serialize_with = "useless_fn")]
    field_1: usize
}

#[derive(Serialize)]
pub struct MyThroughStruct {
    field_1: Option<usize>
}

pub fn useless_fn<S>(_val: &MySerializableStruct, _ser: S) -> Result<S::Ok, S::Error>
where
    S: serde::Serializer
{
    panic!("This is never called and most probably an error")
}

impl From<MySerializableStruct> for MyThroughStruct {
    fn from(val: MySerializableStruct) -> Self {
        Self {
            field_1: Some(val.field_1)
        }
    }
}
```

gitstart:
Closes #2491 

## Fix: Added a compile time warning.
A compile-time warning is displayed when the `into` container attribute is used with other field attributes. [Link to a test program](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b0b05ac54f3065e553d425a3277ae405).

In order to emit a compile-time warning, the `#![feature(proc_macro_diagnostic)]` was added to the `lib.rs` files. This required the use of `proc-macro2` nightly features.
```xml
[dependencies]
# nightly version supports #![feature(proc_macro_diagnostic)]
proc-macro2 = { version = "1.0", features = ["nightly"] }
```

Sample of warning issued:
```rust
warning: #[serde(into)] container attribute will invalidate all field attributes
 --> demo\src\main.rs:8:5
  |
8 |     field_1: usize,
  |     ^^^^^^^

warning: `demo` (bin "demo") generated 1 warning
```

___

Do we need test for this?

### DIFF
--- a/serde_derive_internals/Cargo.toml
+++ b/serde_derive_internals/Cargo.toml
@@ -15,7 +15,8 @@ rust-version = "1.56"
 path = "lib.rs"
 
 [dependencies]
-proc-macro2 = "1.0"
+# nightly version supports #![feature(proc_macro_diagnostic)]
+proc-macro2 = { version = "1.0", features = ["nightly"] }
 quote = "1.0"
 syn = { version = "2.0.28", default-features = false, features = ["clone-impls", "derive", "parsing", "printing"] }
 
--- a/serde_derive_internals/Cargo.toml
+++ b/serde_derive_internals/Cargo.toml
@@ -15,8 +15,7 @@ rust-version = "1.56"
 path = "lib.rs"
 
 [dependencies]
-# nightly version supports #![feature(proc_macro_diagnostic)]
-proc-macro2 = { version = "1.0", features = ["nightly"] }
+proc-macro2 = "1.0"
 quote = "1.0"
 syn = { version = "2.0.28", default-features = false, features = ["clone-impls", "derive", "parsing", "printing"] }
 
--- a/serde_derive/Cargo.toml
+++ b/serde_derive/Cargo.toml
@@ -15,6 +15,8 @@ rust-version = "1.56"
 [features]
 default = []
 deserialize_in_place = []
+# Provide impls for types that require unstable functionality e.g compile warnings
+unstable = []
 
 [lib]
 name = "serde_derive"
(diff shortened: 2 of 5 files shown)


oli-obk:
> https://developers.gitstart.com/client/stockly/ticket/SERDE-1

I cannot view this url without an account.

Please track issues in our issue tracking instead of in another webpage.

serde_derive_internals/Cargo.toml
```diff
@@ -15,7 +15,8 @@ rust-version = "1.56"
 path = "lib.rs"
 
 [dependencies]
-proc-macro2 = "1.0"
+# nightly version supports #![feature(proc_macro_diagnostic)]
+proc-macro2 = { version = "1.0", features = ["nightly"] }
```
oli-obk, line 19: serde cannot use nightly. This is not a change we can possibly accept

kodejuice: > serde cannot use nightly. This is not a change we can possibly accept

Can this warning be an  unstable only feature?  The only tools we found that can warn at compile time require Nightly Rust. The codebase already has the unstable feature flag that enables features that only compile on Nightly Rust. We've added this fix as an unstable feature

gitstart: > serde cannot use nightly. This is not a change we can possibly accept

Can this warning be an  unstable only feature?  The only tools we found that can warn at compile time require Nightly Rust. The codebase already has the unstable feature flag that enables features that only compile on Nightly Rust. We've added this fix as an unstable feature

PR #2637, opened 2023-10-26
Ensure f32 deserialized from f64 and vice versa preserve NaN sign

dtolnay:
`f32 as f64` and `f64 as f32`, if the input is NaN, produces an output NaN with nondeterministic sign. This leads to bugs like TOML `+nan` and `-nan` and YAML `.nan` and `-.nan` being deserialized to f32 with the wrong sign nondeterministically because they're manipulated as f64 internally by the library's Value type.

See https://github.com/rust-lang/miri/issues/3139.

### DIFF
--- a/serde/src/de/impls.rs
+++ b/serde/src/de/impls.rs
@@ -180,6 +180,28 @@ macro_rules! num_as_self {
     };
 }
 
+macro_rules! num_as_copysign_self {
+    ($ty:ident : $visit:ident) => {
+        #[inline]
+        fn $visit<E>(self, v: $ty) -> Result<Self::Value, E>
+        where
+            E: Error,
+        {
+            #[cfg(no_float_copysign)]
+            {
+                Ok(v as Self::Value)
+            }
+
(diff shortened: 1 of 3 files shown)


serde/src/de/impls.rs
```diff
@@ -180,6 +180,28 @@ macro_rules! num_as_self {
...
+    ($ty:ident : $visit:ident) => {
+        #[inline]
+        fn $visit<E>(self, v: $ty) -> Result<Self::Value, E>
+        where
+            E: Error,
+        {
+            #[cfg(any(no_float_copysign, not(feature = "std")))]
+            {
+                Ok(v as Self::Value)
+            }
```
dtolnay, line 193: This produces nondeterministic wrong values if serde is built with "std" feature disabled. https://github.com/rust-lang/rust/issues/50145

I'm not sure whether we can manually implement a correct copysign without going through the LLVM copysign intrinsics. If that were possible, I presume libcore would already have done it?

PR #2639, opened 2023-10-28
Draft: implement attribute to disallow deserialization of struct with named fields from sequence

### LINKED ISSUE #1587: Struct can be deserialized from sequence
When I derive `Deserialize` for a struct containing named fields, I expect that in any format that has a way of expressing named fields (i.e. pretty much every ordinary format except Bincode), it will always be expressed as a map containing named fields. To my surprise, it can also be deserialized from a sequence of field values without field names, even in formats where this is not correct, such as JSON and YAML.

Accepting invalid data is a security risk as well as risk of causing people to rely on an unintended implementation details. This also risks having different implementations that differ in what data they accept, which is also a security risk. Further it bloats programs with visitor methods for parsing invalid data, which would otherwise automatically be removed by the compiler.

In this example, I'd like deserialization to fail, but it incorrectly parses invalid data into a structure. ([Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=3d9882b7cc1321e413064f49050495d2))
```rust

#[derive(Debug, Deserialize)]
struct Person {
    first_name: String,
    last_name: String,
}

fn main() {
    eprintln!("{:#?}", serde_json::from_str::<Person>(r#"["John","Doe"]"#));
}
```
Maybe this is due to implementations of various formats not taking proper care in what data they accept after `deserialize_struct` is called, or maybe it is because of Serde itself having an ambiguous purpose of the `visit_seq` method, and there needs to be a different visitor method that is used only for formats that don't use field names (e.g. Bincode), with a default implementation that delegates to `visit_seq`. Deriving `Deserialize` for structs would then implement this method, but not the `visit_seq` method which also applies to JSON sequences.

Toromyx:
Fixes #1587

- [ ] Review code
- [ ] Decide on attribute name

### DIFF
--- a/serde_derive/src/de.rs
+++ b/serde_derive/src/de.rs
@@ -961,10 +961,10 @@ fn deserialize_struct(
     let field_visitor = deserialize_field_identifier(&field_names_idents, cattrs);
 
     // untagged struct variants do not get a visit_seq method. The same applies to
-    // structs that only have a map representation.
+    // structs that only have a map representation or are deserialized strict_or_some_other_name_ly.
     let visit_seq = match form {
         StructForm::Untagged(..) => None,
-        _ if cattrs.has_flatten() => None,
+        _ if cattrs.has_flatten() || cattrs.is_strict_or_some_other_name() => None,
         _ => {
             let mut_seq = if field_names_idents.is_empty() {
                 quote!(_)
@@ -1090,8 +1090,8 @@ fn deserialize_struct_in_place(
     cattrs: &attr::Container,
 ) -> Option<Fragment> {
     // for now we do not support in_place deserialization for structs that
-    // are represented as map.
-    if cattrs.has_flatten() {
+    // are represented as map or are deserialized strict_or_some_other_name_ly.
+    if cattrs.has_flatten() || cattrs.is_strict_or_some_other_name() {
         return None;
     }
 
@@ -2001,6 +2001,7 @@ fn deserialize_generated_identifier(
         None,
         !is_variant && cattrs.has_flatten(),
         None,
+        cattrs.is_strict_or_some_other_name(),
     ));
 
     let lifetime = if !is_variant && cattrs.has_flatten() {
@@ -2155,6 +2156,7 @@ fn deserialize_custom_identifier(
         fallthrough_borrowed,
         false,
         cattrs.expecting(),
+        false,
     ));
 
     quote_block! {
@@ -2188,6 +2190,7 @@ fn deserialize_identifier(
     fallthrough_borrowed: Option<TokenStream>,
     collect_other_fields: bool,
     expecting: Option<&str>,
+    is_strict_or_some_other_name: bool,
 ) -> Fragment {
     let str_mapping = fields.iter().map(|(_, ident, aliases)| {
         // `aliases` also contains a main name
@@ -2255,7 +2258,7 @@ fn deserialize_identifier(
     };
 
     let visit_other = if collect_other_fields {
-        quote! {
+        Some(quote! {
             fn visit_bool<__E>(self, __value: bool) -> _serde::__private::Result<Self::Value, __E>
             where
                 __E: _serde::de::Error,
@@ -2346,8 +2349,8 @@ fn deserialize_identifier(
             {
                 _serde::__private::Ok(__Field::__other(_serde::__private::de::Content::Unit))
             }
-        }
-    } else {
+        })
+    } else if !is_strict_or_some_other_name {
         let u64_mapping = fields.iter().enumerate().map(|(i, (_, ident, _))| {
             let i = i as u64;
             quote!(#i => _serde::__private::Ok(#this_value::#ident))
@@ -2368,7 +2371,7 @@ fn deserialize_identifier(
             &u64_fallthrough_arm_tokens
         };
 
-        quote! {
+        Some(quote! {
             fn visit_u64<__E>(self, __value: u64) -> _serde::__private::Result<Self::Value, __E>
             where
                 __E: _serde::de::Error,
@@ -2378,7 +2381,9 @@ fn deserialize_identifier(
                     _ => #u64_fallthrough_arm,
                 }
             }
-        }
+        })
+    } else {
+        None
     };
 
     let visit_borrowed = if fallthrough_borrowed.is_some() || collect_other_fields {
--- a/serde_derive/src/internals/attr.rs
+++ b/serde_derive/src/internals/attr.rs
@@ -210,6 +210,7 @@ pub struct Container {
     rename_all_fields_rules: RenameAllRules,
     ser_bound: Option<Vec<syn::WherePredicate>>,
     de_bound: Option<Vec<syn::WherePredicate>>,
+    is_strict_or_some_other_name: bool,
     tag: TagType,
     type_from: Option<syn::Type>,
     type_try_from: Option<syn::Type>,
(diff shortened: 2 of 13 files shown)


Toromyx:
Documentation probably needs to be added to <https://github.com/serde-rs/serde-rs.github.io>, anywhere else?

Mingun:
I would think, that we should deny deserialization of maps from sequences (at least by default, but I prefer at all). It is responsibility of a deserializer to call the correct `Visitor` method to deserialize the type. If `deserialize_struct` / `deserialize_map` was requested, and the data type stores only field values in form of sequence, it does not mean that deserializer should call `visit_seq`. `visit_seq` is intended to provide _sequence_ and deserializer should not expect that map types could be constructed from it. Instead, the deserializer should provide a compatible `MapAccess` implementation, that could return correct keys from existing data.

The important point that should be explicitly stressed is that this `MapAccess` implementation should not try to use `fields` data to decide what key you should return. You can think that it would be a great idea to return `fields[0]`, `fields[1]` and so on for keys, that that's not. It is worst idea, because of two problems:
- serde conversion mechanisms can change calls of `deserialize_struct` to calls of `deserialize_map` (in case of flatten structs);
- the `fields` argument of a `deserialize_struct` has no guaranties that it contains fields of a struct in order of declaration, and moreover, (in the last serde versions?) it also contains field aliases (i.e. its length even does not reflect the number of fields in a struct).

Fortunately, map keys deserialized using internally created special `enum`s each variant of which represents a field and they can be deserialized from numbers 0, 1, 2 in the order of declaration. Aliases exists only for symbolic names, so each number uniquely represent a field. Moreover, this enum is deserialized by requesting `deserialize_identifier`, so the deserializer implementation could (and probably must) use different rules from usual to provide data to `Visitor::visit_u64` method.

So we need several fixes in various serde crates:
- `serde_derive` should stop to generate `visit_seq` for `struct`s
- various crates for binary formats should stop to call `Visitor::visit_seq` when `deserialize_struct` / `deserialize_map` was requested. Instead they should call `Visitor::visit_map` and deserialize keys using special internal deserializer for keys 
- `serde_json` and `serde_yaml` probably shouldn't do anything

In the summary:
1. it is responsibility of a `Visitor` to return errors. `Deserializer`s should avoid to return they own errors, they should instead call most appropriate `Visitor` method, which would return an error
2. it is responsibility of a `Deserializer` to perform type conversions. For example, in XML is no difference in syntax between numbers or strings, so the XML deserializer will perform conversion when number was requested. That means, that result of `deserialize_any` can be different from result of `deserialize_u32` and that is normal. JSON, for example, has a different syntax for strings and numbers and it MUST NOT parse strings into numbers
3. it is responsibility of a type to implement only those of `Visitor` methods that describes it much closely. They should implement the minimal subset to be as much specific as possible. It shouldn't to do conversions from one type to another (such as sequences to maps or strings to numbers). A special types _could_ do such conversions, but that should be done purposely, and not by an automatic derive from `serde_derive`

Mingun:
Some observations for crates from [main](https://serde.rs/#data-formats) serde page:
|Status|Format|Links|
|------|------|-----|
|:warning:|JSON|Incorrectly [uses](https://github.com/serde-rs/json/blob/f56053d6f8cb45f61a71570a22d3272bfeb89909/src/de.rs#L1824-L1835) `visit_seq` when `deserialize_struct` was requested
|:warning:|postcard|Incorrectly [uses](https://github.com/jamesmunns/postcard/blob/a095b49935f7bd1bab04e6c5914ab11c3bbd5fee/src/de/deserializer.rs#L500C18-L510) `visit_seq` when `deserialize_struct` was requested
|:white_check_mark:|CBOR|[uses](https://github.com/enarx/ciborium/blob/2ac91ce8a398bcdbaa250a92831786ffe42052a2/ciborium/src/de/mod.rs#L443-L465) `visit_map`
|:white_check_mark:|YAML|[uses](https://github.com/dtolnay/serde-yaml/blob/66ec2cf7554603051caaf22229d9d227286833b0/src/de.rs#L1660-L1700) `visit_map`
|:white_check_mark:|Msgpack|[uses](https://github.com/3Hren/msgpack-rust/blob/69f90c636705966d8b6855dae6f7cc49afe273ba/rmp-serde/src/decode.rs#L542-L560) `visit_map` (`deserialize_map` / `deserialize_struct` forwarded into `deserialize_any`)
|:white_check_mark:|TOML|[uses](https://github.com/toml-rs/toml/blob/6c8885e8d83047c1bd286e11c28e203b06fcb912/crates/toml_edit/src/de/table.rs#L15-L20) `visit_map` (`deserialize_map` / `deserialize_struct` forwarded into `deserialize_any`)
|:warning:|Pickle|[can use](https://github.com/birkenfeld/serde-pickle/blob/master/src/de.rs#L1108-L1138) `visit_seq` (`deserialize_map` / `deserialize_struct` forwarded into `deserialize_any`)
|:white_check_mark:|RON|[uses](https://github.com/ron-rs/ron/blob/e27d1d5ecd30c8c15cf11b36c5cfbf402dbca0bb/src/de/mod.rs#L667-L727) `visit_map`
|:warning:|BSON|[can use](https://github.com/mongodb/bson-rust/blob/main/src/de/serde.rs#L625-L677) `visit_seq` (`deserialize_map` / `deserialize_struct` forwarded into `deserialize_any`)
|:white_check_mark:|Avro|[uses](https://github.com/apache/avro/blob/807314512736c2a1c0a4203f4eb9b5d84aa021e4/lang/rust/avro/src/de.rs#L490-L513) `visit_map`
|:warning:|JSON5|[can use](https://github.com/callum-oakley/json5-rs/blob/5066f2e0a2726ceaeeb0a7786804787b3dc801f1/src/de.rs#L46-L69) `visit_seq` (`deserialize_map` / `deserialize_struct` forwarded into `deserialize_any`)
|:white_check_mark:|URL|[uses](https://github.com/samscott89/serde_qs/blob/8a13b677f1b99ab78870f8731e75a7642f2d0530/src/de/mod.rs#L235-L252) `visit_map`
|:white_check_mark:|Envy|[uses](https://github.com/softprops/envy/blob/e22035c3b91f999c325a44c0c77f0e6e05750da3/src/lib.rs#L266-L284) `visit_map` (via `serde`'s `MapDeserializer`)
|:white_check_mark:|Envy Store|delegates deserialization to Envy
|:white_check_mark:|S-expressions|[uses](https://github.com/rotty/lexpr-rs/blob/28b0e078f2c70e4ee92eb6f21fa1ceae5eb7c2e7/serde-lexpr/src/value/de.rs#L217-L238) `visit_map`
|:warning:|DBUS|[can](https://github.com/dbus2/zbus/blob/1856c0f348eff0d25fac0921c963cb05b3811c59/zvariant/src/gvariant/de.rs#L148-L149) [use](https://github.com/dbus2/zbus/blob/1856c0f348eff0d25fac0921c963cb05b3811c59/zvariant/src/dbus/de.rs#L156-L157) `visit_seq`
|:warning:|FlexBuffers|[can use](https://github.com/google/flatbuffers/blob/205285c35c164c0368c913b3f611cf8d672bc103/rust/flexbuffers/src/reader/de.rs#L169-L201) `visit_seq`
|:white_check_mark:|Bencode|[uses](https://github.com/P3KI/bendy/blob/b4286e28c2021a80937c7dc5116003295cc093f6/src/serde/de.rs#L348-L371) `visit_map`
|:white_check_mark:|Token streams|[uses](https://github.com/oxidecomputer/serde_tokenstream/blob/baf22ad1af3bf60de474fd018ed2cd458c22e12a/src/serde_tokenstream.rs#L563-L610) `visit_map`
|:white_check_mark:|DynamoDB Items|[uses](https://github.com/zenlist/serde_dynamo/blob/035206616964ea0e072af1d6cc6a7e6220056a16/src/de/deserializer.rs#L182-L192) `visit_map` (`deserialize_struct` forwarded into `deserialize_map`)
|:white_check_mark:|Hjson|[uses](https://github.com/Canop/deser-hjson/blob/e25cd0759565e5aab315fea4620c2d71f9d51e29/src/de.rs#L895-L938) `visit_map`
|:warning:|CSV|[can use](https://github.com/BurntSushi/rust-csv/blob/533d37b7c95490506c374330221d098c876834df/src/deserializer.rs#L507-L529) `visit_seq`

Toromyx:
> * `serde_derive` should stop to generate `visit_seq` for `struct`s

I created a new branch and PR which should implement this: #2640

But since this would be a breaking change, an attribute as solution for version 1 still makes sense in my opinion. One could maybe display a deprecation warning when `Deserialize` is derived without that attribute?

Issue #2629, opened 2023-10-15
Implement Serialize / Deserialize for `core::task::Poll`


LegNeato:
Would it make sense to implement `Serialize` / `Deserialize` for `core::task::Poll`? If so I can probably put up a PR.

`Poll::Pending` is a `unit_variant` and `Poll::Ready<T>` is a `newtype_variant` of T.

PR #2641, opened 2023-10-31
Add support for core::task::Poll

LegNeato:
Fixes https://github.com/serde-rs/serde/issues/2629

### DIFF
--- a/serde/build.rs
+++ b/serde/build.rs
@@ -33,6 +33,12 @@ fn main() {
         println!("cargo:rustc-cfg=no_float_copysign");
     }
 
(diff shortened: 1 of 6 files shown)


LegNeato:
Hmmm, `Poll` was unstable in the MSRV. Shall I put this behind a feature flag? If so, on or off by default?

dtolnay (commented):
Thank you for the PR, but I would prefer not to add these impls. This seems similar to the existing impls on `Result` which are widely considered a mistake to include (#1070). I think `serde(with = "...")`, or converting the `Poll` into a different type like `Option` for serialization, should be adequate in the few places that encounter this use case.

Issue #2643, opened 2023-11-04
Allow internal tag field in untagged variant


robsdedude:
In https://github.com/serde-rs/serde/issues/1161, it was pointed out that serde should check at compile time, whether a tagged variant contains a field that conflicts with the tag. 

I think this should not apply to the untagged variant. Example:

```rust
use serde_derive; // 1.0.190
use serde; // 1.0.190

#[derive(serde::Deserialize)]
#[serde(tag = "kind")]
enum Foo {
    Bar {
        /// cannot have the following here, makes sense
        // kind: String
    },
    #[serde(untagged)]
    Baz {
        /// error: variant field name `kind` conflicts with internal tag
        /// should be allowed, but isn't
        kind: String  
    }
}
```

The only downside I can see is that for serialization, one could create a `Foo::Baz { kind: String::from("Bar") }` which might not follow the shape of `Foo::Bar`. That's unfortunate, but I think the loss of convenience by not allowing this (especially for deserialization) outweighs this.

PR #2646, opened 2023-11-06
Allow internal tag field in untagged variant

robsdedude:
Fixes https://github.com/serde-rs/serde/issues/2643


dtolnay (approved):
Thanks!

PR #2653, opened 2023-11-21
Fix Range{From,To} deserialize mixup

### LINKED ISSUE #2652: Deserialization for RangeFrom and RangeTo doesn't work
It seems that the deserialization for RangeFrom and RangeTo has been switched around such that RangeTo expects a "start" and RangeFrom expects a "end". They are correct in the serialization though. 

### See the following test:

```rust
#[cfg(test)]
mod tests {
    use std::ops::{Range, RangeFrom, RangeTo};

    use serde_json;

    #[test]
    fn deserialize_range_to() {
        let json = r#"{"end": 173}"#;
        let result: Result<RangeTo<u32>, serde_json::Error> = serde_json::from_str(json);
        assert!(result.is_ok(), "Deserialization failed: {:?}", result); // Fails
    }

    #[test]
    fn deserialize_range_from() {
        let json = r#"{"start": 173}"#;
        let result: Result<RangeFrom<u32>, serde_json::Error> = serde_json::from_str(json);
        assert!(result.is_ok(), "Deserialization failed: {:?}", result); // Fails
    }

    #[test]
    fn deserialize_range() {
        let json = r#"{"start": 173, "end": 176}"#;
        let result: Result<Range<u32>, serde_json::Error> = serde_json::from_str(json);
        assert!(result.is_ok(), "Deserialization failed: {:?}", result); // Passes
    }
}
```

It seems to have been introduced in #2471.

emilbonnek:
closes #2652

**Issue TL;DR**
RangeFrom and RangeTo implementations seems to have been switched around and RangeFrom was expecting a 'end' field when it should be expecting a 'start' field.

### DIFF
--- a/serde/src/de/impls.rs
+++ b/serde/src/de/impls.rs
@@ -2509,7 +2509,7 @@ where
     where
         D: Deserializer<'de>,
     {
-        let start = tri!(deserializer.deserialize_struct(
+        let end = tri!(deserializer.deserialize_struct(
             "RangeFrom",
             range_from::FIELDS,
             range_from::RangeFromVisitor {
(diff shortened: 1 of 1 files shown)


dtolnay (approved):
Thank you! Good catch.

Issue #1724, opened 2020-01-20
Title  Case


gilescope:
Am I the only one that has a case for "Human Case" where I could do with capitals and spaces in the RenameRule?

I'd suggest we close this and only implement it if it gets some thumbs up from others that also could do with this.

gilescope:
Closing unless anyone else needs it...

gilescope:
https://crates.io/crates/heck calls this "Title Case"

treble37:
actually @gilescope - I could use this as well, I was dealing with some JSON that had Title Case / Human Case...any chance to reopen?

gilescope:
I’ve seen a few rust case libraries now and everyone definitely calls this “Title Case”.

dtolnay:
For such less widespread case conventions, I would suggest building this as an attribute macro which inserts appropriate `serde(rename = "...")` attributes onto every field/variant in its input.

dzmitry-lahoda:
me too!

dzmitry-lahoda:
it is very use full with yaml!!! so I can avoid scaring people with snakes and pascals and camels. and using same JSON stuff to render human yaml.

PR #2657, opened 2023-11-24
Add "space case" & "SCREAMING SPACE CASE" rename rules

xenoliss:
Add a two new `RenameRule` variants to handle "space case" and "SCREAMING SPACE CASE".


dtolnay (commented):
Thanks for the PR!

I think these are both in the same category as https://github.com/serde-rs/serde/issues/1724; I would prefer not to build this into serde_derive.

PR #2673, opened 2024-01-09
Use shields.io's MSRV badges

Sky9x:
Added in https://github.com/badges/shields/pull/9871

Feel free to change the formatting/color/etc (default color is blue)


dtolnay (approved):
Neat! Thank you for working on this!

PR #2679, opened 2024-01-26
Format invalid float message to show at least one decimal place

Dav1dde:
It can be confusing at first when deserializing (e.g. JSON) and getting an error like:

> invalid type: floating point `3`, expected u32

It shows a valid integer but it is actually a float.

Changes the error to always show at least one decimal place to avoid confusion.


Dav1dde:
Nevermind sorry for the noise, didn't realize `format` does not support formatting with a minimum precision, only maximum.

dtolnay (commented):
Thanks anyway!

I implemented a different fix in https://github.com/serde-rs/json/pull/1107 + https://github.com/serde-rs/serde/pull/2682.

Issue #1626, opened 2019-09-09
#[serde(flatten)] doesn't work with #[serde(default)] for externally tagged enums


slckl:
Given this:
```Rust
#[derive(Deserialize, Debug)]
enum SomeEnum {
    A(u32),
    B(u32),
}

impl Default for SomeEnum {
    fn default() -> Self {
        SomeEnum::B(1)
    }
}

#[derive(Deserialize, Debug)]
struct SomeStruct {
    filler: u32,
    #[serde(default)]
    #[serde(flatten)]
    some_enum: SomeEnum,
}
```

trying to deserialize this yaml:
```Yaml
filler: 1
```
 into SomeStruct panics with a message:
```
no variant of enum SomeEnum found in flattened data
```
I hoped it would pick up the default value specified for SomeEnum instead...

[Playground example.](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=be54aeb46ee80c82bcaa8bced60b3ba7)

Is this a bug or is this not supported?
Docs don't cover this, I tried searching issues, but couldn't find anything quite like this.

dtolnay:
Thanks, I would accept a PR to make this work.

slckl:
Happy to give it a try, but I haven't dug too much into Serde's internals. Any pointers on where to start?

Ploppz:
I need the same behaviour, but with internal tagging. My example: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=79a52563e681076d5e8ee5fcbee7da81

heftig:
FTR, the issue isn't restricted to external tagging; other forms of tagging don't make it work, either.

DzenanJupic:
I also ran into this problem and just tried to solve it. Unfortunately, this is not just a little bugfix.

The bug is caused in `serde_derive::de::deserialize_map`. This function handles flattened fields completely different than 'normal' fields.

#### Normal fields
For normal fields, `deserialize_map` creates a variable `let mut __fieldN: Option<T> = None` where N is the field index (also for maps) and T is the Field type [`serde_derive::de.rs` lines 2381-2389].

[...]

PR #2687, opened 2024-02-02
Support flattened enums with default field attributes

andrewtoth:
This adds a custom error type for `FlatMapDeserializer`, which distinguishes between custom errors and the specific `no variant of enum {name} found in flattened data` error.

This error is then used in `deserialize_map` in `serde_derive` to `match` on the returned error, and if the field has a `default` attribute it uses the `expr_is_missing` to provide a default value.

Fixes https://github.com/serde-rs/serde/issues/1626


andrewtoth:
The tests fail due to `trybuild` using an older version of `serde` or `serde_derive`. This PR only modifies API in the `private` mod, and in `serde_derive`. However, if after this patch both `serde` and `serde_derive` are not upgraded in lockstep then this will break at compile time. I'm not completely sure then, if this can be considered a non-breaking change? If all packages using `serde` and `serde_derive` in a project upgrade in lockstep then there is no external breaking change. Is it a requirement that `serde_derive` and `serde` be backwards compatible with older versions of each other?

serde/src/private/de.rs
```diff
@@ -2587,17 +2587,43 @@ where
 }
 
 #[cfg(any(feature = "std", feature = "alloc"))]
-pub struct FlatMapDeserializer<'a, 'de: 'a, E>(
-    pub &'a mut Vec<Option<(Content<'de>, Content<'de>)>>,
-    pub PhantomData<E>,
-);
+#[cfg_attr(any(feature = "std", feature = "alloc"), derive(Debug))]
```
Mingun, line 2590: Conditional derive seems not to be necessary because the whole struct is under that condition.

serde/src/private/de.rs
```diff
@@ -2587,17 +2587,43 @@ where
...
+pub enum FlatMapDeserializerError {
+    Message(String),
+    NoVariantFoundInFlattenedData(&'static str),
+}
```
Mingun, line 2594: Isn't that possible to keep original error instead of stringify it?
```suggestion
pub enum FlatMapDeserializerError<E> {
    Inner(E),
    NoVariantFoundInFlattenedData(&'static str),
}
```

Mingun (commented):
Interesting approach! I think, you need to add tests with cases from mentioned issues or even better generalize it and cover all possible variants where `FlatMapDeserializer` is used

andrewtoth:
@Mingun thank you for your review and suggestions! I implemented them both.
As for tests, I will happily write them but first could you confirm that the approach here will be accepted?

Mingun:
I don't know, I am the same external contributor as you. Judging by the general reaction of maintainers, one should hardly expect any attention in the near future

johnmave126:
It is interesting that I went on to try to solve another problem revolving flatten in #2751 and we actually ended up on very similar approach.

Issue #2695, opened 2024-02-12
Allow generic keys for serialization.


jun-sheaf:
Some data formats (such as CBOR and Protobuf) use types other than strings for keys. Currently, `serde` only allows struct serialization to contain string values for keys which prohibits this use case.

## Scope

We are only attempting to solve this problem for homogeneous keys.

## Proposed Solution

Other than adding an attribute (e.g. `key = 1`), the main concern comes from the signature of `SerializeStruct::serialize_field` which doesn't allow generics. Deserialization is trivial as it already uses `MapAccess` for deserializing which takes generics.

The solution we propose is adding the following methods to the `SerializeMap` trait:

```rs
fn serialize_key<T>(&mut self, key: &T) -> Result<(), Self::Error> 
  where T: Serialize + ?Sized;
fn serialize_value<T>(&mut self, value: &T) -> Result<(), Self::Error>
  where T: Serialize + ?Sized;
```

To make this non-breaking, the default implementation would just return `Ok(())`.

sosthene-nitrokey:
Why `#[serde(key = )]` and not just extending `#[serde(rename = …)]` like https://github.com/serde-rs/serde/pull/2209?

jun-sheaf:
> Why `#[serde(key = )]` and not just extending `#[serde(rename = …)]` like https://github.com/serde-rs/serde/pull/2209?

Good question! Perhaps because I didn't see that PR :D 

It's probably better to use a new key though because it's often the case that self-describing formats still want the string key over the binary key. This wasn't implemented in my implementation, but it was something I came across while production testing this feature.

sosthene-nitrokey:
One reason in favor of `rename` for integer keys is for enums, where even in `JSON` an integer or a boolean can work: https://github.com/serde-rs/serde/pull/2525

The advantage of using `rename` is that the interaction with `alias` is also pretty easy to understand (assuming that `alias` also supports non-string datatypes).

Anyway this is would be a great feature to have.

PR #2698, opened 2024-02-13
Implement `#[key = ...]`

jrandolf:
This PR implements the ability to use a limited set of literals for keys when serializing and deserializing structs. The motivation is written in https://github.com/serde-rs/serde/issues/2695.

Closes https://github.com/serde-rs/serde/issues/2695

## Limitations

 - We purposefully error for string keys since this is already supported using `rename`.
 - Floats are not supported (Rust doesn't support them for match statements and in practice, this is not done).
 - Negative keys are not supported for the time being.
 - Keys can clash if the user wants to. (Footgun?)

### DIFF
--- a/serde/src/ser/mod.rs
+++ b/serde/src/ser/mod.rs
@@ -1871,6 +1871,25 @@ pub trait SerializeStruct {
         Ok(())
     }
 
+    /// Serialize a struct field consisting of a typed key and value.
+    ///
+    /// This method is similar to `serialize_field`, but it allows the key to be
+    /// a type that is not a string. This is useful for formats that support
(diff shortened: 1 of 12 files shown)


serde/src/ser/mod.rs
```diff
@@ -1871,6 +1871,25 @@ pub trait SerializeStruct {
...
+        key: &K,
+        value: &V,
+    ) -> Result<(), Self::Error>
+    where
+        K: Serialize,
+        V: Serialize,
+    {
+        let _ = key;
+        let _ = value;
+        Ok(())
```
juntyr, line 1890: Shouldn't the default implementation error instead of silently doing nothing?

PR #2714, opened 2024-03-16
Document that `len` does not include skipped struct fields

Marcono1234:
At least this seems to be how serde_derive behaves:
https://github.com/serde-rs/serde/blob/3f43fca90dfff4a8728765867cea85865843c337/serde_derive/src/ser.rs#L335-L341

Explicitly documenting this makes it easier for custom `Serializer` implementations to know what to expect from the `len` value, and makes it clearer for custom `Serialize` implementations to know what to provide as value.


dtolnay (approved):
Thank you!

PR #2728, opened 2024-04-13
Expanded #[serder(default = ...)] capabilities.

Threadzless:
### Before
Giving a field a default value that isn't the same as that type's `Default` required users to define a function for each field.
```
pub struct MyStruct {
    #[serde(default = "my_struct_a_default")]
    a: i32,
    #[serde(default = "my_struct_b_default")]
    b: i32,
}

fn my_struct_a_default() -> i32 {
    12
}
fn my_struct_b_default() -> i32 {
    16
}
```
### Now
Constant expressions may also be used, so long as they are wrapped in parenthesis
```
pub struct MyStruct {
    #[serde(default = (12))]
    a: i32,
    #[serde(default = (16))]
    b: i32,
}
```
The requirement of wrapping expressions in parenthesis is to ensure string values are not mistaken for a path to a function. This ensures full backwards compatibility.

I have also modified some of the tests to verify this new capability, and incremented the version number

### DIFF
--- a/serde/Cargo.toml
+++ b/serde/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "serde"
-version = "1.0.197"
+version = "1.0.198"
 authors = ["Erick Tryzelaar <erick.tryzelaar@gmail.com>", "David Tolnay <dtolnay@gmail.com>"]
 build = "build.rs"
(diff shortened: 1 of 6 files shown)


Threadzless:
updated to match latest version

Issue #2190, opened 2022-03-17
Why "The 'de lifetime should not appear in the type to which the Deserialize impl applies"?


grelative:
Serde's documentation about [lifetime](https://serde.rs/lifetimes.html#the-deserializede-lifetime) mentioned this:
> The 'de lifetime should not appear in the type to which the Deserialize impl applies.

```
 // Do not do this. Sooner or later you will be sad.
 impl<'de> Deserialize<'de> for Q<'de> {

 // Do this instead.
 impl<'de: 'a, 'a> Deserialize<'de> for Q<'a> {
```
But didn't give a explanation, I'm curious about what's the drawback of using `'de` as the lifetime of type which Deserialize impl applies.

Mingun:
Probably https://users.rust-lang.org/t/how-to-end-borrow-in-this-code/72719/3?u=mingun can give you an answer:
> ...
> ```rust
> // `&'static [u8]` in particular implements
> // `for<'any> BufferedInput<'any, ()>`
> impl<'long: 'short, 'short> BufferedInput<'short, ()> for &'long [u8] {
>     fn read(&mut self, _buf: ()) -> &'short [u8] {
>         self
>     }
> }
> ```
> ...
> You may wonder why we had to spell this out, and the answer is that lifetimes in trait parameters are _invariant_ -- they can't automatically shrink, like you're used to with `&[u8]`, say. The above change makes the implementation _covariant_ by breaking up the lifetime of the implemented trait (`'short`) and the lifetime of the implementer (`'long`) -- every concrete `&'long [u8]` gets an implementation for _every_ `'short` that is shorter.
> ...

But I totally agree that such an explanation should be placed in the documentation.

grelative:
> ```rust
> impl<'long: 'short, 'short> BufferedInput<'short, ()> for &'long [u8] {
>     fn read(&mut self, _buf: ()) -> &'short [u8] {
>         self
>     }
> }
> ```

Yeah, this definiton of `BufferedInput` has the benefit of satisfying the HRTB ` for<'any> &'static [u8]: BufferedInput<'any, ()>`.  
But it does not apply to `Deserialize`, there is nothing can represent a lifetime smaller than anyone, so we can not form a HRTB ` for<'any> Q<'???>: Deserialize<'any`  that the following definition will satisfy.
```rust
 impl<'de: 'a, 'a> Deserialize<'de> for Q<'a> {}
```

ia0:
[...]

PR #2744, opened 2024-05-15
Add lifetime container attribute for user-provided lifetime

ia0:
This PR is for discussion and related to #2190. As discussed there, the benefit is not obvious (except for providing more control to the user).

With this PR, a user can use `#[serde(lifetime = 'a)]` to control the lifetime of the implementation of `Deserialize`. In particular, the user is now able to control all the implementation parameters when using `#[serde(bound = "", lifetime = 'a)]`.

See https://github.com/ia0/wasefire/blob/ccf7bc30ac8440b30981b2f39b734de10d1a037c/crates/protocol/src/lib.rs#L65 for an example in real code which gives `impl<'a, T: Direction<'a>> _serde::Deserialize<'a> for Api<'a, T>` when looking with `cargo expand`. The preceding commented line is the solution without this PR.

### DIFF
--- a/serde_derive/src/de.rs
+++ b/serde_derive/src/de.rs
@@ -225,13 +225,15 @@ fn requires_default(field: &attr::Field, _variant: Option<&attr::Variant>) -> bo
 enum BorrowedLifetimes {
     Borrowed(BTreeSet<syn::Lifetime>),
     Static,
+    User(syn::Lifetime),
 }
 
 impl BorrowedLifetimes {
     fn de_lifetime(&self) -> syn::Lifetime {
-        match *self {
+        match self {
             BorrowedLifetimes::Borrowed(_) => syn::Lifetime::new("'de", Span::call_site()),
             BorrowedLifetimes::Static => syn::Lifetime::new("'static", Span::call_site()),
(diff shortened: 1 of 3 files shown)


ia0:
Actually, I think I have a concrete example where it is beneficial (and not just more expressive) to be able to choose the `Deserialize` lifetime parameter from the `impl` parameters. When having a very large `enum Api<'a, 'de: 'a, T: Flavor>` where each variant is of the form `Foo(T::Type<'a, 'de, Foo>)` because otherwise serde will add a bound `T::<'a, 'de, Foo>: Deserialize<'x>` for each variant `Foo` (possibly increasing compilation time).

```rust
#[derive(Deserialize)]
#[serde(bound = "", lifetime = 'de)]
enum Api<'a, 'de: 'a, T: Flavor> {
    Foo(T::Type<'a, 'de, Foo>),
    // many similar lines replacing Foo with something else
}

trait Flavor {
    type Type<'a, 'de: 'a, T: Variant>: 'a + Deserialize<'de>;
}
```

I'll try to come up with concrete code today. The idea is that `Api` is only meant to be a view into serialized data. So it doesn't need any bound, it is statically always deserializable.

PR #2747, opened 2024-05-25
Document tuple impls as fake variadic

dtolnay:
**Before:**

Serialize
Deserialize

**After:**

Serialize
Deserialize


oli-obk:
That's a very cool Rustdoc feature

Issue #1137, opened 2018-01-09
Allow for easy versioning of structures


Ekleog:
I was thinking something like http://www.boost.org/doc/libs/1_36_0/libs/serialization/doc/tutorial.html#versioning could be expressed elegantly with attributes like (unchecked syntax):
```rust
#[derive(Serialize, Deserialize)]
#[serde(version=5, removed_field(version=4, name="old_field", type=String))]
struct Example {
    #[serde(added(version=5, default_value=None))]
    new_field: Option<String>,

    #[serde(added(version=4, default="build_example"))]
    field: String,
}

fn build_example(old_field: &String, new_field: &Option<String>) -> String {
    "OLD ".to_owned() + old_field
}
```

Here, there has been a field added in version 5, which defaults to `None` when deserializing a file generated with a previous version, and in version 4 the field `old_field` has been renamed to `field` and it's required to add `"OLD "` in front of every value encoded with pre-4 versions.

This would avoid the need to write custom serializer/deserializers when having to handle multiple versions of a format, and having something more transparent to the user than https://github.com/serde-rs/serde/issues/745

Then this whole idea requires quite a bit of engineering, I guess, and of thought about the way to pass arguments to `build_example`.

What do you think about it?

Ekleog:
Another caveat with it actually: I don't know what the versioning policy for serde is, but implementing it requires that `serde` adds a version number to all the serialized data (else if someone forgot the `version=0` on the `struct` they wouldn't be able to bump the version for the next iteration of the `struct`), so this may require a major version bump? Or just not have `version` fields by default and forbid upgrading from a non-`version`ned `struct`.

avl:
This would be immensely useful for anyone writing large amounts of data to disk in an evolving program where data structures change but old saved data still needs to be usable.

dtolnay:
I agree this would be valuable! Lots of users want versioning of structures and the ability to evolve over time while continuing to support old data.

I am closing the issue because I believe this would be better explored as a library separate from Serde, possibly built on top of Serde. Once there is a standalone prototype and people generally agree that it meets their needs, we can reconsider whether an modifications in Serde are required to integrate better with the functionality.

avl:
I've written something: https://github.com/avl/savefile

ssokolow:
@avl For the record, this means that `savefile` will *never* be useful to me for any of my projects:

> Features savefile does not have, and will not have:
> 
[...]

PR #2765, opened 2024-06-28
Externalize derive macro implementation

vic1707:
Hi, 
I'm currently trying to make a crate built on top of serde to try to implement a relatively naive idea for struture versioning regarding #1137 as a POC (and hopefully a suitable solution).

For ease of use (both as the developper and user of the crate) I'd like my crate to be able to benefit from serde's derive implementation that is thankfully in a separate function that could be exported.

This would allow my crate (and other ones of the same kind) to be 100% compatible with all of serde derive's attribute and features by manually running the derive function and modifying the AST with `quote` and `syn`.

This would allow crate developers to inject their features direclty on the `Serialize` or `Deserialize` implementation outputed by `serde_derive`, kinda like what I'm doing:

```diff
--- usage.expanded.initial.rs	2024-06-28 23:42:09
+++ usage.expanded.modified.rs	2024-06-28 23:42:09
@@ -4,7 +4,7 @@
 use std::prelude::rust_2021::*;
 #[macro_use]
 extern crate std;
-use serde_derive::Deserialize;
+use my_serde::Deserialize;
 struct Foo {
     name: String,
     age: u8,
@@ -24,6 +24,11 @@
         where
             __D: _serde::Deserializer<'de>,
         {
+            'my_serde: {
+                {
+                    ::std::io::_print(format_args!("HI!!!\n"));
+                };
+            }
             #[allow(non_camel_case_types)]
             #[doc(hidden)]
             enum __Field {

``` 

original file being:
```rs
/* Clippy config */
#![allow(dead_code)]
/* Dependencies */

#[derive(Deserialize)]
struct Foo {
    name: String,
    age: u8,
    #[serde(default)]
    place_holder: String,
}

fn main() {}
```

This is, I think, kinda related to #2021.

To do so I pretty much replicated the `serde_derive_internals` situation.

If this change is ok to you (if its not I can simply use my fork for the time being), I'd like to have your opinion on how you want me to fix the current clippy issues (the lib file currently being the same as for `serde_derive_internals`)
Simply add a bunch of `#[allow()]` on the `lib.rs`, try my best to fix every one of them, alternative ?

#### Notes
English isn't my primary language, I'm sorry if I made mistakes, if anything sounds bad or if I'm hard to understand 😓.


vic1707:
Just released my code for versioning here if you want to take a look at what this PR allows ^^
https://github.com/vic1707/serde-versioning

btw, if you have time, I'd love to hear your opinion the project

vic1707:
May I ask for your review @dtolnay ?

serde_derive_implementation/src
```diff
@@ -0,0 +1 @@
+../serde_derive/src/implementation/
```
oli-obk, line 1: Is this a symbolic link?

oli-obk:
The issues raised in https://github.com/serde-rs/serde/pull/2021 are not really resolved by this.

I think this is a feature needed commonly enough that we could probably quickly figure out a small robust rust language feature to allow you to call proc macros from other proc macros. I think that would be better in general

oli-obk:
Going to close this as I don't see it landing in this form any time soon.

serde_derive_implementation/src
vic1707: yup it was, just like for `serde_derive_internals `
See: https://github.com/serde-rs/serde/blob/master/serde_derive_internals/src

oli-obk:
For now you can create a git repo with this repo as a submodule and create your crate that way. It'll be on you to update then, but that was already true. The main problem with those approaches is keeping in sync with the serde version for the internal API that the derives use

vic1707:
> For now you can create a git repo with this repo as a submodule and create your crate that way.

Haha, I honestly didn't even thought about submodules, thanks for the idea!


---

> The main problem with those approaches is keeping in sync with the serde version for the internal API that the derives use

Yup but that's a burden for libraries developers not for `serde` maintainers right ?
Lib devs would write down which version of `serde` is supported for each versions of their crates just like some already do when depending on/extending `serde_derive` themselves (thinking of crates like `nutype` which, I think, could face similar sync issues)?

(Not trying to get the PR opened again, just genuinely trying to understand, I probably am missing a point 😄)

vic1707:
Oh, I thought you were saying that for serde and the PR itself, yup it could become a pain for me and others if serde changes too much.

I didn't saw it as an issue for me as in #1137 dtolnay said that if an extension of serde becomes a popular/viable solution it could get merged directly in serde and was hoping I could one day get that chance 🙃.

Thanks for the explanations and your time 👍

vic1707:
Hi, sorry to bother you again 🙇 

about

> For now you can create a git repo with this repo as a submodule and create your crate that way.

how do you see this working ? My project https://github.com/vic1707/serde-versioning was already working using git url for the dependency, but couldn't publish to crates due to that.
I tried many approaches to get a working version publishable on crates.io but can't get one to work properly 😓.

I successfully added my fork as a submodule at the root, pointing the dependency to the local path works fine but I can't publish to crates.io since the newly made crate doesn't exist there (expected, same as using the git url).
So I tried multiple solutions using symlinks from and to various folders but none of them satisfies `cargo build` 😓.

Do you have an idea on how I can get it to work so I can do https://github.com/vic1707/serde-versioning/issues/2 😅  ?

oli-obk:
I mean a literal `git submodule`, not a git dependency. Got submodules are not something cargo understands or cares about, it treats them as just normal files

vic1707:
Yeah thats what I tried but the files weren't getting used during publish, I finally found a way to get it to work properly 🥳 

---

Would you be opened to merging half of this PR (in a separate PR ofc) ?
I would like to merge the creation of the `implementation` subfolder in `serde_derive` and move the root files there.
Meaning its the same content as this PR minus the new crate `serde_derive_implementation`.

Merging this tiny change, apart from making the `serde_derive` source a bit cleaner (in my opinion) would allow me close my fork as I can live without `serde_derive_implementation` using the submodule solution and directly depend on serde's HEAD commit, making everything a bit cleaner.

I understand this can be considered a selfish request so I'll wait for your answer before doing any of that.

vic1707:
After re-re-re-...-re reading your comment I think I finally understood what you were talking about
```rs
#[path = "../serde/serde_derive_implementation/src/mod.rs"]
mod serde_derive_implementation;
```

and using it like 
```rs
use crate::serde_derive_implementation::de::expand_derive_deserialize;
```
Is this what you had in mind? 

Unfortunately, due to the symlinks and `crate::`/`$crate::` it causes `cargo build` to fail
```
error: could not compile `serde-versioning` (lib) due to 72 previous errors; 16 warnings emitted
```

vic1707:
But using 
```rs
#[allow(clippy::all)]
#[rustfmt::skip]
#[path = "../serde/serde_derive/src/implementation/mod.rs"]
mod implementation;
#[allow(clippy::all)]
#[rustfmt::skip]
#[path = "../serde/serde_derive/src/internals/mod.rs"]
mod internals;
```

I can get all `cargo build` to work but not `cargo publish` despite adding `"serde/**/*"` to the include list of the `Cargo.toml` 🤔.
I'll continue searching in this direction. But if it works too I'll still ask if it's possible to make the half-PR 😅

oli-obk:
> else I'd have to make 7 `#[path = ".../XX"] mod XX;` one for each module at the root of `serde_derive`)

That seems not too bad to me. If your crate becomes important enough to support it from serde-rs we can always revisit

vic1707:
Ok then, will work with that for now then 👍

PR #2767, opened 2024-07-02
Use the `#[diagnostic::on_unimplemented]` attribute when possible

weiznich:
This change enables the `#[diagnostic::on_unimplemented]` attribute for the `Serialize` and `Deserialize` trait to point the user to the relevant derives and point out that they might want to check crates features for external types by adding the relevant hints an note.

The help text is up for discussion, but I feel that this might help quite a lot of users that struggle why external crates do not implement the serde traits.


oli-obk (approved):

dtolnay (approved):
Thanks!

I've added a test for this in https://github.com/serde-rs/serde/pull/2769.

PR #2769, opened 2024-07-06
Add ui test of unsatisfied serde trait bound

dtolnay:
Test the diagnostic created by https://github.com/serde-rs/serde/pull/2767.

### DIFF
--- a/test_suite/tests/ui/on_unimplemented.stderr
+++ b/test_suite/tests/ui/on_unimplemented.stderr
@@ -6,6 +6,8 @@ error[E0277]: the trait bound `MyStruct: Serialize` is not satisfied
    |     |
    |     required by a bound introduced by this call
    |
+   = note: for local types consider adding `#[derive(serde::Serialize)]` to your `MyStruct` type
+   = note: for types from other crates check whether the crate offers a `serde` feature flag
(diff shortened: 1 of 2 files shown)


test_suite/tests/ui/on_unimplemented.stderr
```diff
@@ -0,0 +1,328 @@
...
+  --> tests/ui/on_unimplemented.rs:21:15
+   |
+21 |     to_string(&MyStruct);
+   |     --------- ^^^^^^^^^ the trait `Serialize` is not implemented for `MyStruct`
+   |     |
+   |     required by a bound introduced by this call
+   |
+   = note: for local types consider adding `#[derive(serde::Serialize)]` to your `MyStruct` type
+   = note: for types from other crates check whether the crate offers a `serde` feature flag
+   = help: the following other types implement trait `Serialize`:
```
oli-obk, line 11: LMAO wat. An exhaustive list of all impls

cc @estebank

dtolnay: This is from passing `--verbose` which is to work around error messages changing based on how many characters long the directory of your git checkout is, which affects the long type name truncation heuristic. See https://github.com/dtolnay/trybuild/issues/269.

PR #2781, opened 2024-07-28
Add support for 128-bit integers in flatten structs and internally tagged enums

### LINKED ISSUE #1682: Untagged enum does not play nice with u128
I would expect both tests pass, but instead, only `test64` passed, while `test128` failed with
> data did not match any variant of untagged enum E128

```rust

#[derive(Debug, PartialEq)]
pub struct S64 {
    pub f: u64,
}

#[serde(untagged)]
#[derive(Debug, PartialEq)]
pub enum E64 {
    S(S64),
}

#[derive(Debug, PartialEq)]
pub struct S128 {
    pub f: u128,
}

#[serde(untagged)]
#[derive(Debug, PartialEq)]
pub enum E128 {
    S(S128),
}


#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn test64() {
        let incoming_str = r#"{"f":1}"#;
        let expected_incoming = E64::S(S64 { f: 1 });
        let incoming = serde_json::from_str(&incoming_str).unwrap();
        assert_eq!(expected_incoming, incoming);
    }

    #[test]
    fn test128() {
        let incoming_str = r#"{"f":1}"#;
        let expected_incoming = E128::S(S128 { f: 1 });
        let incoming = serde_json::from_str(&incoming_str).unwrap();
        assert_eq!(expected_incoming, incoming);
    }
}

```

[playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=bbf13b2389f764ecd695566b01b0a146)

### LINKED ISSUE #2230: Implement `visit_i128` and `visit_u128` for ContentVisitor
I'm working with a [library](https://github.dev/ipld/libipld/blob/cee752633cf7e06455f1b294f3d120cb639dc9c5/core/src/serde/de.rs) that only supports 128-bit integers for its data format. It is a self describing format which means it implements `deserialize_any`. It mostly works fine with serde out of the box because there is a custom visitor impl that implements the `visit_*128` methods.

The issue I have only rears its head when I derive Deserialize on an **untagged enum**. The debugger drops me in a `Content::deserialize`, which calls `deserializer.__deserialize_content(actually_private::T, visitor)` that takes a `ContentVisitor`. 

https://github.com/serde-rs/serde/blob/56bd3694227fa9d02e85ee1493f9709f241995e8/serde/src/private/de.rs#L297-L298

`ContentVisitor` does not implement `visit_i128` and `visit_u128`. Which I assume is because there is no `Content::I128` and `Content::U128` variants.

https://github.com/serde-rs/serde/blob/56bd3694227fa9d02e85ee1493f9709f241995e8/serde/src/private/de.rs#L220-L249

So the default `visit_*128` methods get called instead causing the error I end up getting.

https://github.com/serde-rs/serde/blob/e1c4517335cc5ce440b7d21bd76887c9fd31f5fa/serde/src/de/mod.rs#L1362-L1376

I implemented a quick solution that ought to be less intrusive but it ends up breaking multiple libraries, including `serde_json`, that use the `Unexpected::*` enum. In fact, in my case it broke so many libraries that I have decided to take a different approach to the problem.

https://github.com/serde-rs/serde/compare/master...appcypher:appcypher/content-visitor-i128

128-bit integers have been part of stable Rust for years now and with more people working with 128-bit integers, I think it is worth finding a permanent solution to this problem.

### LINKED ISSUE #2576: ContentDeserializer does not support 128-bit integers
ContentDeserializer does not support 128-bit integers, which causes issues when attempting to use some features with formats that support 128-bit numbers. The following `tag`/`rename` pattern fails using `ciborium` and `serde_cbor`, returning `ErrorImpl { code: Message("i128 is not supported"), offset: 0 }`.

```rust
    #[test]
    fn bug() {
        #[derive(serde::Deserialize, serde::Serialize)]
        struct Foo {
            message: String,
        }

        #[derive(serde::Deserialize, serde::Serialize)]
        struct Bar {
            value: i128,
        }

        #[derive(serde::Deserialize, serde::Serialize)]
        #[serde(tag = "ty")]
        enum MyEnum {
            #[serde(rename = "foo")]
            Foo(Foo),
            #[serde(rename = "bar")]
            Bar(Bar),
        }

        // this works (decode MyEnum::Foo)
        {
            let v = MyEnum::Foo(Foo {
                message: "hello, world!".to_owned(),
            });
            let encoded = serde_cbor::to_vec(&v).unwrap();
            let decoded: MyEnum = serde_cbor::from_slice(&encoded).unwrap();
            match decoded {
                MyEnum::Foo(v) => assert_eq!(&v.message, "hello, world!"),
                _ => assert!(false),
            }
        }

        // this works (decode Bar directly)
        {
            let v = MyEnum::Bar(Bar {
                value: 12345678901234567890,
            });
            let encoded = serde_cbor::to_vec(&v).unwrap();
            let decoded: Bar = serde_cbor::from_slice(&encoded).unwrap();
            assert_eq!(decoded.value, 12345678901234567890);
        }

        // this fails (decode MyEnum::Bar), i128 not supported
        {
            let v = MyEnum::Bar(Bar {
                value: 12345678901234567890,
            });
            let encoded = serde_cbor::to_vec(&v).unwrap();
            let decoded: MyEnum = serde_cbor::from_slice(&encoded).unwrap();
            match decoded {
                MyEnum::Bar(v) => assert_eq!(v.value, 12345678901234567890),
                _ => assert!(false),
            }
        }
    }
```

### LINKED ISSUE #2948: Does this seem to be a bug from deserializing enum?
## serde version info

```text
└── serde_derive v1.0.219 (proc-macro)
    ├── proc-macro2 v1.0.95
    │   └── unicode-ident v1.0.18
    ├── quote v1.0.40
    │   └── proc-macro2 v1.0.95 (*)
    └── syn v2.0.104
        ├── proc-macro2 v1.0.95 (*)
        ├── quote v1.0.40 (*)
        └── unicode-ident v1.0.18

serde v1.0.219
└── serde_derive v1.0.219 (proc-macro) (*)

serde_json v1.0.141
├── itoa v1.0.15
├── memchr v2.7.5
├── ryu v1.0.20
└── serde v1.0.219
    └── serde_derive v1.0.219 (proc-macro)
```

## system info 
```text
stable-x86_64-pc-windows-msvc (active, default)
rustc 1.88.0 (6b00bc388 2025-06-23)
```

## same rust code example
```rust
#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
pub struct Vote {
    vote_address: String,
    vote_count:u64,
}

pub type Votes = Vec<Vote>;


#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
#[serde(untagged)]
pub enum Case {
    VoteWitnessContract {
        owner_address: String,
      
        votes: Votes,
    },
    FreezeBalanceV2Contract{
        frozen_balance:i128,
        owner_address:String,
    }
}


#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
pub struct Content{
    data:Vec<Case>
}

#[test]
fn test_case(){
    let response =  r#"{"data":[{"owner_address":"41cf418b1cd1a610e8b370abc1e428848f82e3a153","votes":[{"vote_address":"4136b8367bdad6b910a86e1d8af121c870d9e12a71","vote_count":900}]},{"frozen_balance":500000000,"owner_address":"41cf418b1cd1a610e8b370abc1e428848f82e3a153"}]}"#;
    serde_json::from_str::<Content>(&response).unwrap();

}

```

## Error info

```text
---- case::test_case stdout ----
thread 'case::test_case' panicked at src\case.rs:33:48:
called `Result::unwrap()` on an `Err` value: Error("data did not match any variant of untagged enum Case", line: 1, column: 251)
```

[Enum Excerpted from official examples `#[serde(untagged)]`](https://serde.rs/enum-representations.html)

Mingun:
This PR depends on https://github.com/serde-rs/test/pull/6 and currently temporary uses my fork with branch that adds necessary support for testing 128-bit integers. It is possible, however, drop the first two commits, but that will mean that 128-bit integers will not be tested. I'd prefer to merge https://github.com/serde-rs/test/pull/6 and then update this PR to use new version of `serde_test`.

Closes #1682, closes #2230, closes #2576, closes #2948


WiSaGaN:
This looks like a pretty clean fix. @dtolnay would you take a look please?

Pzixel:
Ping?

robertprp:
up

vmx:
@Mingun thanks for this PR. I tried it out for my case. It partially fixes the problem I have, hence I comment directly on this PR. In my case I want to be able to deserialize a `i128` into other signed integer types if they fit. Similarly to mapping an `i64` into other integers.

It applies to all signed integer type, but let's make things concrete and use `i32` as an example. The problem is that the `PrimitiveVisitor` for `i32` doesn't implement `visit_i128()`, hence it falls back to the [default implementation](https://github.com/serde-rs/serde/pull/2781/files#diff-416ca0b8884a19daedc85bfe16a4498e2b18c9e13990e2774d36d60bbc22f454R1433), which triggers an `invalid_type` error.

A workaround is adding in the[ `impl_deserialize_num! { i32 …` ](https://github.com/serde-rs/serde/blob/3efdb99096d96b2e95f71d6214595424bb67beec/serde_core/src/de/impls.rs#L391-L397) the `i128:visit_128` to  `int_to_int!(i64:visit_i64);`. 

Does this belong to this PR or would it be a separate PR? Do you think that would be the correct fix for the issue?

I sadly can't provide a simple test case for it. Mine is pretty entangled as it involves even a Deserializer implementation and I wasn't able to create a simpler version.

Mingun:
@vmx, you're right, I should update my PR

bocc:
For those running into this issue: if you have the chance to use a different numeric type, like `bigdecimal`, with it's own deserializer, that should work.

test_suite/Cargo.toml
```diff
@@ -17,5 +17,5 @@ foldhash = "0.2"
 rustversion = "1.0"
 serde = { path = "../serde", features = ["rc"] }
 serde_derive = { path = "../serde_derive", features = ["deserialize_in_place"] }
-serde_test = "1.0.176"
+serde_test = { git = "https://github.com/Mingun/serde-test", branch = "integer128" }
```
25pwn, line 20: This should be changed for upstreaming

Mingun: Of course. [Corresponding PR](https://github.com/serde-rs/test/pull/6) should be merged first.

bourumir-wyngs:
serde-saphyr also depends on this. I have adopted the old serde-yaml test (https://github.com/dtolnay/serde-yaml/issues/391)  that, despite no code shared, reproduces precisely.

PR #2782, opened 2024-07-30
workspace 2018 rust edition

mahmoud-eltahawy:
set all workspace crates edition to 2018


oli-obk:
Unfortunately this breaks our CI tests for past rustc versions. I don't think we can use these workspace features for the forseeable future, if ever.

PR #2785, opened 2024-08-03
Use serialize_entry instead of serialize_key + serialize_value when serialize flatten newtype enum variant

Mingun:
Serializers that reimplements serialize_entry will get benefits from that


dtolnay (approved):
Thanks!

PR #2786, opened 2024-08-03
Fix roundtrip of flatten externally tagged enum unit variant

Mingun:
Unit variant of externally tagged enum can be deserialized when enum is flatten, [but cannot be serialized](
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=67545ca025699ac9894f464ad1dc2405) -- `ser` test failed:
```rust
#![cfg(test)]

use serde::{Deserialize, Serialize}; // 1.0.204;
use serde_json::{from_str, to_string}; // 1.0.121

struct Flatten {
    #[serde(flatten)]
    data: Enum,
}

enum Enum {
    Unit,
}

const JSON: &str = r#"{"Unit":null}"#;

#[test]
fn de() {
    // ok
    assert_eq!(from_str::<Flatten>(JSON).unwrap(), Flatten { data: Enum::Unit });
}

#[test]
fn ser() {
    // called `Result::unwrap()` on an `Err` value: Error("can only flatten structs and maps (got an enum)", line: 0, column: 0)
    assert_eq!(to_string(&Flatten { data: Enum::Unit }).unwrap(), JSON);
}
```

This PR fixes this inconsistency

### DIFF
--- a/serde/src/private/ser.rs
+++ b/serde/src/private/ser.rs
@@ -54,6 +54,7 @@ enum Unsupported {
     Sequence,
     Tuple,
     TupleStruct,
+    #[cfg(not(any(feature = "std", feature = "alloc")))]
     Enum,
 }
 
@@ -70,6 +71,7 @@ impl Display for Unsupported {
             Unsupported::Sequence => formatter.write_str("a sequence"),
             Unsupported::Tuple => formatter.write_str("a tuple"),
             Unsupported::TupleStruct => formatter.write_str("a tuple struct"),
+            #[cfg(not(any(feature = "std", feature = "alloc")))]
             Unsupported::Enum => formatter.write_str("an enum"),
         }
     }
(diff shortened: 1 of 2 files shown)


Mingun:
This PR does not conflicts with #2608, because changed files not moved in #2608, it has pretty obvious test case and reproduction example, so is it seems it shouldn't take too much time for review.

serde/src/private/ser.rs
```diff
@@ -54,6 +54,7 @@ enum Unsupported {
     Sequence,
     Tuple,
     TupleStruct,
+    #[cfg(not(any(feature = "std", feature = "alloc")))]
```
oli-obk, line 57: Why does this only exist without alloc?

Mingun: This variant used only when allocations are not available, rustc emits a warning about unused variant.

oli-obk: Oh 🤦 I forgot to check the impl, thx

PR #2788, opened 2024-08-06
Consolidate and add new tests of internally tagged enums

Mingun:
The goal of this PR is to improve test coverage by testing corner cases that could be break during refactorings and implementing optimizations.

This PR moves all tests for internally tagged enums (those which is defined using `#[serde(tag = "tag")]` attribute) into a dedicated file, removes duplicated tests and organize tests in a logical manner.

This PR does not add tests which is failed without changes in `serde_text` or in `serde` (they are left in #2569). So only refactoring of existing tests and a new tests that already passes.


dtolnay (approved):
Thank you!

PR #2794, opened 2024-08-11
Temporarily ignore collection_is_never_read on FlattenSkipDeserializing

dtolnay:
@Mingun, this lint fires since #2567. Could you take a look at what should be generated for such a struct?

https://github.com/serde-rs/serde/blob/fc55ac70d34221b38672b1583e496011fbae92aa/test_suite/tests/test_gen.rs#L723-L727

```console
error: collection is never read
   --> test_suite/tests/test_gen.rs:723:25
    |
723 |     #[derive(Serialize, Deserialize)]
    |                         ^^^^^^^^^^^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#collection_is_never_read
note: the lint level is defined here
   --> test_suite/tests/test_gen.rs:23:9
    |
23  | #![deny(clippy::collection_is_never_read)]
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    = note: this error originates in the derive macro `Deserialize` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Issue #1529, opened 2019-05-18
Deserializing to a flattened struct calls `deserialize_map`, while deserializing to an already-flat struct does not


peterjoel:
The original problem is that my format is not self-describing, and I need to rely on the data types in the target struct, when deserializing, but I found that _sometimes_ `deserialize_any` was being called on my Deserializer. It turned out that this was whenever I used the `#[serde(flatten)]` attribute.

I have tried to reduce this problem to the minimal necessary code, in [this gist](https://gist.github.com/peterjoel/a41363fea48c4cb529e4a4bf8421ec20). The problem shows itself by the fact that `deserialize_map` is called in the failing test, but is not called in the passing test.

Deserializing this struct works (does not call `deserialize_map`):
```Rust
#[derive(Deserialize)]
#[serde(rename = "Message")]
struct Flattened {
    a: String,
    b: String,
}
```
But deserializing these structs results in a call to `deserialize_map`:
```Rust
#[derive(Deserialize)]
#[serde(rename = "Message")]
struct Nested {
    a: String,
    #[serde(flatten)]
    inner: NestedInner,
}

#[derive(Debug, PartialEq, Deserialize)]
struct NestedInner {
    b: String,
}
```

I also checked that _both_ structures expect the following deserialize tokens:
```Rust
&[
    Token::Struct { name: "Message", len: 5 },
    Token::Str("a"),
    Token::Str("v1"),
    Token::Str("b"),
    Token::Str("99"),
    Token::StructEnd,
]
```
Perhaps I have this wrong, but I would expect this to indicate that both visitors should interact with the deserializer identically.

Full code: https://gist.github.com/peterjoel/a41363fea48c4cb529e4a4bf8421ec20

peterjoel:
Possibly related to #1504?

peterjoel:
This issue adds significant performance cost when using flattened structs with `serde_json`. See: https://gist.github.com/peterjoel/eafb936143a988f6a922efd2b52ebf18

In the test, the struct with `#[serde(flatten)]` takes 3 times longer to deserialize compared to the struct that is already flat.

```
test benches::deserialize_to_flat                 ... bench:          81 ns/iter (+/- 6)
test benches::deserialize_to_nested_and_flattened ... bench:         223 ns/iter (+/- 14)
```

andrey-yantsen:
[...]

PR #2795, opened 2024-08-11
`has_flatten` rework

Mingun:
This PR removes workaround from #2794 in the first commit.

The 2nd commit unifies behaviour for all cases where flatten fields processed in the deserializer. Consequence of that commit: `FlattenSkipDeserializing[DenyUnknown]`
- does not collect data in `Field`, because do not read them anyway
- gets `deserialize_in_place` method
- gets ability to deserialize from sequence (`visit_seq` method)
- uses `deserialize_struct` instead of `deserialize_map`

The 3rd commit gives understandable message when derive failed due to assertion.

The 4th commit removed dead code that here since 2018.

The last commit does the unification of `has_flatten` behaviour for the serializer side. Consequence of that commit: `FlattenSkipSerializing`
- uses `serialize_struct` instead of `serialize_map`

### DIFF
--- a/serde_derive/src/de.rs
+++ b/serde_derive/src/de.rs
@@ -285,13 +285,7 @@ fn deserialize_body(cont: &Container, params: &Parameters) -> Fragment {
                 deserialize_struct(params, fields, &cont.attrs, StructForm::Struct)
             }
             Data::Struct(Style::Tuple, fields) | Data::Struct(Style::Newtype, fields) => {
-                deserialize_tuple(
-                    params,
-                    fields,
-                    &cont.attrs,
-                    cont.attrs.has_flatten(),
-                    TupleForm::Tuple,
-                )
+                deserialize_tuple(params, fields, &cont.attrs, TupleForm::Tuple)
             }
             Data::Struct(Style::Unit, _) => deserialize_unit_struct(params, &cont.attrs),
         }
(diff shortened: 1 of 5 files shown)


dtolnay (commented):
Thanks!

This PR makes `derive(Deserialize)` ignore a `flatten` attribute on fields that have `skip_deserializing`, and `derive(Serialize)` ignore a `flatten` attribute on fields that have `skip_serializing`.

I would have to think whether that is the behavior that should be intended. It means that in the case of `FlattenSkipSerializing` and `FlattenSkipDeserializing` (using names from the test suite), we serialize using `serialize_struct` and deserialize using `deserialize_map`, or serialize using `serialize_map` and deserialize using `deserialize_struct`.

I think some users would expect that if a type is serialized using `serialize_struct`, it should be deserialized using `deserialize_struct`, and if a type is serialized using `serialize_map`, it should be deserialized using `deserialize_map`. I am not sure how sensitive various data formats are to this.

Mingun:
Boundary between `(de)serialize_struct` and `(de)serialize_map` is already blurred when you use `#[serde(flatten)]`. Actually, that is source of pain because `(de)serialize_map` looses information which is important for some formats (see for example #1529). There is no good solution of this problem in current approach which pretends that simply using special `(De)serializer` will be enough to support flattening. The alternative design is outlined in my comment [here](https://github.com/serde-rs/serde/issues/2186#issuecomment-1065875007).

I think, if you explicitly skip only (de)serialization method you understand consequences. I would say that the base expectation that if field is invisible for (de)serialization it does not affects the format.

dtolnay (approved):
Okay &mdash; makes sense.

Thank you!

PR #2797, opened 2024-08-12
Emit an error if a variant of an untagged enum is annoted with `#[serde(rename)]` or `#[serde(alias)]`

### LINKED ISSUE #2787: `#[serde(rename)]` and `#[serde(alias)]` should be forbidden on variants of untagged enums
Both attributes does nothing and just confuse users:
```rust
#[derive(serde::Deserialize)]
#[serde(untagged)]
pub enum Untagged {
    #[serde(rename = "unit", alias = "_1")]
    Unit,

    #[serde(rename = "newtype", alias = "_2")]
    Newtype(u32),

    #[serde(rename = "tuple", alias = "_3")]
    Tuple(u32, u32),

    #[serde(rename = "struct", alias = "_4")]
    Struct { a: u32 },
}
```
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=bc78b6a4d132a2c255db120036d7fa42

Instead an error should be emitted that such attributes (maybe some other too!) does nothing and not allowed for untagged enums.

devnetsec:
Fixes #2787.

### DIFF
--- a/serde_derive/src/internals/ast.rs
+++ b/serde_derive/src/internals/ast.rs
@@ -65,8 +65,10 @@ impl<'a> Container<'a> {
     ) -> Option<Container<'a>> {
         let mut attrs = attr::Container::from_ast(cx, item);
 
+		let untagged = attrs.tag() == &attr::TagType::None;
+		
         let mut data = match &item.data {
-            syn::Data::Enum(data) => Data::Enum(enum_from_ast(cx, &data.variants, attrs.default())),
+            syn::Data::Enum(data) => Data::Enum(enum_from_ast(cx, &data.variants, attrs.default(), untagged)),
             syn::Data::Struct(data) => {
                 let (style, fields) = struct_from_ast(cx, &data.fields, None, attrs.default());
                 Data::Struct(style, fields)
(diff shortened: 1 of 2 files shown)


serde_derive/src/internals/ast.rs
```diff
@@ -141,11 +143,16 @@ fn enum_from_ast<'a>(
...
+    container_is_untagged: bool,
 ) -> Vec<Variant<'a>> {
     let variants: Vec<Variant> = variants
         .iter()
         .map(|variant| {
             let attrs = attr::Variant::from_ast(cx, variant);
+
+			if container_is_untagged && attrs.name().renamed() {
+				cx.error_spanned_by(&variant.ident, "renaming or adding a alias to a variant of an untagged enum does nothing");
+			}
```
Mingun, line 155: Ideally, error should point to the each problematic attribute instead of a variant name. Also, compile tests should be added to `test_suite/tests/ui`

devnetsec: I fixed those two things. Are there any other attributes that need to have error messages like this?

Mingun: Thanks, much better, but it is better to point to the `alias = "..."` and `rename = "..."` parts only, so when you have multiple attributes on a single line, errors pointed to the corresponding parts of that line.

Don't known about other attributes. You may check other if they need improvements.

devnetsec: I made it point towards the `alias` and `rename` parts to be even more clear. I think `#[serde(rename_all)]` may be another attribute that has this issue, but I couldn't get that error to work in 6d725642135e5d34fcde8cda00324b71d66b9e52.

oli-obk:
Can someone run crater-at-home on this to gauge the impact? I worry someone accidentally is using it in a crate that will impact many other crates.

Or switch it to emitting a warning instead (can we do that from proc macros?)

devnetsec:
Unfortunately, I don't think this can be switched to a warning. Compile errors are handled by `serde::internals::ctxt`, which holds Syn errors. Syn uses Rust's `compile_error!` macro to print them. It looks like we're waiting on the nightly-only `Diagnostic` API to stabilize. See [this](https://stackoverflow.com/questions/54392702/how-to-report-errors-in-a-procedural-macro-using-the-quote-macro) Stack Overflow thread. Also, the test suite only compares expected and actual output when compilation fails.

PR #2802, opened 2024-08-15
Support (de-)serializing flattened unit struct

### LINKED ISSUE #2801: Flattening a unit struct should work like flattening unit itself
https://github.com/serde-rs/serde/pull/1874 landed support for flattening the unit type after a report in https://github.com/serde-rs/serde/issues/1873. However, it did not override `deserialize_unit_struct` and `serialize_unit_struct`, so while flattening `()` now works, flattening `struct Unit;` does not. It's not clear whether this was intentional or not, though it seems reasonable to support unit structs the same as unit here.

jonhoo:
Fixes #2801.

### DIFF
--- a/test_suite/tests/test_annotations.rs
+++ b/test_suite/tests/test_annotations.rs
@@ -1816,6 +1816,32 @@ fn test_flatten_unit() {
 }
 
 #[test]
+fn test_flatten_unit_struct() {
+    #[derive(Debug, PartialEq, Serialize, Deserialize)]
+    struct Response<T> {
+        #[serde(flatten)]
+        data: T,
+        status: usize,
+    }
+
+    #[derive(Debug, PartialEq, Serialize, Deserialize)]
+    struct Unit;
+
+    assert_tokens(
+        &Response {
+            data: Unit,
+            status: 0,
+        },
+        &[
+            Token::Map { len: None },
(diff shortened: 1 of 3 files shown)


test_suite/tests/test_annotations.rs
```diff
@@ -1815,6 +1815,32 @@ fn test_flatten_unit() {
     );
 }
 
+#[test]
+fn test_flatten_unit_struct() {
```
Mingun, line 1819: It would be great if new flatten tests were added to the `flatten` module in the end of this file. Just my opinion

dtolnay (approved):
Thanks!

I would accept a follow-up PR to move the `flatten` tests (including the existing ones adjacent to the one being added here).

Mingun:
> Thanks!
> 
> I would accept a follow-up PR to move the `flatten` tests (including the existing ones adjacent to the one being added here).

@dtolnay , great to hear that. What about moving other connected tests to made them located near to each other? #2788 does that for internally tagged enum tests and I working on similar PR for adjacently tagged enum tests.

jonhoo:
Sure thing!
https://github.com/serde-rs/serde/pull/2803

I also tried to trim the names down to avoid as much stuttering now that they're under the `mod flatten`.

PR #2804, opened 2024-08-16
Consolidate and add new tests of adjacently tagged enums

Mingun:
Similar to #2788, this PR moves all tests which tests _top-level_ adjacently tagged enums (those which is defined using `#[serde(tag = "t", content = "c")]` attribute) into a dedicated file, removes duplicated tests and organizes tests in a logical manner. Flatten tests when adjacently tagged enum is flattened into structure I consider foremost as flatten tests, so they left untouched.

Removed tests which tests deserialization of generated `tag` and `content` field names from integers and bytes for each enum variant kind, such tests leaved only for `Unit` variant. This is because they are handled by the either [`TagOrContentFieldVisitor`](https://github.com/serde-rs/serde/blob/1a9ffdbd0c1d4d8dd4f7e686e00747c932a36d3d/serde/src/private/de.rs#L906) or [`TagContentOtherFieldVisitor`](https://github.com/serde-rs/serde/blob/1a9ffdbd0c1d4d8dd4f7e686e00747c932a36d3d/serde/src/private/de.rs#L981), which can be tested only once.

Added tests includes deserialization from sequential representation, so this part of generated code becomes tested:
https://github.com/serde-rs/serde/blob/1a9ffdbd0c1d4d8dd4f7e686e00747c932a36d3d/serde_derive/src/de.rs#L1680-L1708


dtolnay (approved):
Thank you!

These test suite PRs are much appreciated.

PR #2805, opened 2024-08-17
Fix deserialization of empty structs and tuples in untagged enums

Mingun:
This PR is logical continuation of #2788 and #2804, for untagged enums. While I working on it I found a bug that was fixed.

This PR moves all tests which tests _top-level_ untagged enums (those which is defined using `#[serde(untagged)]` attribute) into a dedicated file and adds tests to cover missing methods of `ContentRefDeserializer` that used for deserialization. Flatten tests where untagged enum is flattened into structure I consider foremost as flatten tests, so they left untouched.

Each method in `ContentRefDeserializer` and linked structs was marked with a comment that shows which test(s) cover it. After this work a bug was catched:

```rust
#[derive(Deserialize)]
#[serde(untagged)]
enum Untagged {
    Newtype(ExternallyTagged),
}

#[derive(Deserialize)]
enum ExternallyTagged {
    EmptyTuple(),
    EmptyStruct {},
}
```

The `Untagged` type could not be deserialized because `SeqRefDeserializer` (which is used for deserialization) has a special code to handle empty sequences which was not covered and never worked:
https://github.com/serde-rs/serde/blob/1a9ffdbd0c1d4d8dd4f7e686e00747c932a36d3d/serde/src/private/de.rs#L2228-L2244

The new tests shows that it is impossible to create such a type that would accept `visit_unit()` in line 2234. The condition `len == 0` assumes that type would be able to deserialized from unit, but:
1. struct variants was never able to deserialize from it (they expect only `visit_map` or `visit_seq`)
2. tuple variants even with zero fields expect `visit_seq` only. The suggestion to accept `visit_unit` instead was rejected in #2520

After appling the fix the `MapRefDeserializer` and `SeqRefDeserializer` no longer used so they are removed. Actually, because direct deserialization of maps and sequences from the `ContentRefDeserializer` uses `value::MapDeserializer`, that is just unifies the behaviour:
https://github.com/serde-rs/serde/blob/1a9ffdbd0c1d4d8dd4f7e686e00747c932a36d3d/serde/src/private/de.rs#L1938-L1946
https://github.com/serde-rs/serde/blob/1a9ffdbd0c1d4d8dd4f7e686e00747c932a36d3d/serde/src/private/de.rs#L1967-L1975
`value::MapDeserializer` and `value::SeqDeserializer` are used:
https://github.com/serde-rs/serde/blob/1a9ffdbd0c1d4d8dd4f7e686e00747c932a36d3d/serde/src/private/de.rs#L1685-L1718


Mingun:
@dtolnay , @oli-obk , can you look at this bugfix?

dtolnay (approved):
Thanks!

PR #2816, opened 2024-09-05
Implement serialize/deserialize for core::net instead of std::net

### LINKED ISSUE #2760: `IpAddr` and `SocketAddr` are in `core`, but serde implements only for `std`
Hello, recently `IpAddr` was moved to `core`, but serde still requires `std` feature

MathiasKoch:
Implement serialize/deserialize for core::net instead of std::net, if running rust version newer than 1.77, where core::net was stabilized

Fixes #2760


serde/src/ser/impls.rs
```diff
@@ -769,20 +769,45 @@ impl Serialize for SystemTime {
...
+        let written_len = {
+            let mut w = Wrapper {
+                buf: &mut buffer,
+                offset: 0,
+            };
+            write!(&mut w, "{}", $value).unwrap();
+            w.offset
         };
-        let written_len = buffer.len() - remaining_len;
         let written = &buffer[..written_len];
```
MathiasKoch, line 811: This feels like something that could be written in a more elegant way?

Any inputs would be welcomed!

oli-obk: Why was the wrapper struct needed at all? What std thing did the previous code use that core doesn't have?

MathiasKoch: It results in 
```
error[E0599]: the method `write_fmt` exists for mutable reference `&mut [u8]`, but its trait bounds were not satisfied
   --> serde/src/ser/impls.rs:782:20
    |
782 |             write!(remaining, "{}", $value).unwrap();
    |                    ^^^^^^^^^
...
964 |             serialize_display_bounded_length!(self, MAX_LEN, serializer)
    |             ------------------------------------------------------------
    |             |
    |             method cannot be called on `&mut [u8]` due to unsatisfied trait bounds
    |             in this macro invocation
    |
note: must implement `io::Write`, `fmt::Write`, or have a `write_fmt` method
   --> serde/src/ser/impls.rs:782:20
    |
782 |             write!(remaining, "{}", $value).unwrap();
    |                    ^^^^^^^^^
...
964 |             serialize_display_bounded_length!(self, MAX_LEN, serializer)
    |             ------------------------------------------------------------ in this macro invocation
    = note: the following trait bounds were not satisfied:
            `[u8]: Write`
            which is required by `&mut [u8]: Write`
    = note: this error originates in the macro `serialize_display_bounded_length` (in Nightly builds, run with -Z macro-backtrace for more info)
    ```
    
Not sure if this is me doing something obviously wrong?

oli-obk: There's a Write import guarded with cfg(std) somewhere. Generalize it to work with not(cfg(no_core_ip)), too

MathiasKoch: I am not sure that there is? It seems that `core::fmt::Write` is imported as `FmtWrite`, But this seems like a complete copy of my wrapper (https://github.com/serde-rs/serde/blob/master/serde/src/de/format.rs), so I would assume it is indeed needed, but that I can reuse that one?

oli-obk: Ah yea, that makes sense. Reusing existing abstractions makes sense

dtolnay (approved):
Thank you!

dtolnay:
Published in 1.0.210.

PR #2819, opened 2024-09-06
Remove 'useless' unsafe?

bilaliscarioth:
Hello ?

Sorry, I just don't understand the use of the `std::from_utf8_unchecked` function when you can simply use str::from_utf8. Which doesn't break serde (even with a `cargo test` everything looks ok.

### DIFF
--- a/serde/src/format.rs
+++ b/serde/src/format.rs
@@ -13,7 +13,10 @@ impl<'a> Buf<'a> {
 
     pub fn as_str(&self) -> &str {
         let slice = &self.bytes[..self.offset];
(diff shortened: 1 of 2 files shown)


dtolnay (commented):
This was added deliberately in https://github.com/serde-rs/serde/pull/2001.

bilaliscarioth:
I guess you tested just in case, so I guess I can close this sweater request?

dtolnay:
Yeah &mdash; I am inclined to keep the current implementation.

Thanks anyway for the PR!

PR #2821, opened 2024-09-06
Eliminate closures capturing needlessly many type parameters

dtolnay:
Closures capture every type parameter that is in scope, even ones which cannot make a difference to the behavior of the closure. This was previously causing a whole new MapDeserializer to be monomorphized for every distinct Visitor type that we deserialize from Content or &amp;Content (i.e. internally tagged and untagged enums).

Tested by measuring the size of the rlib produced from the following small library. `cargo clean && cargo build --release && ls -l target/release/librepro.rlib`

```rust
use serde::Deserialize;

#[derive(Deserialize)]
#[serde(untagged)]
pub enum Untagged {
    A {},
    B {},
}

pub fn from_str(json: &str) -> serde_json::Result<Untagged> {
    serde_json::from_str(json)
}
```

Before this PR: 127K. After this PR: 111K. This is 12% smaller.

### DIFF
--- a/serde/src/private/de.rs
+++ b/serde/src/private/de.rs
@@ -476,14 +476,16 @@ mod content {
         where
             D: Deserializer<'de>,
         {
-            Deserialize::deserialize(deserializer).map(|v| Content::Some(Box::new(v)))
+            let v = tri!(Deserialize::deserialize(deserializer));
+            Ok(Content::Some(Box::new(v)))
         }
 
         fn visit_newtype_struct<D>(self, deserializer: D) -> Result<Self::Value, D::Error>
         where
             D: Deserializer<'de>,
         {
-            Deserialize::deserialize(deserializer).map(|v| Content::Newtype(Box::new(v)))
+            let v = tri!(Deserialize::deserialize(deserializer));
+            Ok(Content::Newtype(Box::new(v)))
         }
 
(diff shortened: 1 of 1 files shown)


serde/src/private/de.rs
```diff
@@ -1113,9 +1115,13 @@ mod content {
...
-        let map = content
-            .into_iter()
-            .map(|(k, v)| (ContentDeserializer::new(k), ContentDeserializer::new(v)));
+        fn content_deserializer_pair<'de, E>(
+            (k, v): (Content<'de>, Content<'de>),
+        ) -> (ContentDeserializer<'de, E>, ContentDeserializer<'de, E>) {
+            (ContentDeserializer::new(k), ContentDeserializer::new(v))
+        }
+
+        let map = content.into_iter().map(content_deserializer_pair);
```
Mingun, line 1124: You also can just pass `content.into_iter()` because `Content` implements `IntoDeserializer` which is used in `MapAccess` implementation:
https://github.com/serde-rs/serde/blob/c82f2580b43c5d44f557bf530554081e6fed646d/serde/src/de/value.rs#L1200-L1207

The same implicit conversion can be applied to use of `SeqDeserializer` and, if implement `IntoDeserializer for &Content`, for `ContentRefDeserializer`.

Maybe this also could reduce generated size?

PR #2822, opened 2024-09-07
Remove explicit ContentDeserializer construction in visit functions

dtolnay:
This is a good suggestion from https://github.com/serde-rs/serde/pull/2821#pullrequestreview-2288334462. I was not able to measure any binary size improvement &mdash; the rlib is 111K before and after. But it's worth making the change.

### DIFF
--- a/serde/src/private/de.rs
+++ b/serde/src/private/de.rs
@@ -313,6 +313,17 @@ mod content {
         }
     }
 
+    impl<'a, 'de, E> de::IntoDeserializer<'de, E> for &'a Content<'de>
+    where
+        E: de::Error,
+    {
+        type Deserializer = ContentRefDeserializer<'a, 'de, E>;
+
+        fn into_deserializer(self) -> Self::Deserializer {
+            ContentRefDeserializer::new(self)
+        }
+    }
+
     /// Used to capture data in [`Content`] from other deserializers.
     /// Cannot capture externally tagged enums, `i128` and `u128`.
     struct ContentVisitor<'de> {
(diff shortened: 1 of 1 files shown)


serde/src/private/de.rs
```diff
@@ -1715,16 +1717,10 @@ mod content {
         V: Visitor<'de>,
         E: de::Error,
     {
-        fn content_ref_deserializer_pair<'a, 'de, E>(
+        fn content_ref_deserializer_pair<'a, 'de>(
```
dtolnay, line 1720: I also tried adding:

```rust
impl<'a, A, B> Pair for &'a (A, B) {
    type First = &'a A;
    type Second = &'a B;
    fn split(self) -> (&'a A, &'a B) {
        (&self.0, &self.1)
    }
}
```

to make this function go away, like in the `visit_content_map` case. See #2823.

But somehow that made the rlib go back up from 111K to 124K, negating the benefit from #2821. I do not fully understand why this happens.

Mingun:
If you are worried about the size of the generated code, then maybe merge
- #2564 (besides, it reduces compile time)
- #2811 (besides, it fixes one bug and corrects areas of responsibility of `Content[Ref]Deserializer`). This one removes a lot of generic code, so should expect reducing size, especially when flattening or internally or adjacently tagged enums is used

I rebased both PRs on current master for easing of testing.

Mingun:
Some tests on my computer (the same test library project as in #2821):
|Commit|Size (bytes)|Diff from master (bytes)
|------------|----------------|------------------
|[master](https://github.com/serde-rs/serde/commit/4487cb26c7f86210c25aeb9d04add9a2e51ad74d)|143694|
|[visitor-driven](https://github.com/Mingun/serde/commit/4976c6498cc561708f4f5650c8fab6fafde4667b)|152842|+9148
|[common-field](https://github.com/Mingun/serde/commit/d4aad552cab3e64735605d28ac3c2b99c863f341)|136930|-6764
|both|142462|-1232

PR #2837, opened 2024-10-19
Make compile error message point to field type instead of call site for "not implemented Serialize trait"

Veetaha:
## Before

The error points to the call site of the `Serialize` derive:

## After

The error points to the type of the field directly, which makes it easier to find where exactly in the struct the type that doesn't implement `Serialize` is located.

---

The trick is to use the type of the field in the type hint for the `serialize_field/entry` method call. Previously that call used a `quote_spanned` inheriting the span of the field itself (I'm not sure that actually did anything). Now, instead of that we pass `serialize_field::<#ty>`, and thus the compiler knows to point to `#ty` in the error message.

Interestingly, the code generated by `Deserialize` already follows this pattern and thus doesn't have this problem. So this only fixes `Serialize` codegen.


serde_derive/src/ser.rs
```diff
@@ -1131,7 +1136,13 @@ fn serialize_struct_visitor(
                     #func(&#field_expr, _serde::__private::ser::FlatMapSerializer(&mut __serde_state))?;
                 }
             } else {
-                let func = struct_trait.serialize_field(span);
+                let ty = if field.attrs.serialize_with().is_none() {
```
oli-obk, line 1139: how does this affect errors when using `serialize_with`?

Veetaha: I've checked the difference in error messages locally with `serde(with)` and it doesn't change (e.g. when the signature of the given function is wrong). I can add a ui snapshot test for that today in the evening.

Also, the error for `serde(with)` still points to the macro call site (not to the function in the `serde(with)` attribute). That can also be improved by wrapping the function itself in a `::core::conver::identity<fn(field_types) -> __S>` which gives the type hint for the function signature and the compiler starts complaing that the function is wrong, so it points the error to the function directly. However, this unfortunatelly casts the function to a function pointer...

Mingun: `with` attributes fixed by me in #2558

Veetaha: I wonder what if I attache the span of the type itself to the parameter of `serialize_field/entry`... Maybe it'll yield the same result. I'll experiment with this later today. 

Otherwise I suppose #2558 should land first, because it looks like it may conflict with this PR?

Mingun: Not sure. I'm ready to rebase all my PRs once their fall into conflict state

Veetaha: I've checked how it works. The similar behavior can be achieved if even with the expression `serialize_field::<_>` is used where the span of `_` is assigned to the span of the field type.

Veetaha: When there is no turbofish syntax, it's important that all parts of the field expression have the span of field, because only then will the compiler point to the original field.

```rust
error[E0277]: the trait bound `NoImpls: _::_serde::Serialize` is not satisfied
    --> crates/sandbox/src/main.rs:34:85
     |
34   |             _serde::ser::SerializeStruct::serialize_field(&mut __serde_state, "x2", &self.x2)?;
     |             ---------------------------------------------                           ^^^^^^^^ the trait `_::_serde::Serialize` is not implemented for `NoImpls`
     |             |
     |             required by a bound introduced by this call
```

For example, if all tokens `&`, `self`, `.` and `x2` have the span of the field, you'll get an error pointing to the field in this case:

```rust
error[E0277]: the trait bound `NoImpls: _::_serde::Serialize` is not satisfied
    --> crates/sandbox/src/main.rs:34:85
     |
34   |             _serde::ser::SerializeStruct::serialize_field(&mut __serde_state, "x2", &self.x2)?;
     |             ---------------------------------------------                           ^^^^^^^^ the trait `_::_serde::Serialize` is not implemented for `NoImpls`
     |             |
     |             required by a bound introduced by this call
```

Alothough.. it would be better if the error pointed to the type itself. What makes this harder to achieve is that `quote_spanned` doesn't change the span of the interpolated tokens. So for example, today the `self` token in the field expression uses the callsite span when it's interpolated:

https://github.com/Veetaha/serde/blob/4b3178b053683b8e13f9f551e2a40fbe4a927e43/serde_derive/src/ser.rs#L100-L104

But instead each usage of this token should change the span to the relevant field.

Veetaha:
I've updated the impl a bit and moved the logic if skipping the type hint for `serialize_with` into a function with some comments explaining how it helps with hinting the correct error span.

I've also noticed that if the user passes invalid function to `skip_serializing_if`, the error also points to the callsite of the derive.

To fight with that we need to re-span the field expression to assign the span of the function in `skip_serializing_if` to all the tokens in  `#field_expr`. Right now this code looks like this:

https://github.com/serde-rs/serde/blob/3415619bfdf7867f40f3e30e9224f884c03450e8/serde_derive/src/ser.rs#L1075-L1078

However, just using `quote_spanned` here won't help. We need to re-span the `#field_expr` for this to work.

PR #2841, opened 2024-10-22
Reduce scope of quote_spanned on SerializeWith wrapper

dtolnay:
This leaves more of the generated code accurately identified as generated, avoiding the `dead_code` and `needless_lifetimes` warnings that originated from #2558.

### DIFF
--- a/serde_derive/src/dummy.rs
+++ b/serde_derive/src/dummy.rs
@@ -14,12 +14,7 @@ pub fn wrap_in_const(serde_path: Option<&syn::Path>, code: TokenStream) -> Token
 
     quote! {
         #[doc(hidden)]
-        #[allow(
-            clippy::needless_lifetimes,
-            non_upper_case_globals,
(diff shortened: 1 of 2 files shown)


Mingun:
@dtolnay, for better understanding, could you explain, how this change fixes the warnings? rustc considers `quote!` code blocks as generated and `quote_spanned!` as not generated?

dtolnay:
Tokens with a call_site span are understood as being macro-generated. It is not a difference between quote vs quote_spanned. Quote produces call_site spanned tokens, but if you passed a span to quote_spanned that happened to be call_site, then those tokens also would be considered generated despite coming from a quote_spanned expansion.

More precisely, the span's "resolution" being call_site is the thing that matters, as opposed to the span's "location" being call_site. So `user_span.resolved_at(Span::call_site())` is generated while `user_span.located_at(Span::call_site())` is not generated.

PR #2847, opened 2024-10-22
Hygiene for macro-generated newtype struct deserialization with `with` attr

### LINKED ISSUE #2846: Serde still causing errors even after 1.0.212
Serde 1.0.211 broke some call sites described in #2844 and later fixed in #2845

I believe it is still broken. The number of errors is reduced, but I am still seeing one:

```
...
  Downloaded serde_derive v1.0.212
...
error[E0425]: cannot find value `__e` in this scope
   --> components/datetime/src/provider/calendar/symbols.rs:336:36
    |
336 |                 deserialize_with = "icu_provider::serde_borrow_de_utils::array_of_cow"
    |                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ not found in this scope
```

CC @robertbastian

dtolnay:
Fixes #2846.

### DIFF
--- a/serde_derive/src/de.rs
+++ b/serde_derive/src/de.rs
@@ -875,13 +875,14 @@ fn deserialize_newtype_struct(
 ) -> TokenStream {
     let delife = params.borrowed.de_lifetime();
     let field_ty = field.ty;
+    let deserializer_var = quote!(__e);
 
     let value = match field.attrs.deserialize_with() {
         None => {
             let span = field.original.span();
             let func = quote_spanned!(span=> <#field_ty as _serde::Deserialize>::deserialize);
             quote! {
-                #func(__e)?
+                #func(#deserializer_var)?
             }
         }
         Some(path) => {
@@ -890,7 +891,7 @@ fn deserialize_newtype_struct(
             // on the #[serde(with = "...")]
             //                       ^^^^^
             quote_spanned! {path.span()=>
-                #path(__e)?
+                #path(#deserializer_var)?
             }
         }
     };
@@ -906,7 +907,7 @@ fn deserialize_newtype_struct(
 
     quote! {
         #[inline]
-        fn visit_newtype_struct<__E>(self, __e: __E) -> _serde::__private::Result<Self::Value, __E::Error>
+        fn visit_newtype_struct<__E>(self, #deserializer_var: __E) -> _serde::__private::Result<Self::Value, __E::Error>
         where
             __E: _serde::Deserializer<#delife>,
         {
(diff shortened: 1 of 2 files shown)


oli-obk:
This seems very footgunny, do you know if there is an issue or discussion about making this less painful? It feels to me like a wrong-defaults issue. I'd expect it to always use `.resolved_at(Span::call_site())` unless opted out of.

That said, wouldn't it be cleaner to do that manually here and thus also avoid clippy issues?

dtolnay:
`.resolved_at(Span::call_site())` is almost never a good idea because various code in rustc and clippy looks at "resolved_at == call_site" to decide whether to disregard the "located_at" of a token when reporting an error or suggestion involving it. For example, on top of https://github.com/serde-rs/serde/commit/04bb76bc00354866b5a1ced7c4a411d81632e33b, this code change:

```diff
-    quote_spanned!(serialize_with.span()=> {
+    quote_spanned!(serialize_with.span().resolved_at(Span::call_site())=> {
         #[doc(hidden)]
         struct __SerializeWith #wrapper_impl_generics #where_clause {
```

has these unwanted effects, among others:

```diff
 error[E0277]: the trait bound `&u8: Serializer` is not satisfied
   --> tests/ui/with/incorrect_type.rs:14:10
    |
-14 | #[derive(Serialize, Deserialize)]
-   |          ^^^^^^^^^ the trait `Serializer` is not implemented for `&u8`
-15 | struct W(#[serde(with = "w")] u8, u8);
-   |                         --- required by a bound introduced by this call
+14 |   #[derive(Serialize, Deserialize)]
+   |            ^--------
+   |            |
+   |  __________in this derive macro expansion
+   | |
+15 | | struct W(#[serde(with = "w")] u8, u8);
+   | |                         --^
+   | |_________________________|_|
+   |                           | the trait `Serializer` is not implemented for `&u8`
+   |                           required by a bound introduced by this call
    |
    = help: the following other types implement trait `Serializer`:
              &mut Formatter<'a>
```

```diff
 error[E0061]: this function takes 1 argument but 2 arguments were supplied
   --> tests/ui/with/incorrect_type.rs:15:25
    |
 15 | struct W(#[serde(with = "w")] u8, u8);
    |                         ^^^ unexpected argument #2 of type `__S`
    |
 note: function defined here
   --> tests/ui/with/incorrect_type.rs:9:12
    |
 9  |     pub fn serialize<T, S: Serializer>(_: S) -> Result<S::Ok, S::Error> {
    |            ^^^^^^^^^                   ----
+help: remove the extra argument
+   |
+15 - struct W(#[serde(with = "w")] u8, u8);
+15 + struct W(#[serde(with = )] u8, u8);
+   |
```

oli-obk:
Hmm... if we can make the world of proc macros better by stopping to do that, that seems great. I was working on some span stuff just a few months ago. I'll see if we can remove the PartialEq impl from spans entirely, it's already a footgun just in the compiler, seems best to remove it if its "benign" uses are problematic, for proc macros

PR #2849, opened 2024-10-25
Added default implementations for 'Serializer' and 'Deserializer' trait by raising error if not overridden

AlexSherbinin:


oli-obk:
Can you talk a bit about your motivation for this change? Ideally with some links to examples that are hard, annoying or footgunny without this?

AlexSherbinin:
1. Default method implementations already exist for the `u128` and `i128` types.
2. There is a significant boilerplate involved with using the `Impossible` serializer in `SerializeTuple`, `SerializeSeq`, and similar cases.

I created this PR while transitioning [this crypto library](https://github.com/RustCrypto/password-hashes) from an internal trait for hash function parameter serialization to Serde, and encountered the aforementioned issue with boilerplate.

oli-obk:
> 1. Default method implementations already exist for the `u128` and `i128` types.

That's only due to them being introduced later, so we can't add them without a breaking change or a default impl



> 2\. There is a significant boilerplate involved with using the `Impossible` serializer in `SerializeTuple`, `SerializeSeq`, and similar cases.

These are rather special cases. Most serializers will want to implement all of them, and with default impls it's annoying to make sure you got them all

AlexSherbinin:
> These are rather special cases. Most serializers will want to implement all of them, and with default impls it's annoying to make sure you got them all

Rust analyzer has code action to implement all trait members, even with default impl.

AlexSherbinin:
Will it be merged/closed?

oli-obk:
Let's close it. The few folk who need to write (de)serializers that want to error on most types can also use r-a to generate all methods and then search-and-replace the `todo!()` with an error

PR #2866, opened 2024-12-09
Mark generated `impl de::Visitor` blocks as `#[automatically_derived]`

tdittr:
When using `cargo llvm-cov` on one of my projects I noticed that the `#[derive(Deserialize)]` annotations on my structs were marked as not covered.

This PR adds `#[automatically_derived]` also to the `impl de::Visitor` blocks that are generated. In my tests, they now no longer show up in the coverage as not covered.

`rustc` automatically excludes `impl` blocks marked with `#[automatically_derived]` since https://github.com/rust-lang/rust/pull/120185. 
The generated `impl Deserialize` block is already marked with `#[automatically_derived]` and thus correctly not counted.
However, the visitors and field visitors which are generated within the `deserialize` method were counted towards the coverage.

### DIFF
--- a/serde_derive/src/de.rs
+++ b/serde_derive/src/de.rs
@@ -424,6 +424,7 @@ fn deserialize_unit_struct(params: &Parameters, cattrs: &attr::Container) -> Fra
             lifetime: _serde::__private::PhantomData<&#delife ()>,
         }
 
+        #[automatically_derived]
         impl #de_impl_generics _serde::de::Visitor<#delife> for __Visitor #de_ty_generics #where_clause {
             type Value = #this_type #ty_generics;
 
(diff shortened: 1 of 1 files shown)


oli-obk (approved):
Thanks! That makes sense

dtolnay:
> The generated `impl Deserialize` block is already marked with `#[automatically_derived]` and thus correctly not counted.

There were some Deserialize and other impls (DeserializeSeed, Serialize, remote deserialize/serialize) that were not marked. I have opened https://github.com/serde-rs/serde/pull/2868 to add `#[automatically_derived]` to them all. Please let me know if there were any of those impls intentionally omitted from this PR and should not get `#[automatically_derived]`. I am not familiar with using llvm-cov in my own projects.

tdittr:
Oh interesting, I guess I just missed those. Thanks for also adding the annotation there!

PR #2879, opened 2025-01-08
add `#[allow(deprecated)]` to derive implementations

### LINKED ISSUE #2195: `#[derive(Serialize)` and `#[derive(Deserialize)]` on deprecated types emit deprecation warnings
Given this code ([playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=2b1fded66fee55a3a02d0890683e9fad)):

```rust
#[deprecated]
#[derive(Default, serde::Serialize)]
struct WithSerialize;

#[deprecated]
#[derive(Default, serde::Deserialize)]
struct WithDeserialize;

#[deprecated]
#[derive(Default)]
struct WithoutSerde;

fn main() {
}
```

I expect not to get deprecation warnings. Instead, I get the following:

```
warning: use of deprecated struct `WithSerialize`
 [--> src/main.rs:3:8
](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=22d2042e09ce11dddfcbac0464dd4fe3#)  |
3 | struct WithSerialize;
  |        ^^^^^^^^^^^^^
  |
  = note: `#[warn(deprecated)]` on by default

warning: use of deprecated struct `WithDeserialize`
 [--> src/main.rs:7:8
](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=22d2042e09ce11dddfcbac0464dd4fe3#)  |
7 | struct WithDeserialize;
  |        ^^^^^^^^^^^^^^^

warning: use of deprecated unit struct `WithDeserialize`
 [--> src/main.rs:7:8
](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=22d2042e09ce11dddfcbac0464dd4fe3#)  |
7 | struct WithDeserialize;
  |        ^^^^^^^^^^^^^^^
```

Notice that the derives for `Default` don't trigger any warnings (this seems to have been [fixed](https://github.com/rust-lang/rust/issues/58822) some time ago).

Not aware of any workaround (other than undeprecating), and this leads to a bunch of warnings for things that are deprecated but that we still want to be able to support serialization for.

rcrisanti:
Allow deprecated in the `Serialize`/`Deserialize` derive implementations. This allows you to deprecate structs, enums, struct fields, or enum variants and not get compiler warnings/errors about use of deprecated thing.

Resolves #2195

### DIFF
--- a/serde_derive/src/de.rs
+++ b/serde_derive/src/de.rs
@@ -34,6 +34,7 @@ pub fn expand_derive_deserialize(input: &mut syn::DeriveInput) -> syn::Result<To
         let used = pretend::pretend_used(&cont, params.is_packed);
         quote! {
             #[automatically_derived]
+            #[allow(deprecated)]
             impl #de_impl_generics #ident #ty_generics #where_clause {
                 #vis fn deserialize<__D>(__deserializer: __D) -> #serde::__private::Result<#remote #ty_generics, __D::Error>
                 where
@@ -49,6 +50,7 @@ pub fn expand_derive_deserialize(input: &mut syn::DeriveInput) -> syn::Result<To
 
         quote! {
             #[automatically_derived]
+            #[allow(deprecated)]
             impl #de_impl_generics #serde::Deserialize<#delife> for #ident #ty_generics #where_clause {
                 fn deserialize<__D>(__deserializer: __D) -> #serde::__private::Result<Self, __D::Error>
                 where
--- a/serde_derive/src/ser.rs
+++ b/serde_derive/src/ser.rs
@@ -30,6 +30,7 @@ pub fn expand_derive_serialize(input: &mut syn::DeriveInput) -> syn::Result<Toke
         let used = pretend::pretend_used(&cont, params.is_packed);
         quote! {
             #[automatically_derived]
+            #[allow(deprecated)]
             impl #impl_generics #ident #ty_generics #where_clause {
                 #vis fn serialize<__S>(__self: &#remote #ty_generics, __serializer: __S) -> #serde::__private::Result<__S::Ok, __S::Error>
                 where
@@ -43,6 +44,7 @@ pub fn expand_derive_serialize(input: &mut syn::DeriveInput) -> syn::Result<Toke
     } else {
         quote! {
             #[automatically_derived]
+            #[allow(deprecated)]
             impl #impl_generics #serde::Serialize for #ident #ty_generics #where_clause {
                 fn serialize<__S>(&self, __serializer: __S) -> #serde::__private::Result<__S::Ok, __S::Error>
                 where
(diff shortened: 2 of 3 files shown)


KarstenB:
I just encountered the same issue and was wondering how that could be implemented. While I am perfectly fine with the given PR, I thought about a more generic way. Like:
```rust
#[deprecated]
#[serde(allow = deprecated)]
struct bla {}
```

rcrisanti:
@KarstenB that is definitely doable if that is the desired behavior. Personally I'm not sure I can think of a scenario where I *do* want to see these messages, so I'd at least be in favor of making the default not showing them, with some sort of option to turn them back on.

(For some justification of this, [in rust itself they hide deprecation messages in derives](https://github.com/rust-lang/rust/pull/58994), so in my opinion this would put serde more inline with expected behavior.)

oli-obk:
> (For some justification of this, [in rust itself they hide deprecation messages in derives](https://github.com/rust-lang/rust/pull/58994), so in my opinion this would put serde more inline with expected behavior.)

in theory that check should work for serde automatically, but making that happen at the proc macro level causes lots of problems inside other diagnostics: https://github.com/serde-rs/serde/pull/2847#issuecomment-2430364534

oli-obk:
No... this is mostly me not wanting to add workarounds that are necessary because of other workarounds. w

oli-obk:
Considering the low cost to maintenance this PR adds, gonna merge it as is

dtolnay (changes requested):
I don't like this workaround because it silently suppresses legitimate deprecations on parts of the generated impl that come from user-specified attributes. Example:

```rust
use serde::Deserializer;

#[derive(Deserialize)]
pub struct Struct {
    #[serde(deserialize_with = "deprecated_with")]
    pub field: i32,
}

#[deprecated]
fn deprecated_with<'de, D>(_deserializer: D) -> Result<i32, D::Error>
where
    D: Deserializer<'de>,
{
    unimplemented!()
}
```

```console
warning: use of deprecated function `deprecated_with`
 --> src/main.rs:6:32
  |
6 |     #[serde(deserialize_with = "deprecated_with")]
  |                                ^^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(deprecated)]` on by default
```

Please put `allow(deprecated)` if and only if the input data structure has either `#[deprecated]` or `#[allow(deprecated)]`.

dtolnay (approved):
Thank you!

dtolnay:
Published in serde_derive 1.0.225.

serde_derive/src/internals/deprecated.rs
```diff
@@ -0,0 +1,58 @@
...
+        Ok(quote! { #[allow(deprecated)]})
+    } else {
+        Ok(TokenStream::default())
+    }
+}
+
+/// Determine if an `#[allow(deprecated)]` should be added to the derived impl.
+///
+/// This should happen if the derive input or a variant of the enum (if derive input is an enum)
+/// has on of:
```
Celeo, line 15: Should this be "has **one** of"?

Issue #642, opened 2016-12-17
Add finalizer attribute hook to validate a deserialized structure


erickt:
In https://github.com/servo/rust-url/pull/252, @Manishearth is writing a serde deserializer for urls since parsing a string can be expensive. However, to do this safely from untrusted sources, they need to validate that the data is semantically correct. We unfortunately don't have an easy way of doing this without writing a custom deserializer. It'd be pretty simple to do as an attribute, however. Here's my idea:

```rust
#[derive(Serialize, Deserialize)]
#[serde(finalizer="check_greater_than_10")]
struct GreaterThan10 {
    n: usize,
}

fn greater_than_10<D>(item: GreaterThan10) -> Result<GreaterThan10, D::Error>
    where D: Deserializer
{
    if item.n > 10 {
        Ok(item)
    } else {
        Err(Error::invalid_value("less than or equal to 10"))
    }
}

dtolnay:
This seems relevant: https://github.com/Keats/validator. Let's make sure to support this library in a nice way.

radix:
My hacky workaround is to

1. don't derive Deserialize for my main data type "DataType"
2. copy my main data type to a "FakeDataType" which has the same fields, and make this one derive Deserialize
3. implement Deserialize for DataType by:
   1. invoking deserialize to FakeDataType
   2. validate the result
   3. convert FakeDataType to DataType.

It's a dumb hack but at least it allows me to avoid rewriting all the deser code manually.

dtolnay:
#626 has another use case.

I am on board with implementing exactly the approach in @erickt's comment.

oli-obk:
Another use-case from irc by @alteous

> I would like to obtain a maximum integer value from a tree of structures. The context is these values are offsets into arrays, and I want to check that they are all valid ahead of time. The integer values are wrapped in Index<T> types.

beamspease:
I believe I could also use this for an enum which is either "unknown" or some value fitting a specific pattern, but both are strings.

Mange:
I have the same usecase as in #626. I want to load shapes, and I also want to store some precomputed attributes on them that are derived from the other fields (for example `radius_squared = radius * radius`).

I don't want to serialize or deserialize these fields, but I want them to be present in the finished structs. The shapes could be deeply nested in another structure that is loaded, so it's very messy to try to go down the structures in a mutable fashion and set it on the structs manually.

scottmcm:
[...]

PR #2891, opened 2025-02-11
Support `#[serde(validate = "function")]` for container, compatible with `validator` crate

zao111222333:
This PR aims to fix https://github.com/serde-rs/serde/issues/939, https://github.com/serde-rs/serde/issues/642

The basic idea is only specify validate function in container attribute, since it is the most general way.
One can directly use `validator::Validate::validate` as validation function, so it is not necessary to have field validator in `serde`.
The `Error` needs to be `impl Display`, to convert it by `serde::de::Error::custom`.

For general usage, one can use arbitrary `fn (&T) -> Result<(), impl Display>` function to validate.
``` rust
#[derive(Deserialize)]
#[serde(validate = "validate_struct")]
struct Struct {
    a: u16,
}

fn validate_struct(deserialized: &Struct) -> Result<(), impl Display> {
    if deserialized.a == 0 {
        return Err("field `a` can not be zero");
    }
    Ok(())
}
```

And for `validator` user, there is no extra overhead to do validate.
``` rust
#[derive(validator::Validate, Deserialize)]
#[serde(validate = "validator::Validate::validate")]
struct ValidatorDemo {
    #[validate(email)]
    mail: String,
}
```

The full unit-test & demo see [here](
https://github.com/zao111222333/serde/blob/780b391f2ebdc78de7810a4a998667f031428108/test_suite/tests/test_validate.rs).
And I am glad to write documents if this idea looks good to you :)

### DIFF
--- a/serde_derive/src/de.rs
+++ b/serde_derive/src/de.rs
@@ -26,6 +26,7 @@ pub fn expand_derive_deserialize(input: &mut syn::DeriveInput) -> syn::Result<To
     let params = Parameters::new(&cont);
     let (de_impl_generics, _, ty_generics, where_clause) = split_with_de_lifetime(&params);
     let body = Stmts(deserialize_body(&cont, &params));
+    let vaildated_body = validate_body(&cont, body);
     let delife = params.borrowed.de_lifetime();
     let serde = cont.attrs.serde_path();
 
@@ -40,7 +41,7 @@ pub fn expand_derive_deserialize(input: &mut syn::DeriveInput) -> syn::Result<To
                     __D: #serde::Deserializer<#delife>,
                 {
                     #used
-                    #body
+                    #vaildated_body
                 }
             }
         }
@@ -54,7 +55,7 @@ pub fn expand_derive_deserialize(input: &mut syn::DeriveInput) -> syn::Result<To
                 where
                     __D: #serde::Deserializer<#delife>,
                 {
-                    #body
+                    #vaildated_body
                 }
 
                 #fn_deserialize_in_place
@@ -270,6 +271,18 @@ fn borrowed_lifetimes(cont: &Container) -> BorrowedLifetimes {
         BorrowedLifetimes::Static
     } else {
         BorrowedLifetimes::Borrowed(lifetimes)
+    }
+}
+
+fn validate_body(cont: &Container, body: Stmts) -> TokenStream {
+    if let Some(vaildate) = cont.attrs.vaildate() {
+        quote! {
+            let __body = { #body }?;
+            #vaildate(&__body).map_err(_serde::de::Error::custom)?;
+            Ok(__body)
+        }
+    } else {
+        quote! { #body }
     }
 }
 
--- a/serde_derive/src/internals/attr.rs
+++ b/serde_derive/src/internals/attr.rs
@@ -169,6 +169,7 @@ pub struct Container {
     remote: Option<syn::Path>,
     identifier: Identifier,
     serde_path: Option<syn::Path>,
+    vaildate: Option<syn::ExprPath>,
     is_packed: bool,
     /// Error message generated when type can't be deserialized
     expecting: Option<String>,
@@ -256,6 +257,7 @@ impl Container {
         let mut remote = Attr::none(cx, REMOTE);
         let mut field_identifier = BoolAttr::none(cx, FIELD_IDENTIFIER);
         let mut variant_identifier = BoolAttr::none(cx, VARIANT_IDENTIFIER);
+        let mut vaildate = Attr::none(cx, VALIDATE);
         let mut serde_path = Attr::none(cx, CRATE);
         let mut expecting = Attr::none(cx, EXPECTING);
         let mut non_exhaustive = false;
@@ -486,6 +488,11 @@ impl Container {
                     if let Some(path) = parse_lit_into_path(cx, CRATE, &meta)? {
                         serde_path.set(&meta.path, path);
                     }
+                } else if meta.path == VALIDATE {
+                    // #[serde(validate = "...")]
+                    if let Some(path) = parse_lit_into_expr_path(cx, VALIDATE, &meta)? {
+                        vaildate.set(&meta.path, path);
+                    }
                 } else if meta.path == EXPECTING {
                     // #[serde(expecting = "a message")]
                     if let Some(s) = get_lit_str(cx, EXPECTING, &meta)? {
@@ -539,6 +546,7 @@ impl Container {
             remote: remote.get(),
             identifier: decide_identifier(cx, item, field_identifier, variant_identifier),
             serde_path: serde_path.get(),
+            vaildate: vaildate.get(),
             is_packed,
             expecting: expecting.get(),
             non_exhaustive,
@@ -595,6 +603,10 @@ impl Container {
 
     pub fn remote(&self) -> Option<&syn::Path> {
         self.remote.as_ref()
+    }
+
+    pub fn vaildate(&self) -> Option<&syn::ExprPath> {
+        self.vaildate.as_ref()
     }
 
     pub fn is_packed(&self) -> bool {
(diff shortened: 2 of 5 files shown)


serde_derive/src/de.rs
```diff
@@ -273,6 +274,18 @@ fn borrowed_lifetimes(cont: &Container) -> BorrowedLifetimes {
     }
 }
 
+fn validate_body(cont: &Container, body: Stmts) -> TokenStream {
+    if let Some(vaildate) = cont.attrs.vaildate() {
+        quote! {
+            let body = { #body }?;
+            #vaildate(&body).map_err(_serde::de::Error::custom)?;
```
Mingun, line 281: Misprint
```suggestion
    if let Some(validate) = cont.attrs.validate() {
        quote! {
            let body = { #body }?;
            #validate(&body).map_err(_serde::de::Error::custom)?;
```

zao111222333:
And I notice that the `fn (&T) -> Result<(), impl Display>` way is also compatible with https://github.com/ohkami-rs/serdev

serde_derive/src/de.rs
```diff
@@ -273,6 +274,18 @@ fn borrowed_lifetimes(cont: &Container) -> BorrowedLifetimes {
     }
 }
 
+fn validate_body(cont: &Container, body: Stmts) -> TokenStream {
+    if let Some(validate) = cont.attrs.validate() {
+        quote! {
+            let body = { #body }?;
+            #validate(&body).map_err(_serde::de::Error::custom)?;
```
oli-obk, line 281:
```suggestion
            #validate(&body).map_err(#serde::de::Error::custom)?;
```

this should be using the `cont.attrs.serde_path()`

oli-obk (changes requested):
I'll need to read up on the issue discussions and the extent to which we may want to simplify `validator` crate support

serde_derive/src/de.rs
zao111222333: thx! fixed

zao111222333:
> I'll need to read up on the issue discussions and the extent to which we may want to simplify `validator` crate support

One way to simplify `validator` crate support is a feature-controlled re-export `serde::_validator`, and add `#[serde(validator)]` arrtibute, so that
``` rust
#[derive(Deserialize)]
#[serde(validator)]
struct ValidatorDemo {
    #[validate(email)]
    mail: String,
}
```
will be equivalent to
``` rust
#[derive(serde::_validator::Validate, Deserialize)]
#[serde(validate = "serde::_validator::Validate::validate")]
struct ValidatorDemo {
    #[validate(email)]
    mail: String,
}
```

How do you think that?

zao111222333:
> I'll need to read up on the issue discussions and the extent to which we may want to simplify `validator` crate support

The most user friendly API should like that
``` rust
#[derive(Deserialize)]
#[serde(validator)]
struct ValidatorDemo {
    #[validate(email)]
    mail: String,
}
```
However it needs to expand into follows, assume the `validator` is a block box to us, and we only wrap it up
``` rust
#[derive(validator::Validate, Deserialize)]
#[serde(validate = "validator::Validate::validate")]
struct ValidatorDemo {
    #[validate(email)]
    mail: String,
}
```
And I think it is impossible to make the Deserialize `proc_macro_derive` to introduce another `proc_macro_derive` for the same container ...

zao111222333:
Great, I will do that

test_suite/tests/ui/validate/both_validator_and_validate.rs
```diff
@@ -0,0 +1,11 @@
+use serde_derive::Deserialize;
+
+#[derive(validator::Validate, Deserialize)]
+#[serde(validate = "validator::Validate::validate")]
+#[serde(validator)]
```
oli-obk, line 5: it's not really wrong to have two validation schemes running at the same time. `serde(validate = "..")` could be specified twice, too (add a test?)

zao111222333: You mean the later validate/validator will override the previous one?
I got it, I will do that

zao111222333: You mean user can define a list of validation schemes?
I think that's a good idea.

zao111222333:
Now there can be multiple validations, I add two test cases
1. Multiple `serde(validate = "...")`, see [here](https://github.com/zao111222333/serde/blob/e611d5f12555ed50dfa4b6c605e016f2d3126f08/test_suite/tests/test_validate.rs#L5-L10)
2. Both `serde(validate = "...")` and `serde(validator)`, see [here](https://github.com/zao111222333/serde/blob/e611d5f12555ed50dfa4b6c605e016f2d3126f08/test_suite/tests/test_validate.rs#L201-L207)

Mingun:
What the advantage of using `#[serde(validator)]` over `#[serde(validate = "...")]`? It seems to me that this is unnecessary introduction of implicit dependency.

Because Rust already supports using paths in attributes, why you choosed using string for `validate` values instead of a path?

serde_derive/src/de.rs
```diff
@@ -26,6 +26,7 @@ pub fn expand_derive_deserialize(input: &mut syn::DeriveInput) -> syn::Result<To
     let params = Parameters::new(&cont);
     let (de_impl_generics, _, ty_generics, where_clause) = split_with_de_lifetime(&params);
     let body = Stmts(deserialize_body(&cont, &params));
+    let vaildated_body = validate_body(&cont, body);
```
Mingun, line 29: One more misprint
```suggestion
    let validated_body = validate_body(&cont, body);
```

serde_derive/src/de.rs
```diff
@@ -273,6 +274,36 @@ fn borrowed_lifetimes(cont: &Container) -> BorrowedLifetimes {
...
 
+fn validate_body(cont: &Container, body: Stmts) -> TokenStream {
+    let serde = cont.attrs.serde_path();
+    let mut validations: Vec<TokenStream> = cont
+        .attrs
+        .validate()
+        .iter()
+        .map(|validate| {
+            quote! {
+                #validate(&body).map_err(#serde::de::Error::custom)?;
```
Mingun, line 285: Probably it is better to use `From` trait here and below?

zao111222333:
Hi @Mingun, thank you for your review.
> What the advantage of using `#[serde(validator)]` over `#[serde(validate = "...")]`? It seems to me that this is unnecessary introduction of implicit dependency.

`#[serde(validator)]` is the short of `#[serde(validate = "validator::Validate::validate")]`
`#[serde(validate = "...")]` is the most general way, and I am open for whether or not use `#[serde(validator)]`

> Because Rust already supports using paths in attributes, why you choosed using string for `validate` values instead of a path?

Considering the user interface consistency, I just reuse the internal APIs for `serialize_with` and `deserialize_with`.
If you think we can use a path to do `#[serde(validate = ...)]`, I can change to it.

serde_derive/src/de.rs
zao111222333: You mean the signature validation should be `fn <E: serde::de::Error>(&T) -> Result<(), impl Into<E>>`?

In this case, if we want `#[serde(validate = "validator::Validate::validate")]` to use `validator`, the only way is to add follow code within `serde`, since Rust only allow local impl for `trait E`
``` rust
impl<E: serde::de::Error> From<validator::ValidationErrors> for E { }
```
It will introduce a new dependence for `serde`, I don't know if we can do that

serde_derive/src/de.rs
zao111222333: Sorry! I will fix that

zao111222333:
Hi, I am wondering is there any suggestions for this PR? So that I can modify under that.
Take your time and thank you!

serde_derive/src/de.rs
oli-obk: Yea I think only adding that "dependency" in serde_derive (essentially requiring anyone who uses `#[serde(validator)]` to have a `validator` dependency is fine, but only since it seems like @dtolnay suggested something like that in the linked issue 😆

oli-obk (approved):

PR #2894, opened 2025-02-20
Convert html links to intra-doc links

dtolnay:


serde/src/ser/mod.rs
```diff
@@ -1805,9 +1804,9 @@ pub trait SerializeMap {
     /// care about performance or are not able to optimize `serialize_entry` any
     /// better than this.
     ///
-    /// [`Serialize`]: ../trait.Serialize.html
-    /// [`serialize_key`]: #tymethod.serialize_key
-    /// [`serialize_value`]: #tymethod.serialize_value
+    /// [`Serialize`]: crate::Serialize
```
Mingun, line 1807: Some lines can be removed altogether because corresponding items already in scope. For example, this line

PR #2901, opened 2025-03-05
MSRV (1.56) and edition (2021) bump

oli-obk:
It has been 3 years (the 1.56 release) since rust-version was actually checked in Cargo.toml. So let's upgrade:
* using rust-version set to 1.56 and dropping all the [build.rs](http://build.rs/) tricks we have to support Rust versions older than 3 years,
* switching to the 2021 edition released at the same version

Then we can subsequently, once enough time has passed, release more serde versions that bump rust-version. I do not believe we'd inconvenience anyone, as either they are not updating their dependencies anyway (and almost everyone is bumping MSRV knowingly or unknowingly). While there are LTS systems (like debian and such) that use 5 years (or even more in some cases), these realistically can't update their dependencies without manual cherry picking anyway.

The reason I'm bringing this up now is we're optimistic that const-trait support will land this year. While I don't plan on moving serde to be usable in const contexts in any short timeframe, I want to slowly move us up so that we can make it usable in const contexts in the next 3 years.


oli-obk:
No, this is not a breaking change nor a major version bump. This is a minor version bump, which is not allowed to break things, but only to signal that there is more change happening than in a usual patch release.

serde/build.rs
```diff
@@ -18,38 +18,17 @@ fn main() {
         println!("cargo:rustc-check-cfg=cfg(no_core_error)");
         println!("cargo:rustc-check-cfg=cfg(no_core_net)");
         println!("cargo:rustc-check-cfg=cfg(no_core_num_saturating)");
-        println!("cargo:rustc-check-cfg=cfg(no_core_try_from)");
         println!("cargo:rustc-check-cfg=cfg(no_diagnostic_namespace)");
         println!("cargo:rustc-check-cfg=cfg(no_float_copysign)");
```
dtolnay, line 22:
```suggestion
```

dtolnay (approved):
Thanks!

Issue #2904, opened 2025-03-09
Serde derive macros should use absolute paths


davidzeng0:
For example, it should use
```rust
extern crate serde as _serde;

   ::_serde::stuff;
// ^--------
```

This will squash `#[warn(clippy::absolute_paths)]`

PR #2905, opened 2025-03-09
fix clippy absolute paths warning

davidzeng0:
Fixes https://github.com/serde-rs/serde/issues/2904


davidzeng0:
Hmm looks more like a clippy issue instead

PR #2906, opened 2025-03-09
fix clippy absolute paths warning

davidzeng0:
First attempt https://github.com/serde-rs/serde/pull/2905 failed because [this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=71f25171e7d282a7073985c1f34eb6ef) pattern apparently doesn't work. Alternatively one could replace `_serde` with a runtime `#path`

Fixes https://github.com/serde-rs/serde/issues/2904


dtolnay (approved):
Thanks!

PR #2908, opened 2025-03-10
Implement `#[serde(implied(key = "key", value = "value")]` for adding static values

frectonz:
# The `implied` Attribute

It would be useful to introduce an `implied` attribute in `serde`, which automatically adds a specific key-value pair during serialization and ensures that the key is present during deserialization. This would allow us to automatically include predefined keys (_like `method` in the following example_) in the serialized and deserialized data without manually specifying them a `struct`.

Example:

```rust
#[derive(Debug, Deserialize, Serialize)]
#[serde(implied(key = "method", value = "Target.setDiscoverTargets"))]
struct TargetSetDiscoverTargets {
    id: String,
    params: TargetSetDiscoverTargetsParams,
}
```

### 1. Serialization

When serializing an instance of `TargetSetDiscoverTargets`, `serde` should automatically include the `"method": "Target.setDiscoverTargets"` key-value pair in the resulting JSON:

```json
{
    "method": "Target.setDiscoverTargets",
    "id": "some-id",
    "params": {  }
}
```

### 2. Deserialization

During deserialization, serde should check if the `"method"` key exists in the input and match it against the predefined value (`"Target.setDiscoverTargets"`). If it does not match or is missing, `serde` should raise an error.

```json
{
    "method": "Target.setDiscoverTargets",
    "id": "some-id",
    "params": {  }
}
```

## Alternatives Considered

Manually adding the method field during serialization and validating its existence during deserialization. This approach introduces repetitive code that could be automated with the implied attribute. This also means that the user will have to implement `Serialize` and `Deserialize` themselves.

This PR implements this attribute.

### DIFF
--- a/serde_derive/src/de.rs
+++ b/serde_derive/src/de.rs
@@ -1042,7 +1042,11 @@ fn deserialize_struct(
     let fields_stmt = if has_flatten {
         None
     } else {
-        let field_names = deserialized_fields.iter().flat_map(|field| field.aliases);
+        let implied_fields = cattrs.implied().iter().map(|(key, _)| key);
+        let field_names = deserialized_fields
+            .iter()
+            .flat_map(|field| field.aliases)
+            .chain(implied_fields);
 
         Some(quote! {
             #[doc(hidden)]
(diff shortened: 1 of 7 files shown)


Mingun:
#1448 introduces `#[serde(tag="...")]` on structs. Your needs can be addressed with that code:
```rust
#[derive(Debug, Deserialize, Serialize)]
#[serde(tag = "method", rename = "Target.setDiscoverTargets")]
struct TargetSetDiscoverTargets {
    id: String,
    params: TargetSetDiscoverTargetsParams,
}
```
Is that would enough for you?

frectonz:
Oh nice, #1448 handles the `serialization` part of the problem. It adds the tag key and value as a field when it serializes the struct, but it doesn't check for the tag when it deserializes.

frectonz:
Hi everyone,

I hope you’re doing well! I wanted to follow up on my PR, which has been open for a couple of weeks without any progress. I’d appreciate any thoughts you might have on it and whether there’s anything I can do to facilitate the review process.

Please let me know if any changes or additional context are needed. Looking forward to your feedback!

Also, thank you for your work in serde ❤️ .

ProbstDJakob:
At least for me it is counter-intuitive to use `#[serde(tag = "method", rename = "Target.setDiscoverTargets")]` and besides the lack of the validation during the deserialisation, it also does neither work with struct-like enum variants nor does it support adding multiple "constant" key-value pairs. Thus I would like to see this PR (or something alike) being merged.

PR #2912, opened 2025-03-22
Customizable buffer type (proof of concept)

dtolnay:
This demonstrates a backward compatible way that a format such as serde_json could use its own serde_json::Value as its buffer type for untagged enums and flatten, instead of "Content", solving https://github.com/serde-rs/serde/issues/1183. Uses "return position impl trait in traits" (stabilized in Rust 1.75 by https://github.com/rust-lang/rust/pull/115822) and "precise capturing in traits" (stabilizing in the near future by https://github.com/rust-lang/rust/pull/138128).

### DIFF
--- a/serde/src/de/mod.rs
+++ b/serde/src/de/mod.rs
@@ -1223,6 +1223,12 @@ pub trait Deserializer<'de>: Sized {
         true
     }
 
+    ///
+    fn deserialize_buffer(self) -> Result<impl Buffer<'de, Error = Self::Error>, Self::Error> {
+        let content = crate::__private::de::Content::deserialize(self)?;
+        Ok(crate::__private::de::ContentDeserializer::new(content))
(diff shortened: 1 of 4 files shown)


juntyr:
This would be WONDERFUL, thank you @dtolnay for exploring it <3

Mingun:
I hope, that before landing this, you will look at least an one implementation for a format other than JSON-like. I can help with testing this approach in quick-xml.

juntyr:
I'm a maintainer for RON and we'd also be very interested

juntyr:
@dtolnay We just had another bug report in RON that we cannot fix because of not being able to customize the content/buffer type. Since precise capturing in traits was released in 1.87, would you be able to take another look at this PR?

radu-byte:
Any progress here?

PR #2922, opened 2025-05-08
Add helper CowStrVisitor and CowBytesVisitor types

Mingun:
As all known, when you deserialize [`Cow`](https://doc.rust-lang.org/nightly/alloc/borrow/enum.Cow.html), it will always be deserialized as `Owned`, which makes it difficult to implement effective deserialization when you manually want to deserialize `Cow<str>` or `Cow<[u8]>`. This PR adds two helper types, `CowStrVisitor` and `CowBytesVisitor`, which both serves as a `Visitor` and a `DeserializeSeed` and allows you to read borrowed data from, for example, `MapAccess`:

```rust
use serde::de::{DeserializeSeed, Deserializer, MapAccess, Visitor};
use serde::de::value::CowStrVisitor;
use std::borrow::Cow;

/// A simple XML DOM node.
struct Element<'a> {
    name: Cow<'a, str>,
    children: Vec<Node<'a>>,
}
enum Node<'a> {
    Text(Cow<'a, str>),
    Element(Element<'a>),
}

impl<'de> Visitor<'de> for Element<'de> {
    type Value = Self;

    fn expecting(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
        f.write_str("a map")
    }

    fn visit_map<A>(mut self, mut map: A) -> Result<Self::Value, A::Error>
    where
        A: MapAccess<'de>,
    {
        while let Some(key) = map.next_key_seed(CowStrVisitor)? {
            if "$text" == key {
                let text = map.next_value_seed(CowStrVisitor)?;
                self.children.push(Node::Text(text));
            } else {
                let elem = Element { name: key, children: Vec::new() };
                let elem = map.next_value_seed(elem)?;
                self.children.push(Node::Element(elem));
            }
        }
        Ok(self)
    }
}

impl<'de> DeserializeSeed<'de> for Element<'de> {
    type Value = Self;

    where
        D: Deserializer<'de>,
    {
        deserializer.deserialize_map(self)
    }
}
```

### DIFF
--- a/test_suite/tests/test_borrow.rs
+++ b/test_suite/tests/test_borrow.rs
@@ -7,8 +7,11 @@
     dead_code,
 )]
 
-use serde::de::value::{BorrowedStrDeserializer, MapDeserializer};
-use serde::de::{Deserialize, Deserializer, IntoDeserializer};
+use serde::de::value::{
+    BorrowedBytesDeserializer, BorrowedStrDeserializer, CowBytesVisitor, CowStrVisitor, Error,
+    MapDeserializer,
+};
+use serde::de::{Deserialize, DeserializeSeed, Deserializer, IntoDeserializer};
 use serde_derive::Deserialize;
 use serde_test::{assert_de_tokens, assert_de_tokens_error, Token};
 use std::borrow::Cow;
@@ -137,7 +140,7 @@ fn test_cow() {
     struct BorrowedStr(&'static str);
 
     impl<'de> IntoDeserializer<'de> for BorrowedStr {
-        type Deserializer = BorrowedStrDeserializer<'de, serde::de::value::Error>;
+        type Deserializer = BorrowedStrDeserializer<'de, Error>;
 
         fn into_deserializer(self) -> Self::Deserializer {
             BorrowedStrDeserializer::new(self.0)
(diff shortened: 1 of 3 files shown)


serde/src/de/value.rs
```diff
@@ -1727,6 +1727,292 @@ where
...
+
+    fn expecting(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.write_str("a byte array")
+    }
+
+    fn visit_str<E>(self, v: &str) -> Result<Self::Value, E>
+    where
+        E: de::Error,
+    {
+        Ok(Cow::Owned(v.as_bytes().to_owned()))
```
oli-obk, line 1959: why is this one owned?

oli-obk (approved):
makes sense. Please add a test for it

oli-obk:
They just test compilation, not execution actually producing `Cow::Borrowed`

Mingun:
Ah, yes. Will be soon

test_suite/tests/test_borrow.rs
```diff
@@ -4,8 +4,11 @@
     clippy::used_underscore_binding
 )]
 
-use serde::de::value::{BorrowedStrDeserializer, MapDeserializer};
-use serde::de::{Deserialize, Deserializer, IntoDeserializer};
+use serde::de::value::{
+    BorrowedBytesDeserializer, BorrowedStrDeserializer, CowBytesVisitor, CowStrVisitor, Error,
+    MapDeserializer,
+};
+use serde::de::{Deserialize, DeserializeSeed, Deserializer, IntoDeserializer};
```
Mingun, line 11: `DeserializeSeed` is required to be in scope to be able to call it methods on `Cow...Visitor`s

test_suite/tests/test_borrow.rs
```diff
@@ -159,6 +162,47 @@ fn test_cow() {
     }
 }
 
+#[test]
+fn test_cow_str_visitor() {
+    let de_str = BorrowedStrDeserializer::<Error>::new("borrowed");
+    let de_bytes = BorrowedBytesDeserializer::<Error>::new(b"borrowed");
+
+    // This example shows, that without CowStrVisitor the result is different
+    match Cow::<str>::deserialize(de_str) {
```
oli-obk, line 171: I'm guessing we can't use the new visitor in the Cow impl because that's generic?

oli-obk:
@dtolnay this seems like a good addition to serde. even if it could live out of tree, it's for basic libstd types

bourumir-wyngs:
As it can help with my problem (#3027), I would like to express the support.

bourumir-wyngs:
Changes reviewed, who is holding this? I still have the suppressed test in mine serde-saphyr that requires this change to be fixed.

PR #2929, opened 2025-05-24
Add more tests for flatten enums to increase test coverage

Mingun:
This PR adds more tests for enum flatten into the struct (each in their own commit):
- Ensure, that serialization produces output that can be deserialized
- Ensure, that any order of fields in the serialized form for 
  ```rust
  struct Flatten {
    #[serde(flatten)]
    data: Enum,
    other_fields: ...,
  }
  ```
  can be deserialized
- Ensure, that `Enum::Unit` can be deserialized when `Enum` is flatten like above

### DIFF
--- a/test_suite/tests/test_annotations.rs
+++ b/test_suite/tests/test_annotations.rs
@@ -2939,7 +2939,7 @@ mod flatten {
 
             #[test]
             fn unit_enum_with_unknown_fields() {
-                #[derive(Debug, PartialEq, Deserialize)]
+                #[derive(Debug, PartialEq, Serialize, Deserialize)]
                 struct Flatten {
                     #[serde(flatten)]
                     x: X,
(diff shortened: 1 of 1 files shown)


test_suite/tests/test_annotations.rs
```diff
@@ -2827,55 +2909,190 @@ mod flatten {
...
+                        Token::Str("content"),
+                        Token::Unit,
+                        // outer
+                        Token::Str("outer"),
                         Token::U32(42),
-                        Token::StructEnd,
+                        Token::MapEnd,
+                    ],
+                );
+                // Field order: [content, tag], outer - did not work
```
dtolnay, line 3013: I am unsure how to interpret this "did not work" comment (and another one below). Is this to indicate something that was unsupported by the original implementation of `flatten` that has since become supported in a more recent release of serde? If so, I am not sure that this comment is warranted and providing clarity.

Mingun: I don't remember this now - it's too many time since creating this PR. Probably that did not work at time of writing and then was fixed by other PR. Currently that part of comment can be safely removed

PR #2931, opened 2025-05-24
Unify access to the serde namespace

Mingun:
In other places `_serde` name is used to access the `serde` crate, so will be defined and can be use there


Mingun:
@dtolnay, @oli-obk, could you look at this simple change? If it's unclear, the `_serde` token is introduced in `wrap_in_const` call few lines below here:
https://github.com/serde-rs/serde/blob/7b7e089b8d2f1e42dbb426ebf4f3a35ad72b5beb/serde_derive/src/de.rs#L64-L68

Issue #2932, opened 2025-05-26
Feature: Allow to skip seq access for internally and adjacently tagged enums


Dzordzu:
Both the adjacently tagged and internally tagged representations can be
expressed either via map or sequence syntax. For example:

```rust
#[derive(Debug, Serialize, Deserialize)]
#[serde(tag = "type", content = "x")]
enum MyType {
    Topic(i64), Sidebar(i64)
}

fn main() {}
```

is equivalent to both

```json
[ "Topic", 1 ]
```

and 

```json
{ "type": "Topic", "x": 1 }
```

### Disable the sequence format

In case you want to disable the sequence format, you can use the
`#[serde(seq=false)]`.

```rust
#[derive(Debug, Serialize, Deserialize)]
#[serde(tag = "type", content = "x", seq=false)]
enum MyType {
    Topic(i64), Sidebar(i64)
}

will casue the sequence syntax to be disabled.

Issue #2917, opened 2025-04-23
`[0]` unexpectedly deserializes to first unit enum variant when another variant is untagged


flying-sheep:
See playground: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=9f33531c484c068a91e29f62770a3505

```rust
#![allow(dead_code)]

use serde_derive::{Serialize, Deserialize}; // 1.0.219

#[derive(Debug, Serialize, Deserialize)]
#[serde(tag = "type")]
enum StructuralSubElement {
    Title,
    #[serde(untagged)]
    SubStructure(SubStructure),
}

#[derive(Debug, Serialize, Deserialize)]
#[serde(tag = "type")]
enum SubStructure {
    Topic, Sidebar
}

fn main() -> Result<(), Box<dyn std::error::Error>> {
    let v: StructuralSubElement = serde_json::from_str("[0]")?;
    dbg!(v);
    Ok(())
}
```

this emits

```
[src/main.rs:21:5] v = Title
```

Dzordzu:
Even more ["funny" code snippet](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=69b1edc9f73fd87a7753e5562745ab8c)

```rust
#[derive(Debug, Serialize, Deserialize)]
enum StructuralSubElement {
    Title,
    Other,
    #[serde(untagged)]
    SubStructure(SubStructure),
}

#[derive(Debug, Serialize, Deserialize)]
#[serde(tag = "type")]
enum SubStructure {
    Topic,
    Sidebar,
}

fn main() -> Result<(), Box<dyn std::error::Error>> {
    let v: StructuralSubElement = serde_json::from_str("[0]")?;
    dbg!(&v);
    let serialized = serde_json::to_string_pretty(&v).unwrap();
    println!("{}", serialized);
    Ok(())
}
// STDOUT:
// {
//   "type": "Topic"
// }

```

Dzordzu:
# Research post (can be edited)

## Code snippet

<details open>
<summary>Code snippet</summary>
<br>

```rust
use serde::{Serialize, Deserialize};

#[derive(Debug, Serialize, Deserialize)]
#[serde(tag = "type")]
enum StructuralSubElement {
    Title,
    Other,
    #[serde(untagged)]
    SubStructure(String),
}

fn main() -> Result<(), Box<dyn std::error::Error>> {
    let v: StructuralSubElement = serde_json::from_str("[0]")?;
[...]

PR #2933, opened 2025-05-26
Feature: Allow to skip seq access for internally and adjacently tagged enums

Dzordzu:
## About
Implementation of the #2932
Refers to #2917

## Changes
- **Allowed to add seq=true/false to the serde internally and adjacently tagged enums**
- **Fixed wrong lint**

### DIFF
--- a/serde/src/private/de.rs
+++ b/serde/src/private/de.rs
@@ -12,8 +12,9 @@ use crate::de::{MapAccess, Unexpected};
 #[cfg(any(feature = "std", feature = "alloc"))]
 pub use self::content::{
     Content, ContentDeserializer, ContentRefDeserializer, EnumDeserializer,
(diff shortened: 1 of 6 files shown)


serde/src/private/de.rs
```diff
@@ -867,14 +870,18 @@ mod content {
...
         {
-            let tag = match tri!(seq.next_element()) {
-                Some(tag) => tag,
-                None => {
-                    return Err(de::Error::missing_field(self.tag_name));
-                }
-            };
-            let rest = de::value::SeqAccessDeserializer::new(seq);
-            Ok((tag, tri!(Content::deserialize(rest))))
+            if self.use_seq {
```
Mingun, line 873: I think, it is better to introduce new type and refactor existing `TaggedContentVisitor` to delegate to the new type all except `visit_seq`. Then there will be no performance penalty in runtime due to this check

PR #2935, opened 2025-05-29
Remove code duplication when deriving Deserialize for adjacently tagged enums

Mingun:
Adjacently tagged enums with many struct variants should get most benefit from that.

### DIFF
--- a/serde_derive/src/de.rs
+++ b/serde_derive/src/de.rs
@@ -1601,10 +1601,12 @@ fn deserialize_adjacently_tagged_enum(
         }
     } else {
         quote! {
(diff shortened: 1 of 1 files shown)


Mingun:
@dtolnay, @oli-obk, could you look at this simple optimization?

Mingun:
@dtolnay, @oli-obk mind to merge this very simple optimization (the second commit does the job). Expanding of `variant_arms` variable performed twice which creates duplication. The effect especially visible when struct variants are used, because they generate a lot of code.

dtolnay (approved):
Thank you!

PR #2939, opened 2025-06-06
Remove `actually_private::T`

Mingun:
> This sounds like a change we could do on master and consider in isolation from this PR

As posted by @oli-obk in https://github.com/serde-rs/serde/pull/2608#discussion_r1810195461, this PR removes `actually_private::T` type from signature of the hidden `Deserializer::__deserialize_content` method, like `Visitor::__private_visit_untagged_option` does not have such guard:
https://github.com/serde-rs/serde/blob/2130ba57885357b56e7087e164a33be0288371e4/serde/src/de/mod.rs#L1679-L1687

This guard just makes life harder in some cases without real benefits. If you use private API you already understand consequences and them can be totally acceptable, especially if you depend on the specific version


oli-obk:
Thanks!

PR #2941, opened 2025-06-19
add f16 and f128 support

Vrtgs:
add initial support for f16 and f128, next steps are to add this support in serde_test if this lands

### DIFF
--- a/serde/src/de/mod.rs
+++ b/serde/src/de/mod.rs
@@ -996,6 +996,18 @@ pub trait Deserializer<'de>: Sized {
         Err(Error::custom("u128 is not supported"))
     }
 
+    /// Hint that the `Deserialize` type is expecting a `f16` value.
+    ///
+    /// The default behavior unconditionally returns an error.
+    #[cfg(feature = "unstable")]
+    fn deserialize_f16<V>(self, visitor: V) -> Result<V::Value, Self::Error>
+    where
+        V: Visitor<'de>,
+    {
+        let _ = visitor;
+        Err(Error::custom("f16 is not supported"))
+    }
+
     /// Hint that the `Deserialize` type is expecting a `f32` value.
     fn deserialize_f32<V>(self, visitor: V) -> Result<V::Value, Self::Error>
     where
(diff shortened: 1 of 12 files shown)


test_suite/tests/test_de.rs
```diff
@@ -816,6 +816,16 @@ fn test_nonzero_usize() {
     test(10, &[Token::U64(10)]);
 }
 
+#[test]
+#[cfg(feature = "unstable")]
+#[ignore = "serde_test lacks support for f128"]
```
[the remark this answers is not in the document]
Vrtgs: my bad

serde/src/de/value.rs
```diff
@@ -359,6 +373,7 @@ where
     }
 }
 
+// just why does this not use the `primitive_deserializer` macro??????
```
Mingun, line 376: Because it has special handling for `deserialize_enum`, which other primitived does not.
```suggestion
```

serde/src/integer128.rs
```diff
@@ -1,9 +0,0 @@
-// No longer used. Old versions of serde used this macro for supporting targets
-// that did not yet have 128-bit integer support.
-#[macro_export]
-#[doc(hidden)]
-macro_rules! serde_if_integer128 {
-    ($($tt:tt)*) => {
-        $($tt)*
-    };
-}
```
Mingun, line 9: Why? It was left for purpose, to ot break old code that uses this macro when serde may not support `u128`/`i128`

serde/src/ser/mod.rs
```diff
@@ -625,6 +625,34 @@ pub trait Serializer: Sized {
...
+    ///         serializer.serialize_f16(*self)
+    ///     }
+    /// }
+    /// ```
+    /// The default behavior unconditionally returns an error.
+    #[cfg(feature = "unstable")]
+    fn serialize_f16(self, v: f16) -> Result<Self::Ok, Self::Error> {
+        let _ = v;
+        Err(Error::custom("f16 is not supported"))
+    }
```
Mingun, line 654: Why not forward to `f32` by default? If that could be reasonable

Mingun (commented):
You need to roll back the removal of `integer128`, the maintainers will not approve it.

My reviews means nothing, I do not have any rights in this repository. So probably request my revies is pointless...

serde/src/integer128.rs
Vrtgs: isn't the new msrv 1.5x ???

serde/src/ser/mod.rs
Vrtgs: that was also a thought I had, but every other serialize method is a **required method** so I just took the most extreme option, I could update it if you think otherwise though

serde/src/de/mod.rs
```diff
@@ -1439,6 +1463,17 @@ pub trait Visitor<'de>: Sized {
         ))
     }
 
+    /// The input contains a `f16`.
+    ///
+    /// The default implementation forwards to visit_f64.
```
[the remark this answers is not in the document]
Vrtgs: this break documentation build sadly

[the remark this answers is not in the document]
Mingun: Very strange, that documentation is broken, it should be broken in other places as well. The correct link should be
```
[`visit_f64`](Self::visit_f64)
```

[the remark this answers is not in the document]
Anders429: I think either way should work. You can see that it does link correctly on docs.rs for the other links for this trait. For example: https://docs.rs/serde/latest/serde/de/trait.Visitor.html#method.visit_u64

bioinformatist:
@Vrtgs Any news for this?

vrtgs:
I've implemented everything stated here, but there is no maintainer that gave this attention so far

PR #2946, opened 2025-07-24
serde_derive: Add alias_all attribute to containers

### LINKED ISSUE #1530: Add alias_all attribute for containers similar to rename_all

michelhe:
Resolves #1530 

This adds a new attribute `#[serde(alias_all = "...")]` to containers, which accepts the same rename rules as `#[serde(rename_all = "...")]`. When applied to a container, the alias will be added to all fields (or variants) similarly to how `#[serde(alias = "...")]` works on a single field/variant.

This attribute can be specified multiple times, with the use-case being accepting multiple case styles during de-serialization.

BTW this is my first contribution in this project, so I would appreciate if you take a closer look, and of course help with updating serde.rs documentation if you decide to accept this PR :smile:


test_suite/tests/test_macros.rs
```diff
@@ -864,3 +864,300 @@ fn test_packed_struct_can_derive_serialize() {
...
+}
+
+#[test]
+fn test_alias_all_enum() {
+    #[derive(Debug, PartialEq, Deserialize)]
+    #[serde(alias_all = "kebab-case")]
+    enum TestEnum {
+        VariantOne { field_one: i32 },
+        VariantTwo { field_two: String },
+    }
```
Mingun, line 1026: Isn't this should be applied to variant names instead of field names inside variants? That is, convert to `variant-one` and `variant-two`

michelhe:
Just now realized that there was already an attempt at this in #1972 which was not accepted, missed that when I skimmed through the open issue.

I hope @dtolnay will have a change of heart here 🙏 this feature is really desired by many applications that use serde for loading configuration files.

I'm for once am tired of getting support tickets where the root cause is that the user used the wrong case style in their config files 😅

Our workaround is very not optimal - we first parse to `serde_yaml::Value` and then manually convert all keys to the correct style, and then convert back to yaml string, and deserialize to the Config struct from there.
This is *dangerous* because we lose type information when deserializing to `serde_yaml::Value`, which can lead to cases where strings in the orignal config files that happen to look like floating points (e.g a hexadecimal hash field) lose their meaning and cause deserialization errors when converted to the concrete Config struct.

max-heller:
> Our workaround is very not optimal - we first parse to `serde_yaml::Value` and then manually convert all keys to the correct style, and then convert back to yaml string, and deserialize to the Config struct from there.

FYI, [`serde_yaml::Value` is itself a deserializer](https://docs.rs/serde_yaml/latest/serde_yaml/value/enum.Value.html#impl-Deserializer%3C'de%3E-for-Value), so you should be able to get rid of the `Value` to string step and deserialize your config struct from the `Value`.

test_suite/tests/test_macros.rs
```diff
@@ -864,3 +864,300 @@ fn test_packed_struct_can_derive_serialize() {
...
+    assert_de_tokens(
+        &TestEnum::VariantOne { field_one: 42 },
+        &[
+            Token::Enum { name: "TestEnum" },
+            Token::Str("variant-one"),
+            Token::Struct {
+                name: "VariantOne",
+                len: 1,
+            },
+            Token::Str("field-one"),
```
Mingun, line 1054: I would expect, that `alias_all` for enums will influence only variant names. If you need to appli the same rules to struct variants, you should add attribute to struct variant. If I'm not wrong, it is how `rename_all` attribute work.

PR #2979, opened 2025-09-15
Mark every trait impl for a private type with do_not_recommend

dtolnay:
Rustc was suggesting these private types in diagnostics, with paths like `_::_serde::__private::ser::content::ContentSerializer<E>`.

### DIFF
--- a/serde/build.rs
+++ b/serde/build.rs
@@ -14,12 +14,19 @@ fn main() {
     };
 
     if minor >= 77 {
+        println!("cargo:rustc-check-cfg=cfg(no_diagnostic_namespace)");
         println!("cargo:rustc-check-cfg=cfg(no_serde_derive)");
     }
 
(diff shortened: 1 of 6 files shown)


oli-obk:
Almost looks like we'd want this attribute to work on modules and apply to everything within

compiler-errors:
Or perhaps less invasively, we could allow it to be applied to traits.

Jules-Bertholet:
It should be implied by `doc(hidden)`, as in https://github.com/rust-lang/rust/pull/119151.

dtolnay:
I agree that `doc(hidden)` should have this effect, I made an issue for that last year: https://github.com/rust-lang/rust/issues/132024

Jules-Bertholet:
https://github.com/rust-lang/rust/pull/146617

PR #2991, opened 2025-09-25
Inline serde_core into serde in docsrs mode

dtolnay:
#2608 regressed the rustdoc rendered documentation of `serde` in many ways because rustdoc does not currently handle documentation of cross-crate re-exported items well. For example there is not enough information in the rmeta of dependencies to accurately reconstruct the way that trait associated types are named in signatures. The [A-cross-crate-reexports](https://github.com/rust-lang/rust/labels/A-cross-crate-reexports) label in rust-lang/rust records numerous other issues.

Before #2608:

<p align="center"></p>

After #2608:

<p align="center"></p>

This PR works around the cross-crate re-export issues by inlining the items defined by `serde_core` into `serde` when building documentation for docs.rs.


oli-obk:
oof. Very hacky. I'd prefer to just keep the status quo. It's not that bad and we can use it to prioritize the rustdoc issues more

dtolnay:
I would be grateful if you could expedite some fixes in rustdoc, but in my opinion having a workaround in place until rustdoc gets fixed is less bad than having the documentation remain in the current state for the same duration.

The workaround can be removed as soon as the version of nightly rustdoc used by docs.rs no longer produces less legible / less copy-pasteable documentation without it.

Beyond the associated type issue from the screenshots which is my prevailing concern, here are some other issues you could help with, which I have noticed in the serde documentation with re-exports:

- https://github.com/rust-lang/rust/issues/49339
  The "Source" links at the module level, for example in `serde::de`

* https://github.com/rust-lang/rust/issues/131180
  Affects `serde::ser::StdError` and `serde::de::StdError`

- https://github.com/rust-lang/rust/issues/44306
  The spelling of `Self` in signatures such as `serde::de::value::MapDeserializer::new`

* The use of `core::fmt::Result` vs `core::result::Result<(), core::fmt::Error>` in the documentation of `impl<'a> Serializer for &mut Formatter<'a>`
  - Something else I noticed in this impl, the associated type problem means rustdoc is producing signatures that wouldn't even compile. `-> Result<<&mut Formatter<'a> as Serializer>::SerializeSeq, Error>`
    ```console
    error[E0106]: missing lifetime specifier
       --> serde_core/src/ser/fmt.rs:117:60
        |
    117 |     fn serialize_seq(self, _len: Option<usize>) -> Result<<&mut Formatter<'a> as Serializer>::SerializeSeq, fmt::Error> {
        |                                                            ^ expected named lifetime parameter
        |
        = help: this function's return type contains a borrowed value, but there is no value for it to be borrowed from
    ```

Some other things that are solved by the workaround in this PR, which are not going to be solved in rustdoc and we will need some other workaround (probably easy) if not using this one:

- Tuple impls, which are rendered `impl<'de, T> Deserialize<'de> for (T₁, T₂, …, Tₙ)` in serde_core but not serde

- `doc(cfg)`, such as on `serde_core::de::Visitor::visit_string`, is missing in serde

PR #3003, opened 2025-10-21
List errors that occured while deserializing an untagged enum

spoutn1k:
Puts #773 back on the table. Ran into the issue recently and seems like changes to the code made it a lot easier to fix. This is a naive attempt to create discussion on the topic.

The test changes are as follows:
```
data did not match any variant of untagged enum Untagged:
- A: invalid type: sequence, expected struct variant Untagged::A
- B: invalid type: sequence, expected struct variant Untagged::B
- C: invalid type: sequence, expected unit variant Untagged::C
- D: invalid type: sequence, expected u8
- E: invalid type: sequence, expected a string
- F: invalid length 1, expected tuple variant Untagged::F with 2 elements
```

The error before being:
```
data did not match any variant of untagged enum Untagged
```

Obviously there is a lot to discuss. I wanted to filter the errors to only output error message that came from deeper in the code (filter out structural issues in the enum itself as they could arguably be obvious), and the TODO message seem to imply it was done for toml (?) but this is out of my ability range as I am new to this codebase.

Previous works include #1544, which feature-gate the change, which could also be implemented here.

### DIFF
--- a/test_suite/tests/test_enum_untagged.rs
+++ b/test_suite/tests/test_enum_untagged.rs
@@ -9,7 +9,7 @@
 mod bytes;
 
 use serde_derive::{Deserialize, Serialize};
-use serde_test::{assert_de_tokens, assert_de_tokens_error, assert_tokens, Token};
+use serde_test::{Token, assert_de_tokens, assert_de_tokens_error, assert_tokens};
 use std::collections::BTreeMap;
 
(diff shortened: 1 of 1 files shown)


spoutn1k:
Oops did not see #2376. Sorry for the spam.

PR #3007, opened 2025-11-07
Add TOON to format list

hxphsts:
Adding [TOON](https://github.com/toon-format/spec/blob/main/SPEC.md) and `serde_toon`, a Serde-compatible TOON serialization crate, to the list of supported formats.

### DIFF
--- a/serde/src/lib.rs
+++ b/serde/src/lib.rs
@@ -64,6 +64,7 @@
 //! - [Hjson], a syntax extension to JSON designed around human reading and
 //!   editing. *(deserialization only)*
 //! - [CSV], Comma-separated values is a tabular text file format.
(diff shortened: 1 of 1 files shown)


oli-obk:
Your crate is under a week old. For me personally that does not make it maintained or finished enough to include here. 

There are also 4 more toon serialization crates with similarly low usage statistics and recent first publishes which makes my "there are incentives outside my knowledge area" sense tingle

Let's give y'all some time to bake and find actual users

hxphsts:
Fair, thank you for your response!

The crate itself is complete for its current ser/de purposes, in line with the 1.4 spec of the TOON format.

My only incentive was, of course, to write a Serde-compatible crate for use in my own projects, as I frequently rely on Serde, and the only other available TOON crate when I published was not.

PR #3015, opened 2025-11-29
Add field attribute `on_duplicate` to allow custom handling of duplicate fields

Lencerf:
Currently serde returns error "duplicate field" when deserializing a struct with duplicate field names, for example,

```json
{
    "name": "John Titor",
    "aliases": ["John", "JT"],
    "aliases": ["TimeTravel_0"]
}
```

The new `on_duplicate` annotation takes a `fn<T>(&mut T, T) -> Result<(), E> where E: Error` to allow custom handling of duplicate fields. 

This can be useful to implement a last-one-wins strategy with the following function,

```rust
fn take_last_duplicate<E, T>(parsed: &mut T, another: T) -> Result<(), E>
where
    E: serde::de::Error,
{
    *parsed = another;
    Ok(())
}
```

or implement a merge-duplicates strategy with the following function,

```rust
fn merge_duplicates<E, T>(parsed: &mut Vec<T>, another: Vec<T>) -> Result<(), E>
where
    E: serde::de::Error,
{
    parsed.extend(another);
    Ok(())
}
```

Test cases are added in the 2nd commit.

### DIFF
--- a/test_suite/tests/test_annotations.rs
+++ b/test_suite/tests/test_annotations.rs
@@ -1942,6 +1942,94 @@ fn test_expecting_message_identifier_enum() {
     );
 }
 
+#[derive(Debug, PartialEq, Deserialize)]
+struct DeserializeDenyDuplicate {
+    value: i32,
+}
+
+#[test]
+fn test_deserialize_deny_duplicate() {
+    assert_de_tokens_error::<DeserializeDenyDuplicate>(
+        &[
+            Token::Struct {
+                name: "DeserializeDenyDuplicate",
+                len: 2,
+            },
+            Token::Str("value"),
+            Token::I32(1),
+            Token::Str("value"),
+            Token::I32(2),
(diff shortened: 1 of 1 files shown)


serde_derive/src/internals/symbol.rs
```diff
@@ -13,6 +13,7 @@ pub const DEFAULT: Symbol = Symbol("default");
 pub const DENY_UNKNOWN_FIELDS: Symbol = Symbol("deny_unknown_fields");
 pub const DESERIALIZE: Symbol = Symbol("deserialize");
 pub const DESERIALIZE_WITH: Symbol = Symbol("deserialize_with");
+pub const DESERIALIZE_DUPLICATE: Symbol = Symbol("deserialize_duplicate");
```
Mingun: Probably a better name is `on_duplicate` or `on_duplicate_key`

PR #3025, opened 2026-01-24
Fix panic due to string slicing at char boundary

### LINKED ISSUE #2953: Bug: #[serde(rename_all = "camelCase")] panics on struct fields with non-ASCII (e.g., CJK) identifiers
Hello Serde Team,

I've encountered a compile-time panic when using `#[derive(Deserialize)]` in combination with `#[serde(rename_all = "camelCase")]` on a struct that has fields with non-ASCII identifiers (specifically, Chinese characters).

### Summary

The `serde_derive` procedural macro appears to panic during compilation when it tries to apply the `camelCase` renaming rule to a field name containing multi-byte UTF-8 characters. The error message clearly indicates a byte-indexing issue on a non-char boundary, suggesting the renaming logic might be incorrectly assuming ASCII strings.

The issue does **not** occur if:

- The `#[serde(rename_all = "camelCase")]` attribute is removed.
- The attribute is changed to `#[serde(rename_all = "PascalCase")]`.

This suggests the bug is specific to the implementation of the `camelCase` transformation logic.

### Minimal Reproducible Example (MRE)

Here is the smallest possible Rust project that demonstrates the bug:

**`Cargo.toml`:**

```toml
[package]
name = "serde-cjk-bug-report"
version = "0.1.0"
edition = "2024"

[dependencies]
serde = { version = "1.0.219", features = ["derive"] }
```

**`src/main.rs`:**

```rust
use serde::Deserialize;

// This struct definition causes the compiler to panic.
#[derive(Deserialize, Debug, Clone)]
#[serde(rename_all = "camelCase")]
pub struct TestPayload {
    // Field name with Chinese characters
    pub 项目名称: String,
    pub 项目地址: String,
}

fn main() {
    // The panic happens at compile time, so the main function's content
    // doesn't matter, but it's here to make the example complete.
    println!("This code will not compile.");
}
```

### Compiler Panic Output

When trying to compile the MRE with `cargo build`, the compiler panics with the following output:

```
error: proc-macro derive panicked
  --> src/main.rs:5:10
   |
5  | #[derive(Deserialize, Debug, Clone)]
   |          ^^^^^^^^^^^
   |
   = help: message: byte index 1 is not a char boundary; it is inside '项' (bytes 0..3) of `项目名称`

```

### Expected vs. Actual Behavior

-   **Expected Behavior:** The code should compile successfully. The `camelCase` rule, when applied to a field name like `项目名称` which has no case or separators to modify, should ideally result in no change to the name, or at least not cause a panic.
-   **Actual Behavior:** The compiler panics with a fatal error related to string indexing.

### Analysis

The panic message `byte index 1 is not a char boundary` strongly suggests that the string manipulation logic within the `rename_all = "camelCase"` implementation is incorrectly treating the UTF-8 field identifier as a sequence of bytes rather than characters. It attempts a byte-level index access, which is unsafe for multi-byte strings. A correct implementation should iterate over `chars()` rather than `bytes()`.

### Workarounds

For others encountering this issue, the current workarounds are:

1.  Avoid using non-ASCII field names if `rename_all` is needed.
2.  Use ASCII field names and the `#[serde(rename = "项目名称")]` attribute on each field individually.
3.  Remove the `rename_all` attribute if the exact casing is not critical for your use case.

### Environment

-   **Rust Version:** `rustc 1.89.0 (29483883e 2025-08-04)` 
-   **Serde Version:** `1.0.219`
-   **OS:** Windows 10

Thank you for maintaining this essential crate! I hope this report is helpful in resolving the issue.

ginnyTheCat:
This is a alternative solution to fix #2953 than #3024 by continuing to just support ASCII but not panic in this unwanted case.

Since this only makes code that didn't compile before compile now I would not consider this a breaking change.

### DIFF
--- a/serde_derive/src/internals/case.rs
+++ b/serde_derive/src/internals/case.rs
@@ -59,7 +59,13 @@ impl RenameRule {
             None | PascalCase => variant.to_owned(),
             LowerCase => variant.to_ascii_lowercase(),
             UpperCase => variant.to_ascii_uppercase(),
-            CamelCase => variant[..1].to_ascii_lowercase() + &variant[1..],
+            CamelCase => {
+                if let Some(first) = variant.get(..1) {
+                    first.to_ascii_lowercase() + &variant[1..]
+                } else {
+                    variant.to_owned()
+                }
+            }
             SnakeCase => {
                 let mut snake = String::new();
                 for (i, ch) in variant.char_indices() {
@@ -100,7 +106,11 @@ impl RenameRule {
             }
             CamelCase => {
                 let pascal = PascalCase.apply_to_field(field);
-                pascal[..1].to_ascii_lowercase() + &pascal[1..]
+                if let Some(first) = pascal.get(..1) {
+                    first.to_ascii_lowercase() + &pascal[1..]
+                } else {
+                    pascal
+                }
             }
             ScreamingSnakeCase => field.to_ascii_uppercase(),
             KebabCase => field.replace('_', "-"),
--- a/test_suite/tests/test_macros.rs
+++ b/test_suite/tests/test_macros.rs
@@ -657,10 +657,7 @@ fn test_rename_all() {
     #[serde(rename_all = "snake_case")]
     enum E {
         #[serde(rename_all = "camelCase")]
-        Serialize {
-            serialize: bool,
-            serialize_seq: bool,
-        },
+        Ändern { ändern: bool, serialize_seq: bool },
         #[serde(rename_all = "kebab-case")]
         SerializeSeq {
             serialize: bool,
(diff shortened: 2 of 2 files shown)


afurm:
The `apply_to_field` path returns `pascal` directly when the first character slice is empty, but doesn't this mean an empty string input returns itself rather than applying any transformation? It would be worth adding a test case for an empty string field name in `apply_to_field` to match the `apply_to_variant` coverage.

ginnyTheCat:
I'm confused what transformations you would want to apply to a empty string? But yes, I'm open to a adding a test case ofc once I better understand your intentions.

afurm:
The implementation mirrors `--glob-case-insensitive` cleanly. On the design question in the PR body: keeping this opt-in is the right call. On case-sensitive filesystems (Linux), there are legitimate files named `.HTML` as distinct from `.html`, so `--type html` should NOT match `.HTML` by default. Opt-in preserves that guarantee.

afurm:
Apologies — the previous comment was accidentally posted to the wrong PR. To continue the thread on this PR: consider the case where a struct field is renamed via `#[serde(rename = "")]` — an empty string is a valid rename target. Does `apply_to_field("")` pass it through as-is, or should it be treated as a no-op that falls back to default casing logic?

PR #3037, opened 2026-02-13
serde_derive: rewrite fields_from_ast with less llvm-lines

dishmaker:
Only 251 less llvm-lines on rustc `1.93.0`


dtolnay (approved):
Thanks!

Issue #1889, opened 2020-09-13
Partial update deserialization


kevincox:
It would be very useful if deserialization could be used for partial updates. For example I have a struct and I want to deserialize into it but allow some fields to me missing.

Right now the only real option is to make every field `Option<T>`. This works but makes working with the structure very inconvinent. You will almost always want to create a second structure without all of the `Option` fields and that increases maintenance burden to keep the two in sync.

If you write your own deserializer I think you can accomplish this but again, there is a lot of maintenance burden for this.

I think one way to implement this would be add an option to the `deserialize_in_place` generation to `serde_derive` to not raise an error if fields are missing. This would allow just updating the relevant fields. This is safe because all of the fields were valid before `deserialize_in_place` was called so it is fine to keep them as they were.

Example use cases:

* Configuration merging. This would enable efficiently and easily deep-merging multiple objects together.
* Multiple default values. This is sort of the same as the previous but right now your only options for defaults are setting them in `#[serde(default="some::path")]` which hardcodes them. If you could merge on top of an existing object that existing object becomes the defaults.
* Delta updates. This is useful for HTTP PATCH or other remote communication. It allows you just to send over the changed fields.

dtolnay:
I would prefer not to build anything for this into the serde crate, but it sounds like something that could be implemented outside of serde, covering all of the use cases you listed. Something like:

```rust
let mut deserializer = serde_json::Deserializer::from_str(" { ... } ");
let defaults = MyStruct { ... };
let s: MyStruct = deserialize_with_defaults(&mut deserializer, &defaults)?;
```

kevincox:
I don't think that would be possible because `serde::Deserialize` only has one method and it returns `Result<Self, _>` this means that there is no way to inject a "base" object and no way to deserialize in a way that doesn't return a complete object.

The only way I can possible thing of this working is something like serialize the default object, prepend it to the data, then deserialize. However that is awful for a variety of reasons not which of least is that it won't work for most formats.

I think this would need to be supported in-crate. My first idea without much thought would be something like:

```rust
pub trait DeserializePartial<'de>: Sized {
    type Base;

    fn deserialize_partial<D: serde::Deserializer>(base: Self::Base, deserializer: D) -> Result<Self, D::Error>;
}
```

Ideally `serde_derive` could then be updated to generate the `Base` type automatically. It could be a struct that is a clone of the struct that it is developed on except that every field is wrapped in `Option`. (More efficient systems could also be used).

[...]

PR #3046, opened 2026-02-28
<features> allow partial update

yuanyan3060:


oli-obk:
Is there previous discussion on this topic? What kind of code do you want to enable with this?

This PR needs tests and an explanation of the motivation

yuanyan3060:
> Is there previous discussion on this topic? What kind of code do you want to enable with this?之前有人讨论过这个话题吗？你想通过这个功能启用哪种类型的代码？
> 
> This PR needs tests and an explanation of the motivation这条公关稿需要测试，并且需要解释其动机。

https://github.com/serde-rs/serde/issues/1889 

Maybe I should create a new crate instead of modifying serde_derive.

oli-obk:
Yes, as stated on the issue, this should be in its own crate.

Mingun:
In fact, something like this does `deserialize_in_place` (this is hidden feature, but as you inspect derive code, you may aware of it existence).

Issue #1762, opened 2020-03-26
Adjacently and internally tagged enums for non-self-describing formats


da-x:
It seems that supporting adjacently and internally tagged enums can only work for self-describing formats (see bincode's https://github.com/servo/bincode/issues/272). For some non-self-describing formats, it would be nice to support adjacently tagged structs by treating the tag and content _values_ as a tuple instead of additionally serializing the tag string and the content string (i.e. elision of the tag and content strings from serialization and deserialization).

Kobzol:
It seems that this is also a problem for msgpack-rust (https://github.com/3Hren/msgpack-rust/issues/250), which serializes adjacently tagged enums as tuples.

dtolnay:
I opened #2496 to make progress on one piece of this.

boomshroom:
I ran into this when writing my own format and decided to do some digging. The cause seems to be this:

https://github.com/serde-rs/serde/blob/fa7da4a93567ed347ad0735c28e439fca688ef26/serde_derive/src/de/struct_.rs#L70-L73

[Deserialising adjacently tagged enums defers to the code for untagged enums](https://github.com/serde-rs/serde/blob/fa7da4a93567ed347ad0735c28e439fca688ef26/serde_derive/src/de/enum_adjacently.rs#L41), but then when that in turn calls the struct deserialization, its code is overly defensive and worried about conflicting with other variants, even though this _isn't_ an untagged enum and there's no chance of conflict.

Since the serialisation for struct variants of untagged enums just calls the regular struct serialization code, the only workaround is to serialise all structs as maps instead of tuples.

For unit variants, [it unconditionally checks for the second element when given a tuple](https://github.com/serde-rs/serde/blob/fa7da4a93567ed347ad0735c28e439fca688ef26/serde_derive/src/de/enum_adjacently.rs#L296-L299) even though [none was generated](https://github.com/serde-rs/serde/blob/fa7da4a93567ed347ad0735c28e439fca688ef26/serde_derive/src/ser.rs#L669-L673). The map version however [gracefully accepts the lack of a content field](https://github.com/serde-rs/serde/blob/fa7da4a93567ed347ad0735c28e439fca688ef26/serde_derive/src/de/enum_adjacently.rs#L248-L249) and succeeds with only one element.

PR #3062, opened 2026-04-15
Fix deserializing adjacently tagged unit variants as sequences

boomshroom:
This behavior is copied from the map visitor where this was already fixed.

Partially addresses #1762 and https://github.com/3Hren/msgpack-rust/issues/250. Resolving the struct variant part of the issue will be more tricky due to its codepath involving the untagged variants logic where the current behavior is more desirable.


boomshroom:
Seems this is more complex than I thought. The map version checks if the next field exists before handing over the variant to the seed, but the sequence version uses the variant before it gets the presence of a second field. If `__Field` implemented clone, then this would be easy.

PR #3066, opened 2026-05-16
Fixes Default Literals | Issue #368

IftekharUddin:
[Serde_Default_Expression.pdf](https://github.com/user-attachments/files/27863411/Serde_Default_Expression.pdf)

[serde_default_expr_implementation_report.docx](https://github.com/user-attachments/files/27863408/serde_default_expr_implementation_report.docx)

Adds a new field-level attribute, `default_expr`, that allows an arbitrary 
Rust expression to be used as a missing-field default during deserialization, 
without requiring a dedicated helper function.

## Motivation
The existing `default = "path"` form requires a named callable, forcing 
boilerplate for trivial defaults like literals or simple constructors. 
`default_expr` eliminates that overhead while keeping semantics explicit 
and avoiding overload of the existing `default` attribute.

## Changes
- `symbol.rs`: introduce the `DEFAULT_EXPR` symbol
- `attr.rs`: parse `default_expr` as `syn::Expr`; extend `Default` enum 
  with `Expr(syn::Expr)`; enforce single-default-source conflict rule
- `de.rs`: emit expression directly in missing-field branches via 
  `quote_spanned!`; no `Default` bound added for `default_expr`
- `de/struct_.rs`: exhaustive match coverage for new `Default::Expr` variant

## Tests
- Runtime: `test_default_expr_struct`, `test_default_expr_tuple`, 
  `test_skip_with_expr`
- Compile-fail (nightly): `conflict.rs`, `conflict_path.rs`, 
  `type_mismatch.rs`

More in depth documents have been attached. It goes over the game plan and then the implementation that was followed.

### DIFF
--- a/serde_derive/src/de.rs
+++ b/serde_derive/src/de.rs
@@ -206,7 +206,7 @@ fn build_generics(cont: &Container, borrowed: &BorrowedLifetimes) -> syn::Generi
                     &generics,
                     &parse_quote!(_serde::#private::Default),
                 ),
-                attr::Default::None | attr::Default::Path(_) => generics,
+                attr::Default::None | attr::Default::Path(_) | attr::Default::Expr(_) => generics,
             };
 
             let delife = borrowed.de_lifetime();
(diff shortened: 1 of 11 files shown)


IftekharUddin:
hmmm i thought i checked all the tests, I'll look into that.

IftekharUddin:
got tired, will pick it up later but I'm confused by the test. I'm like 70% sure the type mismatch file issue is that it needs to check that it doesn't auto convert? but this change in the default expression would, auto, convert? Too tired, i'm confusing myself.

IftekharUddin:
#368

PR #3071, opened 2026-06-04
Deserialize Option from value deserializers as Some

### LINKED ISSUE #3050: MapDeserializer does not work with optional fields
This code ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&code=use+serde%3A%3ADeserialize%3B%0Ause+serde%3A%3Ade%3A%3Avalue%3A%3A%7BError%2C+MapDeserializer%7D%3B%0A%0A%23%5Bderive%28Debug%2C+Deserialize%29%5D%0Astruct+Map+%7B%0A++++optional%3A+Option%3CString%3E%2C%0A%7D%0A%0Afn+main%28%29+%7B%0A++++let+iter+%3D+%5B%28%22optional%22%2C+%22value%22%29%5D.into_iter%28%29%3B%0A++++let+deserializer+%3D+MapDeserializer%3A%3A%3C_%2C+Error%3E%3A%3Anew%28iter%29%3B%0A++++println%21%28%22%7B%3A%3F%7D%22%2C+Map%3A%3Adeserialize%28deserializer%29%29%3B%0A%7D%0A); serde 1.0.228):

```rust
use serde::Deserialize;
use serde::de::value::{Error, MapDeserializer};

#[derive(Debug, Deserialize)]
struct Map {
    optional: Option<String>,
}

fn main() {
    let iter = [("optional", "value")].into_iter();
    let deserializer = MapDeserializer::<_, Error>::new(iter);
    println!("{:?}", Map::deserialize(deserializer));
}
```

prints:

```
Err(Error("invalid type: string \"value\", expected option"))
```

but I would expect it to print:

```
Ok(Map { optional: "value" })
```

I am able to work around this by wrapping the struct with `#[serde(flatten)]`, e.g.:

```rust
#[derive(Deserialize)]
struct Flatten<T> {
    #[serde(flatten)]
    inner: T,
}
```

then using `MapDeserializer` to deserialize into `Flatten<Map>` and then accessing the inner value. (Enums wrapping `Map` also appear to work, for what seems like a similar reason.)

koriyoshi2041:
The value deserializers in `serde::de::value` (the ones you get from `IntoDeserializer`, including `MapDeserializer`) forward `deserialize_option` to `deserialize_any`. `deserialize_any` hands the held value straight to the visitor, so for an `Option<T>` target it ends up calling `visit_str`/`visit_u32`/etc. on the option visitor, which only knows `visit_some`/`visit_none`. The result is a spurious error:

```
invalid type: string "value", expected option
```

Minimal repro (from #3050):

```rust
use serde::Deserialize;
use serde::de::value::{Error, MapDeserializer};

#[derive(Debug, Deserialize)]
struct Map {
    optional: Option<String>,
}

let iter = [("optional", "value")].into_iter();
let de = MapDeserializer::<_, Error>::new(iter);
Map::deserialize(de); // Err: invalid type: string "value", expected option
```

A value deserializer always wraps a value that is present, so deserializing it as an option must yield `Some`. This implements `deserialize_option` as `visitor.visit_some(self)` for the scalar, string, bytes, seq, map, and enum value deserializers, matching how self-describing format deserializers (serde_json, toml, ...) behave. `UnitDeserializer` still returns `visit_none`, and the uninhabited `NeverDeserializer` is left unchanged.

Added two regression tests in `test_value.rs`: a direct `IntoDeserializer` -> `Option<T>` case and the exact `MapDeserializer` reproduction from the issue. Ran `cargo fmt --check`, `cargo clippy -p serde_core`, `cargo test -p serde_core`, and the relevant `test_suite` integration tests (stable and `cargo +nightly --features unstable`); all clean.

Fixes #3050.

### DIFF
--- a/serde_core/src/de/value.rs
+++ b/serde_core/src/de/value.rs
@@ -46,6 +46,24 @@ macro_rules! impl_copy_clone {
 
 ////////////////////////////////////////////////////////////////////////////////
 
+// A value deserializer always holds a value that is present, so deserializing
+// it as an `Option` must produce `Some`. Forwarding `deserialize_option` to
+// `deserialize_any` would instead make the visitor receive the held value
+// directly, which fails for `Option<T>` with "invalid type ..., expected
+// option".
+macro_rules! deserialize_option_some {
+    () => {
+        fn deserialize_option<V>(self, visitor: V) -> Result<V::Value, Self::Error>
+        where
+            V: de::Visitor<'de>,
+        {
+            visitor.visit_some(self)
+        }
+    };
+}
+
+////////////////////////////////////////////////////////////////////////////////
+
 /// A minimal representation of all possible errors that can occur using the
 /// `IntoDeserializer` trait.
 #[derive(Clone, PartialEq)]
@@ -291,9 +309,11 @@ macro_rules! primitive_deserializer {
 
             forward_to_deserialize_any! {
                 bool i8 i16 i32 i64 i128 u8 u16 u32 u64 u128 f32 f64 char str
-                string bytes byte_buf option unit unit_struct newtype_struct seq
+                string bytes byte_buf unit unit_struct newtype_struct seq
                 tuple tuple_struct map struct enum identifier ignored_any
             }
+
+            deserialize_option_some!();
 
             fn deserialize_any<V>(self, visitor: V) -> Result<V::Value, Self::Error>
             where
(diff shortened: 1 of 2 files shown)


oli-obk:
Screenshot_20260604-152335_SHIFT Home

Please don't spam LLM generated stuff all over github

PR #3073, opened 2026-06-16
Match untagged unit variant against an empty map or seq

### LINKED ISSUE #1374: Unexpected results from flattened untagged enum with empty variant.
I have the following code (also on [play.rust-lang.org])

```rust
extern crate serde; // 1.0.75
#[macro_use] extern crate serde_derive; // 1.0.75
extern crate toml; // 0.4.6

#[derive(Debug, Deserialize)]
struct Container {
    other: u16,
    #[serde(flatten)]
    inner: DeEnum,
}

#[derive(Debug, Deserialize)]
#[serde(untagged)]
enum DeEnum {
    NonEmptyVariant(NonEmpty),
    EmptyVariant,
}

#[derive(Debug, Deserialize)]
struct NonEmpty {
    key: u16
}

fn main() {
    let success_str = "\
        other = 12
        key = 12
    ";
    let parsed: Result<Container, toml::de::Error> = toml::from_str(success_str);
    println!("{:#?}", parsed.unwrap());
    
    let fail_str = "\
        other = 12
    ";
    let parsed: Result<Container, toml::de::Error> = toml::from_str(fail_str);
    println!("{:#?}", parsed.unwrap());
    // expected to match `DeEnum::EmptyVariant`, fails to match anything.
}
```

I expect that because the first enum variant fails to match, and the second one will match an empty input, that I would get a result with the second variant. However serde fails to parse the second variant.

Is this a bug, or have I misunderstood the documentation?

[play.rust-lang.org]: https://play.rust-lang.org/?gist=b50c5aaa31a8c62f87118472f4eae3e2&version=stable&mode=debug&edition=2015

StefanoD:
Fixes #1374

Generated by Claude:

## Problem

A unit variant of a `#[serde(untagged)]` enum was only reachable when the
buffered content was `unit`/`none`. When the content was an empty map (or empty
seq), the variant could not be matched and deserialization failed — even though
a unit variant logically represents "no data".

This is most visible together with `#[serde(flatten)]`. When a flattened field
has no remaining keys, the flattened enum is handed an **empty map**, not a
unit, so a unit variant behind `flatten` was impossible to reach:

```rust
#[derive(Deserialize)]
struct Container {
    other: u16,
    #[serde(flatten)]
    inner: DeEnum,
}

#[derive(Deserialize)]
#[serde(untagged)]
enum DeEnum {
    NonEmpty(NonEmpty),
    Empty, // <-- never matched when the flattened fields are absent
}

#[derive(Deserialize)]
struct NonEmpty {
    key: u16,
}
```

Deserializing input that lacks `key` failed with *"data did not match any
variant"* instead of producing `DeEnum::Empty`. The documented workaround was to
replace the unit variant `Empty` with an empty struct variant `Empty {}`.

## Cause

Untagged enums buffer the input into a `Content` and try each variant via
`ContentRefDeserializer`. A unit variant is deserialized through
`UntaggedUnitVisitor`, which only implemented `visit_unit` and `visit_none`. For
`Content::Map([])`, `deserialize_any` dispatches to `visit_map`, hitting the
default `visit_map` and erroring with *"invalid type: map, expected unit
variant"*.

## Fix

Teach `UntaggedUnitVisitor` to accept an **empty** map or **empty** seq as unit.
A **non-empty** map or seq is still rejected (`invalid_type`), so data is never
silently dropped when a preceding variant fails for an unrelated reason.

## Tests

Added two tests in `test_suite/tests/test_enum_untagged.rs`:

- `unit_variant_from_empty_map` — an empty map/seq matches the unit variant; the
  struct variant still wins when its fields are present; a non-empty unknown map
  still fails (no silent data loss).
- `flatten_unit_variant` — the exact scenario from the issue, with a flattened
  untagged enum.

Both tests fail before the change and pass after it. The full `serde_test_suite`
is green and `cargo clippy` is clean.

### DIFF
--- a/serde/src/private/de.rs
+++ b/serde/src/private/de.rs
@@ -3039,6 +3039,34 @@ mod content {
         {
             Ok(())
         }
+
+        // An empty seq is treated as unit so that a unit variant of an untagged
+        // enum is reachable for formats that present "no data" as an empty seq.
+        // A non-empty seq must not be silently accepted as a unit variant.
+        fn visit_seq<A>(self, mut seq: A) -> Result<(), A::Error>
+        where
+            A: SeqAccess<'de>,
+        {
+            match tri!(seq.next_element::<IgnoredAny>()) {
+                None => Ok(()),
+                Some(_) => Err(de::Error::invalid_type(Unexpected::Seq, &self)),
+            }
+        }
+
+        // An empty map is treated as unit so that a unit variant of an untagged
+        // enum is reachable. This is what makes a unit variant work behind
+        // `flatten`, where the absence of fields is presented as an empty map
+        // rather than as unit. See https://github.com/serde-rs/serde/issues/1374
+        // A non-empty map must not be silently accepted as a unit variant.
+        fn visit_map<A>(self, mut map: A) -> Result<(), A::Error>
+        where
+            A: MapAccess<'de>,
+        {
+            match tri!(map.next_key::<IgnoredAny>()) {
+                None => Ok(()),
+                Some(_) => Err(de::Error::invalid_type(Unexpected::Map, &self)),
+            }
+        }
     }
 }
 
--- a/test_suite/tests/test_enum_untagged.rs
+++ b/test_suite/tests/test_enum_untagged.rs
@@ -570,6 +570,109 @@ fn contains_flatten_with_integer_key() {
     );
 }
 
+// https://github.com/serde-rs/serde/issues/1374
+//
+// A unit variant of an untagged enum must be reachable when the input is an
+// empty map (or empty seq), not only when it is `unit`/`none`. This shows up in
+// particular together with `flatten`, where "no remaining fields" is presented
+// to the enum as an empty map rather than as a unit.
+#[test]
+fn unit_variant_from_empty_map() {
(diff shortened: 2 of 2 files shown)


oli-obk:
I do not wish to be the reviewer of LLM output. Please write PR descriptions yourself and make sure you understand the code changes yourself. Otherwise I can just run LLMs locally without you in the middle

StefanoD:
It actually doesn't matter where the code came from. It's a matter of being objective.
But you can take over this issue yourself, no problem. I was just addressing a 7.5 years old bug.
I just hope you don't blindly close this issue and keep ignoring the bug report.

zeenix:
> I do not wish to be the reviewer of LLM output. 

That's fair.

> Please write PR descriptions yourself and make sure you understand the code changes yourself.

That assumes the contributor has the time and interest for that. There is an issue and they're spending their tokens to get you a solution to an issue that's been unresolved for 8 years now.

> Otherwise I can just run LLMs locally without you in the middle

Also very fair but you didn't do that and the underlying issue remains unaddressed still.

If you have access to LLMs, why not use it to pre-review the PR to get an idication of the quality and then if it's worth it, you review yourself? Just any idea, up to you.

PR #3078, opened 2026-07-06
Feat: Add   independent_number  support for   rename_all  attributes

### LINKED ISSUE #3074: [Feature Request] Make `#[serde(rename_all = "")]` support independent number
For exampe, current snake case:
```rust
#[derive(Serialize)]
#[serde(rename_all = "snake_case")]
pub enum Chapter {
    Have1Apple, // will be "have1_apple"
}
```

Hope something can do this:
```rust
#[derive(Serialize)]
#[serde(rename_all = "snake_case", independent_number = true)]
pub enum Chapter {
    Have1Apple, // want to be "have_1_apple"
}
```

EbrahimEldesoky:
**Description**

Right now, when Serde converts identifiers between cases, numbers get grouped with whatever letters sit next to them. `Have1Apple` under `snake_case` becomes `have1_apple`, for example. I ran into a case where the target API expected numbers as their own segment instead (`have_1_apple`), and there wasn't a way to get that out of the existing rename rules.

This adds an opt-in `independent_number` flag for `rename_all` and `rename_all_fields` that treats digits as their own boundary, so numbers get split away from the surrounding letters instead of sticking to them.

**Example Usage:**

```rust
#[derive(Serialize)]
#[serde(rename_all = "snake_case", independent_number)]
pub enum Chapter {
    Have1Apple, // -> "have_1_apple"
}

#[derive(Serialize)]
#[serde(rename_all = "kebab-case", independent_number)]
struct Item {
    get2_bananas: bool, // -> "get-2-bananas"
}
```

**Changes:**

- Parses the new `independent_number` attribute in `serde_derive`.
- `RenameRule`'s `SnakeCase` matching now checks for digit/letter boundaries when the flag is set.
- Carries over automatically to `ScreamingSnakeCase`, `KebabCase`, and `ScreamingKebabCase`, so it doesn't need to be set per case style.
- Leaving the flag off keeps the old behavior untouched, so this is fully backwards compatible.
- Added unit and integration tests in `test_macros.rs` covering the new behavior.

**Related Issues**

_(Closes #3074)_

### DIFF
--- a/serde_derive/src/internals/attr.rs
+++ b/serde_derive/src/internals/attr.rs
@@ -138,6 +138,7 @@ fn unraw(ident: &Ident) -> Ident {
 pub struct RenameAllRules {
     pub serialize: RenameRule,
     pub deserialize: RenameRule,
+    pub independent_number: Option<bool>,
 }
 
 impl RenameAllRules {
@@ -147,6 +148,7 @@ impl RenameAllRules {
         Self {
             serialize: self.serialize.or(other_rules.serialize),
             deserialize: self.deserialize.or(other_rules.deserialize),
+            independent_number: self.independent_number.or(other_rules.independent_number),
         }
     }
 }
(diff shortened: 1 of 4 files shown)


EbrahimEldesoky:
is there anybody for review ?

PR #3089, opened 2026-07-24
Clarify deserializer hint semantics

### LINKED ISSUE #1128: Clarify documentation around what "hint" means in the context of Deserializer methods
In particular, "hint" was intended to indicate something that is allowed to be ignored, but is not allowed to be wrong. This is the same meaning of "hint" as in "What is the capital city of France? Hint: the name begins with Z." For example a Visitor that wants a string should not hint that it wants an integer -- data formats are not required to deserialize that correctly even when the input happens to contain a string.

ColumbusLabs:
Closes #1128.

Clarify that data formats may ignore the `deserialize_*` hints, but callers must still provide a hint that matches the input expected by their `Deserialize` implementation. Include a concrete string-versus-integer example matching the issue description.

Validation:
- `cargo test -p serde_core`
- `cargo fmt --all --check`
- `RUSTDOCFLAGS="-D warnings" cargo doc -p serde_core --no-deps`
- `git diff --check`

### DIFF
--- a/serde_core/src/de/mod.rs
+++ b/serde_core/src/de/mod.rs
@@ -909,9 +909,13 @@ where
 /// 2. The various `deserialize_*` methods. Non-self-describing formats like
 ///    Postcard need to be told what is in the input in order to deserialize it.
 ///    The `deserialize_*` methods are hints to the deserializer for how to
-///    interpret the next piece of input. Non-self-describing formats are not
-///    able to deserialize something like `serde_json::Value` which relies on
-///    `Deserializer::deserialize_any`.
+///    interpret the next piece of input. A data format may ignore these hints,
+///    but the hint must not be wrong. For example, a `Deserialize`
+///    implementation expecting string input must not request `deserialize_i64`,
+///    even if a particular format would happen to accept a string through that
+///    method.
+///    Non-self-describing formats are not able to deserialize something like
+///    `serde_json::Value` which relies on `Deserializer::deserialize_any`.
 ///
 /// When implementing `Deserialize`, you should avoid relying on
 /// `Deserializer::deserialize_any` unless you need to be told by the
--- a/serde_core/src/de/mod.rs
+++ b/serde_core/src/de/mod.rs
@@ -909,11 +909,17 @@ where
 /// 2. The various `deserialize_*` methods. Non-self-describing formats like
 ///    Postcard need to be told what is in the input in order to deserialize it.
 ///    The `deserialize_*` methods are hints to the deserializer for how to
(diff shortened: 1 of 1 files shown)


serde_core/src/de/mod.rs
```diff
@@ -909,9 +909,13 @@ where
 /// 2. The various `deserialize_*` methods. Non-self-describing formats like
 ///    Postcard need to be told what is in the input in order to deserialize it.
 ///    The `deserialize_*` methods are hints to the deserializer for how to
-///    interpret the next piece of input. Non-self-describing formats are not
-///    able to deserialize something like `serde_json::Value` which relies on
-///    `Deserializer::deserialize_any`.
+///    interpret the next piece of input. A data format may ignore these hints,
+///    but the hint must not be wrong. For example, a `Deserialize`
+///    implementation expecting string input must not request `deserialize_i64`,
+///    even if a particular format would happen to accept a string through that
```
Mingun: I cannot say that this is good explanation. `Visitor` is responsible for making the final decision, and this is part of `Deserialize`. So it is not a "particular format", but "particular type". `Deserializer` should not reject the request if the hint does not match what it has -- instead, it should delegate this decision to the `Visitor`, calling the appropriate `Visitor` method. Unfortunately, `Content[Ref]Deserializer` does not currently follow this rule, which is fixed in #2811.

I would say that hints are needed in cases where different types are stored in the same way at the storage level. An example is XML. Both the string `"123"` and the number `123` in XML are presented in the same form:
- as an attribute: `str="123" int="123"`
- as an element: `<str>123</str><int>123</int>`

Since the XmlDeserializer stores everything internally as strings, when requesting a number, it should try to convert the string to a number, and if this is not possible, give the string (and not an error). If the `Visitor` is not able to accept the string, it will report error himself which will also (most likely) give a more clear error description.

Compare this to the JSON approach, where strings and numbers differ at the storage level (`"123"` vs `123`), so JsonDeserializer should not try to convert strings to numbers (but _still should_ call `visit_str` if it has a string, but hint requested `deserialize_u64`).

ColumbusLabs: Updated in 5478122b. The docs now explain that hints help formats disambiguate types sharing a storage representation, using XML as the example, and contrast this with JSON representations. They also explicitly make the Visitor responsible for accepting or rejecting the type presented by the deserializer; the compatibility warning is now phrased in terms of the types the Visitor accepts. Verified with cargo test -p serde_core, strict rustdoc, formatting, and diff checks.

oli-obk:
Screenshot_20260725-154018_SHIFT Home



> * git diff --check

Please don't generate PRs across github just because you can.

